PR #4, opened 2022-04-29
feat: enhance Kyma type

adityabhatia:


operator/api/v1alpha1/kyma_types.go
```diff
@@ -20,24 +20,92 @@ import (
...
+
+type KymaState string
+
+// Valid Kyma States
+const (
+	// KymaStateReady signifies Kyma is ready
+	KymaStateReady KymaState = "Ready"
+
+	// KymaStateProcessing signifies Kyma is reconciling
+	KymaStateProcessing KymaState = "Processing"
```
jakobmoellersap, line 58: Why do we need custom states here instead of reusing the states from Apimachinery (e.g. like Deployment with "Progressing")

jakobmoellersap (commented):
Overall very cool! I like how you kept close to the original k8s specs. Still I am wondering wether we can simply include the API Spec from Deployment as is without introducing custom states. I think we can go ahead either way, just want to bring up the discussion!

jakobmoellersap:
Also, https://github.com/kyma-project/kyma-operator/blob/main/operator/pkg/client/listers/operator.kyma-project.io/v1alpha1/kyma.go#L95 errors out on main, do you know why that is? possibly related to api spec changes

jakobmoellersap (approved):

PR #7, opened 2022-05-02
feat: initial code for creation of component CRs

adityabhatia:


operator/controllers/kyma_controller.go
```diff
@@ -39,24 +54,135 @@ type KymaReconciler struct {
...
+			return fmt.Errorf("error during config map unmarshal %w", err)
+		}
+
+		gvr := schema.GroupVersionResource{
+			Group:    componentYaml["group"].(string),
+			Resource: componentYaml["resource"].(string),
+			Version:  componentYaml["version"].(string),
+		}
+
+		if exists, err := r.GetUnstructuredResource(ctx, gvr, componentName, req.Namespace); client.IgnoreNotFound(err) != nil {
```
jakobmoellersap, line 120: why bother checking for exists if you anyhow ignore it?

operator/controllers/kyma_controller.go
```diff
@@ -39,24 +54,135 @@ type KymaReconciler struct {
...
 func (r *KymaReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
-		For(&operatorv1alpha1.Kyma{}).
+	r.Builder = ctrl.NewControllerManagedBy(mgr)
+	return r.Builder.For(&operatorv1alpha1.Kyma{}).
+		//Owns(&corev1.ConfigMap{}).
+		Watches(
+			&source.Kind{Type: &corev1.ConfigMap{}},
+			handler.EnqueueRequestsFromMapFunc(r.findObjectsForConfigMap),
+			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
```
jakobmoellersap, line 179: I think this is gonna happen quite often, so a watch on this predicate is quite expensive, no?

operator/controllers/kyma_controller.go
```diff
@@ -39,24 +54,135 @@ type KymaReconciler struct {
...
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *KymaReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
-		For(&operatorv1alpha1.Kyma{}).
+	r.Builder = ctrl.NewControllerManagedBy(mgr)
+	return r.Builder.For(&operatorv1alpha1.Kyma{}).
+		//Owns(&corev1.ConfigMap{}).
+		Watches(
+			&source.Kind{Type: &corev1.ConfigMap{}},
```
jakobmoellersap, line 177: Why did you use Watches instead of For here?

operator/controllers/kyma_controller.go
```diff
@@ -18,19 +18,34 @@ package controllers
...
 
 // KymaReconciler reconciles a Kyma object
 type KymaReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme  *runtime.Scheme
+	Builder *builder.Builder
 }
 
 //+kubebuilder:rbac:groups=operator.kyma-project.io,resources=kymas,verbs=get;list;watch;create;update;patch;delete
```
jakobmoellersap, line 51: Im wondering how you expanded the RBAC generation without extending the logic here, kubebuilder has inbuilt rbac generation I think

jakobmoellersap (commented):
Awesome 🚀

operator/controllers/kyma_controller.go
adityabhatia: It is sure, but it will be restricted by cache selectors, coming up next :)

operator/controllers/kyma_controller.go
adityabhatia: since it is an externally managed resource and is not the object to be reconciled

operator/controllers/kyma_controller.go
adityabhatia: Do you mean RBAC for ConfigMaps or unstructured resources? 
We cannot annotate unstructured resources for RBAC generation in this case, it has to be explicit. My idea would be to add it with CRD installation.

PR #17, opened 2022-05-05
Improve watching of component reconcilers + plugin manifest operator

adityabhatia:


jakobmoellersap (changes requested):
I think we should get the helm-component in its own operator directly instead of merging here, wdyt?

operator/controllers/kyma_controller.go
```diff
@@ -191,10 +193,13 @@ func (r *KymaReconciler) ReconcileFromConfigMap(ctx context.Context, req ctrl.Re
 						"operator.kyma-project.io/release":         progression.New,
 					},
 				},
-				"spec": componentYaml["spec"],
+				"spec":   componentYaml["spec"],
+				"status": map[string]string{},
```
jakobmoellersap, line 197: this breaks on deep copy, you will have to use `interface{}` here
```suggestion
				"status": map[string]interface{},
```

operator/config/samples/component-integration-installed/rbac/component_rbac.yaml
```diff
@@ -8,6 +8,7 @@ rules:
     resources:
       - serverlessconfigurations
       - istioconfigurations
+      - helms
```
jakobmoellersap, line 11:
```suggestion
      - manifestconfigurations
```

jakobmoellersap: I think we should keep it generic if we later on change our approach, alternative would be `helmconfigurations`

operator/controllers/kyma_controller.go
```diff
@@ -348,18 +372,63 @@ func (r *KymaReconciler) GetUnstructuredResource(ctx context.Context, gvr schema
...
+			var ownerName string
+
+			for _, ownerRef := range ownerRefs {
+				if kymaObj.Kind == ownerRef.Kind {
+					ownerName = ownerRef.Name
+					break
+				}
+			}
+
+			if err := r.Get(context.TODO(), types.NamespacedName{Name: ownerName, Namespace: componentObj.GetNamespace()}, &kymaObj); err != nil {
```
jakobmoellersap, line 403: I think a context here needs to be propagated downstream.

operator/controllers/kyma_controller.go
```diff
@@ -348,18 +372,63 @@ func (r *KymaReconciler) GetUnstructuredResource(ctx context.Context, gvr schema
...
+				return
+			}
+
+			kymaObj.Status.State = operatorv1alpha1.KymaStateReady
+			kymaObj.Status.ObservedGeneration = kymaObj.Generation
+			var condition *operatorv1alpha1.KymaCondition
+
+			for _, existingCondition := range kymaObj.Status.Conditions {
+				if existingCondition.Type == operatorv1alpha1.ConditionTypeReady {
+					condition = &existingCondition
```
jakobmoellersap, line 413: this loop is always taking the last condition right now, I think we could make it break here. The way I see it every state will have one condition. so checking on kymaObj.Status.Conditions could mean 3-4 conditions. That means we will have to find a way to get the right condition. I think we need the technical reason message to be mapped to the reconciler in this case

operator/controllers/kyma_controller.go
```diff
@@ -348,18 +372,63 @@ func (r *KymaReconciler) GetUnstructuredResource(ctx context.Context, gvr schema
...
+				}
+				kymaObj.Status.Conditions = append(kymaObj.Status.Conditions, *condition)
+			}
+
+			condition.LastTransitionTime = &metav1.Time{Time: time.Now()}
+			condition.Message = "successfully installed component type: " + e.ObjectNew.GetObjectKind().GroupVersionKind().String()
+			condition.Reason = "all component installed"
+			condition.Status = operatorv1alpha1.ConditionStatusTrue
+
+			if err := r.Status().Update(context.TODO(), &kymaObj); err != nil {
```
jakobmoellersap, line 429: this needs a retry backoff and a context propagation I think

operator/controllers/kyma_controller.go
adityabhatia: Every state cannot have one condition each IMO. 
I thought we discussed on having one Ready condition / component

operator/config/samples/component-integration-installed/rbac/component_rbac.yaml
adityabhatia: done

operator/controllers/kyma_controller.go
adityabhatia: I removed it completely, since we do not need an empty map anyways

operator/controllers/kyma_controller.go
jakobmoellersap: exactly but if we have one ready condition / component then we will find here 5 conditions for 5 components in one kyma status. Right now the logic would exit with the last ready condition. If we have the setup of 3 conditions with the conditions set to "ready, not ready, ready", and you wanted to change the second condition, then you would only change the last one to ready, not causing any change, no?

operator/controllers/kyma_controller.go
```diff
@@ -122,23 +122,26 @@ func (r *KymaReconciler) GetConfigMap(ctx context.Context, component string) (*c
...
 	for _, component := range kymaObj.Spec.Components {
+		componentNames = append(componentNames, component.Name)
 		configMap, err := r.GetConfigMap(ctx, component.Name)
 		if err != nil {
 			logger.Error(err, "component mapping ConfigMap read error, will not re-queue resource %s", req.NamespacedName.String())
 			return err
 		}
 
 		componentName := component.Name + "-name"
+		//+ uuid.New().String()
```
jakobmoellersap, line 144:
```suggestion
```

operator/controllers/kyma_controller.go
```diff
@@ -329,37 +349,57 @@ func (r *KymaReconciler) updateKymaStatus(ctx context.Context, kyma *operatorv1a
...
+	r.Builder = ctrl.NewControllerManagedBy(mgr)
+	return r.Builder.For(&operatorv1alpha1.Kyma{}).Complete(r)
+}
 
-	dynamicClient, err := dynamic.NewForConfig(config)
+func (r *KymaReconciler) ComponentChangeHandler(e event.UpdateEvent, q workqueue.RateLimitingInterface) {
+	objectBytes, err := json.Marshal(e.ObjectNew)
 	if err != nil {
-		return nil, err
+		panic(err)
```
jakobmoellersap, line 361: do we have no smart way to not panic here? this will cause a crash right?

jakobmoellersap (approved):

PR #19, opened 2022-05-06
update crs to get updated dynamically based on versioned configmaps

jakobmoellersap:


operator/config/samples/component-integration-installed/config_map_istio_2.0.yaml
```diff
@@ -0,0 +1,19 @@
...
+    "operator.kyma-project.io/controller-name": "istio"
+    "operator.kyma-project.io/release": "v2.0.2"
+data:
+  istio: |-
+    kind: IstioConfiguration
+    group: kyma.kyma-project.io
+    resource: istioconfigurations
+    version: v1alpha1 
+    spec:
+      foo: "created From Dynamic Client for {{installOperation .}} to {{installTargetVersion .}}"
```
pbochynski, line 18: In the kyma controller, you use labels for `applied-as` and `release`. I think we should stick to labels only and not try to render custom resources with templating. Such templates introduce coupling and are hard to trace/refactor when you rename some variables.

operator/controllers/kyma_controller.go
```diff
@@ -220,6 +208,30 @@ func (r *KymaReconciler) ReconcileFromConfigMap(ctx context.Context, req ctrl.Re
 	return nil
 }
 
+func (r *KymaReconciler) GetTemplatedComponent(templateName, componentTemplate string, release *KymaReleaseProgressionInfo) (map[string]interface{}, error) {
```
pbochynski, line 211: I am not sure if we should template components. It can become fragile. You provide the same information in labels.

operator/config/samples/component-integration-installed/config_map_istio_2.0.yaml
jakobmoellersap: Had to test the upgrade somehow :D 

Can remove this without issues, question however: How we are going to inject values in the spec later if not through a template? 

I cannot imagine there will be always a (purely) static configuration for these config maps. Especially for dynamic generation of CRs, I think we need a way to adapt values and create the chart based on some data on the environment or landscape. If its not templating, we should think of something else that fits. We could also think of doing that simply in the component reconciler, but then we limit the CR definitions here to be purely static.

Towards your concern: I think Labels are even harder to trace than Spec changes as they do not introduce a generational change in the status.

Let me know if you still think we should remove it, and we can proceed accordingly

adityabhatia: {{installTargetVersion .}} = release label from versioned config map
{{installOperation .}} = upgrade/downgrade
{{installOriginVersion .}}  = release label on current component CR

Looking at this I don't think we necessarily require templating. 
Maybe I didn't understand the use case behind templating here, so correct me if I missed something.

operator/controllers/kyma_controller.go
```diff
@@ -350,11 +362,11 @@ func (r *KymaReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		Watches(
-			&source.Kind{Type: &corev1.ConfigMap{}},
-			handler.EnqueueRequestsFromMapFunc(r.findObjectsForConfigMap),
-			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
-		).
+		//Watches(
+		//	&source.Kind{Type: &corev1.ConfigMap{}},
+		//	handler.EnqueueRequestsFromMapFunc(r.findObjectsForConfigMap),
+		//	builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
+		//).
```
adityabhatia, line 369: please rebase :)

operator/config/samples/component-integration-installed/config_map_istio_2.0.yaml
jakobmoellersap: okay before we get into another discussion, ill just remove it :D we can just readd it if we need it later on

operator/controllers/kyma_controller.go
```diff
@@ -133,61 +158,43 @@ func (r *KymaReconciler) ReconcileFromConfigMap(ctx context.Context, req ctrl.Re
...
-		res.SetGroupVersionKind(schema.GroupVersionKind{
-			Group:   componentYaml["group"].(string),
-			Kind:    componentYaml["kind"].(string),
-			Version: componentYaml["version"].(string),
-		})
+		res, err := r.GetUnstructuredResource(ctx, schema.GroupVersionResource{
+			Group:    componentYaml["group"].(string),
+			Resource: componentYaml["resource"].(string),
+			Version:  componentYaml["version"].(string),
+		}, componentName, req.Namespace)
```
adityabhatia, line 184: Maybe we could keep the kubebuilder provided r.client.Get() for standardisation?

PR #37, opened 2022-05-19
switch from ConfigMap templates to ModuleTemplate

jakobmoellersap:
This means we dont have to calculate hashes and can rely on generation + channel instead of resourceversion + hash


operator/config/samples/component-integration-installed/operator_v1alpha1_moduletemplate_istio_rapid.yaml
```diff
@@ -0,0 +1,18 @@
...
+    "operator.kyma-project.io/channel": "rapid"
+spec:
+  data:
+    istio: |-
+      kind: IstioConfiguration
+      group: kyma.kyma-project.io
+      resource: istioconfigurations
+      version: v1alpha1
+      spec:
+        foo: "Rapid Speed!"
```
adityabhatia, line 18: should be revised, without component name + spec inside spec

operator/config/samples/component-integration-installed/operator_v1alpha1_moduletemplate_istio_stable.yaml
```diff
@@ -0,0 +1,18 @@
...
+    "operator.kyma-project.io/channel": "stable"
+spec:
+  data:
+    istio: |-
+      kind: IstioConfiguration
+      group: kyma.kyma-project.io
+      resource: istioconfigurations
+      version: v1alpha1
+      spec:
+        foo: "Stable Speed!"
```
adityabhatia, line 18: should be revised, without component name + spec inside spec
**-- same for other moduletemplate resources**

operator/api/v1alpha1/moduletemplate_types.go
```diff
@@ -0,0 +1,57 @@
...
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// ModuleTemplateSpec defines the desired state of ModuleTemplate
+type ModuleTemplateSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	Data map[string]string `json:"data,omitempty"`
```
adityabhatia, line 31:
~~~~~~
Data TemplateData `json:"data,omitempty"`


type TemplateData struct {
      Kind string
      Group string
      Resource string
      Version string
      Settings map[string]interface{} // for component CR spec
}

operator/pkg/release/lookup.go
```diff
@@ -99,9 +99,9 @@ func (c *channelConfigMapLookup) Lookup(ctx context.Context) (*corev1.ConfigMap,
 	}
 
 	if actualChannel != c.channel {
-		log.FromContext(ctx).Info(fmt.Sprintf("using %s (instead of %s) for component %s", actualChannel, c.channel, c.component.Name))
+		log.FromContext(ctx).V(0).Info(fmt.Sprintf("using %s (instead of %s) for component %s", actualChannel, c.channel, c.component.Name))
 	} else {
-		log.FromContext(ctx).Info(fmt.Sprintf("using %s for component %s", actualChannel, c.component.Name))
+		log.FromContext(ctx).V(0).Info(fmt.Sprintf("using %s for component %s", actualChannel, c.component.Name))
```
adityabhatia, line 104: isnt't verbosity level by default 0?

operator/pkg/release/lookup.go
```diff
@@ -3,29 +3,29 @@ package release
 import (
```
adityabhatia, line 3: All configmap instances should be renamed in this module

operator/pkg/watch/template_change.go
```diff
@@ -0,0 +1,120 @@
...
+	ctx context.Context,
+) func(event.CreateEvent, workqueue.RateLimitingInterface) {
+	watch := h.Watch(ctx)
+	return func(e event.CreateEvent, queue workqueue.RateLimitingInterface) {
+		requests := watch(e.Object)
+		for _, request := range requests {
+			queue.AddRateLimited(request)
+		}
+	}
+}
```
adityabhatia, line 120: are these functions used anywhere, maybe I missed it?

adityabhatia (dismissed):

operator/api/v1alpha1/moduletemplate_types.go
jakobmoellersap: map[string]interface is not a valid spec unfortunately, i tried this but it didnt work out as I expected since it doesnt compile

operator/pkg/release/lookup.go
jakobmoellersap: True, left it by accident from a bit of debugging, Ill remove it

operator/pkg/watch/template_change.go
jakobmoellersap: These functions can be used as replacement of the mapfunc approach and I left them there to switch between the approaches for testing, will be removed

adityabhatia (approved):
Great stuff 🚀

PR #38, opened 2022-05-23
Implement finalizers

adityabhatia:


operator/pkg/watch/template_change.go
```diff
@@ -50,7 +51,8 @@ func (h *TemplateChangeHandler) Watch(ctx context.Context) handler.MapFunc {
 			Name:      template.GetName(),
 		}
 		for _, kyma := range kymas.Items {
-			if kyma.Spec.Channel != v1alpha1.Channel(channel) {
+			if kyma.Spec.Channel != channel {
+				// what if the channel belongs to the component only?
```
jakobmoellersap, line 55: This case is uncovered, good catch. I think I will have to adjust the check to also check for the component condition. I think we can do this from L60 on and just move the channel check down.

operator/controllers/kyma_controller.go
```diff
@@ -112,17 +114,18 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
 	return r.updateKymaStatus(ctx, kyma, operatorv1alpha1.KymaStateProcessing, "initial state")
 }
 
-func (r *KymaReconciler) HandleProcessingState(ctx context.Context, logger *logr.Logger, kyma *operatorv1alpha1.Kyma, templates release.TemplateLookupResultsByName) error {
+func (r *KymaReconciler) HandleProcessingState(ctx context.Context, logger *logr.Logger, kyma *operatorv1alpha1.Kyma) error {
 	logger.Info("processing " + kyma.Name)
 
+	templates, err := release.GetTemplates(r, ctx, kyma)
+	if err != nil {
+		return r.KymaStatus().UpdateStatus(ctx, kyma, operatorv1alpha1.KymaStateError, "templates could not be fetched")
```
jakobmoellersap, line 126: I think you should use  `updateKymaStatus` here or change all calls to use the Status Handler directly

jakobmoellersap: Also this error we should probably propagate somehow into the message since its quite important to know why the template fetching failed (e.g. because of multiple candidates)

operator/controllers/kyma_controller.go
```diff
@@ -137,29 +140,76 @@ func (r *KymaReconciler) HandleProcessingState(ctx context.Context, logger *logr
...
 }
 
-func (r *KymaReconciler) HandleDeletingState(_ context.Context) error {
-	return nil
-}
+func (r *KymaReconciler) HandleDeletingState(ctx context.Context, logger *logr.Logger, kyma *operatorv1alpha1.Kyma) (bool, error) {
+	templates, err := release.GetTemplates(r, ctx, kyma)
+	if err != nil {
+		return false, r.KymaStatus().UpdateStatus(ctx, kyma, operatorv1alpha1.KymaStateError,
+			"deletion cannot proceed - templates could not be fetched")
```
jakobmoellersap, line 147: Same as above: we should probably propagate the error message for transparency.

operator/controllers/kyma_controller.go
```diff
@@ -137,29 +140,76 @@ func (r *KymaReconciler) HandleProcessingState(ctx context.Context, logger *logr
...
-		logger.Info("skipping reconciliation for " + kyma.Name + ", already reconciled!")
+		desiredComponentStruct := &lookupResult.Template.Spec.Data
+		desiredComponentStruct.SetName(component.Name + "-name")
+		desiredComponentStruct.SetNamespace(kyma.GetNamespace())
+
+		actualComponentStruct := desiredComponentStruct.DeepCopy()
+
+		if err = r.Get(ctx, client.ObjectKeyFromObject(actualComponentStruct), actualComponentStruct); err == nil {
+			// component CR still exists
+			logger.Info(fmt.Sprintf("deletion cannot proceed - waiting for component CR %s to be deleted for %s",
```
jakobmoellersap, line 165: I think we should write this as a DeletionCondition into the status instead of just logging here, then we can immediately see in the Kyma why the deletions is still processing, so not only "ReadyCondition" but also "DeletionCondition". What do you think?

operator/controllers/kyma_controller.go
```diff
@@ -137,29 +140,76 @@ func (r *KymaReconciler) HandleProcessingState(ctx context.Context, logger *logr
...
+				desiredComponentStruct.GetName(), client.ObjectKeyFromObject(kyma)))
+			return true, nil
```
jakobmoellersap, line 167: Instead of returning a bool I think we should introduce a new Type of Error and check for the Err similar to `client.IsNotFound(err)`

jakobmoellersap (changes requested):
Just smaller things, really cool that you took out the templateconfigstatus for an inplace lookup and error condition 👍

adityabhatia:
> Bit hard to read but works I guess 😐 👍
It was a quick implementation, will try to improve in the next iteration

PR #39, opened 2022-05-23
fix: template update

adityabhatia:


jakobmoellersap (approved):
I think this still doesnt gover the case when the channel of a component got updates but I will test during the day

PR #40, opened 2022-05-24
Fix component naming and some other bugs

ruanxin:
- add missing package for dockerfile
- fix ServiceAccount naming, missing prefix
- change manifest naming

### DIFF
--- a/operator/pkg/util/kyma_util.go
+++ b/operator/pkg/util/kyma_util.go
@@ -47,7 +47,7 @@ func GetUnstructuredComponentFromTemplate(templates release.TemplateLookupResult
 	}
 
 	desiredComponentStruct := &lookupResult.Template.Spec.Data
-	desiredComponentStruct.SetName(componentName + "-name")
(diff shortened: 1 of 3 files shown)


operator/pkg/util/kyma_util.go
```diff
@@ -47,7 +47,7 @@ func GetUnstructuredComponentFromTemplate(templates release.TemplateLookupResult
 	}
 
 	desiredComponentStruct := &lookupResult.Template.Spec.Data
-	desiredComponentStruct.SetName(componentName + "-name")
+	desiredComponentStruct.SetName(componentName + kyma.Name)
```
ruanxin, line 50: To Reviewer: this change avoid unique component name when deploy multiple kyma objects, and due to current load test setup, we relay on `-` to split the component name to adapt into component<->target cluster mapping, so here I haven't use `-` as a separator (just want to avoid maintenance difference between branches).

jakobmoellersap (approved):

PR #41, opened 2022-06-03
Implement listener component

khlifi411:
**Changes proposed in this PR**

registered listener module with the controller manager to react on SKR events sent by the watcher


operator/controllers/kyma_controller.go
```diff
@@ -416,6 +419,15 @@ func (r *KymaReconciler) SetupWithManager(setupLog logr.Logger, mgr ctrl.Manager
 		handler.EnqueueRequestsFromMapFunc(r.TemplateChangeHandler().Watch(context.TODO())),
 		builder.WithPredicates(predicate.GenerationChangedPredicate{}))
 
+	controllerBuilder = controllerBuilder.Watches(
```
jeremyharisch, line 422: We need to check if this does not trigger the `reconcile` function. Since the Builder only works if the `forInput` is not empty. This input is set in line 379. Then if an event happens, the respective handler function gets triggered and then the reconcile function. We are currently handling the events in the same way in the watcher, and there we encounter exactly this problem.

operator/pkg/listener/watcher_events.go
```diff
@@ -0,0 +1,147 @@
...
+		//Label component CR with a current timestamp
+		//when an event is received from the SKR watcher
+		componentName := genericEvent.Object.GetName()
+		clusterName := genericEvent.Object.GetClusterName()
+		h.logger.WithValues(
+			"component", componentName,
+			"cluster-name", clusterName,
+		).Info("started dispatching event")
+
+		kymaCR, err := h.getKymaCR(ctx, clusterName)
```
jakobmoellersap, line 40: I think it's a bad practice to call the k8s API in the Handler, Ideally this should only requeue and the queue should then backout mentioning that the Kyma doesn't exist

jakobmoellersap: Then you also dont need to do the error handling here and the handler gets way shorter

operator/pkg/listener/watcher_events.go
```diff
@@ -0,0 +1,147 @@
...
+		if err != nil {
+			h.logger.WithValues(
+				"component", componentName,
+				"cluster-name", clusterName,
+			).Error(err, "failed to get Kyma CR for component")
+			return
+		}
+
+		componentCR, err := h.getComponentCR(ctx, kymaCR, componentName, clusterName, q)
```
jakobmoellersap, line 49: Same for the component, this failure should be handled in the reconciliation loop imo

jakobmoellersap: When you draw out all these failure conditions and just put it in the reconciliation loop I think we could achieve a better scaling for many requests since the querying of the api server will be naturally limited by the informer queue

operator/pkg/listener/watcher_events.go
```diff
@@ -0,0 +1,147 @@
...
+		).Info("successfully triggered component reconciliation loop")
+	}
+}
+
+func (h *WatcherEventsHandler) updateComponentCRLabels(ctx context.Context, componentCR *unstructured.Unstructured, kymaCR *v1alpha1.Kyma, q workqueue.RateLimitingInterface) error {
+	componentLabels := componentCR.GetLabels()
+	labelValue := strings.ReplaceAll(time.Now().UTC().Format(time.RFC3339), ":", ".")
+	componentLabels[labels.ListenerLastUpdated] = labelValue
+	componentCR.SetLabels(componentLabels)
+	err := h.Update(ctx, componentCR, &client.UpdateOptions{})
```
jakobmoellersap, line 78: I think we should never update a component CR Label through a Change Watcher with the K8s API. I think it could make more sense to instead issue a Kubernetes Event or set an Annotation instead. But even this is dangerous since many requests can lead to overloading the api server with label updates.

operator/pkg/listener/watcher_events.go
```diff
@@ -0,0 +1,147 @@
...
+	return &kymaCRsForCluster.Items[0], nil
+}
+
+func (h *WatcherEventsHandler) handleComponentNotFoundError(ctx context.Context, kymaCR *v1alpha1.Kyma, componentKey client.ObjectKey, q workqueue.RateLimitingInterface) error {
+	latestKymaCR := &v1alpha1.Kyma{}
+	if err := h.Get(ctx, client.ObjectKeyFromObject(kymaCR), latestKymaCR); err != nil {
+		return fmt.Errorf("could not get updated Kyma CR for component: %s", err)
+	}
+
+	if kymaCR.Status.State != v1alpha1.KymaStateProcessing {
```
jakobmoellersap, line 140: This is a big assumption, You won't know what the state of the Kyma will be in when it gets taken out of the  for processing. Why not requeue it every time you receive an event? Is this a load issue?

jakobmoellersap (commented):
I think if I understand the logic correctly, this logic will not actually be able to react to the change of the component directly but instead only trigger the reconciliation of Kyma. 

Would the component operator need to manually determine if a change was necessary? Right now I do not see a way how an additional module can be added / removed from kyma based on this implementation. Since you do not touch Kyma, it will trigger a Reconciliation but immediately have the same state again as before right?

What I am referring to is that a Kyma Installation might not have the definition for a module in place and then the watcher will pickup a configmap. Then this will trigger a reconciliation but since the module is not actually listed in the kyma spec, nothing will happen I think

operator/pkg/listener/watcher_events.go
jeremyharisch: The updating of the label has nothing do to with what we want to achieve, this is just a very temporary thing, to see that the listener is able to interact with the CR. In the end the listener should NOT set any label, it should patch the Component CR instead.

operator/pkg/listener/watcher_events.go
khlifi411: Totally agree! Thanks for bringing it up 😄 
As we discussed today, all of these concerns will be deprecated by the new design, for which I pushed a first draft implementation a few minutes ago. Feel free to have another look 👍

operator/pkg/listener/watcher_events.go
khlifi411: As mentioned above. The new design approach resolves the above concerns!

operator/pkg/listener/watcher_events.go
khlifi411: As mentioned above. The new design approach resolves the above concerns!

operator/controllers/kyma_controller.go
khlifi411: Concern resolved after we tested it together 😄

jakobmoellersap:
sorry but this pr now has quite a few conflicts. since you dont add too much it should mostly be a rebase though. very sorry for the inconvenience :(

khlifi411:
> sorry but this pr now has quite a few conflicts. since you dont add too much it should mostly be a rebase though. very sorry for the inconvenience :(

No Problem 😄 My PR is way small. I should be able to rebase in no time 👍

operator/go.mod
```diff
@@ -3,13 +3,30 @@ module github.com/kyma-project/kyma-operator/operator
...
 	github.com/onsi/ginkgo v1.16.5
-	github.com/onsi/gomega v1.17.0
-	k8s.io/api v0.23.0
-	k8s.io/apimachinery v0.23.0
-	k8s.io/client-go v0.23.0
-	sigs.k8s.io/controller-runtime v0.11.0
+	github.com/onsi/gomega v1.18.1
+	go.uber.org/zap v1.19.1
+	golang.org/x/time v0.0.0-20220210224613-90d013bbcef8
+	k8s.io/api v0.24.2
```
jakobmoellersap, line 12: This for sure is not correct. All these dependencies HAVE to stick to v0.23.0 I would expect only the kyma-listener dependency to be upgraded. All other things will be upgraded in the go1.18 migration that is currently being prepared in another PR.

jakobmoellersap (changes requested):
LGTM :) There is one change I request as we shouldnt bump all versions from k8s deps in this PR since we will need a different PR to handle this correctly (together with script changes and go1.18 migration).

operator/controllers/kyma_controller.go
```diff
@@ -406,6 +411,16 @@ func (r *KymaReconciler) SetupWithManager(mgr ctrl.Manager, options controller.O
 		}
 	}
 
+	controllerBuilder = controllerBuilder.Watches(
+		&source.Kind{Type: &operatorv1alpha1.ModuleTemplate{}},
+		handler.EnqueueRequestsFromMapFunc(watch.NewTemplateChangeHandler(r).Watch(context.TODO())),
+		builder.WithPredicates(predicate.GenerationChangedPredicate{}))
+
+	// here we define a watch on secrets for the kyma operator so that the cache is picking up changes
+	controllerBuilder = controllerBuilder.Watches(&source.Kind{Type: &v1.Secret{}}, handler.Funcs{})
```
jeremyharisch, line 420: Pls remove - maybe rebase issue

jeremyharisch, line 420: Pls remove - maybe rebase issue

jeremyharisch, line 420: Pls remove - maybe rebase issue

operator/controllers/kyma_controller.go
```diff
@@ -406,6 +411,16 @@ func (r *KymaReconciler) SetupWithManager(mgr ctrl.Manager, options controller.O
 		}
 	}
 
+	controllerBuilder = controllerBuilder.Watches(
```
jeremyharisch, line 414: 414 - 420 Pls remove - maybe rebase issue

operator/config/manager/gateway.yaml
```diff
@@ -0,0 +1,71 @@
...
+kind: Gateway
+metadata:
+  name: kyma-gateway
+  namespace: default
+spec:
+  selector:
+    istio: ingressgateway
+  servers:
+    - hosts:
+        - '*'
```
jakobmoellersap, line 11: Dangerous Gateway configuration

operator/config/manager/kustomization.yaml
```diff
@@ -8,3 +9,9 @@ configMapGenerator:
 - files:
   - controller_manager_config.yaml
   name: manager-config
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
```
jakobmoellersap, line 15: why do you need this kustomization?

operator/controllers/kyma_controller.go
```diff
@@ -53,6 +56,8 @@ type RequeueIntervals struct {
 	Waiting time.Duration
 }
 
+const componentName = "kyma"
```
jakobmoellersap, line 59: can we make it so that it instead uses Kind? then theres already a constant available in types

operator/controllers/kyma_controller.go
```diff
@@ -416,6 +421,8 @@ func (r *KymaReconciler) SetupWithManager(mgr ctrl.Manager, options controller.O
 		}
 	}
 
+	controllerBuilder = listener.RegisterListenerComponent(logger, mgr, controllerBuilder, listenerAddr, &handler.EnqueueRequestForObject{}, componentName)
```
jakobmoellersap, line 424: why do you need explicit enqueueforobject here, wouldn't that be done through your handler?

operator/config/manager/gateway.yaml
jeremyharisch: I agree, but since we do not have a configured host for this PoC, we need to use the wildcard at this time fo the PoC. If we have a host in place, we definitely need to adapt it. Same goes for the VirtualService

operator/controllers/kyma_controller.go
```diff
@@ -419,6 +425,8 @@ func (r *KymaReconciler) SetupWithManager(mgr ctrl.Manager, options controller.O
...
```
adityabhatia, line 428: 1. controllerBuilder is passed as a pointer, so the value should not be overridden 
2. it doesn't make sense to pass the manager around. Rather you should get the channel from your listener you want to watch on
3. You listener should have its own logger instance
4. Instead of  `handler.EnqueueRequestForObject{} `use `handler.Funcs{GenericFunc: yourGenericFunc}`
Channel internally calls this, please check kubebuilder's implementation. 

`yourGenericFunc` will receive an eventHandler and a queue to add the kyma object too

adityabhatia (changes requested):

operator/controllers/kyma_controller.go
khlifi411: The Event Handler after the last refactor is a parameter which the caller of the `RegisterListenerComponent()` method should specify for future proofing. For the time being, it is a simple `EnqueueRequestForObject{}` for the `kyma` and `manifest` operators.

jakobmoellersap: Hmm, I really think this is bad practice to handle the manager around and then to the enqueue inside. I think you should instead work with the controllerBuilder (if at all)

operator/Makefile
```diff
@@ -70,7 +70,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
 .PHONY: docker-build
-docker-build: test ## Build docker image with the manager.
+docker-build: #test ## Build docker image with the manager.
```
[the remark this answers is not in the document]
jakobmoellersap: Please do not uncomment this or remove the skip from the test logic

operator/config/samples/secret/gardener-secret-gen.sh
```diff
@@ -0,0 +1,12 @@
+cat <<EOF > ./../component-integration-installed/secret.yaml
+apiVersion: v1
```
jakobmoellersap, line 2: this secret is already available in a different script

operator/controllers/kyma_controller.go
```diff
@@ -71,6 +76,7 @@ func (r *KymaReconciler) GetEventRecorder() record.EventRecorder {
 //+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
 //+kubebuilder:rbac:groups=operator.kyma-project.io,resources=moduletemplates,verbs=get;list;watch;create;update;patch;onEvent;delete
 //+kubebuilder:rbac:groups=operator.kyma-project.io,resources=moduletemplates/finalizers,verbs=update
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch;create;update;patch;onEvent;delete
```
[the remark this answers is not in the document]
jakobmoellersap: CRDs are currently necessary, but in the runtime for create/update/patch/delete, for control plane we only need get / list / watch

operator/controllers/kyma_controller.go
```diff
@@ -463,6 +467,13 @@ func (r *KymaReconciler) SetupWithManager(mgr ctrl.Manager, options controller.O
 				builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}))
 	}
 
+	//register listener component
+	runnableListener, eventChannel := listener.RegisterListenerComponent(listenerAddr, strings.ToLower(v1alpha1.KymaKind))
+	//watch event channel
+	controllerBuilder.Watches(eventChannel, &handler.EnqueueRequestForObject{})
```
jakobmoellersap, line 473: still using enqueuerequestforobject instead of the custom change handler

khlifi411: I think the handler should not be in the responsibility of the listener implementation. See README file in this [PR](https://github.com/khlifi411/kyma-listener/pull/2). I just provided this as a sample implementation in the PR. If you have other logic for the handler, I will remove this change, and you should provide your handler logic when invoking `Watches()`

adityabhatia (approved):

operator/controllers/kyma_controller.go
khlifi411: As discussed in our meeting, the custom handler logic will be included in the next iteration 👍

jakobmoellersap (approved):

PR #42, opened 2022-06-13
test: setup simple failure test of no module CR

jakobmoellersap:


adityabhatia (approved):

operator/controllers/kyma_controller.go
```diff
@@ -166,6 +166,13 @@ func (r *KymaReconciler) HandleDeletingState(ctx context.Context, logger *logr.L
 
 		if err = r.Get(ctx, client.ObjectKeyFromObject(actualComponentStruct), actualComponentStruct); err == nil {
 			// component CR still exists
+
+			//TODO remove skip for hardcoded components
+			if strings.Contains(actualComponentStruct.GetName(), "istio") ||
+				strings.Contains(actualComponentStruct.GetName(), "serverless") {
+				continue
+			}
```
adityabhatia, line 174: do we need these components, since they were removed from the Kyma resource?

jakobmoellersap: I removed them for testing but I think we should start removing anything in that regard. I guess we can leave it out.

adityabhatia (approved):
🔥

PR #52, opened 2022-06-15
Adapt component custom check

adityabhatia:


operator/config/samples/component-integration-installed/operator_v1alpha1_kyma.yaml
```diff
@@ -9,4 +9,10 @@ spec:
...
     - name: manifest
-      channel: stable
\ No newline at end of file
+      channel: stable
+      settings:
+        - apiVersion: operator.kyma-project.io/v1alpha1
+          kind: Kyma
+          name: kyma-sample
+          namespace: default
+          state: "Processing"
```
jakobmoellersap, line 18: desiredState maybe? state always sounds like active state

jakobmoellersap, line 18: desiredState maybe? state always sounds like active state

jakobmoellersap (dismissed):

operator/config/samples/component-integration-installed/operator_v1alpha1_kyma.yaml
adityabhatia: we can revisit this in future, as this will also involve adaptations on the manifest-op side.
For now I have renamed `settings` to `customStates`  and adapted this on the manifest-op side https://github.com/kyma-project/manifest-operator/pull/19

jakobmoellersap (approved):

PR #56, opened 2022-06-21
feat: setup moduletemplate cache

jakobmoellersap:
Cache Key is generated from Controller-Name (e.g. manifest) + Channel (stable), as there can be at most one template for this combination. If a module change is detected, the template is purged from the cache and will be dynamically recalculated the next time a kyma is reconciled for this channel. This will happen right after the invalidation since all Kymas connected to the channel with the module will be requeued.


jakobmoellersap:
We use the internal system cache lookup for kubebuilder

PR #58, opened 2022-06-27
Refactored condition handling

jakobmoellersap:
Requires https://github.com/kyma-project/kyma-operator/pull/57

this reduces the amount of status updates and reconciliation loops by interacting with the "Module" abstraction which combines all module settings, the current component as well as the channel into a struct that is used in the reconciler. Doing this, we can refactor a lot of state handling and only update when its really necessary, not after every change on the status. E.g. now when modules are already up to date or have state changes, we only update once for all conditions and immediately check for their state, saving 1 reconcile loop.


operator/config/samples/component-integration-installed/secret-gardener-k8s.yaml
```diff
@@ -0,0 +1,10 @@
...
...VUd0V05pdDNNR3RNVjFCalEwRjNSVUZCWVU1RFRVVkJkMFJuV1VSV1VqQlFRVkZJTDBKQlVVUkJaMGR0VFVFNFIwRXhWV1JGZDBWQ0NpOTNVVVpOUVUxQ1FXWTRkMGhSV1VSV1VqQlBRa0paUlVaR1dGWXJkblZUTWxOVlRXWjJZazlxVmpGblJITTJkMWRRZFRSTlFUQkhRMU54UjFOSllqTUtSRkZGUWtOM1ZVRkJORWxDUVZGQ2NYTTJZVzFHYUZaUFpVaFRSRVZNYzFOR1JEa3JkRGs1U1hOa05EbFJhVzlIY2t4aEx6Z3ZlR2wzZDFGcVFpczJjZ3BwT1VSWVVGVktNR2s1UlcxVk5XTkNaa2RUZW1aSFRqRXhhbEZPVVc1VWVrTlJVakV2VUc5bFRqTk5hMHd3YkVkSGRtbFhXV3AzY1ZkWVprMVFhbG9yQ21rMk9YWlJXVlp5UWxGWFRubGlVMjl6VVVoMWRYTnliWEpYV0ZSQ2JFaGtORWhyY0d4M1RUbFhiMHMzZWt0MVZFUm1Sa0pzVDFSRE9XcDVXVE4xTWpFS1pVaEthM2xoV20wNVFYa3JWbXRtVDA5eGNsSkJVelZuYVZkVlNGSmpVMnAwVnpCT2FtbERibkY0ZUU1QlQxZzBWbFZaT0ZkUldtbElZM0ZJVTBsRFdncFdiVzF5WnpSME5sZDFaRWRQWTNWcWRFODRiakptT1RsdkswdzVjVWwxYVZoRFZYTmxiazlYU1RSR2QzUk1Ra05XWmtKaWIycFVkR1ZSS3pZelMxUXZDbUp0ZGxRMVJHOVlLMEowU2twMmRWRlpkRzFHUVhwSGJFWkRUSGQxVEZWUVJGVm9aQW90TFMwdExVVk9SQ0JEUlZKVVNVWkpRMEZVUlMwdExTMHRDZz09CnVzZXJzOgogIC0gbmFtZTogc2hvb3QtLWplbGx5ZmlzaC0tam0tdGVzdC0xLXRva2VuCiAgICB1c2VyOgogICAgICB0b2tlbjogPi0KICAgICAgICBwR1hUVnpFU2MzR1F4SHNaSk1pTEVWT0FOWXYwU1ExdUlnOXZTVkdhVk0zOTBqbm51N1duYmFXcEgwOVZKbU5oM2RqeDZRdzU5cVRWNko0cVhhRXplS0E4OGZmc2hTN2JpeURoNHhCRGZHRnZ4S3NUcWZzWE81NzFwRFo4UTZOdQoK
```
jakobmoellersap, line 10: this is an already rotated config for sample purposes, its not actually used anywhere and we can keep it.

operator/controllers/kyma_controller.go
```diff
@@ -121,283 +122,282 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
-			return false, err
-		}
-
-		if err = r.Get(ctx, client.ObjectKeyFromObject(actualComponentStruct), actualComponentStruct); err == nil {
+	modules, err := util.ParseTemplates(kyma, templates)
+	if err != nil {
+		return false, err
+	}
+	for name, module := range modules {
+		if err = r.Get(ctx, client.ObjectKeyFromObject(module.Unstructured), module.Unstructured); err == nil {
```
ruanxin, line 214: If we use `r.Get(ctx, client.ObjectKeyFromObject(module.Unstructured), module.Unstructured);`
does it mean we still expect component CR exists in control-plane cluster? I thought we moved it to target cluster or?

jakobmoellersap: Component CR exists on both clusters, the target cluster is only synchronized

operator/controllers/kyma_controller.go
```diff
@@ -121,283 +122,282 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
 		"resource", client.ObjectKeyFromObject(kyma))
-
 	if err := remote.RemoveFinalizerFromRemoteKyma(ctx, r, client.ObjectKeyFromObject(kyma)); client.IgnoreNotFound(err) != nil {
 		return false, err
 	}
 	logger.Info("removed remote finalizer",
 		"resource", client.ObjectKeyFromObject(kyma))
-
-	// remove finalizer
 	controllerutil.RemoveFinalizer(kyma, labels.Finalizer)
```
ruanxin, line 233: Assume remote cluster is gone, and the client received not found error, and we ignored this error, which means we can remove the finalizer of kyma (local). 
Is it possible in this case, in the above `for name, module := range modules` loop, we enter into modules error state which prevent this remote unreachable kyma to be deleted.
What I means is should we introduce some logic if the remote kyma is not found, we force remove kyma and all related modules without error check.

jakobmoellersap: The logic means that if the object can not be found remotely, it will not try to remove the finalizer as it considers it already deleted and it will behave like normal. So your logic is actually covered (I think)

operator/pkg/util/kyma_util.go
```diff
@@ -26,20 +33,38 @@ func SetComponentCRLabels(unstructuredCompCR *unstructured.Unstructured, compone
 	unstructuredCompCR.SetLabels(labelMap)
 }
 
-func CopyComponentSettingsToUnstructuredFromResource(resource *unstructured.Unstructured, component operatorv1alpha1.ComponentType) {
-	if len(component.CustomStates) > 0 {
-		resource.Object["spec"].(map[string]interface{})["customStates"] = component.CustomStates
+func CopySettingsToUnstructuredFromResource(resource *unstructured.Unstructured, settings []operatorv1alpha1.Settings) {
+	if len(settings) > 0 {
+		resource.Object["spec"].(map[string]interface{})["customStates"] = settings
```
adityabhatia, line 38: we need to rethink this logic. 
settings should ideally be used for spec override
and customStates for state handling of non-native resource

operator/controllers/kyma_controller.go
```diff
@@ -121,283 +122,282 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
-	updateRequired, err := r.checkAndUpdateComponentConditions(ctx, kyma, templates)
+	statusUpdateRequiredFromCreation, err := r.CreateOrUpdateModules(ctx, kyma, modules)
 	if err != nil {
+		message := fmt.Sprintf("Component CR creation error: %s", err.Error())
+		logger.Info(message)
+		r.Event(kyma, "Warning", "ReconciliationFailed", fmt.Sprintf("Reconciliation failed: %s", message))
+		statusErr := status.Helper(r).UpdateStatus(ctx, kyma, operatorv1alpha1.KymaStateError, message)
+		if statusErr != nil {
+			return statusErr
+		}
```
adityabhatia, line 176: this way the external `err` is separated from the one below. But your wish :) 
```suggestion
		if err := status.Helper(r).UpdateStatus(ctx, kyma, operatorv1alpha1.KymaStateError, message); err != nil {
			return err
		}
```

adityabhatia (dismissed):
Great improvement🔥 
some optional things, not necessary

operator/pkg/util/kyma_util.go
jakobmoellersap: For sure, I think we need to extract this logic into a separate key and then align on 2 keys for state checks and settings.

adityabhatia (approved):

PR #65, opened 2022-06-30
Improve secret generate script

ruanxin:
- skip unit tests
- increase resource limits for kyma-operator

### DIFF
--- a/operator/config/manager/manager.yaml
+++ b/operator/config/manager/manager.yaml
@@ -51,10 +51,10 @@ spec:
         # More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
         resources:
           limits:
-            cpu: 500m
-            memory: 128Mi
+            cpu: 4000m
+            memory: 5Gi
           requests:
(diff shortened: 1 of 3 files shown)


operator/config/samples/secret/k3d-secret-gen.sh
```diff
@@ -1,4 +1,6 @@
-cat <<EOF > ./../component-integration-installed/secret.yaml
+#! /bin/bash
```
adityabhatia, line 1: Please rename to something like kubeconfig-secret-gen.sh

operator/config/manager/manager.yaml
```diff
@@ -51,10 +51,10 @@ spec:
...
           limits:
-            cpu: 500m
-            memory: 128Mi
+            cpu: 4000m
+            memory: 5Gi
           requests:
-            cpu: 10m
-            memory: 64Mi
+            cpu: 2000m
+            memory: 2Gi
```
adityabhatia, line 58: should this be a part of this PR?

PR #67, opened 2022-06-30
add golangci yaml

ruanxin:


ruanxin:
pjtester log, golint output generated: https://storage.googleapis.com/kyma-prow-logs/pr-logs/pull/kyma-project_kyma-operator/31/ruanxin_test_of_prowjob_pre-main-kyma-operator-lint/1542554071783706624/build-log.txt

jakobmoellersap (dismissed):

ruanxin:
The linter issue haven't resolved, will create other issue to cover it after team discussion.

"exhaustruct"
"gochecknoglobals"
"gochecknoinits"
"goerr113"
"ireturn"
"wrapcheck"

operator/pkg/remote/kyma_synchronization_context.go
```diff
@@ -166,48 +200,57 @@ func (c *KymaSynchronizationContext) SynchronizeRemoteKyma(ctx context.Context,
...
 		return true, nil
-	} else if c.controlPlaneKyma.Status.ObservedGeneration != c.controlPlaneKyma.GetGeneration() {
+	}
+
+	if c.controlPlaneKyma.Status.ObservedGeneration != c.controlPlaneKyma.GetGeneration() {
 		// control plane got updated, runtime on cluster is using the wrong base instance for customization
+		//nolint:godox
 		// TODO this now requires custom merge logic, but for now we reapply the control plane version
 		remoteKyma.Spec = c.controlPlaneKyma.Spec
 	} else if remoteKyma.Status.State != c.controlPlaneKyma.Status.State {
```
ruanxin, line 223: Hi @jakobmoellersap, can you double check if here is an `else if` or another `if` check? By reading the code, it seems should be another `if` check, because the condition is not same as before, but I'm not sure your initial thoughts.

operator/controllers/kyma_controller.go
```diff
@@ -317,32 +342,38 @@ func (r *KymaReconciler) SyncConditionsWithModuleStates(ctx context.Context, kym
...
-func (r *KymaReconciler) CreateOrUpdateModules(ctx context.Context, kyma *operatorv1alpha1.Kyma, modules util.Modules) (bool, error) {
+func (r *KymaReconciler) CreateOrUpdateModules(ctx context.Context, kyma *operatorv1alpha1.Kyma,
+	modules util.Modules,
+) (bool, error) {
 	logger := log.FromContext(ctx).WithName(client.ObjectKey{Name: kyma.Name, Namespace: kyma.Namespace}.String())
 	kymaSyncNecessary := false
+
 	for name, module := range modules {
 		// either the template in the condition is outdated (reflected by a generation change on the template)
-		//or the template that is supposed to be applied changed (e.g. because the kyma spec changed)
```
jakobmoellersap, line 325: please leave this comment in, I tend to forget it myself quite often :')

operator/main.go
```diff
@@ -62,27 +76,17 @@ func init() {
...
-		"determines the duration after which an already successfully reconciled Kyma is enqueued for checking "+
-			"if it's still in a consistent state.")
-	flag.DurationVar(&requeueFailureInterval, "requeue-failure-interval", 10*time.Second,
-		"determines the duration after which a failing reconciliation is retried and "+
-			"enqueued for a next try at recovering (e.g. because an Remote Synchronization Interaction failed)")
-	flag.DurationVar(&requeueWaitingInterval, "requeue-waiting-interval", 3*time.Second,
-		"etermines the duration after which a pending reconciliation is requeued "+
-			"if the operator decides that it needs to wait for a certain state to update before it can proceed "+
-			"(e.g. because of pending finalizers in the deletion process)")
+	defineFlag(metricsAddr, probeAddr, maxConcurrentReconciles,
```
jakobmoellersap, line 88: If you want to set it up like this you can just create the variables inside the function instead of passing pointers!

operator/pkg/release/lookup.go
```diff
@@ -20,26 +20,32 @@ type TemplateInChannel struct {
 
 type TemplatesInChannels map[string]*TemplateInChannel
 
-func GetTemplates(ctx context.Context, c client.Reader, k *operatorv1alpha1.Kyma) (TemplatesInChannels, error) {
+func GetTemplates(ctx context.Context, c client.Reader, kymaObj *operatorv1alpha1.Kyma) (TemplatesInChannels, error) {
```
jakobmoellersap, line 23:
```suggestion
func GetTemplates(ctx context.Context, c client.Reader, kyma *operatorv1alpha1.Kyma) (TemplatesInChannels, error) {
```

operator/pkg/release/lookup.go
```diff
@@ -119,33 +117,57 @@ func (c *channelTemplateLookup) WithContext(ctx context.Context) (*TemplateInCha
...
 	return &TemplateInChannel{
-		Template: &templateList.Items[0],
+		Template: &template,
 		Channel:  &actualChannel,
 		Outdated: false,
 	}, nil
 }
 
-func NewMoreThanOneTemplateCandidateErr(component operatorv1alpha1.ComponentType, candidateTemplates []operatorv1alpha1.ModuleTemplate) error {
+func (c *channelTemplateLookup) createDesiredChannel() operatorv1alpha1.Channel {
```
jakobmoellersap, line 148:
```suggestion
func (c *channelTemplateLookup) getDesiredChannel() operatorv1alpha1.Channel {
```

operator/pkg/remote/kyma_synchronization_context.go
jakobmoellersap: please do not make this to an if, otherwise the logic will try to first update the status and then the actual object. its meant to use 2 reconcile loops here. So thats why theres an elif control flow. I think we can rewrite it but then the code needs to be structured differently

operator/pkg/watch/template_change.go
```diff
@@ -53,38 +54,50 @@ func (h *TemplateChangeHandler) Watch(ctx context.Context) handler.MapFunc {
...
+		requests = handleRequest(ctx, kymas, channel, controller, templateNamespacedName, requests)
 
-			if !requeueKyma {
-				continue
-			}
+		return requests
+	}
+}
+
+func handleRequest(ctx context.Context, kymas *v1alpha1.KymaList, channel v1alpha1.Channel,
```
jakobmoellersap, line 63: What check caused this change? For me its not much easier to read this way tbh...

ruanxin: pkg/watch/template_change.go:25: Function 'Watch' is too long (68 > 60) ([funlen](https://github.com/ultraware/funlen)), basically complains long method

ruanxin: I actually just use IDE to extracted a method, tbh. Do you have any suggestions? maybe define several more concrete methods.

operator/main.go
ruanxin: But how to use them? return all of them back?

operator/controllers/kyma_controller.go
ruanxin: Yes, bring back already, I assume it's accidentally removed, initial idea is resolve the complain that there must be space between comment symbol and actual text

operator/pkg/watch/template_change.go
jakobmoellersap: what I usually do is use one function per loop, that makes things a bit easier, but the issue i have is with the amount of parameters it passes.

operator/main.go
jakobmoellersap: yes exactly, just define a struct with all the config values like on the old reconciler and then translate the config struct into the Reconciler struct by manually assigning values

ruanxin: oh, how stupid I am.. I'm sometimes a bit doubt myself the capability of my code skill now, shouldn't relay too much on IDE auto refactoring in the future..

operator/pkg/watch/template_change.go
ruanxin: extracted a piece of code as more readable method

jakobmoellersap (approved):

jakobmoellersap:
@Tomasz-Smelcerz-SAP could you (re-)sign the CLA here? Ill add the label back as well.

PR #71, opened 2022-07-06
Fix linter issue

ruanxin:


operator/pkg/remote/kyma_synchronization_context.go
```diff
@@ -187,9 +189,8 @@ func (c *KymaSynchronizationContext) CreateOrFetchRemoteKyma(ctx context.Context
 }
 
 func (c *KymaSynchronizationContext) SynchronizeRemoteKyma(ctx context.Context,
-	remoteKyma *operatorv1alpha1.Kyma,
+	remoteKyma *operatorv1alpha1.Kyma, recorder record.EventRecorder,
```
jakobmoellersap, line 192: why do you change the recorder away from the context?

ruanxin: The reason is about that eventing_adapter, check here https://github.com/kyma-project/kyma-operator/pull/71/files#r914643808

jakobmoellersap (approved):

PR #84, opened 2022-07-12
Fix deletion propagation

jakobmoellersap:
This removes the custom deletion logic and adds a proper controller reference instead of a simple owner reference. We use a proper controller handler now with the difference being that we only propagate changes on state change on the manifest on excessive reconciles. The interesting part is that now we also do not need custom deletion logic anymore since both background and foreground deletion automatically get propagated correctly into the controlled manfiests. In addition the watch works as expected. 

One note: During Debugging I noticed that if we change the spec of the manifest, the state field does not change so it is not reconciled by kyma. Maybe we need to adjust this behavior and reconcile on every change instead (resource version change)


operator/pkg/watch/restricted_enqueue_request_for_owner.go
```diff
@@ -0,0 +1,227 @@
+package watch
```
adityabhatia, line 1: Maybe I missed something, but I still so not see a good enough reason to implement the EventHandler interface. We can instead use the inbuild handler.Funcs{
UpdateFunc: ..
CreateFunc: ..
DeleteFunc: ..
GenericFunc: .. (not required)
}

operator/config/samples/component-integration-installed/moduletemplates/resources.yaml
```diff
@@ -12,4 +12,12 @@ relation: 'local'
...
\ No newline at end of file
+  path: "./data/nginx-ingress" # path is relative to the current resource file
+---
+name: 'busybox'
+type: 'helm-chart'
+relation: 'local'
+input:
+  type: 'dir'
+  compress: true
+  path: "./data/busybox" # path is relative to the current resource file
```
adityabhatia, line 23: should we also remove this

operator/config/samples/component-integration-installed/moduletemplates/data/.gitignore
```diff
@@ -1 +1,2 @@
-nginx-ingress
\ No newline at end of file
+nginx-ingress
+busybox
```
adityabhatia, line 2: and this

adityabhatia (approved):
🔥

PR #85, opened 2022-07-13
Migrated listener dependency

khlifi411:
**Changes proposed in this PR**

Migrated listener dependency to `kyma-watcher` repo


jeremyharisch (dismissed):

operator/main.go
```diff
@@ -98,10 +98,10 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
-	setupManager(flagVar, controllers.NewCacheFunc(), scheme)
+	setupManager(flagVar, controllers.NewCacheFunc(), scheme, &opts)
 }
 
-func setupManager(flagVar *FlagVar, newCacheFunc cache.NewCacheFunc, scheme *runtime.Scheme) {
+func setupManager(flagVar *FlagVar, newCacheFunc cache.NewCacheFunc, scheme *runtime.Scheme, opts *zap.Options) {
```
jakobmoellersap, line 104: instead of passing zap, I think we should pass logr.Logger since it is used everywhere else. also you can simply use ctrl.LoggerFrom in your code and it will use the singleton from the set logger. That way you dont have to pass it around in the code on every parameter :)

jakobmoellersap (changes requested):
+1 but the lint stuff will probably cause some conflicts depending on which PR goes in first

operator/main.go
khlifi411: Thanks for the tip! I will refactor it after this [PR](https://github.com/kyma-project/kyma-watcher/pull/12) gets merged. :)

operator/controllers/kyma_controller.go
```diff
@@ -375,52 +376,61 @@ func (r *KymaReconciler) CreateOrUpdateModules(ctx context.Context, kyma *v1alph
...
-				return false, err
-			}
-			if outdatedOverride {
-				kyma.RefreshOverride(name)
-			}
-			logger.Info("successfully updated module CR")
-			status.Helper(r).SyncReadyConditionForModules(kyma, parsed.Modules{name: module},
-				v1alpha1.ConditionStatusFalse, "updated condition for module CR")
-			return true, nil
+			return r.createAndUpdateNotFoundModule(ctx, kyma, module, logger, outdatedOverride, name)
```
jakobmoellersap, line 379: Lint issue fixed with proper refactoring in https://github.com/kyma-project/kyma-operator/pull/84

jakobmoellersap (changes requested):
I fixed all these lint issues so I think we should wait for https://github.com/kyma-project/kyma-operator/pull/84 to merge in. After that you probably wont need to do all of these refactorings

jakobmoellersap (approved):
🔥

PR #92, opened 2022-07-22
Create CRD State Validation in Validation Webhook

jakobmoellersap:
Throws Field Validation Errors if state is not properly defined in a cr that is used inside a moduletemplate

Currently using assumed created Name for Name of CRDs as there are no existing label selectors for CRDs and field selectors are not supported in CRDs for fetching Plurals or other GVK Info: https://github.com/kubernetes/kubernetes/issues/53459


adityabhatia (dismissed):
lgtm 🚀

docs/starting-operator-with-webhooks.md
```diff
@@ -0,0 +1,44 @@
+# Enabling Webhooks in Kyma Operator
+
+To make local testing easier, webhooks are disabled by default. To enable webhooks running with the operator,
+you will have to change some kustomization.yaml files as well as introduce a flag that will enable the webhook server:
+
+
+In `config/crd/kustomization.yaml`:
+
```
ruanxin, line 8: the related config under `operator/config/default/kustomization.yaml` also need to be uncommented, and cert manager needs to deployed. Maybe mention this tutorial is also an option: https://book.kubebuilder.io/cronjob-tutorial/running-webhook.html

ruanxin (approved):

PR #93, opened 2022-07-22
Remove Overrides from Kyma CR

jakobmoellersap:
Maybe even makes sense to remove settings, will double check this and submit a follow-up pr if necessary


adityabhatia (approved):
I like when features in doubt are removed 🔥

PR #94, opened 2022-07-25
Poc kyma operator ii   refactor module template management

ruanxin:


operator/pkg/parsed/template_to_module.go
```diff
@@ -74,12 +74,22 @@ func TemplatesToModules(
...
+	return moduleName + kymaName
+}
+
 func NewModule(
 	template *v1alpha1.ModuleTemplate,
 	verification signature.Verification,
 ) (*unstructured.Unstructured, error) {
-	component := &template.Spec.Data
-
+	component := template.Spec.Data.DeepCopy()
```
ruanxin, line 85: @jakobmoellersap , I read your PR, here just directly return for control plane module, but I think eventually to treat them (manifest/module CR) identically should no harm, we can discuss on this.

operator/config/samples/component-integration-installed/crd/controlplanemodule_types.go
```diff
@@ -0,0 +1,98 @@
+/*
```
ruanxin, line 1: To reviewer: This crd is here for testing

operator/controllers/suite_test.go
```diff
@@ -76,14 +77,22 @@ var _ = BeforeSuite(func() {
 
 	manifestCrd := &v1.CustomResourceDefinition{}
 	res, err := http.DefaultClient.Get(
-		"https://raw.githubusercontent.com/adityabhatia/manifest-operator/overrideApiChanges/api/config/crd/bases/component.kyma-project.io_manifests.yaml") //nolint:lll
+		"https://raw.githubusercontent.com/kyma-project/manifest-operator/main/operator/config/crd/bases/component.kyma-project.io_manifests.yaml") //nolint:lll
```
ruanxin, line 80: To Reviewer: the manifests crd is based on the latest main, some fields has been updated.

operator/api/v1alpha1/kyma_types.go
```diff
@@ -250,11 +259,19 @@ type KymaCondition struct {
...
+	ModuleInfo ModuleInfo `json:"moduleInfo,omitempty"`
+
 	// Timestamp for when Kyma last transitioned from one status to another.
 	// +optional
 	LastTransitionTime *metav1.Time `json:"lastTransitionTime,omitempty"`
 }
 
+type ModuleInfo struct {
+	// Generation tracks the active Generation of the Deployed Module CR.
+	Generation int64 `json:"generation,omitempty"`
```
jakobmoellersap, line 272: Why would you need to track the generation of the module CR? Since you only want to react to state changes those would not invoke a generation increment, but a resource version increment instead

operator/config/samples/component-integration-installed/operator_v1alpha1_moduletemplate_cp-module.yaml
```diff
@@ -0,0 +1,62 @@
...
+        - baseUrl: ghcr.io/fake/signed
+          componentNameMapping: urlPath
+          type: ociRegistry
+      resources:
+        - access:
+            digest: sha256:9d37fb366e4371b44c94496f3db276bd6f3a1019c9d44077b689c68ec6486c1e
+            type: localOciBlob
+          name: crds
+          relation: local
+          type: yaml
```
jakobmoellersap, line 41: crds are potentially many yamls, thus we should probably not use yaml since its not one file.

operator/controllers/kyma_controller.go
```diff
@@ -368,6 +368,10 @@ func (r *KymaReconciler) CreateOrUpdateModules(ctx context.Context, kyma *v1alph
 			return true, nil
 		}
 		update := func() (bool, error) {
+			if module.Template.Spec.Target == v1alpha1.TargetControlPlane {
```
jakobmoellersap, line 371: This logic does not need the filter I think since there should be no case in which the target is for the control-plane and the reconciliation is triggered.

operator/pkg/parsed/module.go
```diff
@@ -88,5 +88,13 @@ func (m *Module) UpdateStatusFromCluster(ctx context.Context, clnt client.Client
 
 	m.Unstructured.Object["status"] = unstructuredFromServer.Object["status"]
 	m.Unstructured.SetResourceVersion(unstructuredFromServer.GetResourceVersion())
+	m.Unstructured.SetGeneration(unstructuredFromServer.GetGeneration())
```
jakobmoellersap, line 91: also here, to set the generation typically is a pattern that does not make sense since the generation will only change for spec changes. If you ever needed to set that I think there is a state management issue hidden somewhere

operator/pkg/parsed/template_to_module.go
jakobmoellersap: Agreed. I think this is what we should aim for, NewModule should just return the unstructured no matter the type

operator/pkg/parsed/template_to_module.go
```diff
@@ -74,12 +74,22 @@ func TemplatesToModules(
...
+	if template.Spec.Target == v1alpha1.TargetRemote {
+		resource := template.Spec.Data.DeepCopy()
+		if err := mergeIntoSpecResource(resource, component); err != nil {
+			return nil, err
+		}
+		component.SetKind("Manifest")
```
jakobmoellersap, line 91: Why do you set kind but not group/version?

operator/pkg/watch/restricted_enqueue_request_for_owner.go
```diff
@@ -162,6 +162,10 @@ func (e *RestrictedEnqueueRequestForOwner) getOwnerReconcileRequestFromOwnerRefe
 			e.Log.Error(err, "error getting owner")
 		}
 
+		// In case component state can't be updated due to operator error, still possible enqueue by Generation update.
+		if componentOld.GetGeneration() != componentNew.GetGeneration() {
```
jakobmoellersap, line 166: A state update will not trigger a generation change, only a spec update. If you cannot update the state and the generation still changed, I think it is safer to just requeue instead of tracking the generation. We can discuss this I think

jakobmoellersap (commented):
Looks great already, just some questions regarding the Generation tracking of the Module CR mainly.

operator/config/samples/component-integration-installed/operator_v1alpha1_moduletemplate_cp-module.yaml
ruanxin: what type do you suggest we define here, a compressed file? I'm thinking it can be an aggregated yaml include all, which also easy for deploy.

operator/controllers/kyma_controller.go
ruanxin: I think by the current implementation, although reconciliation by watch the template change is prevented within `template_change.go`,  The template outdated check is still happens in `HandleConsistencyChanges` and `HandleProcessingState`, so the update will be triggered in the following `if module.TemplateOutdated`. If we don't do template outdated check for control plane CR, then this update will not happen. But I think even in this case, put a check here as a defensive statement is also fine.

operator/pkg/parsed/module.go
ruanxin: removed

operator/pkg/parsed/template_to_module.go
ruanxin: group version is already copied from template.Spec.Data by `resource` and  `apiVersion`, this two fields should be mandatory defined in module template, or am I wrong here?

operator/pkg/watch/restricted_enqueue_request_for_owner.go
ruanxin: removed

operator/api/v1alpha1/kyma_types.go
ruanxin: removed

operator/config/samples/component-integration-installed/operator_v1alpha1_moduletemplate_cp-module.yaml
jakobmoellersap: Well in OCI usually you use a compressed directory in a VFS, cli already has an implementation for that for the charts since they are multiple files so we could simply reuse that here. Of course this still allows you to define a folder with only one huge aggregated yaml in if we want to go with your approach later on

operator/controllers/kyma_controller.go
```diff
@@ -183,7 +180,7 @@ func (r *KymaReconciler) HandleProcessingState(ctx context.Context, kyma *v1alph
 	var err error
 	var modules parsed.Modules
 	// these are the actual modules
-	modules, err = r.GetModules(ctx, kyma, false)
```
jakobmoellersap, line 186: This false check means it will not actually try to check for outdated templates during processing which I think resolves the issue discussed above

operator/controllers/kyma_controller.go
```diff
@@ -256,12 +253,8 @@ func (r *KymaReconciler) HandleConsistencyChanges(ctx context.Context, kyma *v1a
...
```
jakobmoellersap: this should have said true, not sure why its on false here

jakobmoellersap (commented):
The last changes on the templateOutdated check do make sense but I think there was a bug introduced since the GetModules call in the ConsistencyCheck was called with false for the template check, that should not have happened actually

ruanxin:
| /test pre-main-kyma-operator-tests
-- | --

.gitignore
```diff
@@ -79,3 +79,7 @@ out.txt
 
 # ignore vscode delve debug binary
 */__debug_bin
+
+.run/
+/operator/config/crd/bases/component.kyma-project.io_kcpmodules.yaml
```
jakobmoellersap, line 84: why is this in gitignore?

operator/controllers/kyma_controller.go
```diff
@@ -418,16 +410,35 @@ func (r *KymaReconciler) CreateModule(ctx context.Context, name string, kyma *v1
...
 		return fmt.Errorf("error occurred while updating module from settings: %w", err)
 	}
 	// set labels
 	module.ApplyLabels(kyma, name)
-	// update the spec
-	module.Unstructured.Object["spec"] = module.Template.Spec.Data.Object["spec"]
-	if err := r.Update(ctx, module.Unstructured, &client.UpdateOptions{}); err != nil {
-		return fmt.Errorf("error updating custom resource of type %s %w", name, err)
+
+	if module.GetOwnerReferences() == nil {
```
jakobmoellersap, line 436: this will break if there is an owner reference that comes from a different controller, maybe we should introduce an error case here

operator/go.mod
```diff
@@ -5,72 +5,73 @@ go 1.18
 require (
 	github.com/gardener/component-spec/bindings-go v0.0.65
 	github.com/go-logr/logr v1.2.3
-	github.com/imdario/mergo v0.3.12
+	github.com/imdario/mergo v0.3.13
 	github.com/kyma-project/kyma-watcher/listener v0.0.0-20220714080306-99a0a5162c8b
+	github.com/kyma-project/manifest-operator/operator v0.0.0-20220726131454-83e937e2eb36
```
jakobmoellersap, line 10: Why is this dependency in? I think you always use string-based GVKs?

operator/pkg/parsed/module.go
```diff
@@ -88,5 +88,6 @@ func (m *Module) UpdateStatusFromCluster(ctx context.Context, clnt client.Client
 
 	m.Unstructured.Object["status"] = unstructuredFromServer.Object["status"]
 	m.Unstructured.SetResourceVersion(unstructuredFromServer.GetResourceVersion())
+	m.Unstructured.SetOwnerReferences(unstructuredFromServer.GetOwnerReferences())
```
jakobmoellersap, line 91: With this the Method Name probably isnt the best anymore, maybe something like UpdateModuleFromCluster?

operator/pkg/test/util.go
```diff
@@ -0,0 +1,60 @@
...
+	if data.GetKind() != "" {
+		moduleTemplate.Spec.Data = data
+	}
+	return &moduleTemplate, nil
+}
+
+func readModuleTemplate(module v1alpha1.Module, moduleTemplate *v1alpha1.ModuleTemplate) error {
+	var template string
+	switch module.ControllerName {
+	case "manifest":
```
jakobmoellersap, line 43: Wouldnt this be an option without introducing our flag actually? So the remote flag would be set by controller label instead?

jakobmoellersap (commented):
Looks great, some questions but overall nice implementation

.gitignore
ruanxin: it's because that `make manifests` will scan all folders, and generate crds under this `crd/bases/` folder, but actually this kcpmodule is just for testing purpose (generated by `make test-crd`). I haven't find a good way to prevent this happen, so just add it to gitignore..

operator/pkg/test/util.go
ruanxin: You mean just use one template for all case?

jakobmoellersap: Currently in our code we use a flag in the module template to determine wether to install it as a central module or not. Instead we can just base it on the controller. If the controller is manifest, then we know to wrap it in a manifest CR

ruanxin: ok, understood, but I think conceptually it's better introduce that target flag, although by current design, the manifest operator is the de facto which can do the distinguish. But to bind this two aspect together maybe a bit rash, it's like a temporary workaround to me, for example, if later we introduce another operator to manage remote modules, this assumption will not fit anymore.

operator/controllers/kyma_controller.go
ruanxin: add a new function `ContainsExpectedOwnerReference` to check if contains kyma as owner. btw, we allow other ownerreference exists, right?

operator/go.mod
ruanxin: wait until we have a decision if introduce manifest operator as dependency.

operator/pkg/parsed/module.go
ruanxin: renamed

jakobmoellersap (approved):

PR #98, opened 2022-07-29
Virtual Sync of KCP and SKR Modules

jakobmoellersap:
This PR prepares a dynamic Synchronization from SKR and KCP Modules based on a super set of modules. There is also a new flag to determine wether to create the SKR Kyma with default modules or not


jakobmoellersap:
/hold

still need to extend test cases

ruanxin (approved):

PR #101, opened 2022-08-02
module deletion

ruanxin:
- add logic to delete not exist modules
- add moduleInfo to kyma conditions to track current module information


operator/controllers/kyma_controller.go
```diff
@@ -212,7 +231,7 @@ func (r *KymaReconciler) HandleProcessingState(ctx context.Context, kyma *v1alph
 
 	// if the ready condition is not applicable, but we changed the conditions, we still need to issue an update
 	if statusUpdateRequiredFromCreation || statusUpdateRequiredFromSync {
```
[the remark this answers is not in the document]
jakobmoellersap: The reason why this is done is to prevent too many reconciliation loops, otherwise we would have to reconcile n times for 1 kyma with n modules for the status updates. Ideally I wanted to avoid that, and the easiest way to solve the issue at the time was to just collect all updates and return at the end. That way I reduce the amount of reconciliations heavily. Ideally I would like to stick with that

operator/controllers/kyma_controller.go
```diff
@@ -201,6 +205,21 @@ func (r *KymaReconciler) HandleProcessingState(ctx context.Context, kyma *v1alph
...
 
+	notExistsModuleInfos := kyma.GetNotExistsModuleInfos()
+	if len(notExistsModuleInfos) > 0 {
+		requireSync, err := r.DeleteNotExistsModulesAndUpdateConditions(ctx, &kyma.Status, notExistsModuleInfos)
+		if err != nil {
+			return r.UpdateStatusFromErr(ctx, kyma, v1alpha1.KymaStateError,
+				fmt.Errorf("error while syncing conditions during deleting non exists modules: %w", err))
+		}
+		if requireSync {
+			if err := status.Helper(r).UpdateStatusForExistingModules(
```
jakobmoellersap, line 216: after we update a status we should requeue again to not run into inconsistency errors

operator/controllers/kyma_controller.go
```diff
@@ -503,3 +525,56 @@ func (r *KymaReconciler) GetModules(ctx context.Context, kyma *v1alpha1.Kyma) (p
...
+func (r *KymaReconciler) deleteModule(ctx context.Context, moduleInfo *v1alpha1.ModuleInfo) error {
+	module := unstructured.Unstructured{}
+	module.SetNamespace(moduleInfo.Namespace)
+	module.SetName(moduleInfo.Name)
+	module.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   moduleInfo.GroupVersionKind.Group,
+		Version: moduleInfo.GroupVersionKind.Version,
+		Kind:    moduleInfo.GroupVersionKind.Kind,
+	})
+	return r.Delete(ctx, &module, &client.DeleteOptions{})
```
jakobmoellersap, line 579: what if the module is already deleted, will this call still succeed? Also your delete is async due to you not setting the deletion propagation to foreground, is this on purpose? If it is, I think you are missing the check that the module really has been deleted before removing it from the condition.

jakobmoellersap (changes requested):
I have a few questions regarding ModuleInfo since I think we should refactor this. Other than that the deletion is for me a bit of a question mark since its done without blocking the code and thus async without doing a check wether the module really was deleted.

operator/controllers/kyma_controller.go
ruanxin: refactored

operator/controllers/kyma_controller.go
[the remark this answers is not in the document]
ruanxin: refactored to do update in the end with additional `statusUpdateRequiredFromDeletion`

operator/controllers/kyma_controller.go
```diff
@@ -206,6 +210,16 @@ func (r *KymaReconciler) HandleProcessingState(ctx context.Context, kyma *v1alph
 			fmt.Errorf("error while syncing conditions during processing: %w", err))
 	}
 
+	status.Helper(r).SyncModuleInfo(kyma, modules)
+
+	statusUpdateRequiredFromDeletion := r.UpdateAndRemoveNotExistsModuleInfo(ctx, kyma)
```
jakobmoellersap, line 215:
```suggestion
	statusUpdateRequiredFromDeletion := r.UpdateStatusModuleInfos(ctx, kyma)
```

operator/controllers/kyma_controller.go
```diff
@@ -206,6 +210,16 @@ func (r *KymaReconciler) HandleProcessingState(ctx context.Context, kyma *v1alph
 			fmt.Errorf("error while syncing conditions during processing: %w", err))
 	}
 
+	status.Helper(r).SyncModuleInfo(kyma, modules)
```
jakobmoellersap, line 213: Why do you need a dedicated update on remove if you already have a sync here? IMO the next few lines are hard to understand in terms of naming

operator/controllers/kyma_controller.go
```diff
@@ -504,3 +522,71 @@ func (r *KymaReconciler) GetModules(ctx context.Context, kyma *v1alpha1.Kyma) (p
...
+func (r *KymaReconciler) UpdateAndRemoveNotExistsModuleInfo(ctx context.Context, kyma *v1alpha1.Kyma) bool {
+	requireUpdateCondition := false
+	moduleInfoMap := kyma.GetModuleInfoMap()
+	moduleInfos := kyma.GetNotExistsModuleInfos()
+	if len(moduleInfos) == 0 {
+		return false
+	}
+	for i := range moduleInfos {
+		moduleInfo := moduleInfos[i]
+		_, err := r.getModule(ctx, moduleInfo)
```
jakobmoellersap, line 535: I think this is not well optimized since you do a get request for all modules. Instead you could group them by GVK and then do a list request with a label selector filter, and then map it back to the existing module infos. For now should be okay though

operator/controllers/kyma_controller.go
```diff
@@ -504,3 +522,71 @@ func (r *KymaReconciler) GetModules(ctx context.Context, kyma *v1alpha1.Kyma) (p
...
+	module.SetName(moduleInfo.Name)
+	module.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   moduleInfo.GroupVersionKind.Group,
+		Version: moduleInfo.GroupVersionKind.Version,
+		Kind:    moduleInfo.GroupVersionKind.Kind,
+	})
+	return r.Delete(ctx, &module, &client.DeleteOptions{})
+}
+
+func (r *KymaReconciler) getModule(ctx context.Context,
```
jakobmoellersap, line 581: This logic is used heavily also for the create/update check, maybe refactor to use this method there?

operator/controllers/kyma_controller_test.go
```diff
@@ -84,35 +84,44 @@ var _ = Describe("Kyma with empty ModuleTemplate", func() {
...
-			moduleTemplates = append(moduleTemplates, template)
+			moduleTemplates[module.Name] = template
 		}
 	})
 
 	It("CR should be recreated after delete", func() {
 		By("CR created")
 		for _, activeModule := range moduleTemplates {
-			Eventually(ModuleExist(kyma, activeModule), timeout, interval).Should(Succeed())
+			Eventually(ModuleExist(kyma, activeModule), timeout, interval).Should(BeTrue())
```
jakobmoellersap, line 124:
```suggestion
			Eventually(ModuleExists(kyma, activeModule), timeout, interval).Should(BeTrue())
```

jakobmoellersap (changes requested):
Hey there, I think were close to the merge here :) just smaller naming issues other than that Logic LGTM 🚀

operator/controllers/kyma_controller.go
```diff
@@ -115,7 +119,7 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	}
 
 	// create a remote synchronization context, and update the remote kyma with the state of the control plane
-	if kyma.Spec.Sync.Enabled {
+	if kyma.Status.State != v1alpha1.KymaStateDeleting && kyma.Spec.Sync.Enabled {
```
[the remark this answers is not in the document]
ruanxin: make sense, I think this `Status.State != v1alpha1.KymaStateDeleting` better put into `replaceWithVirtualKyma` as a precondition to prevent update happens when kyma in deleting state.

operator/controllers/kyma_controller.go
ruanxin: If you check the SyncModuleInfo definition, it's just sync all module defined in `spec.modules`, and this sync doesn't require to call apiserver, that's why it happens both in `HandleConsistencyChanges` `HandleProcessingState`. 

And next few lines 
`UpdateStatusModuleInfos`: check if module resource still exists in cluster, if not exists, return true so that triggering `UpdateStatusForExistingModules` to update `status.moduleInfos`
`DeleteNoLongerExistingModules`: delete module which not decared in `spec.modules` but in `status.moduleInfos`
The above functions both require to call apiserver, only happens in `HandleProcessingState`

operator/controllers/kyma_controller.go
ruanxin: completely agree, I think we need a list request and do updating conditions, templateInfos( when it move out of conditions) all at once, currently by sync conditions, it's also get request one by one, here will be a performance issue.

operator/controllers/kyma_controller.go
jakobmoellersap: With Sync we actually mean copy over then right? Maybe we change that name then to make it clearer?

For UpdateStatusModuleInfos I would maybe name it to UpdateModuleInfoStatus
for DeleteNoLongerExistingMoudles I think its okay :)

operator/controllers/kyma_controller.go
ruanxin: indeed, replaced in two places, but also related to performance concern, we should not use it for get all modules (and unfortunately current are all for this purpose)

operator/pkg/parsed/module.go
```diff
@@ -59,25 +55,12 @@ func (m *Module) StateMismatchedWithCondition(condition *v1alpha1.KymaCondition)
 		condition.TemplateInfo.Channel != m.Template.Spec.Channel
 }
 
-// UpdateModuleFromCluster update the module with necessary information (status, ownerReference) based on
-// an interaction with a client that is connected to a cluster. It will wrap any error returned from the client,
-// so checking for a k8s error can be achieved with Unwrap.
-func (m *Module) UpdateModuleFromCluster(ctx context.Context, clnt client.Client) error {
```
ruanxin, line 65: Just want to mention, it's actually a hidden cyclic dependency.

jakobmoellersap (approved):

operator/pkg/parsed/module.go
ruanxin: what I mean is previously by using this `UpdateModuleFromCluster`, it introduced `module` to `KymaReconciler`, and with this `client.Client` param, it potentially introduce `KymaReconciler` into `module`, which become cyclic depended, it' will not complain by the compiles although, since KymaReconciler used as Client interface, otherwise I believe it's cyclic import.

PR #103, opened 2022-08-04
Setup Sync of KCP ModuleTemplates into SKR ConfigMap

jakobmoellersap:


operator/controllers/kyma_controller_test.go
```diff
@@ -239,5 +240,8 @@ var _ = Describe("Kyma sync into Remote Cluster", func() {
 		for _, activeModule := range moduleTemplates {
 			Eventually(ModuleExists(kyma, activeModule), timeout, interval).Should(BeTrue())
 		}
+
+		By("Remote Module Catalog created")
+		Eventually(RemoteCatalogExists(runtimeClient, kyma), "30s", interval).Should(Succeed())
```
ruanxin, line 245: let's keep time format consistent

operator/controllers/module_template_controller.go
```diff
@@ -0,0 +1,78 @@
...
+	logger := log.FromContext(ctx)
+	logger.Info("Catalog Sync loop starting for", "resource", req.NamespacedName.String())
+
+	// check if kyma resource exists
+	kyma := &v1alpha1.Kyma{}
+	if err := r.Get(ctx, req.NamespacedName, kyma); err != nil {
+		// we'll ignore not-found errors, since they can't be fixed by an immediate
+		// requeue (we'll need to wait for a new notification), and we can get them
+		// on deleted requests.
+		logger.Info(req.NamespacedName.String() + " got deleted!")
```
ruanxin, line 39: This log message is a bit misleading, or? I think only if it's not found error, we can safely say it's got deleted.

operator/pkg/remote/kyma_synchronization_context.go
```diff
@@ -285,3 +289,68 @@ func GetRemoteObjectKey(kyma *operatorv1alpha1.Kyma) client.ObjectKey {
...
+		return err
+	}
+	create = k8serrors.IsNotFound(err)
+
+	if catalog.Data == nil {
+		catalog.Data = make(map[string]string)
+	}
+
+	for _, moduleTemplate := range moduleTemplates.Items {
+		moduleTemplate := &moduleTemplate
```
ruanxin: with this way, it still copy a moduleTemplate, which is not necessary.

operator/pkg/remote/kyma_synchronization_context.go
```diff
@@ -262,26 +265,91 @@ func (c *KymaSynchronizationContext) ReplaceWithVirtualKyma(kyma *operatorv1alph
...
+			return err
+		}
+
+		catalog.Data[moduleName] = string(yml)
+	}
+
+	if create {
+		return c.runtimeClient.Create(ctx, catalog)
+	}
+	return c.runtimeClient.Update(ctx, catalog)
```
ruanxin, line 345: Is it better we introduce moduleTemplate generation into configmap? Then after first `.Get`, we can compare the generation in catalog and current moduleTemplate, if no change, we don't need to do `.Update` here.

operator/controllers/module_template_controller.go
jakobmoellersap: You are right, I missed this. Will add an exception case to the err here as I forgot

operator/pkg/remote/kyma_synchronization_context.go
jakobmoellersap: https://github.com/kyma-project/kyma-operator/pull/103/commits/419dd905606805d9e26a67bfd52f7891d0d6678e#diff-bc1bb91f1f01617c2d3693314155dcfad6e5a612788e40ec921f8236d0a5a11cR323

I think you reviewed based on old state

operator/pkg/remote/kyma_synchronization_context.go
jakobmoellersap: ConfigMap has no incrementing generation I think, no?

ruanxin: no, I mean save moduleTemplate generation into ConfigMap as additional data, same as what you did for `Target`, `Channel`, then unmarshal it for comparing

operator/pkg/remote/kyma_synchronization_context.go
```diff
@@ -342,7 +354,22 @@ func (c *KymaSynchronizationContext) CreateOrUpdateModuleTemplateCatalog(
...
+		return c.runtimeClient.Update(ctx, catalog)
+	}
+	return nil
+}
+
+func DoesModuleTemplateNeedUpdateInCatalog(catalog *corev1.ConfigMap, template *v1alpha1.ModuleTemplate) bool {
+	moduleName := template.GetLabels()[v1alpha1.ModuleName]
+	if catalog.Data[moduleName] != "" {
+		oldCatalogEntry := &CatalogEntry{}
+		if err := yaml.Unmarshal([]byte(catalog.Data[moduleName]), oldCatalogEntry); err != nil ||
```
ruanxin, line 367:
```suggestion
		if err := yaml.Unmarshal([]byte(catalog.Data[moduleName]), oldCatalogEntry); err != nil &&
```

jakobmoellersap: I did the Logic OR on purpose as I do not want the Update to be ignored in case one entry is failing. Do you disagree with this?

ruanxin: oh, sorry, my mistake, I'm previously thinking this `err != nil` will lead to `oldCatalogEntry.Generation` null pointer, but actually in this case, exactly should use OR.

ruanxin (approved):

PR #105, opened 2022-08-08
Refact module template info

ruanxin:


operator/pkg/release/lookup.go
```diff
@@ -40,18 +40,19 @@ func GetTemplates(ctx context.Context, c client.Reader, kyma *operatorv1alpha1.K
...
-					lookupResult.ModuleTemplate.Spec.Channel != condition.TemplateInfo.Channel {
+	for moduleName, lookupResult := range templates {
+		for i := range k.Status.ModuleInfos {
+			moduleInfo := &k.Status.ModuleInfos[i]
+			if moduleInfo.ModuleName == moduleName && lookupResult != nil {
+				if lookupResult.ModuleTemplate.GetGeneration() != moduleInfo.TemplateInfo.Generation ||
+					lookupResult.ModuleTemplate.Spec.Channel != moduleInfo.TemplateInfo.Channel {
 					logger.Info("detected outdated template",
-						"condition", condition.Reason,
+						"templateInfo", moduleInfo.ModuleName,
```
jakobmoellersap, line 50:
```suggestion
						"module", moduleInfo.ModuleName,
```

operator/.golangci.yaml
```diff
@@ -40,12 +40,14 @@ linters-settings:
...
       - Client
+      - record.EventRecorder
 issues:
   max-issues-per-linter: 0
   max-same-issues: 0
   fix: true
 output:
   sort-results: true
 run:
+  timeout: 5m
```
jakobmoellersap, line 51: Can we make the timeout run on something lower like 2 mins?

operator/api/v1alpha1/kyma_types.go
```diff
@@ -224,11 +218,11 @@ type ModuleInfo struct {
 	// ModuleName is the unique identifier of the module.
 	ModuleName string `json:"moduleName"`
 
-	// Channel is the current deployed module channel
-	Channel Channel `json:"channel"`
-
-	// GroupVersionKind is the current deployed module gvk
-	GroupVersionKind metav1.GroupVersionKind `json:"gvk"`
+	// Additional Information when the condition is bound to a ModuleTemplate. It contains information about the last
+	// parsing that occurred and will track the state of the parser ModuleTemplate in Context of the Installation.
```
jakobmoellersap, line 222:
```suggestion
	// parsing that occurred and will track the state of the parsed ModuleTemplate in the context of installation.
```

operator/api/v1alpha1/kyma_types.go
```diff
@@ -224,11 +218,11 @@ type ModuleInfo struct {
...
+	// This will update when Channel, Profile or the ModuleTemplate used in the Condition is changed.
```
jakobmoellersap, line 223:
```suggestion
	// This will update when Channel, Profile or the ModuleTemplate used in the Condition is changed.
```
What is this? We should probably change this sentence

operator/pkg/status/kyma_handler.go
```diff
@@ -105,6 +95,20 @@ func (k *Kyma) SyncModuleInfo(kyma *operatorv1alpha1.Kyma, modules parsed.Module
...
+	kyma *operatorv1alpha1.Kyma,
+	moduleName string,
+) (*operatorv1alpha1.TemplateInfo, error) {
+	for i := range kyma.Status.ModuleInfos {
+		moduleInfo := &kyma.Status.ModuleInfos[i]
+		if moduleInfo.ModuleName == moduleName {
+			return &moduleInfo.TemplateInfo, nil
+		}
+	}
+	// should not happen
```
jakobmoellersap, line 108: why should this not happen? If there is a mutation done from outside the operator this can happen, right?

jakobmoellersap (changes requested):

operator/.golangci.yaml
ruanxin: just make sure we are on same picture, this timeout configured as a threshold value for total duration for this lint check, it's not means if some failure happens, it wait 5m before finish. so it depends on the power of the vm allocated, and the default value is `1m`, by current code base we already reach this threshold, I think here should configure a reasonable big enough time.

operator/api/v1alpha1/kyma_types.go
ruanxin: will update in next commit

operator/pkg/status/kyma_handler.go
ruanxin: theoretically it will happen, but nearly impossible, this method happens after `if module.TemplateOutdated` check, and this  TemplateOutdated relays on moduleInfo contains TemplateInfo, if the mutation you mean is manually removed those moduleInfo entry, it will not detect template is outdated.

operator/.golangci.yaml
jakobmoellersap: I think having over 1 minute for a static code check is already unacceptable in this case. Is this normal as a duration for the codebase? Ideally I dont want to have any job running longer than 5 mins max and a lint shouldnt take that much. If its just so bad performance-wise that we have to increase it im fine

ruanxin: Yes, I got your point, we can consider increase the machine power of this lint job running, currently the duration time is around 1m32s, and you can see it in the [job history ](https://status.build.kyma-project.io/job-history/gs/kyma-prow-logs/pr-logs/directory/pre-main-kyma-operator-lint), when the code base is stable enough, I think we can find a reasonable required config for this job machine.

ruanxin: good news! find `concurrency` param, configured to 8 reduced the time

jakobmoellersap (approved):

PR #109, opened 2022-08-09
Gherkin tests

tobiscr:
First draft of Gherkin test cases.

@ameteiko : JFYI :)


operator/test/cucumber_test.go
```diff
@@ -0,0 +1,129 @@
+package test_test
```
jakobmoellersap, line 1: this package name looks interesting :D 

anyhow, would it make sense to introduce a separate module for these tests? I think since they are e2e, they should be kept apart from our unit tests

tobiscr: Agree, makes sense to move them into a separate module. Will do that...

jakobmoellersap:
As discussed, will move into fast-integration-tests

PR #111, opened 2022-08-10
feat (template-operator): init implementation for sample resource

adityabhatia:


samples/template-operator/operator/controllers/sample_controller.go
```diff
@@ -85,20 +109,118 @@ func (r *SampleReconciler) HandleInitialState(ctx context.Context, sampleResourc
...
+	}
+
+	// Use manifest library client to install a sample chart
+	ready, err := manifestClient.Install(manifestLib.InstallInfo{
+		Ctx: ctx,
+		ChartInfo: &manifestLib.ChartInfo{
+			ChartPath:   sampleResource.Spec.ChartPath,
+			ReleaseName: sampleResource.Spec.ReleaseName,
+		},
+		RemoteInfo: custom.RemoteInfo{
```
jakobmoellersap, line 133: I think we should prepare the sample to run as a runtime / skr operator not a control plane operator, in this case, we wouldnt need this right?

samples/template-operator/operator/controllers/sample_controller.go
```diff
@@ -85,20 +109,118 @@ func (r *SampleReconciler) HandleInitialState(ctx context.Context, sampleResourc
...
+		},
+		ResourceInfo: manifestLib.ResourceInfo{
+			// base operator resource to be passed for custom checks
+			BaseResource: &unstructured.Unstructured{Object: sampleObjUnstructured},
+		},
+		CheckFn: func(context.Context, *unstructured.Unstructured, *logr.Logger, custom.RemoteInfo) (bool, error) {
+			// your custom logic here to set ready state
+			return true, nil
+		},
+		CheckReadyStates: false,
```
jakobmoellersap, line 147: I think we should default this to true since most people want to have a ready check, probably we can use a similar default readyness as the declarative pattern with services/deployments etc

samples/template-operator/operator/controllers/sample_controller.go
```diff
@@ -85,20 +109,118 @@ func (r *SampleReconciler) HandleInitialState(ctx context.Context, sampleResourc
...
+	}
+
+	manifestClient, err := r.getManifestClient(logger, sampleResource.Spec.ChartFlags)
+	if err != nil {
+		sampleResource.Status.State = v1alpha1.SampleStateError
+		return r.Client.Status().Update(ctx, sampleResource)
+	}
+
+	// Use manifest library client to install a sample chart
+	readyToBeDeleted, err := manifestClient.Uninstall(manifestLib.InstallInfo{
```
jakobmoellersap, line 192: can we refactor the installInfo to be the same for processing and deleting state? a lot of boilerplate otherwise

samples/template-operator/operator/controllers/sample_controller.go
```diff
@@ -61,13 +72,26 @@ func (r *SampleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+	if controllerutil.AddFinalizer(sampleResource, deletionFinalizer) {
+		return ctrl.Result{}, r.Client.Update(ctx, sampleResource)
+	}
+
 	switch sampleResource.Status.State {
 	case "":
 		return ctrl.Result{}, r.HandleInitialState(ctx, sampleResource)
 	case v1alpha1.SampleStateProcessing:
-		return ctrl.Result{}, r.HandleProcessingState(ctx, sampleResource)
+		return ctrl.Result{}, r.HandleProcessingState(ctx, sampleResource, &logger)
```
jakobmoellersap, line 92: the logger you do not need to pass, you can just call `logger := log.FromContext(ctx)`

jakobmoellersap (commented):
Looks great, just some small refactorings! Thanks for updating the Readmes as well!

samples/template-operator/operator/controllers/sample_controller.go
adityabhatia: For now this is required to know the destination client, the name is a bis misleading I agree. I will refactor this in the manifest-library improve interface part.

samples/template-operator/operator/controllers/sample_controller.go
adityabhatia: this verifies ready state for all 'native' installed resources, supported by helm. I will enable it.

jakobmoellersap (approved):

PR #112, opened 2022-08-12
setup local module list sync

jakobmoellersap:
This PR refactors the Catalog Synchronization to also work in local mode when Synchronization is disabled. It also enables switching the module list sync off with a flag:
```
flag.BoolVar(&flagVar.enableModuleCatalog, "enable-module-catalog", true,
		"Enabling the Module Catalog Synchronization for Introspection of available Modules based on ModuleTemplates.")
``` 
Setting this to disabled will completely shut off the second control loop
It also removes the profile as we want to define this in the Module CRs individually


operator/pkg/catalog/sync.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func NewSync(client client.Client, settings Settings) *Sync {
+	return &Sync{Catalog: New(client, settings)}
+}
+
+func (s *Sync) Cleanup(
+	ctx context.Context,
+) error {
+	return s.Catalog.Delete(ctx)
```
ruanxin, line 24: I think here not found error should be ingored

operator/controllers/module_catalog_controller.go
```diff
@@ -39,10 +42,9 @@ func (r *ModuleTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		// on deleted requests.
 		if k8serrors.IsNotFound(err) {
 			logger.Info(req.NamespacedName.String() + " got deleted!")
-			return ctrl.Result{}, nil
+			return ctrl.Result{}, catalogSync.Cleanup(ctx)
```
ruanxin, line 45: Do we have race condition when both controller handle kyma resource? I'm not sure how exactly it behaviors, but I assume when the kyma out of worker queue, both controller received it. Then imaging a kyma in deleting state enter into reconcile, first in the module_catalog_controller, the `Get` request still can fetch this kyma, then in parallel, in kyma_controller, the finalizer get removed, then this kyma get deleted by apiserver, in this case, it will not enter into worker queue again, right? Then how this catalog cleanup can be triggered?

jakobmoellersap: In this case at most it will be deferred by the cache interval of the informer cache, which by default is set to 60 minutes in kubebuilder projects. In this case the delete is triggered and then after an hour the cache will be reset leading to another call that will trigger the not found case on the module catalog. In this case most likely the cluster is gonna be deprovisioned before we can cleanup the catalog though.

ruanxin: Let's merge it first, I'll try to do a simulate for this case, it's kind of interesting to me.

ruanxin (approved):

PR #117, opened 2022-08-17
Refactor status.condition

ruanxin:
- refactor condition to only present overall module state


operator/api/v1alpha1/condition_builder.go
```diff
@@ -0,0 +1,46 @@
...
+	return &metav1.Condition{
+		Type:               string(ConditionTypeReady),
+		Status:             cb.Status,
+		LastTransitionTime: metav1.Time{Time: time.Now()},
+		Reason:             string(cb.Reason),
+		Message:            cb.generateMessage(),
+	}
+}
+
+func (cb *ConditionBuilder) generateMessage() string {
```
jakobmoellersap, line 36: why is the condition builder code in the api package? I would have assumed this should go into a dedicated `machinery` module or into the kyma controller `pkg` folder. We also have some other candidates for this I think

operator/api/v1alpha1/kyma_types.go
```diff
@@ -342,57 +322,50 @@ func init() {
...
+	for _, condition := range kyma.Status.Conditions {
+		if condition.Reason == string(ConditionReasonModulesIsReady) {
+			found = true
 		}
 	}
+	if !found {
+		builder := NewConditionBuilder()
+		builder.SetStatus(metav1.ConditionFalse)
+		builder.SetReason(ConditionReasonModulesIsReady)
+		newCondition := builder.Build()
```
jakobmoellersap, line 336: why use a builder pattern if you anyhow have to use them all at once. Anyhow I think this is more functional Builder style so the right way would be
```suggestion
		newCondition := NewConditionBuilder().WithStatus(metav1.ConditionFalse).WithReason(ConditionReasonModulesIsReady)
```
formatted over multiple lines.

A much more common way in go to write builders is with varags like so:

```
BuildCondition(
  WithStatus(metav1.ConditionFalse),
  WithReason(ConditionReasonModulesIsReady)
)
```
pick your poison :)

operator/config/crd/bases/operator.kyma-project.io_kymas.yaml
```diff
@@ -123,31 +123,68 @@ spec:
                 description: List of status conditions to indicate the status of a
                   ServiceInstance.
                 items:
-                  description: KymaCondition describes condition information for Kyma.
+                  description: "Condition contains details for one aspect of the current
```
jakobmoellersap, line 126: this looks broken, is this correct?

jakobmoellersap (changes requested):
Great Refactoring! Just a few nits, but overall 💯

operator/api/v1alpha1/condition_builder.go
ruanxin: This builder is only used by kyma_types, ideally should not expose to controller, that's why put it in `api` package.
Even controller will use it, I don't think it's controller's responsible to own the implementation, it's part of the contract that api defines and providing to controller.
I assume other candidates you mean those functions with `Kyma` as receiver exists in `kyma_types`, then yes, I also think there should extract to other place, and normally those types definition files only have struct.

operator/api/v1alpha1/kyma_types.go
ruanxin: Use builder pattern here just want to reduce and encapsulate some condition params initialization, like `message`, `LastTransitionTime`.
Good, I take the suggestion to change the style to that chained style, I think the official name for this style is `Fluent interface`, but for that common way, I think the problem to me is lost the benefit to do verification in `build()` step, although I don't implement any yet:)

operator/config/crd/bases/operator.kyma-project.io_kymas.yaml
ruanxin: this is automatically generated, I think it's a controller-gen problem, it reads the description from `metav1.Condition`, and don't know why quote the string, but it's not broken, still valid string.

operator/api/v1alpha1/kyma_types.go
ruanxin: But back to you first point, by the current state, I think it's more common change it to normal initialization, but maybe later iteration, some params like `type`, `message` will be exposed and can be optional, with this builder pattern is a bit flexible.

operator/api/v1alpha1/condition_builder.go
jakobmoellersap: Lets continue with this now then and after this, do another refactoring for our own machinery package :)

operator/pkg/status/kyma_handler.go
```diff
@@ -109,70 +97,21 @@ func (k *Kyma) GetTemplateInfoForModule(
...
-			}
-		}
+	if err == nil && isValidState(state) {
+		return operatorv1alpha1.State(state)
 	}
+	return operatorv1alpha1.StateError
+}
 
-	return updateRequired, nil
+func isValidState(state string) bool {
```
jakobmoellersap, line 111: I think this should move into the machinery pkg later on. For now, how about we move it to the API package?

jakobmoellersap (commented):
Awesome stuff, just some nitpicking from my side, after that we can merge it in I think 💯

operator/pkg/status/kyma_handler.go
ruanxin: Just want to clarify `machinery` pkg you mentioned here first. 
I see two options:
1.  refactoring the package structure similar to `k8s.io/apimachinery` and `k8s.io/api`, then `/api` only have resource type definition, all other existing code in the current `/api` can be reorganized under `/apimachinery/pkg/` 
2. only have `/apimachinery`, the type definition go to `/apimachinery/pkg/api`
Or do you have different view in your mind?

ruanxin: And yes, this code should definitely not be here.

ruanxin: I think I just leave this here, ideally `SyncModuleInfo` should also go to apimachinery.

operator/controllers/kyma_controller.go
```diff
@@ -428,6 +388,20 @@ func initModuleFromServer(module *parsed.Module) *unstructured.Unstructured {
 	return &unstructuredFromServer
 }
 
+func (r *KymaReconciler) SyncModuleStatus(ctx context.Context, modules parsed.Modules) {
```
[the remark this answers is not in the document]
ruanxin: no, I don't think so, status helper is designed update `kyma.status` since it has a `StatusWriter`, but this is module status. But after revisit the controller, the module related code is too much, I'm trying to extract them to another pkg.

operator/api/v1alpha1/kyma_types.go
```diff
@@ -342,57 +322,46 @@ func init() {
...
-		}
-		if !found {
-			newCondition := KymaCondition{
-				Type:    ConditionTypeReady,
-				Status:  ConditionStatusFalse,
-				Reason:  module.Name,
-				Message: NewModuleMessage,
-			}
-			kyma.Status.Conditions = append(kyma.Status.Conditions, newCondition)
+func (kyma *Kyma) InitModuleConditions() {
```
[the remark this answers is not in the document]
ruanxin: After a second thought, I think I'll delete it, it designs to initialize all conditions, to have placeholder, but at the moment, we only have one condition, and it will be anyone create/update in `SyncConditionsWithModuleStates`, and I don't think later we will add more condition into it tbh.

operator/api/v1alpha1/kyma_types.go
```diff
@@ -342,57 +322,46 @@ func init() {
...
-		conditionsMap[condition.Reason] = &conditionExistsPair{exists: false, condition: condition}
-	}
-
-	for i := range kyma.Spec.Modules {
-		module := &kyma.Spec.Modules[i]
-		if _, exists := conditionsMap[module.Name]; exists {
-			conditionsMap[module.Name].exists = true
+		if condition.Reason == string(reason) {
+			isNewReason = false
+			if condition.Status != newCondition.Status || condition.Type != newCondition.Type {
```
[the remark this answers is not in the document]
ruanxin: The problem is that `LastTransitionTime`, according to the definition, it's the last time the condition transitioned from one status to another, so if the status are not changed, we shouldn't update the LastTransitionTime.

operator/api/v1alpha1/kyma_types.go
```diff
@@ -342,57 +322,46 @@ func init() {
...
-// TODO: drop this after condition.Reason != module.Name.
-func (kyma *Kyma) FilterNotExistsConditions() bool {
-	conditionsMap := make(map[string]*conditionExistsPair)
-	updateRequired := false
+func (kyma *Kyma) UpdateCondition(reason KymaConditionReason, status metav1.ConditionStatus) {
+	builder := NewConditionBuilder()
+	newCondition := builder.SetReason(reason).SetStatus(status).Build()
+	isNewReason := true
 	for i := range kyma.Status.Conditions {
 		condition := &kyma.Status.Conditions[i]
```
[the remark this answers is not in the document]
ruanxin: https://github.com/kyma-project/kyma-operator/pull/117#discussion_r950036549

operator/api/v1alpha1/kyma_types.go
```diff
@@ -342,57 +322,46 @@ func init() {
...
+	found := false
+	for _, condition := range kyma.Status.Conditions {
+		if condition.Reason == string(ConditionReasonModulesIsReady) {
+			found = true
 		}
 	}
+	if !found {
+		builder := NewConditionBuilder()
+		newCondition := builder.SetStatus(metav1.ConditionFalse).SetReason(ConditionReasonModulesIsReady).Build()
+		kyma.Status.Conditions = append(kyma.Status.Conditions, *newCondition)
```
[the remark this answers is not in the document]
ruanxin: oh, thanks to this hint! my original intention here is to save additional copy. But after research a bit how gc works in go, it's actually not a good idea return pointer here. In high chance, this returned data will trigger compile escape analysis to put it in heap, and need gc to clean it up later, and if just return value, normally the data copied still stay in stack and originally will be overwritten immediately. But I don't know what you mean regarding dereference, I mean the data will not lost if saved in heap and referenced within Status.Conditions, if kyma is still in memory, I don't think gc will delete it.

operator/api/v1alpha1/kyma_types.go
```diff
@@ -342,57 +322,46 @@ func init() {
...
+func (kyma *Kyma) InitModuleConditions() {
+	found := false
+	for _, condition := range kyma.Status.Conditions {
+		if condition.Reason == string(ConditionReasonModulesIsReady) {
+			found = true
 		}
 	}
+	if !found {
+		builder := NewConditionBuilder()
+		newCondition := builder.SetStatus(metav1.ConditionFalse).SetReason(ConditionReasonModulesIsReady).Build()
```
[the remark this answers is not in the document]
ruanxin: related function `InitModuleConditions` deleted

operator/api/v1alpha1/condition_builder.go
```diff
@@ -0,0 +1,48 @@
...
+	if cb.Reason == ConditionReasonModulesIsReady {
+		if cb.Status == metav1.ConditionTrue {
+			return "all modules are in ready state"
```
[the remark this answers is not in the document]
ruanxin: I created const for these messages, but imo I don't think it's a good idea to compare the message in any case, instead we should use status and reason, they are identifiers for program, but message is for human readable purpose.

operator/controllers/kyma_controller.go
```diff
@@ -317,42 +316,16 @@ func (r *KymaReconciler) HandleConsistencyChanges(ctx context.Context, kyma *v1a
...
-		} else if conditionsUpdated, err = status.Helper(r).
-			UpdateConditionFromComponentState(name, module, kyma); err != nil {
-			// if the component update failed we have to stop and retry until it succeeded
-			break
-		}
-
-		if !statusUpdateRequired {
-			// if any given condition was updated, we have to make sure to trigger a status update later on
-			statusUpdateRequired = conditionsUpdated
+func (r *KymaReconciler) SyncConditionsWithModuleStates(kyma *v1alpha1.Kyma) {
```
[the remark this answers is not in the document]
ruanxin: done

operator/api/v1alpha1/kyma_types.go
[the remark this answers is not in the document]
jakobmoellersap: Deref just means copying from heap :) Anyhow You changed it so its all good

jakobmoellersap (approved):
Deserves a 🎖️

PR #125, opened 2022-08-21
fix: rbac issues with kubebuilder generated image

adityabhatia:


jakobmoellersap (approved):
Well have to tackle that todo next

PR #138, opened 2022-08-30
Refactoring reconcile logic

ruanxin:
- replace HandleConsistencyChanges with HandleProcessingState


jakobmoellersap (approved):
I think we should rename some of the methods to indicate they dont always update the resources and call the API-Server, other than that, Great Job!

PR #139, opened 2022-08-30
Change namespace and namePrefix

ruanxin:
- add PeerAuthentication


operator/config/prometheus/pa.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: security.istio.io/v1beta1
```
jakobmoellersap, line 1: this is an istio resource in prometheus we should avoid this if possible and create another baes / overlay

jakobmoellersap (approved):

PR #142, opened 2022-08-31
Refactor Module Sync out of Controller Loop

jakobmoellersap:


operator/controllers/kyma_controller.go
```diff
@@ -191,18 +190,19 @@ func (r *KymaReconciler) HandleInitialState(ctx context.Context, kyma *v1alpha1.
...
 	modules, err = r.GenerateModulesFromTemplate(ctx, kyma)
 	if err != nil {
 		return r.UpdateStatusFromErr(ctx, kyma, v1alpha1.StateError,
 			fmt.Errorf("error while fetching modules during processing: %w", err))
 	}
 
-	statusUpdateRequiredFromCreation, err := r.CreateOrUpdateModules(ctx, kyma, modules)
+	sync := sync.New(r)
+
+	statusUpdateRequiredFromModuleSync, err := sync.Sync(ctx, kyma, modules)
```
ruanxin, line 205: maybe a more specific name like `SyncModule`? align with `SyncModuleInfo`

jakobmoellersap: the package is already called `module/sync` so I think it would be enough, if you dont think this is the case, we can rename it.

ruanxin (approved):

operator/controllers/kyma_controller.go
ruanxin: tbh I want to observe the meaning of the method directly from name, but not through package folder, especially here is an sub package. e.g: like the `parse.GenerateModulesFromTemplates` you defined is very expressive although it's also under `module`, but anyway, just a small habit concern.

PR #152, opened 2022-09-01
Add kcp dashboard

ruanxin:
- add kcp dashboard into repo
- add script to generate grafana configmap for auto provision


operator/grafana/controller-resources-metrics-configmap.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: controller-resources-metrics-dashboard
+  namespace: kcp-system
```
jakobmoellersap, line 5: shouldnt these be in the default namespace and then be automatically corrected by the kustomization? I am concerned why we put these outside of the kustomization

jakobmoellersap (changes requested):

operator/grafana/controller-resources-metrics-configmap.yaml
ruanxin: make sense, will change it 👍

operator/grafana/generate-configmap.sh
```diff
@@ -0,0 +1,15 @@
+#! /bin/bash
```
jakobmoellersap, line 1: why not put the entire `operator/grafana` directory into the config directory? If its not included in the kustomization, it is ignored.

jakobmoellersap (changes requested):
One last, nit then we can merge 🏅

operator/grafana/generate-configmap.sh
ruanxin: I think config it's better just have resources yaml files, by the definition here https://book-v1.book.kubebuilder.io/basics/project_creation_and_structure.html
```
config/...

Kubebuilder creates yaml config for installing the CRDs and related objects under config/.
```
and this plugin also lack of the feature to config output (maybe another PR to support it😄)

ruanxin: the generated configmap already in /config directory, which include the grafana dashboard content as `data:`, and they are under kustomization control.
the file in `operator/grafana` is just pure grafana dashboard json, it can be used for manually import.

jakobmoellersap (approved):

PR #156, opened 2022-09-05
Refactor controller logic

ruanxin:
- put more thing into module/sync

### DIFF
--- a/operator/controllers/kyma_controller.go
+++ b/operator/controllers/kyma_controller.go
@@ -26,7 +26,6 @@ import (
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/kyma-project/lifecycle-manager/operator/api/v1alpha1"
@@ -61,8 +60,6 @@ type KymaReconciler struct {
 	record.EventRecorder
 	RequeueIntervals
 	signature.VerificationSettings
-	RateQPS   int
-	RateBurst int
 }
 
 //nolint:lll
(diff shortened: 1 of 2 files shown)


operator/pkg/module/sync/runner_impl.go
```diff
@@ -169,3 +171,36 @@ func stateFromUnstructured(obj *unstructured.Unstructured) v1alpha1.State {
 	}
 	return v1alpha1.StateError
 }
+
+func (r *runnerImpl) UpdateModuleInfos(ctx context.Context, kyma *v1alpha1.Kyma) bool {
```
jakobmoellersap, line 175: I think this can become unexported
```suggestion
func (r *runnerImpl) updateModuleInfos(ctx context.Context, kyma *v1alpha1.Kyma) bool {
```

operator/pkg/module/sync/runner_impl.go
```diff
@@ -156,7 +157,8 @@ func (r *runnerImpl) SyncModuleInfo(_ context.Context, kyma *v1alpha1.Kyma, modu
 			kyma.Status.ModuleInfos = append(kyma.Status.ModuleInfos, latestModuleInfo)
 		}
 	}
-	return updateRequired
+	statusUpdateRequiredFromDeletion := r.UpdateModuleInfos(ctx, kyma)
+	return updateRequired || statusUpdateRequiredFromDeletion
```
jakobmoellersap, line 161: I think we should avoid this or statements if possible, they are really hard to read. Is it possible to document this?

operator/controllers/kyma_controller.go
```diff
@@ -208,24 +205,15 @@ func (r *KymaReconciler) HandleProcessingState(ctx context.Context, kyma *v1alph
...
-	if err != nil {
-		return r.UpdateStatusFromErr(ctx, kyma, v1alpha1.StateError,
-			fmt.Errorf("error while syncing conditions during processing: %w", err))
-	}
 
-	statusUpdateRequiredFromDeletion := r.UpdateModuleInfos(ctx, kyma)
 	moduleInfos := kyma.GetNoLongerExistingModuleInfos()
-	err = r.DeleteNoLongerExistingModules(ctx, moduleInfos)
-	if err != nil {
+	if err := r.DeleteNoLongerExistingModules(ctx, moduleInfos); err != nil {
```
jakobmoellersap, line 211: I think this needs a comment in the control loop,otherwise people wont understand why we need to check the delete case here

jakobmoellersap (changes requested):
Just smaller nits, other than that LGTM 🚀

operator/pkg/module/sync/runner_impl.go
ruanxin: I think this or statement is hard to avoid, the status updated determined by two parts, if the state of existing modules get changed or if module get deleted, I refactored this code again, please check if it's fine for you.

operator/controllers/kyma_controller.go
ruanxin: added a comment

operator/pkg/module/sync/runner_impl.go
ruanxin: done

jakobmoellersap (approved):

Issue #162, opened 2022-09-06
Propagate Target to proper Remote Flag during Manifest Installation


jakobmoellersap:
For installing in single-cluster mode, we need to enable the Module Manager for local installation. To make this work we will have to propagate the remote flag into the Manifest. For now there is only one path where we propagate the remote flag. However we need to verify if we also use local installations for central components. Then we need to propagate the remote flag to false. This has to be used in the single cluster mode to target the local cluster instead of using a Secret.

AC:

- [x] When running in Local Mode (e.g. in K3d) with only one cluster, the Manifest should be installed with `remote: false`, this was enabled by https://github.com/kyma-project/module-manager/pull/89 in the scope of https://github.com/kyma-project/module-manager/issues/78 and now allows module-manager to install local cluster resources
- [x] When running in KCP Mode (managed) with 2 clusters, the Manifest should be installed with `remote: true`
- [ ] Adjusted for the different workloads, the Kyma will have to differentiate if it should directly create the Module CR ( which we will now call `Extension Mode` ) and just track its status, or if we even install the component centrally (in which case we fully manage the component in the control plane as a module). In both cases, we need a `ModuleTemplate`. However, in Extension Mode the `descriptor` is not used since we only parse the Module CR and assume the CRD and the operator are preinstalled (thus called `Extension`).
- [x] By default, assume we do NOT run in Extension Mode, and that we create the Manifest every time

jakobmoellersap:
After discussion with @ruanxin we decided to no longer support extension mode and instead introduce a new CRD if the Use Case comes up again. This avoids the Problem that the ModuleTemplate with Extension Use Case would need to contain an `ModuleTemplate` without `.spec.descriptor` since it would be ignored

jakobmoellersap:
Implemented with removal of Extension support

PR #163, opened 2022-09-06
Always use Manifest for Rendering instead of native Module CRs

jakobmoellersap:
Implements part of https://github.com/kyma-project/lifecycle-manager/issues/162

This always uses the `Manifest`CR in the future to install a resource. It will not use native Module CRs anymore. If we later on want to introduce that again, we will create another CR instead of ModuleTemplate for this.


operator/pkg/module/parse/merger.go
```diff
@@ -0,0 +1,40 @@
...
+func mergeTargetIntoSpec(remote v1alpha1.Target, component *unstructured.Unstructured) error {
+	if err := mergo.Merge(&component.Object,
+		map[string]any{"remote": ConvertTargetToRemote(remote)},
+		mergo.WithAppendSlice); err != nil {
+		return fmt.Errorf("error while merging the remote attribute into the spec: %w", err)
+	}
+	return nil
+}
+
+func ConvertTargetToRemote(remote v1alpha1.Target) bool {
```
ruanxin, line 31: I actually would just propagate this target to manifest, and introduce a new target value, like `TargetLocal`)  but since the logic in manifest already [implemented](https://github.com/kyma-project/module-manager/pull/89/files), I'm fine with it.
But do you think if it make sense to add a new target value, like `TargetLocal`? I think it make the kyma resource a bit intuitive to understand

operator/controllers/kyma_controller_helper_test.go
```diff
@@ -119,9 +119,7 @@ func SKRModuleExistWithOverwrites(kymaName string, moduleTemplate *v1alpha1.Modu
 
 func getModule(kymaName string, moduleTemplate *v1alpha1.ModuleTemplate) (*unstructured.Unstructured, error) {
 	component := moduleTemplate.Spec.Data.DeepCopy()
-	if moduleTemplate.Spec.Target == v1alpha1.TargetRemote {
-		component.SetKind("Manifest")
-	}
+	component.SetKind("Manifest")
```
ruanxin, line 122: since we changed the manifest-operator to module-manager, maybe better to define it as a const and give an explainable name (I don't have a good name candidate yet, something like `ManagedModuleKind`), it's hard to understand this Manifest for others if don't know the history I think.

operator/controllers/kyma_controller_test.go
```diff
@@ -84,19 +84,21 @@ var _ = Describe("Kyma with empty ModuleTemplate", func() {
...
+		kcpModule *v1alpha1.Module
+	)
 	moduleTemplates := make(map[string]*v1alpha1.ModuleTemplate)
-	var skrModule *v1alpha1.Module
-	var kcpModule *v1alpha1.Module
 	BeforeAll(func() {
 		kyma = NewTestKyma("kyma-test-recreate")
 		skrModule = &v1alpha1.Module{
-			ControllerName: "manifest",
+			ControllerName: "manifest", // this is a module for SKR that should be installed by manifest operator
```
ruanxin, line 96:
```suggestion
			ControllerName: "manifest", // this is a module for SKR that should be installed by module-manager
```

operator/controllers/kyma_controller_test.go
```diff
@@ -84,19 +84,21 @@ var _ = Describe("Kyma with empty ModuleTemplate", func() {
...
 			Name:           "skr-module",
 			Channel:        v1alpha1.ChannelStable,
 		}
 		kcpModule = &v1alpha1.Module{
-			ControllerName: "kcp-operator",
+			ControllerName: "manifest", // this is a module for KCP that should be installed by manifest operator
```
ruanxin, line 101:
```suggestion
			ControllerName: "manifest", // this is a module for KCP that should be installed by module-manager
```

operator/pkg/module/parse/template_to_module.go
```diff
@@ -98,13 +98,14 @@ func NewModule(
...
-		}
-		component.SetKind("Manifest")
+	resource := template.Spec.Data.DeepCopy()
+	if err := mergeResourceIntoSpec(resource, component); err != nil {
+		return nil, err
+	}
+	if err := mergeTargetIntoSpec(template.Spec.Target, component); err != nil {
+		return nil, err
 	}
+	component.SetKind("Manifest")
```
ruanxin, line 108: same here: https://github.com/kyma-project/lifecycle-manager/pull/163/files#r963498435

operator/controllers/kyma_controller_helper_test.go
jakobmoellersap: Since we plan to refactor this to make use of the real type instead of unstructured, I think this is only temporary and we can leave it in

operator/pkg/module/parse/template_to_module.go
jakobmoellersap: See above comment

operator/pkg/module/parse/merger.go
jakobmoellersap: I think if we want to do that, this should be a separate PR for refactoring to not make it too big here. (it also changes the CRD after all and all samples)

ruanxin: ok, created a proposal issue for tracking this, we can consider it later. https://github.com/kyma-project/lifecycle-manager/issues/167

ruanxin (approved):

PR #175, opened 2022-09-07
lifecycle-manager: WatcherCR initial setup

jeremyharisch:
- Upgraded to latest listener, to have correct enqueuing in place for KymaCRs
- Introduced WatcherCR in kustomize
- Moved Service to Watcher kustomize layer
- Deactivated Watcher kustomize layer, in order to avoid current integration failures (Will be activated after runtime-watcher is integrated into lifecycle-manager)


operator/Dockerfile
```diff
@@ -24,4 +24,5 @@ WORKDIR /
 COPY --from=builder /workspace/manager .
 USER 65532:65532
 
+EXPOSE 8082
```
jakobmoellersap, line 27: Do we have to expose the port here? Isnt this simply optional?

operator/config/watcher/watcher.yaml
```diff
@@ -0,0 +1,23 @@
+# labels and selectors will be added using kustomize
+apiVersion: operator.kyma-project.io/v1alpha1
+kind: Watcher
+metadata:
+  name: watcher
```
jakobmoellersap, line 5: this will conflict with other watcher crs created. you should call it
```suggestion
  name: controller-manager-watcher
  namespace: system
```
that way its creating these resources with the default values and will replace them together with the kustomization in `manager` and `default`

operator/config/watcher/watcher.yaml
```diff
@@ -0,0 +1,23 @@
...
+spec:
+  componentName: lifecycle-manager
+  contractVersion: v1
+  gvrsToWatch: [] # needs to be removed after removing it in the runtime-WatcherCRD
+  serviceInfo:
+    serviceName: lifecycle-manager-event-service
```
jakobmoellersap, line 11:
```suggestion
    serviceName: lifecycle-manager-event-service
```
this is prejudiced against the prefix `lifecycle-manager`. You will have to write a custom patch to inject the correct name here. (see CERT-MANAGER section in `default`)

operator/config/watcher/watcher.yaml
```diff
@@ -0,0 +1,23 @@
...
+    servicePort: 8082
+
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: event-service
+spec:
+  ports:
```
jakobmoellersap, line 20: This service is missing a selector

jakobmoellersap (dismissed):

operator/config/watcher/watcher.yaml
jeremyharisch: Thats why I added the comment, that the selector and labels will be added using kustomize. If you call `make deploy` you will see that the selectors will be added

operator/Dockerfile
jeremyharisch: I looked it up in the official doc
```
The EXPOSE instruction does not actually publish the port. It functions as a type of documentation between the person who builds the image and the person who runs the container, about which ports are intended to be published.
```
Thus, yes it is just optional and I will remove it

operator/config/watcher/watcher.yaml
jakobmoellersap: yes but as above, the services should be appliable without kustomize in its default form. This is why if you run without the transformer it will fail

jeremyharisch: I will add them to make it work, if kustomize is not being used

operator/config/watcher/watcher.yaml
```diff
@@ -2,21 +2,25 @@
...
 ---
 apiVersion: v1
 kind: Service
 metadata:
   name: event-service
+  namespace: system
 spec:
+  selector:
+    app.kubernetes.io/component: lifecycle-manager.kyma-project.io
   ports:
```
adityabhatia, line 23: why not use the existing service in `config/manager/manager.yaml`?

adityabhatia (approved):

PR #176, opened 2022-09-08
Refactor commonlabels

ruanxin:
- add commonlabel override


jakobmoellersap (changes requested):
IMO we should leave the default selectors in so that its possible to still k apply -f these resources in the default configuration as much as possible. Otherwise we will have so many overlays and patches that it gets as unmaintainable as helm value overrides.

ruanxin:
> IMO we should leave the default selectors in so that its possible to still k apply -f these resources in the default configuration as much as possible. Otherwise we will have so many overlays and patches that it gets as unmaintainable as helm value overrides.

I'm still thinking we should prevent directly do `k apply -f`, the final resource should be generated by `kustomize` always, but maybe I'm not get the point of the maintenance efforts, let's have a talk when you back.

jakobmoellersap (approved):

PR #179, opened 2022-09-09
feat (template-operator): enable Wait (and Timeout) by default

adityabhatia:


samples/template-operator/operator/controllers/sample_controller.go
```diff
@@ -108,6 +109,8 @@ func (m *ManifestResolver) Get(obj types.BaseCustomObject) (types.InstallationSp
 			ConfigFlags: types.Flags{
 				"Namespace":       chartNs,
 				"CreateNamespace": true,
+				"Wait":            true,
```
jakobmoellersap, line 112: IMO these flags should be set by the library explictly with preset constants to enable these flags. Anyhow something that can easily be introduced later. Nevertheless these flags need documentation

jakobmoellersap (approved):

samples/template-operator/operator/controllers/sample_controller.go
adityabhatia: `declarative.WithResourcesReady(true)` should do that trick. This is a custom implementation and requires no `timeout` or `Wait`, it simply reconciles with a requeue interval until are resources are removed. In the declarative library only the requeue for deletion has to be implemented and it should work as is.

adityabhatia: Actually `declarative.WithResourcesReady(true)` should be able to handle it without an explicit `Wait` of `Timeout`. Just that re-queuing is not implemented for `Deleting` state yet. Part of this [item](https://github.com/kyma-project/module-manager/issues/96). Once this is implemented, it should workout out as is.

jakobmoellersap: Then I would say we implement Deleting first and then have the default ready right?

adityabhatia:
We can close this PR since https://github.com/kyma-project/module-manager/pull/100/files#diff-61aac7494c63b7864591f7c2d7702773ea640ab18f03ef7e4ff72e33cb6ba47fR117 coupled with the implementation at  https://github.com/kyma-project/module-manager/pull/98, should solve the issue.

PR #190, opened 2022-09-16
Adjust grafana resource and remove proxy metric service

ruanxin:


operator/config/grafana/controller-resources-metrics-configmap.yaml
```diff
@@ -2,10 +2,10 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: controller-resources-metrics-dashboard
-  namespace: system
+  namespace: kyma-system
```
jakobmoellersap, line 5: I think this should be kcp-system

operator/config/grafana/controller-runtime-metrics-configmap.yaml
```diff
@@ -2,10 +2,10 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: controller-runtime-metrics-dashboard
-  namespace: system
+  namespace: kyma-system
```
jakobmoellersap, line 5: same here

operator/config/grafana/kustomization.yaml
```diff
@@ -4,6 +4,6 @@ kind: Component
 resources:
 - controller-resources-metrics-configmap.yaml
 - controller-runtime-metrics-configmap.yaml
-- kcp-dashboard-configmap.yaml
+- manager-configmap.yaml
 generatorOptions:
-  disableNameSuffixHash: true
+  disableNameSuffixHash: true
```
jakobmoellersap, line 9: what does this do?

operator/config/grafana/manager-configmap.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: manager-dashboard
+  namespace: kyma-system
```
jakobmoellersap, line 5: and again namespace

operator/config/grafana/controller-resources-metrics-configmap.yaml
jakobmoellersap: or rather "system" as it gets replaced in the kustomization

operator/config/grafana/kustomization.yaml
ruanxin: as the name described, it will stop add an random name suffix, without this option, the name will become something like `name-chmdhgkb7t`

operator/config/grafana/controller-resources-metrics-configmap.yaml
ruanxin: this is required by KCP grafana, it only check configmap in `kyma-system`.

operator/config/default/kustomization.yaml
```diff
@@ -38,12 +38,35 @@ components:
...
+        path: /metadata/namespace
+        value: kyma-system
+  - target:
+      version: v1
+      kind: ConfigMap
+      name: controller-runtime-metrics-dashboard
+    patch: |-
+      - op: replace
+        path: /metadata/namespace
+        value: kyma-system
```
ruanxin, line 60: I have to write this patches specifically to overwrite the namespace for those dashboard configmap

jakobmoellersap (approved):

Issue #195, opened 2022-09-20
ModuleTemplate MutatingWebhookConfiguration rejects resource if .spec.data is a CR that is not contained in Control Plane


jakobmoellersap:
When applying ModuleTemplates, there is a validating Webhook running checks against a Module Template before accepting it into the cluster. However with this check, it also verifies the .status.state resource of the CRD. The issue here is that the CRD is currently still in the ModuleTemplate and this means that it must not have been installed in the KCP.

This currently leads to the limitation that all CRDs that are referenced in the ModuleTemplate have to be present in KCP as well. This is a heavily problematic side-effect as these should only be present in SKR for the Module to be installed correctly. Worst case the CRDs will have to be applied during the module installation.

- [x] No longer rely on CRDs in the Cluster during the webhook check in https://github.com/kyma-project/lifecycle-manager/blob/main/operator/api/v1alpha1/moduletemplate_webhook.go
- [ ] Find a good way to deduce the correct CRD based on the KCP first, if it doesnt find it, allow a "lenient" mode that is enabled by default which does not validate the CRD
- [ ] Find a way to derive the correct CRD based on the GVK in spec.data and introspect the module-template layers to find the correct crd.

jakobmoellersap:
The PR mentioned above will at least introduce the above-mentioned "lenient" mode until we find a better solution by introspecting the module template

jakobmoellersap:
We decided to remove the Validation as we expect it to be done at build time on the module

PR #197, opened 2022-09-20
feat: introduce strict-module-template-verification, disabled by default

jakobmoellersap:
this will skip CRD verification on the CR against `status.state` by default as we currently have no stable way to verify the CRD

Also introduces a test separation that allows to restart the webhook server during tests (here so i can test it both with enabled and disabled validation without having to teardown the testenv environment)

Workaround to solve blocking creation of CR in https://github.com/kyma-project/lifecycle-manager/issues/195 until we find a good validation since we disable the CRD check by default. We can enable it again later if we find a good validation way. This means we no longer need it in the control plane


operator/api/v1alpha1/moduletemplate_webhook.go
```diff
@@ -92,8 +99,10 @@ func (r *clusterAwareModuleTemplateValidator) validate(ctx context.Context, temp
 		allErrs = append(allErrs, err)
 	}
 
-	if err := r.validateCR(ctx, template); err != nil {
-		allErrs = append(allErrs, err)
+	if r.StrictModuleTemplateVerification {
+		if err := r.validateCR(ctx, template); err != nil {
```
ruanxin, line 103: what is the use case needs to validate CR? single cluster mode? 
Why not remove this validation, the template defines the CRD in component descriptor, then have a logic to validate CR by this template, the process flow is kind of wrong to me. Theoretically, I can define a module template with invalid CRD, it can pass this validateCR step, but the CR is meaningless.

jakobmoellersap: I think defining a module template with invalid crd is (should be) coverd by kyma cli. If this is not the case (e.g. manual creation) of course you are right. WDYT @ruanxin ?

ruanxin: since cli should do the job, then IMO I suggest we remove this validation to simplify our webhook logic.

jakobmoellersap: Agree, I removed the validation and will adjust the PR

ruanxin (approved):

PR #209, opened 2022-09-23
Create Github Action for Smoke Tests to work around DIND problem

jakobmoellersap:


.github/workflows/test.yml
```diff
@@ -0,0 +1,43 @@
...
+      with:
+        cache: true
+        go-version-file: 'tests/smoke/go.mod'
+        cache-dependency-path: 'tests/smoke/go.sum'
+
+    - name: Set up k3d/kyma for ${{ matrix.provision-type }}
+      if: ${{ matrix.provision-type }} == "kyma-cli"
+      run: |
+        wget -qO - https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | TAG=v5.4.6 bash
+        wget -q https://storage.googleapis.com/kyma-cli-unstable/kyma-linux
```
adityabhatia, line 33: Have you looked into [wget](https://github.com/marketplace/actions/github-action-for-wget) Github action?

jakobmoellersap: Its using a 3 year old docker image and is just a shell script for alpine. I figured I just use the inbuilt wget from ubuntu instead until we have to run multi-arch

ruanxin (approved):

PR #213, opened 2022-09-26
feat: setup control-plane compatible deployment

jakobmoellersap:


operator/config/rbac/leader_election_role.yaml
```diff
@@ -14,7 +14,7 @@ rules:
   - watch
   - create
   - update
-  - patch
+  - patches
```
ruanxin, line 17: certainly batch renaming issue, or?

PR #214, opened 2022-09-26
Introduce Rate-Limiter for template Operator

jeremyharisch:
This PR introduces a rate limiter for the template operator. The rate limiter is based on both an overall (token bucket) and per-item (exponential) rate limiter.

Furthermore, an environment test was set up for the template operator, covering the basic use-case.


samples/template-operator/operator/controllers/sample_controller.go
```diff
@@ -117,3 +127,13 @@ func (m *ManifestResolver) Get(obj types.BaseCustomObject) (types.InstallationSp
 		},
 	}, nil
 }
+
+// TemplateRateLimiter implements a rate limiter for a client-go.workqueue.  It has
+// both an overall (token bucket) and per-item (exponential) rate limiting.
+func TemplateRateLimiter(failureBaseDelay time.Duration, failureMaxDelay time.Duration,
```
jakobmoellersap, line 133: Lets introduce a `RateLimitArgs` flag here

jakobmoellersap (changes requested):
Please look at above changes and try to write an envtest if possible

samples/template-operator/operator/controllers/sample_controller.go
jeremyharisch: QQ: Do you want a flag which turns off the rate limiting completely? Since all input values for the rate limtier are already configured by flags. Have a look here: https://github.com/kyma-project/lifecycle-manager/pull/214/files#diff-f7a109f64839bd112a58f412858bd1543effde751d03b04028d0d8a330d5d2b4R124-R131

jakobmoellersap: Im just talking about a struct to pass to TemplateRateLimiter so that you do not have to pass 4 args into setupWithManager

adityabhatia:
I have created this PR https://github.com/jeremyharisch/kyma-operator/pull/1
[envtest](https://book.kubebuilder.io/reference/envtest.html) have no controllers running so you have to set resources to ready manually or use a custom resource..

samples/template-operator/operator/controllers/sample_controller.go
```diff
@@ -104,16 +126,25 @@ func (m *ManifestResolver) Get(obj types.BaseCustomObject) (types.InstallationSp
...
 		ChartFlags: types.ChartFlags{
 			ConfigFlags: types.Flags{
-				"Namespace":       chartNs,
-				"CreateNamespace": true,
-			},
-			SetFlags: types.Flags{
-				"nameOverride": nameOverride,
+				"Namespace":                chartNs,
+				"CreateNamespace":          true,
+				"DisableOpenAPIValidation": true,
```
adityabhatia, line 135: this will break the `module-chart` aka `redis` scenario. Please make these flags configurable as part of `SetupWithManager` ideally

adityabhatia (changes requested):
`busybox` chart should be placed inside a `test` folder or something similar, but not inside the operator root directory. 
Also the `flags` for Manifest client should be `configurable`.

samples/template-operator/operator/Makefile
```diff
@@ -8,7 +8,7 @@ ENVTEST_K8S_VERSION = 1.24.1
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v4.5.6
 CONTROLLER_TOOLS_VERSION ?= v0.9.2
-GOLANG_CI_LINT_VERSION ?= v1.46.2
+GOLANG_CI_LINT_VERSION ?= v1.48.0
```
jakobmoellersap, line 11:
```suggestion
GOLANG_CI_LINT_VERSION ?= v1.50.0
```

jakobmoellersap (changes requested):
Great stuff, I however still have some questions:

Correct me if im wrong, but This test does not actually test for rate limiting, right? Also the busy box chart we should probably not hard code into our samples here but instead download it for tests on demand. Wdyt?

jeremyharisch:
> Great stuff, I however still have some questions:
> 
> Correct me if im wrong, but This test does not actually test for rate limiting, right? Also the busy box chart we should probably not hard code into our samples here but instead download it for tests on demand. Wdyt?

Yes, you are right - The implemented test only tests the happy path and does not consider the RateLimiting. I am currently thinking about a way how to properly test the RateLimiting in this envTest

jeremyharisch:
The envTest for the RateLimiter is still missing. I will do a research how to test it properly, until then this PR is on hold

jeremyharisch:
I have now spent some time to find a proper way of testing the `workqueue.RateLimiter`, and I have come to the conclusion that there is no reasonable way without too much overhead to test this functionality.

We are here relying on a basic builder option given by the `sigs.k8s.io/controller-runtime/pkg/controller` pkg and used [internally by the controller](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L336). The used rate limiters are imported from the `k8s.io/client-go/util/workqueue` pkg. Both packages are well known and tested.

For a proper testing, in our test-suite, we would need to analyse the `Queue` from [Controller](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L61) - But since the `Controller` is a [private field](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/builder/controller.go#L60) of the `Builder` and no read-functions are given, we do not have a proper access for analysing.

In addition, I have asked the kubebuilder-community how this is testable without any responses until now (asked one week ago).


All in all, I propose to not test this functionality in our suite due to the explained reasons.

jakobmoellersap (approved):

adityabhatia (approved):

PR #224, opened 2022-09-29
add KCP watcher controller and tests

adityabhatia:
- Integration with lifecycle manager
- Reiterate implementation and refactor test (improvement areas could be covered in follow-up PRs)
- KCP Watcher can be enabled via flag `enable-kcp-watcher`, which is by default set to `true`
- SKR watcher chart moved to `operator/internal/charts`
- Necessary implementation added to `Dockerfile`


adityabhatia:
Was this working before? https://github.com/kyma-project/lifecycle-manager/pull/224/files#diff-caa6bb0b6f058fb6a0a3d3421cac26d99698d1edd6aec7757dd8b45c1b87ac67R386

operator/api/v1alpha1/watcher_types.go
```diff
@@ -0,0 +1,203 @@
...
+
+// WatcherCondition describes condition information for Watcher.
+type WatcherCondition struct {
+	// Type is used to reflect what type of condition we are dealing with.
+	// Most commonly WatcherConditionTypeReady it is used as extension marker in the future.
+	Type WatcherConditionType `json:"type"`
+
+	// Status of the Watcher Condition.
+	// Value can be one of ("True", "False", "Unknown").
+	Status WatcherConditionStatus `json:"status"`
```
ruanxin, line 105: please add `+kubebuilder:validation:Enum`

operator/api/v1alpha1/watcher_types.go
```diff
@@ -0,0 +1,203 @@
...
+type WatcherSpec struct {
+	// ServiceInfo describes the service information of the operator
+	ServiceInfo Service `json:"serviceInfo"`
+
+	// LabelsToWatch describes the labels that should be watched
+	LabelsToWatch map[string]string `json:"labelsToWatch"`
+
+	// Field describes the subresource that should be watched
+	// Value can be one of ("spec", "status")
+	Field FieldName `json:"field"`
```
ruanxin, line 39: please add `+kubebuilder:validation:Enum`

operator/api/v1alpha1/kyma_types.go
```diff
@@ -380,10 +380,10 @@ func IsValidState(state string) bool {
 func (kyma *Kyma) SyncConditionsWithModuleStates() {
 	conditionReason := ConditionReasonModulesAreReady
 	conditionStatus := metav1.ConditionTrue
-	for i := range kyma.Status.ModuleInfos {
-		moduleInfo := kyma.Status.ModuleInfos[i]
+	for _, moduleInfo := range kyma.Status.ModuleInfos {
```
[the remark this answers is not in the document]
adityabhatia: yes that's what I was wondering, that either the pointer is missing or this is unnecessary. 
But what about this? https://github.com/kyma-project/lifecycle-manager/pull/224#issuecomment-1262341595

operator/api/v1alpha1/watcher_types.go
adityabhatia: had not pushed the crd until now, it's also in

operator/main.go
```diff
@@ -197,6 +205,14 @@ func defineFlagVar() *FlagVar {
 	flag.BoolVar(&flagVar.enableModuleCatalog, "enable-module-catalog", true,
 		"Enabling the Module Catalog Synchronization for Introspection of "+
 			"available Modules based on ModuleTemplates.")
+	flag.BoolVar(&flagVar.enableKcpWatcher, "enable-kcp-watcher", true,
```
ruanxin, line 208: By default should disable watcher
```suggestion
	flag.BoolVar(&flagVar.enableKcpWatcher, "enable-kcp-watcher", false,
```

operator/main.go
```diff
@@ -197,6 +205,14 @@ func defineFlagVar() *FlagVar {
...
+		"Enabling KCP Watcher to reconcile Watcher CRs created by KCP run operators")
+	flag.StringVar(&flagVar.skrWatcherPath, "skr-watcher-path", "./internal/skr/chart/skr-webhook",
```
ruanxin, line 210:
```suggestion
	flag.StringVar(&flagVar.skrWatcherPath, "skr-watcher-path", "skr-webhook",
```
This is not correct path after deployed with docker

operator/main.go
```diff
@@ -197,6 +205,14 @@ func defineFlagVar() *FlagVar {
...
+		"The path to the skr watcher chart.")
+	flag.StringVar(&flagVar.vsName, "virtual-svc-name", "kcp-events",
```
ruanxin, line 212: Please change to dynamically create this VirtualService, otherwise we need to predefined a `VirtualService` and provide dummy Spec (it will failed the validation webhook in gardener cluster if not provide), more probmatic is keeping the name and namespace consistent between these two flags and the manifest. Agreed this solution with @jakobmoellersap, @khlifi411

operator/controllers/watcher_controller.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+func (r *WatcherReconciler) HandleErrorState(ctx context.Context, obj *v1alpha1.Watcher) error {
+	return r.updateWatcherCRStatus(ctx, obj, v1alpha1.WatcherStateProcessing, "observed generation change")
+}
+
+func (r *WatcherReconciler) HandleReadyState(ctx context.Context, logger logr.Logger,
+	obj *v1alpha1.Watcher,
+) error {
+	if obj.Generation != obj.Status.ObservedGeneration {
```
ruanxin, line 200: This condition will prevent CR get changes after state become Ready

operator/config/watcher/watcher.yaml
```diff
@@ -1,18 +1,18 @@
...
   labelsToWatch:
-    "watched-by": "lifecycle-manager"
+    "operator.kyma-project.io/watched-by": "lifecycle-manager"
   field: "spec"
   serviceInfo:
-    name: event-service
+    name: lifecycle-manager-event-service
     port: 8082
-    namespace: system
+    namespace: kcp-system
```
ruanxin, line 15: Please double check if this namespace really needed, otherwise drop it. I checked the code, basically there is no real usage just print out a hostname. The reason is this namespace is not controlled by Kustomize overwrite by default, we need to add additional patch to keep it consisten.
```
 func destinationHost(serviceName, serviceNamespace string) string {
	return fmt.Sprintf("%s.%s.svc.cluster.local", serviceName, serviceNamespace)
}

**Review by ruanxin — CHANGES_REQUESTED** [review]
Existing Bug:

- Watcher update get 

```
1.664546347144869e+09	ERROR	kcp-system/lifecycle-manager-watcher	failed to update watcher cr to ready status	{"controller": "watcher", "controllerGroup": "operator.kyma-project.io", "controllerKind": "Watcher", "Watcher": {"name":"lifecycle-manager-watcher","namespace":"kcp-system"}, "namespace": "kcp-system", "name": "lifecycle-manager-watcher", "reconcileID": "5d854d01-a3ff-4fec-8ca9-0f6a7f3ae91f", "error": "Operation cannot be fulfilled on watchers.operator.kyma-project.io \"lifecycle-manager-watcher\": the object has been modified; please apply your changes to the latest version and try again"}
```

- skr-webook helm chart not deployed correctly
  - skr-webook deployed in default namespace
  - kcp.IP 
```
1.6645464436851554e+09	ERROR	skr-webhook		{"error": "Post \"http://kcp.ip/v1/lifecycle-manager/event\": dial tcp: lookup kcp.ip on 100.104.0.10:53: no such host"}
github.com/kyma-project/runtime-watcher/skr/internal.(*Handler).sendRequestToKcp
```

operator/main.go
ruanxin: overwrite added [here](https://github.com/kyma-project/lifecycle-manager/pull/224/files#diff-28823c7f87387cdc0cd68c606bbca94a8e4c675690d3b9c6635a052a9b58caccR10-R16)

operator/config/watcher/watcher.yaml
khlifi411: The namespace is needed because when configuring the destination rule in the virtual service, we need to include the namespace in order for the virtual service to know how to route the traffic to the service in case of different namespace, see:
https://github.com/istio/istio.io/blob/4b89c302aea85cf7041db7617984aadcb08599b4/content/en/docs/concepts/traffic-management/index.md?plain=1#L240-L252

khlifi411: Resolving after huddle 👍

PR #231, opened 2022-10-04
Dynamically create Virtual Service resource for watcher reconciler

khlifi411:


operator/internal/custom/istio_client.go
```diff
@@ -15,46 +17,84 @@ import (
...
-	if err != nil {
+	var err error
+	var virtualService *istioclientapi.VirtualService
+	virtualService, err = c.NetworkingV1beta1().
+		VirtualServices(c.IstioGateway.Namespace).
+		Get(ctx, virtualServiceName, metav1.GetOptions{})
+	if client.IgnoreNotFound(err) != nil {
 		return nil, fmt.Errorf("failed to fetch virtual service %w", err)
 	}
+	if kerrors.IsNotFound(err) {
```
ruanxin, line 59: consider rename it to `apierrors`, that's how usually the naming defined in [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/interfaces.go#L22)

operator/main.go
```diff
@@ -213,6 +214,8 @@ func defineFlagVar() *FlagVar {
 		"The name of the Istio virtual service to be updated.")
 	flag.StringVar(&flagVar.vsNamespace, "virtual-svc-namespace", metav1.NamespaceDefault,
 		"The namespace of the Istio virtual service to be updated.")
+	flag.StringVar(&flagVar.istioGateway, "istio-gateway", "default/kyma-gw",
```
ruanxin, line 208: do we really need these value to be configured as flag? what's the purpose to introduce these 3 overwrites?

operator/internal/custom/istio_client.go
```diff
@@ -15,46 +17,84 @@ import (
...
 	istioclient.Interface
+	// IstioGateway string
+	IstioGateway client.ObjectKey
 }
 
-func NewVersionedIstioClient(cfg *rest.Config) (*IstioClient, error) {
+func NewVersionedIstioClient(cfg *rest.Config, istioGw string) (*IstioClient, error) {
+	gwStrings := strings.Split(istioGw, "/")
+	if len(gwStrings) != 2 {
+		return nil, fmt.Errorf("error validating gateway config: expected (namespace/name), got: (%s)", istioGw)
```
ruanxin, line 34: Can we assume this Gateway is always deployed in `default` namespace with name `kyma-gateway` (or consider other name to avoid potential naming conflict)? How about fix the name and namespace, what will be the side effect? Then we can get rid of this parsing logical.

operator/internal/custom/istio_client.go
```diff
@@ -15,46 +17,84 @@ import (
...
+	_, err := c.NetworkingV1beta1().
+		Gateways(c.IstioGateway.Namespace).
+		Get(ctx, c.IstioGateway.Name, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("error getting configured istio gateway: %w", err)
+	}
+	virtualSvc := &istioclientapi.VirtualService{}
+	virtualSvc.SetName(virtualServiceName)
+	virtualSvc.SetNamespace(c.IstioGateway.Namespace)
+	virtualSvc.Spec.Gateways = append(virtualSvc.Spec.Gateways, c.IstioGateway.Name)
```
ruanxin, line 80: Same here, can we fix this Namespace instead of read for config.

operator/internal/custom/istio_client.go
jakobmoellersap: I think the gateway name should in general be discovered and if there are more than one, we should use a label selector for getting the correct gateway. Reason for this is that I think it can vary in name between the different KCP environments.

khlifi411: Hi @jakobmoellersap, I think you're right. I have created a ticket for your suggestion:
#234 
Please have a look and check if the ACs are to the point 👍

operator/internal/custom/istio_client.go
```diff
@@ -15,8 +16,10 @@ import (
 )
 
 const (
-	firstElementIdx = 0
-	contractVersion = "1"
+	firstElementIdx    = 0
+	contractVersion    = "1"
```
ruanxin, line 20: use `v1` is more fit for version definition.

adityabhatia: correct, please change the template accordingly

operator/internal/custom/istio_client.go
jakobmoellersap: Added a suggestion to the AC

operator/internal/custom/istio_client.go
khlifi411: Applied it [here](https://github.com/kyma-project/lifecycle-manager/pull/231/commits/10183a46319e726d4351465c872a45443ebd123e)

ruanxin (approved):

PR #232, opened 2022-10-04
Refactor load_test as dedicate kustomize build instead of component

ruanxin:
- add load-test-deploy make command

### DIFF
--- a/operator/config/load_test/kustomization.yaml
+++ b/operator/config/load_test/kustomization.yaml
@@ -1,9 +1,37 @@
-apiVersion: kustomize.config.k8s.io/v1alpha1
-kind: Component
-
 resources:
   - metrics_cluster_role_binding.yaml
   - metrics_role.yaml
   - metrics_role_binding.yaml
+  - ../manager
 generatorOptions:
   disableNameSuffixHash: true
+
+# Adds namespace to all resources.
+namespace: kcp-system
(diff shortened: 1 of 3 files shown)


jakobmoellersap (changes requested):
I think this needs an explanation of why it is there and how it can be used in the main kustomization.yaml

Other than that, if its not used for active development it might be wort not setting a makefile target for this. Your choice on this though.

Generally LGTM :)

operator/config/load_test/kustomization.yaml
```diff
@@ -1,9 +1,37 @@
-apiVersion: kustomize.config.k8s.io/v1alpha1
-kind: Component
-
 resources:
```
ruanxin, line 1:
```suggestion
# The main purpose of this dedicate kustomization file is separate resources configuration for load test from default setup, enable the ability to configure additional components.
resources:
```

ruanxin:
> I think this needs an explanation of why it is there and how it can be used in the main kustomization.yaml
> 
> Other than that, if its not used for active development it might be wort not setting a makefile target for this. Your choice on this though.
> 
> Generally LGTM :)

Make sense, I removed the makefile command, and added a comment to explain the purpose of that kustomization file, feel free add suggestions to the changes if needed.

jakobmoellersap (approved):

PR #233, opened 2022-10-04
!feat: add semver-version based skew handling

### LINKED ISSUE #193: Do not propagate downgrade of Module Version when downgrading channel in ModuleTemplate
When creating ModuleTemplates, it is mandatory to define the `spec.channel` field. This is responsible for defining different release channels: https://github.com/kyma-project/community/pull/717

For propagating correct release channels, it is important to not propagate downgrades as Kyma Downgrades are not supported.
This is because it takes lots of maintenance effort to correctly handle downgrades for every scenario.

Currently, lifecycle-manager propagates every change to the release channel of a module in kyma. However, this leads to downgrade propagation which we do not want (see above).

Ideally once the module was upgraded to a newer channel (e.g. rapid with v1.17.0) and we then downgrade to stable (v.1.1.0) then it will not downgrade until (stable has reached a version >v1.17.0)

- [x] Allow Upgrades as usual
- [x] Deny Downgrades as long as there is a smaller version in the channel that is tracked than the installed version

jakobmoellersap:
This adds channel skew handliing for semantic version differences in channesl. This is done to catch a case, in which a Kyma instance was created with a Module in a specific version (e.g. 2.1.0), but the module was already installed with a higher version of that module before (e.g. 3.0.0). This can happen if the channel configuration changed.

This is a regular scenario for cases where the user downgrades a previous module installation in the stability (e.g. from fast->stable). This will lead to the module not being downgraded (which is an unsupported versioning path), and instead it will wait for the stable branch to catchup.

If the user wants to completely downgrade a module instead (e.g. because of bugs), he will have to first uninstall a module, and then reinstall the module in the desired channel (e.g. remove the module from kyma in fast channel, and then add it again in stable). This will cause a reinstallation.

It also removes Observed Generation, as it was never used or referenced in checks since we switched to ModuleTemplate Parsing + Remote Sync on every Reconciliation.

Pending Tasks:

- [ ] Write Tests for Skew Cases and Ignore Cases
- [ ] Extract into interface to make unit-testable

Fixes #193


operator/api/v1alpha1/condition_builder.go
```diff
@@ -31,24 +33,47 @@ func (cb *ConditionBuilder) SetReason(reason KymaConditionReason) *ConditionBuil
...
+			return MessageModuleCatalogIsSynced
+		case metav1.ConditionUnknown:
+			fallthrough
+		case metav1.ConditionFalse:
+			fallthrough
+		default:
+			return MessageModuleCatalogIsOutOfSync
+		}
+	default:
+		return "no detailed message available as reason is unknown to API"
```
ruanxin, line 77: should we put this return state also as an const?
If you not use this default case, and return it in the end, then you can get rid of each `fallthrough` in above `case`. That's not a typical usage of fallthrough I think.

jakobmoellersap: I tried this and then the linter said no to me :(

jakobmoellersap: Should I adjust it so it doesnt use this linter setting anymore?

ruanxin (approved):

PR #241, opened 2022-10-06
feat: move kcp watcher to lifecycle-manager

adityabhatia:


operator/config/default/kustomization.yaml
```diff
@@ -23,7 +23,7 @@ components:
   - ../crd
   - ../rbac
   # [WATCHER] To enable the watcher, uncomment all the sections with [WATCHER]
-  #- ../watcher
+  - ../watcher
```
adityabhatia, line 26: comment out before merge!!

ruanxin (dismissed):

PR #259, opened 2022-10-11
fix: e2e and docs issues 

adityabhatia:


operator/config/watcher/operator_v1alpha1_watcher.yaml
```diff
@@ -10,7 +10,7 @@ spec:
     "operator.kyma-project.io/watched-by": "lifecycle-manager"
   field: "spec"
   serviceInfo:
-    name: klm-event-service
+    name: event-service
```
ruanxin, line 13: `klm-` is the namePrefix for KCP deployment, if you replace it with `lifecycle-manager-event-service`, the name not fit.

operator/config/istio/gataway.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: networking.istio.io/v1beta1
 kind: Gateway
 metadata:
-  name: lifecycle-manager-kyma-gateway
+  name: kyma-gateway
```
ruanxin, line 5: This name has to be `lifecycle-manager-kyma-gateway` until watcher can dynamically using existing gateway, please check the patch I introduced in this PR https://github.com/kyma-project/lifecycle-manager/pull/262/files#diff-6ff3fbcd21a686f5d672eb6d639fb2b9c018e668eb3dd2588589f09465659186R6

If you change to this way, in KCP deployment, this gateway will rename to `klm-kyma-gateway`

operator/config/istio/kustomization.yaml
```diff
@@ -10,5 +10,14 @@ resources:
 generatorOptions:
   disableNameSuffixHash: true
 
-patchesStrategicMerge:
-  - patches/istio_namespace_labels.yaml
+patches:
```
ruanxin, line 13: I have an solution in this PR already, not only namespace, but also name should be changed. https://github.com/kyma-project/lifecycle-manager/pull/262/files#diff-6ff3fbcd21a686f5d672eb6d639fb2b9c018e668eb3dd2588589f09465659186R1-R6

operator/config/watcher/patches/watcher_service_patch.yaml
```diff
@@ -0,0 +1,3 @@
+- op: replace
+  path: /spec/serviceInfo/name
+  value: lifecycle-manager-event-service
```
ruanxin, line 3: same as what I mentioned in module-manager pr, the nameprefix is different for KCP deployment, we can't hard coded to lifecycle-manager-event-service

ruanxin (changes requested):
Please revisit the comments I left.

operator/config/watcher/operator_v1alpha1_watcher.yaml
adityabhatia: But in a non-KCP deployment scenario it doesn't work, since the prefix is `lifecycle-manager`. Whatever prefix it is should not be a part of this file.

operator/config/istio/gataway.yaml
adityabhatia: The patch you proposed doesn't help the `default` Kustomize build unfortunately

operator/config/istio/kustomization.yaml
adityabhatia: Name is automatically corrected using `namePrefix`

operator/config/watcher/patches/watcher_service_patch.yaml
adityabhatia: ok then we need to do it on the base layer.

operator/config/istio/gataway.yaml
adityabhatia: The end result comes out to be` lifecycle-manager-lifecycle-manager-kyma-gateway` !!

ruanxin (approved):

PR #260, opened 2022-10-12
Implement SKR chart removal for last watcher CR deletion

khlifi411:
## Proposed changes:

- Uninstall chart while reconciling the deletion of the last watcher CR


operator/Makefile
```diff
@@ -174,6 +174,9 @@ informer-gen: ## Download informer-gen
 lister-gen: ## Download lister-gen
 	GOBIN=$(LOCALBIN) go install k8s.io/code-generator/cmd/lister-gen@v0.23.4
 
+.PHONY: resolve
```
ruanxin, line 177: what's the reason to introduce this?

khlifi411: It's just a shortcut for us instead of `go mod tidy` 😄

operator/internal/deploy/webhook_chart.go
```diff
@@ -127,8 +143,14 @@ func installOrRemoveChartOnSKR(ctx context.Context, restConfig *rest.Config, rel
...
 		if !uninstalled {
-			//nolint:goerr113
-			return fmt.Errorf("waiting for skr webhook resources to be deleted")
+			return ErrSKRWebhookWasNotRemoved
+		}
+		ready, err := ops.VerifyResources(deployInfo)
+		if err != nil {
+			return fmt.Errorf("failed to verify webhook resources: %w", err)
+		}
+		if ready {
```
ruanxin, line 152: `ready` or `!ready` ?

khlifi411: verify resources returns `ready=true` when resources are there.So I am expecting it to be `false`.

adityabhatia (changes requested):
Implementation looks great, but please adapt module-manager interface as mentioned here https://github.com/kyma-project/module-manager/#sample-usage
and tests are missing

khlifi411:
> Implementation looks great, but please adapt module-manager interface as mentioned here https://github.com/kyma-project/module-manager/#sample-usage
> and tests are missing

I have created a ticket to take care of adapting the module-manager interface to match the usage guide #268 
Please check the ACs and update them as necessary 👍

adityabhatia (approved):

PR #262, opened 2022-10-12
Allow skrwebhook resource limit configurable

ruanxin:


operator/config/watcher/operator_v1alpha1_watcher.yaml
```diff
@@ -10,7 +10,7 @@ spec:
     "operator.kyma-project.io/watched-by": "lifecycle-manager"
   field: "spec"
   serviceInfo:
-    name: lifecycle-manager-event-service
+    name: klm-event-service
```
adityabhatia, line 13: should not be hard coded

PR #276, opened 2022-10-18
feat: refactor module catalog to sync into runtime

jakobmoellersap:
Creates an SSA-based Remote Catalog Sync that is syncing all ModuleTemplates into the Runtime. It creates/updates all ModuleTemplates that are not existing, whose generation is outdated, or after a given force Interval. It deletes all templates not occuring in the control-plane.


operator/pkg/catalog/remote_catalog.go
```diff
@@ -0,0 +1,228 @@
...
+		// update it
+		controlPlaneSyncedGen, _ := strconv.Atoi(remote.Annotations[v1alpha1.LastSyncGenerationControlPlane])
+		controlPlaneGen := (&controlPlaneList.Items[controlPlaneIndex]).GetGeneration()
+		runtimeSyncedGen, _ := strconv.Atoi(remote.Annotations[v1alpha1.LastSyncGenerationRuntime])
+
+		if int64(controlPlaneSyncedGen) != controlPlaneGen {
+			remote.Annotations[v1alpha1.LastSyncGenerationControlPlane] = strconv.Itoa(int(controlPlaneGen))
+			diffToApply = append(diffToApply, remote)
+		} else if int64(runtimeSyncedGen) != remote.GetGeneration() {
+			remote.Annotations[v1alpha1.LastSyncGenerationRuntime] = strconv.Itoa(int(remote.GetGeneration()))
```
ruanxin, line 158: the remote generation will get increased after apply the diff.

operator/pkg/catalog/remote_catalog.go
```diff
@@ -0,0 +1,257 @@
...
+		controlPlaneGen := (&controlPlaneList.Items[controlPlaneIndex]).GetGeneration()
+		runtimeSyncedGen, _ := strconv.Atoi(remote.Annotations[v1alpha1.LastSyncGenerationRuntime])
+
+		if int64(controlPlaneSyncedGen) != controlPlaneGen {
+			remote.Annotations[v1alpha1.LastSyncGenerationControlPlane] = strconv.Itoa(int(controlPlaneGen))
+			(&controlPlaneList.Items[controlPlaneIndex]).Spec.DeepCopyInto(&remote.Spec)
+			diffToApply = append(diffToApply, remote)
+		} else if int64(runtimeSyncedGen) != remote.GetGeneration() {
+			(&controlPlaneList.Items[controlPlaneIndex]).Spec.DeepCopyInto(&remote.Spec)
+			remote.Annotations[v1alpha1.LastSyncGenerationRuntime] = strconv.Itoa(int(remote.GetGeneration() + 1))
```
ruanxin, line 170: better add a comment to explain this `+1`

jakobmoellersap: Its in the comment of the main method https://github.com/kyma-project/lifecycle-manager/pull/276/files#diff-b247ded7d56054153c2091874b28f2131d1b537bc921a7cbb7c6a82354f6b16aR106-R113

operator/pkg/catalog/remote_catalog.go
```diff
@@ -0,0 +1,257 @@
...
+func (c *RemoteCatalog) Cleanup(ctx context.Context) error {
+	moduleTemplatesRuntime := &v1alpha1.ModuleTemplateList{}
+	err := c.runtimeClient.List(ctx, moduleTemplatesRuntime)
+	// it can happen that the ModuleTemplate CRD is not existing in the Remote Cluster, then no cleanup is necessary
+	if meta.IsNoMatchError(err) {
+		return nil
+	}
+
+	for i := range moduleTemplatesRuntime.Items {
+		if err := c.runtimeClient.Delete(ctx, &moduleTemplatesRuntime.Items[i]); err != nil {
```
ruanxin, line 251: I think we should ignore notfound error

operator/pkg/catalog/remote_catalog.go
```diff
@@ -0,0 +1,257 @@
...
+		return c.runtimeClient.Create(ctx, &v1extensions.CustomResourceDefinition{
+			ObjectMeta: metav1.ObjectMeta{Name: crd.Name, Namespace: crd.Namespace}, Spec: crd.Spec,
+		})
+	}
+
+	if err != nil {
+		return err
+	}
+
+	// crd.SetResourceVersion(crdFromRuntime.GetResourceVersion())
```
ruanxin, line 235: remove this code?

operator/pkg/catalog/remote_catalog.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = c.runtimeClient.Get(ctx, client.ObjectKey{
+		Name: fmt.Sprintf("%s.%s", plural, v1alpha1.GroupVersion.Group),
+	}, crdFromRuntime)
+
+	if k8serrors.IsNotFound(err) {
+		return c.runtimeClient.Create(ctx, &v1extensions.CustomResourceDefinition{
```
ruanxin, line 226: here if the create return success, doesn't mean crd generated, right? I think here it's more safe not return by creation, but let this not found error return in the later step.
```suggestion
		_ := c.runtimeClient.Create(ctx, &v1extensions.CustomResourceDefinition{
```

operator/pkg/catalog/remote_catalog.go
```diff
@@ -0,0 +1,257 @@
...
+func (c *RemoteCatalog) Delete(
+	ctx context.Context,
+) error {
+	var moduleTemplatesRuntime *v1alpha1.ModuleTemplateList
+	err := c.runtimeClient.List(ctx, moduleTemplatesRuntime)
+	if err != nil {
+		return err
+	}
+	for i := range moduleTemplatesRuntime.Items {
+		if err := c.runtimeClient.Delete(ctx, &moduleTemplatesRuntime.Items[i]); err != nil {
```
ruanxin, line 200: should we ignore not found error?

operator/pkg/catalog/remote_catalog.go
```diff
@@ -0,0 +1,257 @@
...
+
+		if err := c.runtimeClient.Patch(
+			ctx, diff, patch, c.settings.SSAPatchOptions,
+		); err != nil {
+			return fmt.Errorf("could not apply module template diff: %w", err)
+		}
+	}
+
+	for _, diff := range diffDelete {
+		if err := c.runtimeClient.Delete(ctx, diff); err != nil {
```
ruanxin, line 90: ignore not found?

operator/controllers/kyma_controller_helper_test.go
```diff
@@ -221,24 +220,98 @@ func RemoteKymaExists(remoteClient client.Client, kymaName string) func() error
...
+
+		return true
+	}
+}
+
+func ModuleTemplatesLabelsCountMatch(
+	clnt client.Client, kyma *v1alpha1.Kyma, count int,
+) func() bool {
+	return func() bool {
+		for _, module := range kyma.Spec.Modules {
```
ruanxin, line 266: I think we should refactor this check use not `Eventually` but only `Expect`, the `kyma` here is static, right? Then we can refactor the error check to `Expect(err).ShouldNot(HaveOccurred())` so that for troubleshooting a bit easy for spot the place where error occurred, otherwise we only get this function failed, but don't know exactly which line.

operator/controllers/kyma_controller_helper_test.go
```diff
@@ -221,24 +220,98 @@ func RemoteKymaExists(remoteClient client.Client, kymaName string) func() error
...
+				return err
+			}
+		}
+
+		return nil
+	}
+}
+
+func ModuleTemplatesLastSyncGenMatches(clnt client.Client, kyma *v1alpha1.Kyma) func() bool {
+	return func() bool {
```
ruanxin, line 240: same as here: https://github.com/kyma-project/lifecycle-manager/pull/276/files#r1004338714

operator/controllers/kyma_controller_helper_test.go
```diff
@@ -221,24 +220,98 @@ func RemoteKymaExists(remoteClient client.Client, kymaName string) func() error
...
-	clnt client.Client,
-	kyma *v1alpha1.Kyma,
-) (*v1.ConfigMap, error) {
-	catalog := &v1.ConfigMap{}
-	catalog.SetName(controllers.CatalogName)
-	catalog.SetNamespace(kyma.GetNamespace())
-	err := clnt.Get(ctx, client.ObjectKeyFromObject(catalog), catalog)
-	if err != nil {
-		return nil, err
+func ModuleTemplatesExist(clnt client.Client, kyma *v1alpha1.Kyma) func() error {
```
ruanxin, line 223: same as here: https://github.com/kyma-project/lifecycle-manager/pull/276/files#r1004338714

operator/controllers/kyma_controller_helper_test.go
ruanxin: no, I'm wrong, the actual resource you want to check here is dynamic, then it's better we return the error instead of true to know exactly what error occurred for troubleshooting purpose.

operator/controllers/kyma_controller_helper_test.go
```diff
@@ -221,24 +220,98 @@ func RemoteKymaExists(remoteClient client.Client, kymaName string) func() error
...
+				return false
+			}
+
+			descriptor, err := template.Spec.GetDescriptor()
+			if err != nil {
+				return false
+			}
+
+			if len(descriptor.GetLabels()) != count {
+				return false
```
ruanxin, line 281: Then here we should define a new error

operator/pkg/catalog/remote_catalog.go
jakobmoellersap: In this case the error will show that the CRD does not exist but still create the CRD which I find confusing tbh. Why would it need to error out here. Instead if the CRD is not yet ready when the create is called it will error there and then retry.

ruanxin: my main concern is after this `CreateModuleTemplateCRDInRuntime`, it tries to list moduleTemplate, but I think in the initial step, during the create crd, the `List` will alway error out if we not return that not found error. For me, it's better make sure CRD exists before doing List, but anyway, it will return error and do correctly in next reconcile, it's fine keep the current code.
```
if err := c.CreateModuleTemplateCRDInRuntime(ctx, v1alpha1.ModuleTemplateKind.Plural()); err != nil {
			return err
		}
		err = c.runtimeClient.List(ctx, moduleTemplatesRuntime)
```

jakobmoellersap: I just verified this and it will only error out if the API Server is too slow in processing the CRD request async. While it can happen its not a "must".
Anyhow I also checked the errors and it was a really weird error when returning the previous err instead of the one from the list request. Considering this I would prefer to keep the current code

ruanxin (approved):

PR #277, opened 2022-10-18
Refactor watcher and kyma reconcilers to improve load test performance

khlifi411:
AC:
- [x] Dedicate controller for Watcher CR
  - [x] In the reconcile, do virtualservice update 
  - [x] Config 1 worker to avoid concurrent write
- [x] Migrate  skrwebhook installation logic into Kyma controller
  - [x] Skrwebhook installation state as kyma condition
- [x] Make sure ca cert between Secret and ValidationWebhookConfigurations consistent:
  > Done by removing secret lookup from SKR web-hook chart [here](https://github.com/kyma-project/lifecycle-manager/pull/277/commits/882016e097acb69d423a56408d9adc6a0036c962)
  > Verification steps:
  ```sh
  k get secret skr-webhook-tls -o json | jq -r '.data["ca.crt"]' | base64 --decode > skr-ca-cert.pem && \
  k get validatingwebhookconfigurations.admissionregistration.k8s.io skr-webhook -o json | jq -r '.webhooks[0].clientConfig.caBundle' | base64 --decode > skr-ca-bundle.pem && \
  diff -u skr-ca-cert.pem skr-ca-bundle.pem
  ```
- [x] Make sure web-hook deployment pods are restarted when chart is updated


operator/controllers/kyma_controller.go
```diff
@@ -279,6 +303,15 @@ func (r *KymaReconciler) HandleProcessingState(ctx context.Context, kyma *v1alph
 
 func (r *KymaReconciler) HandleDeletingState(ctx context.Context, kyma *v1alpha1.Kyma) (bool, error) {
 	logger := log.FromContext(ctx)
+	if r.EnableKcpWatcher {
+		if err := r.RemoveWebhookChart(ctx, kyma, r.RemoteClientCache, r.Client); client.IgnoreNotFound(err) != nil {
+			return false, fmt.Errorf(": %w", err)
+		}
+
+		if chartRemoved := r.IsSkrChartRemoved(ctx, kyma, r.RemoteClientCache, r.Client); !chartRemoved {
+			return true, nil
```
ruanxin, line 312: if removed, don't return, otherwise following deletion will not processing, and please add a logger here

khlifi411: Here, I am returning when chart is not removed. I have refactored it to be more obvious 😄 
Sure, I added a debug log before returning 👍

operator/controllers/kyma_controller.go
```diff
@@ -392,3 +426,14 @@ func (r *KymaReconciler) deleteModule(ctx context.Context, moduleStatus *v1alpha
 	manifest.SetName(moduleStatus.Name)
 	return r.Delete(ctx, &manifest, &client.DeleteOptions{})
 }
+
+func (r *KymaReconciler) SetSKRChartManager(skrChartConfig *SkrChartConfig) error {
+	//nolint:goerr113
+	if skrChartConfig == nil {
+		return fmt.Errorf("skr chart config is not set")
```
ruanxin, line 433: this one convert to error.new, not ignore by lint.

operator/controllers/testwithwatcher/kyma_controller_test.go
```diff
@@ -0,0 +1,149 @@
...
+const (
+	namespace                      = "default"
+	webhookConfigNameTpl           = "%s-webhook"
+	servicePathTpl                 = "/validate/%s"
+	specSubresources               = "*"
+	statusSubresources             = "*/status"
+	expectedWebhookNamePartsLength = 4
+)
+
+var _ = Describe("Kyma with multiple module CRs in remote sync mode", Ordered, func() {
```
ruanxin, line 30: I rewrite this test in https://github.com/kyma-project/lifecycle-manager/pull/294/files#diff-d449b17c7b4dc1b913e796f0eb376a9aab8994bee23bc46857fc3c812130b0baR35-R62

ruanxin (commented):
In the Performance testing, dynamically render watcher helm resources is the bottleneck, however, simply caching the manifest will not resolve this problem, and brings new bugs, Followup blocking issues is mentioned here:
https://github.com/kyma-project/module-manager/issues/153
https://github.com/kyma-project/module-manager/issues/151

jakobmoellersap (commented):
Mostly Refactoring Asks and smaller understanding problems. Main thing we need to consider is if we really want

- Separate / Disconnected Flags for Remote Sync and SKR Webhook Deployment
- Webhook deployment in processing state or in the pre-sync or if we want to move the sync context somehow into the processing state.
- Separate "testhelper" and "testwithwatcher" packages and names

operator/main.go
```diff
@@ -299,16 +302,28 @@ func setupKymaReconciler(
 	intervals controllers.RequeueIntervals,
 	options controller.Options,
 ) {
+	fileInfo, err := os.Stat(flagVar.skrWatcherPath)
```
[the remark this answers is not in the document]
khlifi411: Agree. As sync is Kyma CR specific, I have enabled this check only if `flagVar.enableKcpWatcher` is true.

operator/main.go
```diff
@@ -324,17 +339,14 @@ func setupKcpWatcherReconciler(
...
 		setupLog.Error(err, "failed to read local skr chart")
 	}
-	watcherConfig := &controllers.WatcherConfig{
-		WebhookChartPath:       flagVar.skrWatcherPath,
-		SkrWebhookMemoryLimits: flagVar.skrWebhookMemoryLimits,
-		SkrWebhookCPULimits:    flagVar.skrWebhookCPULimits,
-	}
+	// set MaxConcurrentReconciles to 1 to avoid concurrent writes on
+	// the Istio virtual service resource the WatcherReconciler is managing
+	options.MaxConcurrentReconciles = 1
```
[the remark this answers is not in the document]
khlifi411: As the watcher reconcile loop reconciles one virtual service resource (to avoid istio routing side effects when having many virtual services), we have decided to disable concurrent writes for it to avoid conflicts and due to the fact that it reconciles few watcher CRs we deemed this a suitable solution.

operator/internal/deploy/chart_manager.go
```diff
@@ -0,0 +1,186 @@
...
+	modulelib "github.com/kyma-project/module-manager/operator/pkg/manifest"
+	corev1 "k8s.io/api/core/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	k8syaml "sigs.k8s.io/yaml"
+)
+
+type SKRChartManager struct {
```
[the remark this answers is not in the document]
khlifi411: Implemented it [here](https://github.com/kyma-project/lifecycle-manager/pull/277/commits/2b80146fed3a4260477c504161617088af890f2c)

operator/main.go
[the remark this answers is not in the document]
jakobmoellersap: What I do not understand is why you cannot concurrently work with the same virtual service cr over the API. which error would you receive and why would that not be possible concurrently? I think there is a misunderstanding on my side maybe but I think you should be able to work with a single virtual service resource from many loops at once where one will win and others should back of and retry with conflict.

operator/controllers/watcher_controller.go
```diff
@@ -92,124 +83,8 @@ func (r *WatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, r.Update(ctx, watcherObj)
```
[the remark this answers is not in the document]
khlifi411: It was not intended. Fixed it [here](https://github.com/kyma-project/lifecycle-manager/pull/277/commits/2b80146fed3a4260477c504161617088af890f2c)

operator/controllers/kyma_controller.go
```diff
@@ -392,3 +429,13 @@ func (r *KymaReconciler) deleteModule(ctx context.Context, moduleStatus *v1alpha
 	manifest.SetName(moduleStatus.Name)
 	return r.Delete(ctx, &manifest, &client.DeleteOptions{})
 }
+
+func (r *KymaReconciler) SetSKRChartManager(skrChartConfig *SkrChartConfig) error {
```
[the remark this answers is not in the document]
khlifi411: Implemented it [here](https://github.com/kyma-project/lifecycle-manager/pull/277/commits/2b80146fed3a4260477c504161617088af890f2c)

operator/controllers/watcher_controller.go
```diff
@@ -73,17 +62,19 @@ func (r *WatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	logger.Info("Reconciliation loop starting for", "resource", req.NamespacedName.String())
```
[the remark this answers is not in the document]
khlifi411: Fixed it [here](https://github.com/kyma-project/lifecycle-manager/pull/277/commits/2b80146fed3a4260477c504161617088af890f2c)

operator/controllers/watcher_controller.go
```diff
@@ -92,124 +83,8 @@ func (r *WatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-	case v1alpha1.WatcherStateProcessing:
-		obj.AddOrUpdateReadyCondition(v1alpha1.ConditionStatusFalse, msg)
-	case v1alpha1.WatcherStateDeleting:
-		obj.AddOrUpdateReadyCondition(v1alpha1.ConditionStatusFalse, msg)
-	case v1alpha1.WatcherStateError:
-		obj.AddOrUpdateReadyCondition(v1alpha1.ConditionStatusFalse, msg)
-	}
-
-	return r.Status().Update(ctx, obj.SetObservedGeneration())
+	return ctrl.Result{RequeueAfter: r.RequeueIntervals.Failure},
```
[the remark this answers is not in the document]
khlifi411: Fixed it [here](https://github.com/kyma-project/lifecycle-manager/pull/277/commits/2b80146fed3a4260477c504161617088af890f2c)

operator/internal/deploy/webhook_chart.go
```diff
@@ -4,74 +4,61 @@ import (
...
-	servicePathTpl                      = "/validate/%s"
-	webhookNameTpl                      = "%s.operator.kyma-project.io"
-	ReleaseName                         = "skr"
-	specSubresources                    = "*"
-	statusSubresources                  = "*/status"
-	configuredWebhooksDeletionThreshold = 1
-	expectedWebhookNamePartsLength      = 4
-	IstioSytemNs                        = "istio-system"
-	IngressServiceName                  = "istio-ingressgateway"
+	ModeInstall            = Mode("install")
```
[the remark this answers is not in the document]
jakobmoellersap: I think the mode is not necessary as there should be constants (this is present also in declarative lib I think right?) things like istio-system etc should have constants in the istio api. If not i agree we need them.

operator/internal/charts/skr-webhook/templates/deployment.yaml
```diff
@@ -30,6 +30,7 @@ spec:
     metadata:
       labels:
         app: skr-webhook
+        triggerLabel: "{{.Values.triggerLabel}}"
```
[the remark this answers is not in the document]
khlifi411: Fixed it [here](https://github.com/kyma-project/lifecycle-manager/pull/277/commits/2b80146fed3a4260477c504161617088af890f2c)

operator/controllers/watcher_controller.go
```diff
@@ -71,26 +74,35 @@ func (r *WatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-	return ctrl.Result{RequeueAfter: r.RequeueIntervals.Failure},
-		r.UpdateVirtualServiceConfig(ctx, watcherObj)
+	err := r.UpdateVirtualServiceConfig(ctx, watcherObj)
+	if err != nil {
+		return ctrl.Result{RequeueAfter: r.RequeueIntervals.Failure}, err
+	}
+	return ctrl.Result{RequeueAfter: r.RequeueIntervals.Success}, nil
 }
 
 func (r *WatcherReconciler) SetIstioClient() error {
```
jakobmoellersap, line 103: Same as for the Other setter I think you can simply make this entire call in SetupWithManager

operator/internal/deploy/chart_manager.go
```diff
@@ -18,24 +18,26 @@ import (
...
 	webhookChartPath             string
 	skrWebhookMemoryLimits       string
 	skrWebhookCPULimits          string
 	kcpAddr                      string
 	enableWebhookPreInstallCheck bool
 }
 
-func NewSKRChartManager(chartPath, memoryLimits, cpuLimits string, enableWebhookPreInstallCheck bool) *SKRChartManager {
-	return &SKRChartManager{
+func NewSKRWebhookChartManager(chartPath, memoryLimits, cpuLimits string,
```
jakobmoellersap, line 29: I think memoryLimits and cpuLimits are probably not necessary to be put as separate args. Maybe we create a config Struct for this manager?

operator/main.go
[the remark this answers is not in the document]
ruanxin: Regarding this, I think it's better just use 1 worker for watcher loop, because for each individual module operator, we have one dedicate watcher CR, so in total, we probably only have 20 watcher CRs, one worker can sufficiently handle it, and don't have to deal with concurrent write to virtual service although eventually the write will success.

operator/controllers/kyma_controller.go
```diff
@@ -279,6 +306,16 @@ func (r *KymaReconciler) HandleProcessingState(ctx context.Context, kyma *v1alph
 
 func (r *KymaReconciler) HandleDeletingState(ctx context.Context, kyma *v1alpha1.Kyma) (bool, error) {
 	logger := log.FromContext(ctx)
+	if r.EnableKcpWatcher {
```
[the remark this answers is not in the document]
ruanxin: To introduce this flag is purely based on the concern to align with that `enable-kcp-watcher` flag, if the flag is not enabled, then this logic should not be included in kyma_controller. But I think you suggestion to introduce an additional flag in `kyma.Spec.Sync` for control watcher is a better approach, similar as what we did for catalog sync.

operator/internal/deploy/chart_manager.go
```diff
@@ -18,24 +18,26 @@ import (
 	k8syaml "sigs.k8s.io/yaml"
 )
 
-type SKRChartManager struct {
+type SKRWebhookChartManager struct {
```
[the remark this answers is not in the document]
khlifi411: The `internal` package wasn't introduced in this PR. But the intention of using it was to restrict anyone from reusing the `deploy` and `custom` packages. But we can decide to drop it and move the code to the `remote` package under `pkg` as you suggested.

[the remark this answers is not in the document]
jakobmoellersap: I think this is worth a small refactoring issue then, would be great if you could create a follow-up. For the rest we are good here I think :)

operator/main.go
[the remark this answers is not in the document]
jakobmoellersap: Under this argument I agree. Could we document this decision in code so that its clear what the expectation is? otherwise people will wonder why we decided on this.

operator/controllers/setup.go
```diff
@@ -72,6 +71,14 @@ func (r *KymaReconciler) SetupWithManager(mgr ctrl.Manager, options controller.O
 	return nil
 }
 
+func deployChartConfig(skrChartConfig *SkrChartConfig) deploy.ManagerConfig {
```
jakobmoellersap, line 74: This function is not necessary, you just create a struct, so i think you can simply create it in setupWithManager.

jakobmoellersap (commented):
Some minor style nits. Also some job is failing so you may want to have another look here. Also you need to rebase since we had another merge to main from module manager updates.

operator/Makefile
```diff
@@ -57,15 +57,15 @@ generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
 
 .PHONY: vet
 vet: ## Run go vet against code.
-	go vet ./...
+	go vet -tags testing ./...
```
jakobmoellersap, line 60: hmm looking at this, a dedicated build tag also does not seem to be really beautiful. Do you know the best practice here? It would be good to stick to convention here and Ive never seen this style of test imports.

khlifi411: I agree. But after reading this: https://stackoverflow.com/questions/31794141/can-i-create-shared-test-utilities
I think this is the better option

operator/controllers/kyma_controller.go
```diff
@@ -253,7 +272,14 @@ func (r *KymaReconciler) HandleProcessingState(ctx context.Context, kyma *v1alph
 		return r.UpdateStatusWithEventFromErr(ctx, kyma, v1alpha1.StateError,
 			fmt.Errorf("error while syncing conditions during deleting non exists modules: %w", err))
 	}
-
+	statusUpdateRequiredFromSKRWebhookSync := false
+	if r.EnableKcpWatcher {
```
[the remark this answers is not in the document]
khlifi411: I have reused the syncCtx in the deleting part. In the processing logic there is no syncCtx to reuse.

operator/controllers/setup.go
```diff
@@ -24,8 +24,8 @@ import (
 )
 
 // SetupWithManager sets up the Kyma controller with the Manager.
-func (r *KymaReconciler) SetupWithManager(mgr ctrl.Manager,
-	options controller.Options, listenerAddr string,
+func (r *KymaReconciler) SetupWithManager(mgr ctrl.Manager, options controller.Options, listenerAddr string,
+	skrChartConfig *SkrChartConfig,
```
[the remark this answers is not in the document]
khlifi411: Sure 👍 
Let's do it as a follow-up. I have created this [issue](https://github.com/kyma-project/lifecycle-manager/issues/296) for it. Feel free to define the ACs.

operator/internal/deploy/webhook_chart.go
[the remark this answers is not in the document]
khlifi411: The mode will be removed soon as we want to implement separate methods for installing and for removing the chart in a follow-up issue. for the istio consts I looked the packages that we already import and couldn't find any.

operator/internal/deploy/chart_manager.go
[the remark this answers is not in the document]
khlifi411: https://github.com/kyma-project/lifecycle-manager/issues/297

operator/Makefile
jakobmoellersap: Considering this and also some other posts, I would suggest we revert to your original suggestion but with slightly different naming as per recommendation of https://go.dev/doc/go1.4#internalpackages

In this, you should create an "internal/testing" package that cannot be used by other modules and will not be built by the runtime

operator/controllers/kyma_controller.go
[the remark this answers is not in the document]
jakobmoellersap: the syncCtx is created even before going into processing. I think this might need a bigger refactoring since some parts that require the sync context need the kyma sync to be enabled while others do not.

operator/controllers/kyma_controller.go
```diff
@@ -280,11 +297,16 @@ func (r *KymaReconciler) HandleProcessingState(ctx context.Context, kyma *v1alph
 func (r *KymaReconciler) HandleDeletingState(ctx context.Context, kyma *v1alpha1.Kyma) (bool, error) {
 	logger := log.FromContext(ctx)
 
-	if kyma.Spec.Sync.Enabled {
+	if kyma.Spec.Sync.Enabled { //nolint:nestif
```
jakobmoellersap, line 300: please refactor to not have to ignore no-lint

operator/controllers/kyma_controller_remote_sync_test.go
```diff
@@ -47,28 +48,29 @@ var _ = Describe("Kyma with multiple module CRs in remote sync mode", Ordered, f
...
 		By("skr-module-client created in kcp")
-		Eventually(ModuleExists(kyma.GetName(), skrModuleFromClient.Name), Timeout, Interval).Should(BeTrue())
+		Eventually(ModuleExists(kyma.GetName(), skrModuleFromClient.Name),
+			testhelper.Timeout, testhelper.Interval).Should(BeTrue())
 	})
 })
 
 var _ = Describe("Kyma sync into Remote Cluster", Ordered, func() {
-	kyma := NewTestKyma("kyma-test-remote-skr")
+	kyma := testhelper.NewTestKyma("kyma-test-remote-skr")
```
jakobmoellersap, line 68: I think you could use a "." import for testhelper so you dont have tos specify it for every call.

operator/controllers/kyma_controller.go
```diff
@@ -253,7 +263,14 @@ func (r *KymaReconciler) HandleProcessingState(ctx context.Context, kyma *v1alph
 		return r.UpdateStatusWithEventFromErr(ctx, kyma, v1alpha1.StateError,
 			fmt.Errorf("error while syncing conditions during deleting non exists modules: %w", err))
 	}
-
+	statusUpdateRequiredFromSKRWebhookSync := false
+	if r.EnableKcpWatcher && kyma.Spec.Sync.Enabled {
```
jakobmoellersap, line 267: Is there a case where the watcher loop has to run even if the Kyma CR sync is disabled? E.g. for module manager notifications?

khlifi411: We don't want to install the chart for kymas that have sync disabled because the watcher is not relevant for them.

operator/controllers/kyma_controller.go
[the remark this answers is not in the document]
khlifi411: Sure, let's do it in [this](https://github.com/kyma-project/lifecycle-manager/issues/296) follow-up issue.

operator/main.go
```diff
@@ -275,6 +278,8 @@ func defineFlagVar() *FlagVar {
 		"The resources.limits.memory for skr webhook.")
 	flag.StringVar(&flagVar.skrWebhookCPULimits, "skr-webhook-cpu-limits", "0.1",
 		"The resources.limits.cpu for skr webhook.")
+	flag.BoolVar(&flagVar.enableWebhookPreInstallCheck, "enable-webhook-pre-install-check", false,
```
jakobmoellersap, line 281: is this webhook for the skr? if so i think we should rename it to contain skr. Also looking at flag wars, we can maybe simplify to "enable-watcher" since it completely disables all watcher logics.

operator/main.go
```diff
@@ -299,11 +304,29 @@ func setupKymaReconciler(
 	intervals controllers.RequeueIntervals,
 	options controller.Options,
 ) {
+	if flagVar.enableKcpWatcher {
+		fileInfo, err := os.Stat(flagVar.skrWatcherPath)
+		if err != nil || !fileInfo.IsDir() {
+			setupLog.Error(err, "failed to read local skr chart")
```
jakobmoellersap, line 310: this will panic if its a directory since err is nil

operator/main.go
```diff
@@ -299,11 +304,29 @@ func setupKymaReconciler(
 	intervals controllers.RequeueIntervals,
 	options controller.Options,
 ) {
+	if flagVar.enableKcpWatcher {
+		fileInfo, err := os.Stat(flagVar.skrWatcherPath)
```
jakobmoellersap, line 308: This will always fail by default when starting locally since the path of the dockerfile and the path locally do not align. I would prefer if this can be fixed by using the same relative path in the dockerfile so the default happy path does not error out.

To run locally currently I have to run with
```
--enable-kcp-watcher --skr-watcher-path internal/charts/skr-webhook
```

Considering the path here. I do not understand why its in internal? Isnt this a pure go package?

jakobmoellersap (commented):
I noticed that in the `watcher` kustomize target, there is a patch saying 

```yaml
- op: add
  value: --enable-kcp-watcher=false
```

this is probably wrong since the watcher kustomization should enable the watcher. Not sure how you currently test it but this is not working out of the box I think...

operator/internal/deploy/chart_manager.go
```diff
@@ -0,0 +1,222 @@
...
+		customConfigKey:         string(bytes),
+	}, nil
+}
+
+func (m *EnabledSKRWebhookChartManager) resolveKcpAddr(ctx context.Context, kcpClient client.Client) (string, error) {
+	if m.kcpAddr != "" {
+		return m.kcpAddr, nil
+	}
+	// Get external IP from the ISTIO load balancer external IP
+	loadBalancerService := &corev1.Service{}
```
jakobmoellersap, line 156: This implementation is too opinionated on load balancer only. I think we should at least offer node-port to allow implementation locally.

operator/internal/deploy/chart_manager.go
```diff
@@ -0,0 +1,222 @@
...
+	bytes, err := k8syaml.Marshal(chartCfg)
+	if err != nil {
+		return nil, err
+	}
+
+	kcpAddr, err := m.resolveKcpAddr(ctx, kcpClient)
+	if err != nil {
+		return nil, err
+	}
+	return map[string]interface{}{
```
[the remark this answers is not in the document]
jakobmoellersap: This is because stringifiedManifest is not cached based on `SetFlags`, bug in ModuleManager. I think this is a known cache limitation @adityabhatia but we should still make the cache optional so that people can disable it if they notice bugs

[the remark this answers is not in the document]
jakobmoellersap: @adityabhatia FYI

operator/internal/deploy/chart_manager.go
```diff
@@ -0,0 +1,222 @@
...
+			return fmt.Errorf("failed to uninstall webhook config: %w", err)
+		}
+		if !uninstalled {
+			return ErrSKRWebhookWasNotRemoved
+		}
+
+		return nil
+	}
+	if enableWebhookPreInstallCheck {
+		// TODO(khlifi411): verify webhook configuration with watchers' configuration before re-installing the chart
```
jakobmoellersap, line 204: Resolve TODO

operator/internal/deploy/chart_manager.go
```diff
@@ -0,0 +1,222 @@
...
+	skrCfg, err := remote.GetRemoteRestConfig(ctx, kcpClient, client.ObjectKeyFromObject(kyma),
+		kyma.Spec.Sync.Strategy)
+	if err != nil {
+		return true, err
+	}
+	argsVals, err := m.generateHelmChartArgs(ctx, kcpClient)
+	if err != nil {
+		return true, err
+	}
+	// TODO(khlifi411): make sure that validating-webhook-config resource is in sync with the secret configuration
```
jakobmoellersap, line 95: Resolve TODO

operator/internal/deploy/chart_manager.go
```diff
@@ -0,0 +1,222 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	k8syaml "sigs.k8s.io/yaml"
+)
+
+var ErrExpectedExactlyOneSKRConfig = errors.New("expected exactly one SKR config")
+
+type SKRWebhookChartManager interface {
+	InstallWebhookChart(ctx context.Context, kyma *v1alpha1.Kyma,
```
jakobmoellersap, line 25: instead of passing the Sync Context, you could pass it into the struct during creation and not have to pass it as function arg in the interface.
Same goes for the cache

operator/internal/deploy/chart_manager.go
```diff
@@ -0,0 +1,222 @@
...
+	}, nil
+}
+
+func (m *EnabledSKRWebhookChartManager) resolveKcpAddr(ctx context.Context, kcpClient client.Client) (string, error) {
+	if m.kcpAddr != "" {
+		return m.kcpAddr, nil
+	}
+	// Get external IP from the ISTIO load balancer external IP
+	loadBalancerService := &corev1.Service{}
+	if err := kcpClient.Get(ctx, client.ObjectKey{Name: IngressServiceName, Namespace: IstioSytemNs},
```
jakobmoellersap, line 157: This get call is not using the informer cache and is thus extremely slow. You should add a watches on services to handle deletion and error correctly for Watcher CR. In general you are missing a state on Watcher CR if the Virtual Service has been created/updated correctly

operator/internal/deploy/chart_manager.go
```diff
@@ -0,0 +1,222 @@
...
+		ready, err := modulelib.ConsistencyCheck(&logger, deployInfo, nil, nil)
+		if err != nil {
+			return fmt.Errorf("failed to verify webhook resources: %w", err)
+		}
+		if ready {
+			logger.V(1).Info("chart resources already installed, nothing to do!")
+			return nil
+		}
+	}
+	installed, err := modulelib.InstallChart(&logger, deployInfo, nil, nil)
```
jakobmoellersap, line 214: You are not passing a chart cache here and this will cause performance issues during rendering

operator/main.go
khlifi411: This flag was used to toggle a check done before installing the chart to prevent overriding it if the same chart resources are installed already when the `enable-watcher` flag is true but I think we do not need it anymore. I am going to remove it 👍

operator/main.go
khlifi411: I have migrated the skr chart from the internal folder and aligned the relative path in the dockerfile to be the same as the default `--skr-watcher-path` flag value.

operator/internal/deploy/chart_manager.go
khlifi411: As discussed. This is a bigger refactoring which will be done [here](https://github.com/kyma-project/lifecycle-manager/issues/301)

operator/internal/deploy/chart_manager.go
khlifi411: As discussed. This is a bigger refactoring which will be done [here](https://github.com/kyma-project/lifecycle-manager/issues/301)

operator/internal/deploy/chart_manager.go
khlifi411: As discussed. This is a bigger refactoring which will be done [here](https://github.com/kyma-project/lifecycle-manager/issues/302)

operator/internal/deploy/chart_manager.go
khlifi411: Agree! This should be fixed when chart cache becomes optional, right?

operator/internal/deploy/chart_manager.go
khlifi411: As discussed. This is a bigger refactoring which will be done here:
#303

operator/internal/deploy/chart_manager.go
jakobmoellersap: Not sure, I think chart cache should always be used. Please correct me here if wrong @adityabhatia

operator/internal/deploy/chart_manager.go
jakobmoellersap: The Service Lookup is not part of a bigger factory and can be fixed by doing a watch with a label IMO. Please share your opinion @ruanxin @adityabhatia

operator/internal/deploy/chart_manager.go
jakobmoellersap: Why is this part of bigger refactoring? @adityabhatia @ruanxin

operator/internal/deploy/chart_manager.go
jakobmoellersap: Why is this part of bigger refactoring? @adityabhatia @ruanxin

operator/internal/deploy/chart_manager.go
jakobmoellersap: THIS is part of the bigger refactoring :D

operator/internal/deploy/chart_manager.go
khlifi411: As discussed, I will double check if the service is included in the cache and will add it if it's not 👍

khlifi411: I have checked and it is added [here](https://github.com/kyma-project/lifecycle-manager/blob/d8f59928a280ac807b3c8b3b6f17df38c7461252/operator/controllers/util.go#L18-L20)

operator/internal/deploy/chart_manager.go
khlifi411: I tried to pass it on creation and realised that it needs further refactoring.
Let's do it in [this](https://github.com/kyma-project/lifecycle-manager/issues/296) backlog item. 👍

jakobmoellersap (changes requested):
Please Rebase

jakobmoellersap (changes requested):
Please Rebase and fix tests

jakobmoellersap (approved):

PR #291, opened 2022-10-29
upgrade: module-manager@main and adjust implementation

adityabhatia:


operator/internal/deploy/webhook_config_test.go
```diff
@@ -163,3 +170,55 @@ func createKymaCR(kymaName string) *v1alpha1.Kyma {
 		},
 	}
 }
+
+func isChartRemoved(ctx context.Context, k8sClient client.Client) func() bool {
```
jakobmoellersap, line 174: Why does this test helper exist twice?

jakobmoellersap (commented):
Test and lint failure, would be cool if you could double check

operator/controllers/kyma_controller_remote_sync_test.go
```diff
@@ -45,19 +47,20 @@ var _ = Describe("Kyma with multiple module CRs in remote sync mode", Ordered, f
...
+			testutils.Interval).Should(Succeed())
 		remoteKyma, err := GetKyma(runtimeClient, kyma.GetName())
 		Expect(err).ShouldNot(HaveOccurred())
 
 		By("add skr-module-client to remoteKyma.spec.modules")
 		remoteKyma.Spec.Modules = []v1alpha1.Module{
 			*skrModuleFromClient,
 		}
-		Eventually(runtimeClient.Update(ctx, remoteKyma), timeout, interval).Should(Succeed())
+		Eventually(runtimeClient.Update(ctx, remoteKyma), testutils.Timeout, testutils.Interval).Should(Succeed())
```
jakobmoellersap, line 59: maybe we should import with `. "github.com/kyma-project/lifecycle-manager/operator/internal/testutils"` so that we dont have to use the package name everytime? What do you think? Then you dont need to adjust all tests here.

adityabhatia: The tests will anyways be adjusted by switching to an exported variable from test utils.

jakobmoellersap (approved):

Issue #301, opened 2022-11-07
Refactor SKR watcher installation to use native Kubernetes client


khlifi411:
**Description**
Using the native Kubernetes client to install the SKR watcher should be implemented and relying on the manifest library should be removed.

**ACs**

- [x] Remove dependency between SKR chart installation and the manifest library and use native Kubernetes client instead.
- [x] Use trigger label with `secretResourceVersion` value to ensure pods restart when chart is redeployed.
- [x] Use sync context on Kyma reconciler setup and pass it to `ChartManager` method calls as needed
- [x] The SKR web-hook's installation should leverage SSA to be idempotent.

And ensure that:
- [x] the `caBundle` configured for each deployed `webhook` configuration is the same as the `ca.crt` field of the `tls-secret`

**Related discussions**

https://github.com/kyma-project/lifecycle-manager/pull/277#discussion_r1014820030
https://github.com/kyma-project/lifecycle-manager/pull/277#discussion_r1014819992
https://github.com/kyma-project/lifecycle-manager/pull/277#discussion_r1011751082

**PR**
#313

jakobmoellersap:
Please fill TBD fields @khlifi411

PR #313, opened 2022-11-15
Refactor SKR watcher installation to use native Kubernetes client

khlifi411:
implementation of #301


adityabhatia:
Updated module-manager lib with the latest commit

controllers/kyma_controller.go
```diff
@@ -202,23 +203,25 @@ func (r *KymaReconciler) syncModuleCatalog(ctx context.Context, syncContext *rem
...
+func (r *KymaReconciler) stateHandling(ctx context.Context, kyma *v1alpha1.Kyma,
+	syncContext *remote.KymaSynchronizationContext,
+) (ctrl.Result, error) {
 	log.FromContext(ctx).Info("syncing state", "state", string(kyma.Status.State))
 	switch kyma.Status.State {
 	case "":
 		return ctrl.Result{}, r.HandleInitialState(ctx, kyma)
 	case v1alpha1.StateProcessing:
-		return ctrl.Result{Requeue: true}, r.HandleProcessingState(ctx, kyma)
+		return ctrl.Result{Requeue: true}, r.HandleProcessingState(ctx, kyma, syncContext)
```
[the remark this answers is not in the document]
adityabhatia: The `ctx` is set during controller start, so that is not a good option, even if use unique kyma keys as keys for `context.WithValue(...)`

pkg/deploy/chart_manager.go
```diff
@@ -6,26 +6,28 @@ import (
...
+
 var ErrExpectedExactlyOneSKRConfig = errors.New("expected exactly one SKR config")
 
 type SKRWebhookChartManager interface {
-	InstallWebhookChart(ctx context.Context, kyma *v1alpha1.Kyma,
-		remoteClientCache *remote.ClientCache, kcpClient client.Client) (bool, error)
-	RemoveWebhookChart(ctx context.Context, kyma *v1alpha1.Kyma,
-		syncCtx *remote.KymaSynchronizationContext) error
+	InstallWebhookChart(ctx context.Context, syncContext *remote.KymaSynchronizationContext) (bool, error)
+	RemoveWebhookChart(ctx context.Context, syncContext *remote.KymaSynchronizationContext) error
```
adityabhatia, line 30: the file is called `chart_manager.go`, with a `SKRWebhookChartManager`, with method names `InstallWebhookChart`, too much duplicacy. Could we come up with better names?

pkg/deploy/chart_manager.go
```diff
@@ -74,81 +76,97 @@ type SkrChartConfig struct {
...
-func (m *EnabledSKRWebhookChartManager) InstallWebhookChart(ctx context.Context, kyma *v1alpha1.Kyma,
-	remoteClientCache *remote.ClientCache, kcpClient client.Client,
+func (m *EnabledSKRWebhookChartManager) InstallWebhookChart(ctx context.Context,
+	syncContext *remote.KymaSynchronizationContext,
 ) (bool, error) {
-	skrClient, err := remote.NewRemoteClient(ctx, kcpClient, client.ObjectKeyFromObject(kyma),
-		kyma.Spec.Sync.Strategy, remoteClientCache)
+	logger := logf.FromContext(ctx)
+	kymaObjKey := client.ObjectKeyFromObject(syncContext.ControlPlaneKyma)
+	skrCfg, err := remote.GetRemoteRestConfig(ctx, syncContext.ControlPlaneClient, kymaObjKey,
```
adityabhatia, line 89: can we store the `rest.Config` also so the cache along with `client.Client`, since they belong together and we do not have to do a lookup for the secret everytime

skr-webhook/templates/deployment.yaml
```diff
@@ -30,7 +30,7 @@ spec:
     metadata:
       labels:
         app: skr-webhook
-        operator.kyma-project.io/pod-restart-trigger: "{{.Values.triggerLabel}}"
+        operator.kyma-project.io/pod-restart-trigger: {{uuidv4}}
```
adityabhatia, line 33: `pod-restart-trigger ` this label we do not need anymore, as we discussed the parsed Manifest will be cached. Could you please remove the trigger label logic?

skr-webhook/templates/rbac.yaml
```diff
@@ -11,20 +11,20 @@ metadata:
   namespace: {{.Release.Namespace}}
   name: kyma-reader
 rules:
-- apiGroups: ["operator.kyma-project.io"] # "" indicates the core API group
-  resources: ["*"]
-  verbs: ["get", "watch", "list"]
+  - apiGroups: [ "operator.kyma-project.io" ] # "" indicates the core API group
+    resources: [ "*" ]
+    verbs: [ "get", "watch", "list" ]
```
adityabhatia, line 16: this formatting was fine before

skr-webhook/templates/rbac.yaml
```diff
@@ -11,20 +11,20 @@ metadata:
...
   namespace: {{.Release.Namespace}}
 subjects:
-# You can specify more than one "subject"
-- kind: ServiceAccount
-  name: {{.Release.Name}}-webhook-sa
-  namespace: {{.Release.Namespace}}
+  # You can specify more than one "subject"
+  - kind: ServiceAccount
+    name: {{.Release.Name}}-webhook-sa
+    namespace: {{.Release.Namespace}}
```
adityabhatia, line 27: this formatting was fine before

skr-webhook/values.yaml
```diff
@@ -2,15 +2,13 @@
...
-image: 
+image:
   # -- Image repository
-  repository: "eu.gcr.io/kyma-project"
+  repository: "europe-docker.pkg.dev/kyma-project/dev"
   # -- Image name
   name: "runtime-watcher-skr"
   # -- Image tag
-  tag: "latest"
+  tag: "PR-80"
```
adityabhatia, line 11: please revert to latest

pkg/deploy/webhook_chart.go
```diff
@@ -1,61 +1,122 @@
...
+	baseRes := &unstructured.Unstructured{}
+	baseRes.SetLabels(map[string]string{
+		modulelabels.CacheKey: kymaObjKey.String(),
+	})
+	baseRes.SetNamespace(metav1.NamespaceDefault)
+	baseRes.SetName(staticWatcherConfigName)
+	return baseRes
+}
+
+func CheckWebhookCABundleConsistency(ctx context.Context, skrClient client.Client, kymaObjKey client.ObjectKey) error {
```
adityabhatia, line 99: As far as I see `CheckWebhookCABundleConsistency` is only used in tests. Then why is it part of this file?

pkg/deploy/webhook_chart.go
```diff
@@ -1,61 +1,122 @@
...
-	DeploymentNameTpl      = "%s-webhook"
-	triggerLabelTimeFormat = "200601021504050700"
+	customConfigKey         = "modules"
+	IstioSytemNs            = "istio-system"
+	IngressServiceName      = "istio-ingressgateway"
+	DeploymentNameTpl       = "%s-webhook"
+	releaseNameTpl          = "%s-%s-skr"
+	staticWatcherConfigName = "static-watcher-config-name"
+	WebhookConfigNameTpl    = "%s-webhook"
+	SecretNameTpl           = "%s-webhook-tls" //nolint:gosec
```
adityabhatia, line 32: remove lint, by renaming `SecretNameTpl`

adityabhatia, line 32: `%s-webhook` is used 3 times. Could we re-use?

controllers/withwatcher/kyma_controller_test.go
```diff
@@ -21,90 +27,212 @@ import (
...
+	ErrWebhookNamePartsNumberMismatch  = errors.New("webhook name dot separated parts number mismatch")
+	ErrManagedByLabelNotFound          = errors.New("managed-by label not found")
+	ErrModuleNameMismatch              = errors.New("module name mismatch")
+	ErrSvcPathMismatch                 = errors.New("service path mismatch")
+	ErrWatchLabelsMismatch             = errors.New("watch labels mismatch")
+	ErrStatusSubResourcesMismatch      = errors.New("status sub-resources mismatch")
+	ErrSpecSubResourcesMismatch        = errors.New("spec sub-resources mismatch")
+)
 
+var _ = Describe("Kyma with multiple module CRs in remote sync mode", Ordered, Serial, func() {
```
adityabhatia, line 49: Ordered - I thought we decided to get rid of them?
Is setting Serial by mistake?

controllers/withwatcher/watcher_controller_test.go
```diff
@@ -48,37 +48,26 @@ func allCRsDeleted() func(customIstioClient *istio.Client) {
...
 		}
 		// verify
-		Eventually(isCrDeletionFinished(), Timeout, Interval).Should(BeTrue())
+		Eventually(isWatcherCrDeletionFinished(), Timeout, Interval).Should(BeTrue())
 		Eventually(isVsRemoved(suiteCtx, customIstioClient)).Should(BeTrue())
 	}
 }
 
-var _ = Describe("Watcher CR scenarios", Ordered, func() {
+var _ = Describe("Watcher CR scenarios", Ordered, Serial, func() {
```
adityabhatia, line 56: Ordered - I thought we decided to get rid of them?
Is setting Serial by mistake?

controllers/withwatcher/kyma_controller_test.go
```diff
@@ -21,90 +27,212 @@ import (
...
+			Namespace: metav1.NamespaceDefault,
+			Name:      deploy.ResolveSKRChartResourceName(deploy.WebhookConfigNameTpl, kymaObjKey),
+		}, webhookConfig), Timeout, Interval).Should(Succeed())
+		Expect(isWebhookConfigured(watcherCrForKyma, webhookConfig, kyma.Name)).To(Succeed())
+	})
+
+	It("kyma reconciler re-installs watcher helm chart when webhook CA bundle is not consistent", func() {
+		By("updating webhook config with corrupt CA bundle data")
+		Expect(webhookConfig.Webhooks).NotTo(BeEmpty())
+		webhookConfig.Webhooks[0].ClientConfig.CABundle = []byte(dummyCaBundleData)
```
adityabhatia, line 77: are you updating this `dummyCaBundleData`. As far as I see it's only set on the variable and not in the testenv cluster?

controllers/withwatcher/kyma_controller_test.go
```diff
@@ -21,90 +27,212 @@ import (
...
+		By("updating kyma channel to trigger its reconciliation")
+		kyma.Spec.Channel = v1alpha1.ChannelRapid
+		Expect(controlPlaneClient.Update(suiteCtx, kyma)).To(Succeed())
+		By("waiting for the kyma update event to be processed")
+		time.Sleep(2 * time.Second)
```
adityabhatia, line 82: this should not be required

controllers/withwatcher/kyma_controller_test.go
```diff
@@ -21,90 +27,212 @@ import (
...
+		watcherCrForKyma.Spec.LabelsToWatch[labelKey] = labelValue
+		Expect(controlPlaneClient.Update(suiteCtx, watcherCrForKyma)).To(Succeed())
+		By("waiting for watcher CR labelsToWatch to be updated")
+		Eventually(isWatcherCrLabelUpdated(client.ObjectKeyFromObject(watcherCrForKyma),
+			labelKey, labelValue), Timeout, Interval).Should(BeTrue())
+		By("updating kyma channel to trigger its reconciliation")
+		kyma.Spec.Channel = v1alpha1.ChannelFast
+		Expect(controlPlaneClient.Update(suiteCtx, kyma)).To(Succeed())
+		By("waiting for the kyma update event to be processed")
+		time.Sleep(2 * time.Second)
```
adityabhatia, line 101: could we please get rid of `time.Sleep`?

controllers/withwatcher/kyma_controller_test.go
```diff
@@ -21,90 +27,212 @@ import (
...
+			Name:      deploy.ResolveSKRChartResourceName(deploy.WebhookConfigNameTpl, kymaObjKey),
+		}, webhookConfig)).To(Succeed())
+		Expect(isWebhookConfigured(watcherCrForKyma, webhookConfig, kyma.Name)).To(Succeed())
+
+		caBundleValueBeforeUpdate := webhookConfig.Webhooks[0].ClientConfig.CABundle
+		By("updating kyma channel to trigger its reconciliation")
+		kyma.Spec.Channel = v1alpha1.ChannelRapid
+		Expect(controlPlaneClient.Update(suiteCtx, kyma)).To(Succeed())
+		By("waiting for the kyma update event to be processed")
+		time.Sleep(2 * time.Second)
```
adityabhatia, line 116: could we remove `time.Sleep`?

pkg/deploy/chart_manager.go
```diff
@@ -6,26 +6,28 @@ import (
...
+const caCertificateSecretKey = "ca.crt"
+
 var ErrExpectedExactlyOneSKRConfig = errors.New("expected exactly one SKR config")
 
 type SKRWebhookChartManager interface {
-	InstallWebhookChart(ctx context.Context, kyma *v1alpha1.Kyma,
-		remoteClientCache *remote.ClientCache, kcpClient client.Client) (bool, error)
-	RemoveWebhookChart(ctx context.Context, kyma *v1alpha1.Kyma,
-		syncCtx *remote.KymaSynchronizationContext) error
+	InstallWebhookChart(ctx context.Context, syncContext *remote.KymaSynchronizationContext) (bool, error)
```
adityabhatia, line 29: I see that InstallWebhookChart is always called both during Processing and Ready States. I some webhook resource is missing during `Ready` State, it force applies. Could you write a test for this use case? I have my doubts it works..

controllers/kyma_controller.go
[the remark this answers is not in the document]
jakobmoellersap: Just because you use ctx during controller start does not mean the values set during the control-loop propagate upstream or cross boundaries.

If you call `context.WithValue(ctx, key, value)` in `Reconcile`, it will only pass the object in the context used within one reconciliation, thus it is safe to use even with a single key, there will be no references after Reconcile is over, and it will also be GCed normally.

pkg/deploy/chart_manager.go
jakobmoellersap: Looking at the comment above: Force apply should never be a regular case imo and we should only use 3-way merge. Is that happening / possible right now?

pkg/deploy/chart_manager.go
```diff
@@ -6,26 +6,28 @@ import (
...
+	RemoveWebhookChart(ctx context.Context, syncContext *remote.KymaSynchronizationContext) error
 }
 
 type DisabledSKRWebhookChartManager struct{}
```
jakobmoellersap, line 33: Imo this implementation should change to not use a `Disabled` struct. It should not be created in the first place. This kind of design reminds me a lot of Java interface implementations...

controllers/kyma_controller.go
```diff
@@ -286,15 +289,13 @@ func (r *KymaReconciler) HandleProcessingState(ctx context.Context, kyma *v1alph
...
 	logger := log.FromContext(ctx)
 
-	if kyma.Spec.Sync.Enabled {
-		syncContext, err := remote.InitializeKymaSynchronizationContext(ctx, kyma, r.Client, r.RemoteClientCache)
-		if err != nil {
-			return false, fmt.Errorf("remote sync initialization failed: %w", err)
-		}
-		if err = r.RemoveWebhookChart(ctx, kyma, syncContext); err != nil {
+	if kyma.Spec.Sync.Enabled && syncContext != nil {
+		if err := r.RemoveWebhookChart(ctx, syncContext); err != nil {
```
jakobmoellersap, line 298: why has remove web hook chart no dedication switch between recoverable and non-recoverable errors? e.g. a chart installation is not something that can necessarily be recovered simply by waiting (and returning true, nil as you do)

pkg/deploy/webhook_chart.go
```diff
@@ -1,61 +1,122 @@
...
-	ModeUninstall          = Mode("uninstall")
-	customConfigKey        = "modules"
-	ReleaseName            = "skr"
-	IstioSytemNs           = "istio-system"
-	IngressServiceName     = "istio-ingressgateway"
-	DeploymentNameTpl      = "%s-webhook"
-	triggerLabelTimeFormat = "200601021504050700"
+	customConfigKey         = "modules"
+	IstioSytemNs            = "istio-system"
+	IngressServiceName      = "istio-ingressgateway"
```
jakobmoellersap, line 27: I still think this should be configurable in the controller via flag, same as the namespace. An Istio-installation can look different than this and it heavily opinionated

skr-webhook/templates/rbac.yaml
```diff
@@ -11,20 +11,20 @@ metadata:
   namespace: {{.Release.Namespace}}
   name: kyma-reader
 rules:
-- apiGroups: ["operator.kyma-project.io"] # "" indicates the core API group
-  resources: ["*"]
-  verbs: ["get", "watch", "list"]
+  - apiGroups: [ "operator.kyma-project.io" ] # "" indicates the core API group
+    resources: [ "*" ]
```
jakobmoellersap, line 15: The star import is not strict enough, why is this here?

skr-webhook/values.yaml
```diff
@@ -2,15 +2,13 @@
 sidecar: true
 # -- Replicas
 replicas: 1
-# -- Trigger pod restart
-triggerLabel: ""
-image: 
+image:
   # -- Image repository
-  repository: "eu.gcr.io/kyma-project"
+  repository: "europe-docker.pkg.dev/kyma-project/dev"
```
jakobmoellersap, line 7: please also revert this to the proper registry

controllers/watcher_controller.go
```diff
@@ -66,7 +66,7 @@ func (r *WatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 	watcherObj := &v1alpha1.Watcher{}
 	if err := r.Get(ctx, client.ObjectKey{Name: req.Name, Namespace: req.Namespace}, watcherObj); err != nil {
-		logger.Error(err, "Failed to get reconciliation object")
+		logger.V(1).Info("Failed to get reconciliation object")
```
jakobmoellersap, line 69: Why is V1 hardcoded here?

pkg/deploy/chart_manager.go
```diff
@@ -74,81 +76,97 @@ type SkrChartConfig struct {
...
-	if err != nil {
-		return true, err
+	if !installed {
+		return true, ErrSKRWebhookNotInstalled
 	}
-	kyma.UpdateCondition(v1alpha1.ConditionReasonSKRWebhookIsReady, metav1.ConditionTrue)
+	syncContext.ControlPlaneKyma.UpdateCondition(v1alpha1.ConditionReasonSKRWebhookIsReady, metav1.ConditionTrue)
+	logger.Info("successfully installed webhook chart",
+		"release-name", skrWatcherInstallInfo.ChartInfo.ReleaseName)
+	logger.V(1).Info("following modules were installed",
```
jakobmoellersap, line 110: again, why is V1 hardcoded?

skr-webhook/templates/deployment.yaml
khlifi411: I think we still need it. when the set flags' values change, only the `webhook` config and secret will change and the pods will not be restarted. unless this label value changes.

controllers/withwatcher/kyma_controller_test.go
khlifi411: I am not sure we can. I tried different post checks without success. what do you suggest?

controllers/kyma_controller.go
jakobmoellersap: agree on this, no way for us here.

controllers/kyma_controller.go
```diff
@@ -231,48 +232,20 @@ func (r *KymaReconciler) HandleInitialState(ctx context.Context, kyma *v1alpha1.
...
-	modules, err = r.GenerateModulesFromTemplate(ctx, kyma)
-	if err != nil {
-		return r.UpdateStatusWithEventFromErr(ctx, kyma, v1alpha1.StateError,
-			fmt.Errorf("error while fetching modules during processing: %w", err))
-	}
-
-	runner := sync.New(r)
-
-	statusUpdateRequiredFromModuleSync, err := runner.Sync(ctx, kyma, modules)
+	statusUpdateRequiredFromModulesSync, err := r.syncModules(ctx, kyma)
```
adityabhatia, line 235: any particular reason to move this out?

pkg/deploy/template_chart_manager.go
```diff
@@ -0,0 +1,196 @@
...
+		oldResource.SetGroupVersionKind(resource.GroupVersionKind())
+		err := syncContext.RuntimeClient.Get(ctx, resourceObjKey, oldResource)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return true, fmt.Errorf("failed to get webhook %s: %w", resource.GetKind(), err)
+		}
+		if apierrors.IsNotFound(err) {
+			if err := syncContext.RuntimeClient.Create(ctx, resource, defaultFieldOwner); err != nil {
+				return true, fmt.Errorf("failed to create webhook %s: %w", resource.GetKind(), err)
+			}
+		}
```
adityabhatia, line 126: you can directly use `syncContext.RuntimeClient.Create`. If the resource exists it will anyways result in a conflict and report an error

controllers/kyma_controller.go
khlifi411: I moved this part to a separate method because of the cyclomatic complexity linting error. see https://status.build.kyma-project.io/view/gs/kyma-prow-logs/pr-logs/pull/kyma-project_lifecycle-manager/313/pull-lifecycle-mgr-lint/1613491447321858048

pkg/deploy/template_chart_manager.go
```diff
@@ -0,0 +1,196 @@
...
+	for _, resource := range resources {
+		resourceObjKey := client.ObjectKeyFromObject(resource)
+		oldResource := &unstructured.Unstructured{}
+		oldResource.SetGroupVersionKind(resource.GroupVersionKind())
+		err := syncContext.RuntimeClient.Get(ctx, resourceObjKey, oldResource)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return true, fmt.Errorf("failed to get webhook %s: %w", resource.GetKind(), err)
+		}
+		if apierrors.IsNotFound(err) {
+			if err := syncContext.RuntimeClient.Create(ctx, resource, defaultFieldOwner); err != nil {
```
adityabhatia, line 123: why are you passing a defaultFieldOwner? 
You're using client-side apply here.

pkg/deploy/template_chart_manager.go
```diff
@@ -0,0 +1,196 @@
...
+				return true, fmt.Errorf("failed to create webhook %s: %w", resource.GetKind(), err)
+			}
+		}
+		// completely replace old resource with new resource
```
adityabhatia, line 127: considering that the resource was freshly created, why is it being updated right after. Somehow this doesn't make sense.

pkg/deploy/template_chart_manager.go
```diff
@@ -0,0 +1,196 @@
...
+		oldResource := &unstructured.Unstructured{}
+		oldResource.SetGroupVersionKind(resource.GroupVersionKind())
+		err := syncContext.RuntimeClient.Get(ctx, resourceObjKey, oldResource)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get webhook %s: %w", resource.GetKind(), err)
+		}
+		if err == nil {
+			if err := syncContext.RuntimeClient.Delete(ctx, resource); err != nil {
+				return fmt.Errorf("failed to delete webhook %s: %w", resource.GetKind(), err)
+			}
```
adityabhatia, line 171: you can directly delete the resource without calling `Get()`

pkg/deploy/template_chart_manager.go
```diff
@@ -0,0 +1,196 @@
...
+
+func NewSKRWebhookTemplateChartManager(kcpRestConfig *rest.Config, config *SkrChartManagerConfig,
+) (*SKRWebhookTemplateChartManager, error) {
+	resolvedKcpAddr, err := resolveKcpAddr(kcpRestConfig, config)
+	if err != nil {
+		return nil, err
+	}
+	return &SKRWebhookTemplateChartManager{
+		config:  config,
+		cache:   &sync.Map{},
```
adityabhatia, line 57: can we rename `cache` to something like `configCache`

pkg/deploy/template_chart_manager.go
```diff
@@ -0,0 +1,196 @@
...
+		Values:      chartArgValues,
+	})
+}
+
+func (m *SKRWebhookTemplateChartManager) cachedConfig(kymaObjKey client.ObjectKey,
+	chartArgValues map[string]interface{},
+) bool {
+	cachedArgValues, ok := m.cache.Load(kymaObjKey)
+	if !ok {
+		m.cache.Store(kymaObjKey, chartArgValues)
```
adityabhatia, line 83: why are you string chartConfig by kyma name? AFAIK, all kymas will have the same config. Shouldn't we simplify this just store config?

adityabhatia (changes requested):
Looks good overall, apart from the comments, inline comments are missing.

Makefile
```diff
@@ -185,6 +185,10 @@ lister-gen: ## Download lister-gen
 fmt: ## Run go fmt against code.
 	go fmt ./...
 
+.PHONY: resolve
+resolve: ## Run go mod tidy against code.
```
jakobmoellersap, line 189: could we try to get rid of these resolves down the line? I would like to avoid having to pin golang versions like we had to for the old reconciler

go.mod
```diff
@@ -7,11 +7,11 @@ require (
 	github.com/gardener/component-spec/bindings-go v0.0.70
 	github.com/go-logr/logr v1.2.3
 	github.com/go-logr/zapr v1.2.3
-	github.com/kyma-project/module-manager v0.0.0-20221207164018-ddf69229acb6
-	github.com/kyma-project/module-manager/operator v0.0.0-20221125150159-5e84c4855249
+	github.com/kyma-project/module-manager v0.0.0-20221214222100-348f6d59ed0b
 	github.com/kyma-project/runtime-watcher/listener v0.0.0-20221129135108-60e6093eba9d
 	github.com/onsi/ginkgo/v2 v2.5.0
 	github.com/onsi/gomega v1.24.1
+	github.com/slok/go-helm-template v0.3.0
```
jakobmoellersap, line 14: this still includes the helm libraries, did we not agree to get rid of the entire chart?

jakobmoellersap (commented):
just wondering in general why we don't entirely remove helm from the dependency list. If you want to do that in a new step, also fine by me

go.mod
khlifi411: Yes. But we will do it in a separate ticket: #346

pkg/deploy/template_chart_manager.go
```diff
@@ -0,0 +1,146 @@
...
+		resourceObjKey := client.ObjectKeyFromObject(resource)
+		oldResource := &unstructured.Unstructured{}
+		oldResource.SetGroupVersionKind(resource.GroupVersionKind())
+		if err := skrClient.Get(ctx, resourceObjKey, oldResource); err != nil {
+			return fmt.Errorf("failed to get webhook %s: %w", resource.GetKind(), err)
+		}
+		resource.SetResourceVersion(oldResource.GetResourceVersion())
+		if err := skrClient.Update(ctx, resource); err != nil {
+			return fmt.Errorf("failed to replace webhook %s: %w", resource.GetKind(), err)
+		}
```
adityabhatia, line 142: maybe you can simply use patch here.

pkg/deploy/template_chart_manager.go
```diff
@@ -0,0 +1,146 @@
...
+		return fmt.Errorf("failed to create webhook %s: %w", resource.GetKind(), err)
+	}
+	if resourceAlreadyExists {
+		resourceObjKey := client.ObjectKeyFromObject(resource)
+		oldResource := &unstructured.Unstructured{}
+		oldResource.SetGroupVersionKind(resource.GroupVersionKind())
+		if err := skrClient.Get(ctx, resourceObjKey, oldResource); err != nil {
+			return fmt.Errorf("failed to get webhook %s: %w", resource.GetKind(), err)
+		}
+		resource.SetResourceVersion(oldResource.GetResourceVersion())
```
adityabhatia, line 139: why setting this explicitly?

khlifi411: In CSA, without the resource version field set, the `Update()` call fails. As agreed already in the daily, this implementation will be replaced with the SSA implementation.

pkg/deploy/template_chart_manager.go
khlifi411: As already discussed in the daily, SSA will replace this implementation.

jakobmoellersap (commented):
LGTM, good stuff! waiting for other reviews

pkg/deploy/template_chart_manager.go
```diff
@@ -0,0 +1,120 @@
...
+	manifest, err := renderChartToRawManifest(ctx, kymaObjKey, m.config.WebhookChartPath, chartArgValues)
+	if err != nil {
+		return true, err
+	}
+	resources, err := getRawManifestUnstructuredResources(manifest)
+	if err != nil {
+		return true, err
+	}
+	for _, resource := range resources {
+		err := syncContext.RuntimeClient.Patch(ctx, resource, client.Apply, client.ForceOwnership, skrChartFieldOwner)
```
jakobmoellersap, line 70: would you kindly update this to a concurrent implementation using ErrGroup? otherwise we will have to wait for each resource patch to complete, which develops bottlenecking here

pkg/deploy/template_chart_manager.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	resources, err := getRawManifestUnstructuredResources(manifest)
+	if err != nil {
+		return err
+	}
+	for _, resource := range resources {
+		if err := syncContext.RuntimeClient.Delete(ctx, resource); err != nil && !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to delete webhook %s: %w", resource.GetKind(), err)
+		}
+	}
```
adityabhatia, line 104: does it make sense to implement similar errGrp here, so that deletion of following resources are not blocked by one failing?

adityabhatia (approved):

Issue #234, opened 2022-10-05
Istio Gateway resource should be discoverable using a label 


khlifi411:
## Description:
Currently, the watcher reconciler expects that an `istio` gateway resource with the name `lifecycle-manager-gateway` is deployed in the default namespace.
Due to the fact that this can vary between the different KCP environments, a better approach would be to implement a more generic way for resolving the name and namespace of the gateway resource. The watcher reconciler uses it to configure the virtual service resource it is managing.

Edit: In this Issue we don't configure the gateway itself. It's only about finding a proper gateway and configuring Virtual Service for the Watcher using that Gateway.

## ACs:
- [ ] Design more flexible Gateway lookup process: for example a label selector to resolve the `istio` gateway resource used by the custom `istio` client.
- [ ] Make the Gateway configuration more dynamic - command line flags require application restart to change. Maybe the Gateway configuration can be fetched from the Watcher object? 
- ~~[ ] Use `operator.kyma-project.io/ingress-for` + controller name on the gateway to determine the controller that should be used~~
- [ ] If there are more than 1 candidate, emit a warning-event for the watcher object and default to the first found gateway
- [ ] If there is no gateway candidate, return an error (stop processing) and output a warning-event for the watcher object that there is no ingress for events

## Related:
_Originally posted by @jakobmoellersap in https://github.com/kyma-project/lifecycle-manager/pull/231#discussion_r987619930_

Tomasz-Smelcerz-SAP:
I have a question about the AC @khlifi411:
"Use operator.kyma-project.io/ingress-for + controller name on the gateway to determine the controller that should be used"
I assume this means, that `spec.selector.app` should be set to the `operator.kyma-project.io/ingress-for-<controller-name>` on the gateway object.
But we don't create any `gateway` object directly in the codebase. I assume then that this should be done in the yaml files / kustomization definitions of the project.

jeremyharisch:
@Tomasz-Smelcerz-SAP the Gateway will be created at Deploy-Time. I am currently working on that. You do not have to implement the Gateway for this task 👍

khlifi411:
> @khlifi411 I have a question about the AC:
> "Use operator.kyma-project.io/ingress-for + controller name on the gateway to determine the controller that should be used"
> I assume this means, that `spec.selector.app` should be set to the `operator.kyma-project.io/ingress-for-<controller-name>` on the gateway object.
[...]

PR #319, opened 2022-11-22
Lookup Istio gateway by label

Tomasz-Smelcerz-SAP:
**Description**

The Istio Gateway for dynamically created VirtualService is now selected by label instead of fixed name.
See: #234 

- Replace CLI flag `gateway-name` with `gateway-ns-name` that accepts unique name in the format: `namespace/name`
- Add new CLI flag `gateway-selector` that accepts a  label selector query as in the K8s _LIST_ REST API
- Add new `gateway` object to the Watcher CR. The object has two properties: `namespacedName` and `labelSelector` that corresponds to the CLI flags.
- Define basic validations for the new CLI flags and the Watcher CR `gateway` object fields.
- CLI flags for the gateway lookup are merged during reconciliation with the Watcher CR configuration. Watcher CR values have  precedence over CLI flags
- namespaced name of the gateway has precedence over label selector for lookup, because it's more specific
- Add warnings as described in #234 
- Provide the default value for gateway selector CLI flag: `operator.kyma-project.io/watcher-gateway=default`
- Add new file in the main package: `flags.go` to remove flag management from `main.go` - `main.go` becomes too big.
- **Related Issue(s)**
#234


controllers/withwatcher/suite_with_watcher_test.go
```diff
@@ -79,7 +79,7 @@ const (
 	webhookChartPath       = "../../skr-webhook"
 	istioResourcesFilePath = "../../config/samples/tests/istio-test-resources.yaml"
 	virtualServiceName     = "kcp-events"
-	gatewayName            = "lifecycle-manager-kyma-gateway"
+	gatewaySelector        = "operator.kyma-project.io/gateway=default"
```
jakobmoellersap, line 82: could we use the constants available in v1alpha1 package instead of hardcoding the values?

main.go
```diff
@@ -280,7 +280,7 @@ func defineFlagVar() *FlagVar {
 		"The resources.limits.cpu for skr webhook.")
 	flag.StringVar(&flagVar.virtualServiceName, "virtual-svc-name", "kcp-events",
 		"Name of the virtual service resource to be reconciled by the watcher control loop.")
-	flag.StringVar(&flagVar.virtualServiceName, "gateway-name", "lifecycle-manager-kyma-gateway",
+	flag.StringVar(&flagVar.gatewaySelector, "gateway-selector", "operator.kyma-project.io/gateway=default",
```
jakobmoellersap, line 283: Also here it should have values bound to v1alpha1

pkg/istio/client.go
```diff
@@ -63,30 +67,59 @@ func (c *Client) getVirtualService(ctx context.Context) (*istioclientapi.Virtual
...
-		return nil, fmt.Errorf("error getting configured istio gateway: %w", err)
+		return nil, err
 	}
+
 	virtualSvc := &istioclientapi.VirtualService{}
 	virtualSvc.SetName(c.virtualServiceName)
 	virtualSvc.SetNamespace(metav1.NamespaceDefault)
-	virtualSvc.Spec.Gateways = append(virtualSvc.Spec.Gateways, c.gatewayName)
+	if gateway != nil {
+		gwKey := apitypes.NamespacedName{Namespace: gateway.Namespace, Name: gateway.Name}
```
jakobmoellersap, line 85:
```suggestion
		gwKey := client.ObjectKeyFromObject(gateway)
```

pkg/istio/client.go
```diff
@@ -63,30 +67,59 @@ func (c *Client) getVirtualService(ctx context.Context) (*istioclientapi.Virtual
...
+		Gateways(metav1.NamespaceAll).
+		List(ctx, lo)
+
+	if err != nil {
+		return nil, fmt.Errorf("error looking up Istio gateway with label %q: %w", c.gatewaySelector, err)
+	}
+
+	if len(gateways.Items) == 0 {
+		c.logger.Info("Warning: No matching Istio gateways found", "labelSelector", c.gatewaySelector)
+		return nil, nil
```
jakobmoellersap, line 112: why is this an info warning and not an error for lookup?

pkg/istio/client.go
```diff
@@ -63,30 +67,59 @@ func (c *Client) getVirtualService(ctx context.Context) (*istioclientapi.Virtual
...
+	}
+
+	if len(gateways.Items) > 1 {
+		gwKey := apitypes.NamespacedName{Namespace: gateways.Items[0].Namespace, Name: gateways.Items[0].Name}
```
jakobmoellersap, line 116:
```suggestion
		gwKey := client.ObjectKeyFromObject(gateways.Items[0])
```

pkg/istio/client.go
```diff
@@ -63,30 +67,59 @@ func (c *Client) getVirtualService(ctx context.Context) (*istioclientapi.Virtual
 	return virtualService, nil
 }
 
-func (c *Client) createVirtualService(ctx context.Context, watcher *v1alpha1.Watcher,
+func (c *Client) CreateVirtualService(ctx context.Context, watcher *v1alpha1.Watcher,
```
jakobmoellersap, line 70:
```suggestion
func (c *Client) createVirtualService(ctx context.Context, watcher *v1alpha1.Watcher,
```

main.go
jakobmoellersap: Why is the gateway-selector hard-coded in the app? Wouldn't it make sense to put it in the Kyma/Watcher instead?

controllers/withwatcher/suite_with_watcher_test.go
Tomasz-Smelcerz-SAP: The reason is primarily my lack of knowledge of the v1alpha1 package :) In addition gatewayName was hardcoded here before. I'll define a new constant in the v1alpha1 with a default value for this flag.

main.go
Tomasz-Smelcerz-SAP: `Wouldn't it make sense to put it in the Kyma/Watcher instead?` I am not sure if I understand... I defined a constant for it in the v1alpha1 package. Is that enough?

pkg/istio/client.go
Tomasz-Smelcerz-SAP: This comes from the issue Acceptance Criteria. @khlifi411 could you confirm?

pkg/istio/client.go
Tomasz-Smelcerz-SAP: Thanks for catching that, leftover from testing...

pkg/istio/client.go
khlifi411: Actually, the ACs were updated by @jakobmoellersap [here](https://github.com/kyma-project/lifecycle-manager/pull/231#discussion_r987619930) 😅
@jakobmoellersap would be great if you adjust the ACs 👍

main.go
jakobmoellersap: I think what I meant is that the selector probably is not something we should define at operator-startup, but dynamically based on a setting in the Kyma or Watcher CR. WDYT?

pkg/istio/client.go
jakobmoellersap: The ACs are correct. However imo, the proper handling here should return the error and send a Kubernetes event to the resource with the warning. This is also what I meant in the ACs. Simply logging an error in the controller is not visible to most users deploying the controller. They will only look at the CR. If you call `logger.Info` it will only be visible in controller-logs. The ACs are still fitting imo.

jeremyharisch: Agree in this point with @jakobmoellersap

jeremyharisch (commented):
https://github.com/kyma-project/lifecycle-manager/issues/234#issuecomment-1324131357

pkg/istio/client.go
khlifi411: @jakobmoellersap I agree about your description above 👍 
I still think it's unclear (at least for me) in the ACs if `output a warning` means `fire a warning event` or `log a warning`.

Tomasz-Smelcerz-SAP: @khlifi411 @jakobmoellersap  Same for me, I understood this AC as "continue processing instead of firing an error, but log the warning".

api/v1alpha1/operator_labels.go
```diff
@@ -24,6 +24,8 @@ const (
 	// WatchedByLabel defines a redirect to a controller that should be getting a notification
 	// if this resource is changed.
 	WatchedByLabel = OperatorPrefix + Separator + "watched-by"
+	//DefaultIstioGatewaySelector defines a default label selector for a Gateway to configure a VirtualService for the Watcher
+	DefaultIstioGatewaySelector = OperatorPrefix + Separator + "watcher-gateway=default"
```
jakobmoellersap, line 28: Why did you choose to not build the selector with `metav1`?

api/v1alpha1/watcher_types.go
```diff
@@ -20,6 +20,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	WatcherControllerName = "watcher"
```
jakobmoellersap, line 24: I think it was a bad idea of me to originally include lifecycle manager in the API types. Ideally we should not add more controllers to the API types I believe.

jakobmoellersap (commented):
looks good, just one more comment that the `watcher` controller should probably not be included in the API types. After that, for me LGTM

jakobmoellersap (commented):
Please verify lint and implement comments, after that LGTM

api/v1alpha1/operator_labels.go
Tomasz-Smelcerz-SAP: What would be a benefit of it?

api/v1alpha1/watcher_types.go
Tomasz-Smelcerz-SAP: OK, I did that because I've seen this pattern, although I also found it strange. I will move it somewhere outside the API package.

main.go
Tomasz-Smelcerz-SAP: I did as you suggested, It's taken from the Watcher CR now, if defined.

api/v1alpha1/operator_labels.go
jakobmoellersap: Ideally you can then provide an entire selector also as metav1 selector in the spec of watcher, similar to template selectors in a deployment. What do you think? If you're still against it as a default, which I find fair, I think we should at least consider putting a metav1 selector for the spec of the Watcher

api/v1alpha1/watcher_types.go
```diff
@@ -45,6 +50,24 @@ const (
...
+	// +kubebuilder:validation:MaxLength=127
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?[/][a-z]([-a-z0-9]*[a-z0-9])?$`
+	NamespacedName string `json:"namespacedName,omitempty"`
+
+	// LabelSelector allows to select the Gateway using label selectors as defined in the K8s LIST API.
+	// Ignored if NamespacedName is set.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=512
+	LabelSelector string `json:"labelSelector,omitempty"`
```
jakobmoellersap, line 68: This is the selector I was referencing above

jakobmoellersap (commented):
Looks good overall. Just a small concern regarding the selector, after that LGTM :) great stuff! 🥇

jakobmoellersap (approved):

jeremyharisch:
great implementation 🔥

Issue #326, opened 2022-11-25
Dynamically fetch VirtualService Hosts from Gateway


jeremyharisch:
Implementing this issue should be done after the following PR has been merged: https://github.com/kyma-project/lifecycle-manager/pull/319

**Description**
During the Watcher reconcile loop the corresponding `Istio.VirtualServices` will be created and configured. The configuration of the VirtualServices are part of the spec of the WatcherCRs, besides the `VirtualsService.Hosts`. The Hosts should be automatically fetched from the corresponding Istio. Gateway instead of a [hardcoded value](https://github.com/kyma-project/lifecycle-manager/blob/main/pkg/istio/client.go#L81).

**ACs:**
- [x] Remove hardcoded `vsHost` value
- [x] Fetch Hosts list from corresponding Istio.Gateway
- [x] Include fetched hosts into the VirtualService configuration
- [x] Determine whether the current approach is sufficient (change to the Gateway won't trigger reconciliation) if not, a follow-up or a fix should be created
- [x] Have a dedicated success interval for the Watcher
- [x] For each Watcher,  create a dedicated virtual service, it will be implemented not here, but in the following issue: https://github.com/kyma-project/lifecycle-manager/issues/336

ruanxin:
Where is this hardcoded value come from, can you provide a bit more information here? @jeremyharisch 
```
vsHost                     = "listener.kyma.cloud.sap" 
```

ruanxin:
management-plane-config PR: https://github.tools.sap/kyma/management-plane-config/pull/2678

jakobmoellersap:
I believe with the latest changes we have everything in place and I do not think we need additional follow-up, thus closing

PR #335, opened 2022-12-21
Dynamically fetch virtual service hosts from gateway

ruanxin:
related issue: https://github.com/kyma-project/lifecycle-manager/issues/326


api/v1alpha1/watcher_types.go
```diff
@@ -52,18 +52,8 @@ const (
...
-	// +kubebuilder:validation:MinLength=3
-	// +kubebuilder:validation:MaxLength=127
-	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?[/][a-z]([-a-z0-9]*[a-z0-9])?$`
-	NamespacedName string `json:"namespacedName,omitempty"`
-
 	// LabelSelector allows to select the Gateway using label selectors as defined in the K8s LIST API.
-	// Ignored if NamespacedName is set.
-	// +kubebuilder:validation:Optional
-	LabelSelector *metav1.LabelSelector `json:"selector,omitempty"`
+	LabelSelector *metav1.LabelSelector `json:"selector"`
```
jakobmoellersap, line 56: if this is  no longer optional, I do not think we need a pointer here.

controllers/watcher_controller.go
```diff
@@ -69,35 +71,45 @@ func (r *WatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	watcherObj := &v1alpha1.Watcher{}
 	if err := r.Get(ctx, client.ObjectKey{Name: req.Name, Namespace: req.Namespace}, watcherObj); err != nil {
 		logger.Error(err, "Failed to get reconciliation object")
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		return ctrl.Result{Requeue: true}, client.IgnoreNotFound(err)
```
jakobmoellersap, line 74: if you have a no found it will do error == nil but still requeue. I think this will infinetly requeue then.

controllers/watcher_controller.go
```diff
@@ -69,35 +71,45 @@ func (r *WatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-			return ctrl.Result{Requeue: true}, err
+			return ctrl.Result{}, err
 		}
-		return ctrl.Result{}, nil
+		return ctrl.Result{Requeue: true}, nil
 	}
 
 	// check finalizer on native object
 	if !controllerutil.ContainsFinalizer(watcherObj, watcherFinalizer) {
 		controllerutil.AddFinalizer(watcherObj, watcherFinalizer)
```
jakobmoellersap, line 93: you can check if the finalizer was added or not by using the bool returned here and then base your return on this instead of requeue

controllers/withwatcher/watcher_controller_helper_test.go
```diff
@@ -80,54 +85,58 @@ func createWatcherCR(managerInstanceName string, statusOnly bool) *v1alpha1.Watc
...
-			Name:      fmt.Sprintf("%s-sample", component),
-			Namespace: metav1.NamespaceDefault,
-		}, watcherCR)
-		if err != nil {
-			continue
-		}
-
-		watchers = append(watchers, watcherCR)
+	if err := controlPlaneClient.List(suiteCtx, watcherList); err != nil {
+		return watchers
```
jakobmoellersap, line 99: this should imo not be ignored to avoid flaky tests

controllers/withwatcher/watcher_controller_helper_test.go
```diff
@@ -80,54 +85,58 @@ func createWatcherCR(managerInstanceName string, statusOnly bool) *v1alpha1.Watc
...
-	if len(watcherObjKeys) == 0 {
-		return func(g Gomega) bool {
-			watchers := listTestWatcherCrs(controlPlaneClient)
-			return len(watchers) == 0
-		}
-	}
-	return func(g Gomega) bool {
-		err := controlPlaneClient.Get(suiteCtx, watcherObjKeys[0], &v1alpha1.Watcher{})
-		return apierrors.IsNotFound(err)
+	return err
```
jakobmoellersap, line 112: if the cr is present then the get call will not error out. This means that it will return nil even if it is still there

pkg/istio/client.go
```diff
@@ -27,23 +26,18 @@ const (
 )
 
 var (
-	vsHost                     = "listener.kyma.cloud.sap" //nolint:gochecknoglobals
-	errNoGatewayConfigured     = errors.New("error processing Watcher: No istio gateway configured")
-	errCantFindMatchingGateway = errors.New("can't find matching Istio Gateway")
+	errNoGatewayConfigured        = errors.New("error processing Watcher: No istio gateway configured")
```
jakobmoellersap, line 29: to make the errors assertable, I think they should always be exported.

pkg/testutils/utils.go
```diff
@@ -29,7 +29,7 @@ const (
 	letterBytes        = "abcdefghijklmnopqrstuvwxyz"
 	defaultBufferSize  = 2048
 	httpClientTimeout  = 2 * time.Second
-	Timeout            = time.Second * 10
+	Timeout            = time.Second * 20
```
jakobmoellersap, line 32: why was this necessary?

controllers/watcher_controller.go
```diff
@@ -69,35 +71,45 @@ func (r *WatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	if apierrors.IsNotFound(err) {
+		if _, err := r.IstioClient.CreateVirtualService(ctx, watcherObj); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to create virtual service %w", err)
+		}
+		return ctrl.Result{Requeue: true}, nil
+	}
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	err = r.IstioClient.UpdateVirtualServiceConfig(ctx, watcherObj, virtualService)
```
jakobmoellersap, line 110: can we encapsulate Create/Update into a single serverside apply instead of using the Istio Client?

controllers/watcher_controller.go
```diff
@@ -69,35 +71,45 @@ func (r *WatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 	if err != nil {
-		return ctrl.Result{Requeue: true}, err
+		return ctrl.Result{}, err
 	}
```
Tomasz-Smelcerz-SAP, line 113: I think it's better to extract such domain-specific (and low-level) logic from the controller loop, focusing here only on error handling, requeue policy, finalizers, timestamps etc.
I would try to move this sequence into a method inside `istio/client`, as it used to be before.

pkg/istio/client.go
```diff
@@ -69,46 +63,38 @@ func NewVersionedIstioClient(cfg *rest.Config, config Config, recorder record.Ev
...
+func (c *Client) CreateVirtualService(ctx context.Context, watcher *v1alpha1.Watcher,
 ) (*istioclientapi.VirtualService, error) {
 	if watcher == nil {
 		return &istioclientapi.VirtualService{}, nil
 	}
 
-	gateway, err := c.lookupGateway(ctx, watcher)
+	virtualSvc := &istioclientapi.VirtualService{}
+	virtualSvc.SetName(c.config.VirtualServiceName)
+	virtualSvc.SetNamespace(metav1.NamespaceDefault)
```
Tomasz-Smelcerz-SAP, line 84: Isn't the task about creating a distinct VirtualService For every watcher?
It looks like we're always creating the same one: name=c.config.VirtualServiceName, namespace="default"

flags.go
```diff
@@ -68,14 +54,6 @@ func defineFlagVar() *FlagVar { //nolint:funlen
...
 	flag.StringVar(&flagVar.virtualServiceName, "virtual-svc-name", "kcp-events",
 		"Name of the virtual service resource to be reconciled by the watcher control loop.")
-	flag.Var(newNamespacedNameVar(&flagVar.gatewayNamespacedName), "gateway-ns-name",
-		"Namespaced name of the gateway resource that the virtual service will use. "+
-			"Format: <namespace>/<name>. Example: \"my-namespace/my-gateway\".")
-	flag.Var(newLabelSelectorVar(&flagVar.gatewaySelector).
-		withDefault(operatorv1alpha1.DefaultIstioGatewaySelector()),
-		"gateway-selector", "Label selector of the gateway resource that the virtual service will use. "+
-			"Ignored if gateway-ns-name flag is specified. Format: K8s label selector expression. "+
-			"Example: \"label1=value1,label2=value2\"")
```
Tomasz-Smelcerz-SAP, line 78: I think the system now should ensure we have a distinct VirtualService for different Watcher CRs?
If so, I'd rather remove the flags in exact opposite way: Gateway selectors can be defined statically with CLI flags, because we don't need distinct gateways for different Watcher CRs.
But the VirtualServices must be different, so defining their names with a single static flag doesn't make sense.

api/v1alpha1/watcher_types.go
ruanxin: yes, also changed `WatcherSpec.Gateway` to non optional, I think now for each Watcher, it must have a related gateway.

controllers/watcher_controller.go
ruanxin: sure, this requeue is not need 👍

controllers/watcher_controller.go
ruanxin: I completely agree with the idea that for the control loop, it should not contains low level details. But for this case, I don't think what you mentioned here should consider as low level, to me, the function this `istioClient` defined is already abstracted essential high-level policy, it's the policy this watcher control loop should follows (1. check if VirtualService exists, 2. update VirtualService config), low level detail has been already encapsulated. The problem here maybe this `istioClient` is not designed abstract enough (basically it can be a generic client, but not specific to istio), which leaves the impression that the dependency inversion principle is violated here.

flags.go
ruanxin: Regarding VirtualService, you are right, current design not make sense to just use a single VirtualService, but as I mentioned in the related issues https://github.com/kyma-project/lifecycle-manager/issues/326, we will refactoring it to dedicated VirtualService in this followup issue https://github.com/kyma-project/lifecycle-manager/issues/336, but not here.

Regarding Gateway, if we only consider Gateway for our components (KLM and KMM), to configure it as a flag could make sense, but KLM also have to take care of other components' Watcher, for those components, we are not limiting the gateway setup, it's up to the teams to define their contract, what KLM needs to do is just build the connection between gateway and VirtualService, so use Gateway selector which defined in Watcher CR is more generic approach, that's the reason remove those two flags.

pkg/istio/client.go
ruanxin: no, this will be handled in the followup issue, and it's not high priority as we discussed in daily. https://github.com/kyma-project/lifecycle-manager/issues/336

controllers/withwatcher/watcher_controller_helper_test.go
ruanxin: agree

controllers/withwatcher/watcher_controller_helper_test.go
ruanxin: Yes, I changed the condition to check `if !apierrors.IsNotFound(err)`, to make sure only the not found case can continue.

pkg/istio/client.go
ruanxin: done

pkg/testutils/utils.go
ruanxin: bring it back

controllers/watcher_controller.go
Tomasz-Smelcerz-SAP: @ruanxin I still think this is too low level for the controller-loop.
IMO in the `Reconcile()` function we should only have something like: `ReconcileVirtualServices()` call (meaning: what to do), instead of embedding the necessary steps (meaning: how to do it), as it is now.

For example, the current implementation goes somewhat like this: `Get -> IfNotExistsCreate -> ElseUpdate`.
It could be replaced by, perhaps more efficient, ServerSideApply from K8s.
If we want to introduce such change, we would have to make a lot of changes in the Reconcile() implementation, which shouldn't be the case.
The knowledge of how we perform Istio configurations changes, and how many steps are necessary, shouldn't be a part of the controller loop logic.
At least that is my opinion on the correct responsibility split in the K8s operators, a general pattern.

Having said that, The code is readable and it's easy to refactor if necessary, so I can accept the PR.

pkg/istio/client.go
```diff
@@ -69,46 +62,38 @@ func NewVersionedIstioClient(cfg *rest.Config, config Config, recorder record.Ev
...
+
+	gateways, err := c.LookupGateways(ctx, watcher)
 	if err != nil {
 		return nil, err
 	}
+	for i := range gateways {
```
Tomasz-Smelcerz-SAP, line 89: Just a style preference, but don't you think it's better to collect all the gateways in the `[]string` slice first, and then just assign the ready slice the `virtualSvc.Spec.Gateways` - once?

pkg/istio/client.go
```diff
@@ -118,60 +103,58 @@ func (c *Client) createVirtualService(ctx context.Context, watcher *v1alpha1.Wat
...
+			labelSelector, err)
 	}
 
 	if len(gateways.Items) == 0 {
 		c.eventRecorder.Event(watcher, "Warning", "WatcherGatewayNotFound",
 			"Watcher: Gateway for the VirtualService not found")
-		return nil, fmt.Errorf("%w. Label selector: %q", errCantFindMatchingGateway, lss)
-	}
-
-	if len(gateways.Items) > 1 {
```
Tomasz-Smelcerz-SAP, line 166: Why is this removed? According to the https://github.com/kyma-project/lifecycle-manager/issues/234, we should only support one gateway. Is the contract changed?

api/v1alpha1/watcher_types.go
```diff
@@ -52,18 +51,8 @@ const (
 
 // GatewayConfig is used to select an Istio Gateway object in the cluster.
 type GatewayConfig struct {
-	// NamespacedName takes precedence over LabelSelector if configured. Format to use: "namespaceName/gatewayName"
-	// TODO: Add validation: <name><slash><namespace>
-	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:MinLength=3
-	// +kubebuilder:validation:MaxLength=127
-	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?[/][a-z]([-a-z0-9]*[a-z0-9])?$`
-	NamespacedName string `json:"namespacedName,omitempty"`
```
Tomasz-Smelcerz-SAP, line 61: any particular reason why we exclude direct configuration for the gateway? It's certainly less error-prone than configuration by label selector - so I'd say it's safer.

pkg/istio/client.go
ruanxin: Yes, here the original AC is not a valid concern IMO, I exchanged this idea with @jakobmoellersap already. And if you check the istio virtual service definition https://istio.io/latest/docs/reference/config/networking/virtual-service/, `Spec.Gateways` is a list, it's not necessary to allow only one, and the decision in the AC to use the first find one is also a bit premature:
First, I don't think it can always fetch the same gateway if have multiple selected, the list is not sorted.
Second, if any of the gateway configured wrong and this gateway get selected, by only emitting warning but not stop the process could be even more hard to do troubleshooting.

I think I mentioned about this in daily, for clarification, any specific concern to only allow one gateway? @khlifi411

controllers/watcher_controller.go
```diff
@@ -76,28 +78,36 @@ func (r *WatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			return ctrl.Result{}, fmt.Errorf("failed to create virtual service %w", err)
 		}
-		return ctrl.Result{}, nil
+		return ctrl.Result{Requeue: true}, nil
+	}
+	if err != nil {
+		return ctrl.Result{}, err
 	}
-	err := r.IstioClient.UpdateVirtualServiceConfig(ctx, watcherObj)
+	err = r.IstioClient.UpdateVirtualServiceConfig(ctx, watcherObj, virtualService)
```
Tomasz-Smelcerz-SAP, line 108: The controller should bring the cluster to the desired state.
If the Vs does not exist, we create with the desired state.
But if it exists, we end up here. The `UpdateVirtualServiceConfig()` method does NOT configure the Vs to the desired state in terms of gateways, hosts, etc. As far as I can see it only updates `spec.http` attribute.

api/v1alpha1/watcher_types.go
ruanxin: IMO, on one hand, it's valid concern to allow multiple options to choose gateway, but on the other hand, it's not necessary to define and offering such lose contract if we can restrict them. Take this as an example, by allowing multiple way to choose gateway, then certainly we need to concern the priories for those options, in the original implementation, `NamespacedName` will overwrite selector, which confusing end user if they defined a selector but find out it's not working. This confusing also apples to error-prone, basically if the `NamespacedName` configured wrong, then the selector is useless, we will return error.

The previous concern to allow selecting by name is mainly because gateway is not managed by us in the beginning, so in case the permission problem (we can't add labels to the gateway), to choose by name make sense. But since now the gateway are fully controlled by us, it's not necessary to offer so much options. 


We can choose any of them of cause, but use label selector to select resource is more natural way, that's the reason drop the `NamespacedName`.

pkg/istio/client.go
ruanxin: ok, I also extracted it to a function, please check.

controllers/watcher_controller.go
ruanxin: Yes, good catch! the update gateway is missing here.

pkg/istio/client.go
khlifi411: I have no concerns from my side about having the vs pointing to many gateways.

api/v1alpha1/watcher_types.go
Tomasz-Smelcerz-SAP: As for the "API simplicity", i.e: "keep the API minimal" - you are right.
We could resolve all other problems, but - at this point - it makes no sense to introduce sophisticated configuration for simple things.
:+1:

controllers/watcher_controller.go
Tomasz-Smelcerz-SAP: @ruanxin If we use SSA, then the "GET" step is not necessary. SSA works (kind of) like: create-or-update by design.

Tomasz-Smelcerz-SAP (approved):

PR #340, opened 2023-01-04
First Setup of dynamic PKI

jeremyharisch:
**Short Summary:**
This PR enforces an mTLS connection between a Remote cluster (SKR) and the control-plane cluster (KCP).

The [docs](https://github.com/kyma-project/lifecycle-manager/pull/340/files#diff-34361f12f6e93aca001114ecaf58dcee6248694dd93fa68533f24d6a9f94dde9) have been updated and thus can be used for testing this PR.
Please build your own docker image and use it for deploying the KLM, since the latest image does not include the changes from this PR. Or you can also use my build image to test it.
```
# Use this in Step 4
make local-deploy-with-watcher-secured IMG=jaythedevil666/klm:latest
```

**The Setup**
In the KCP the SREs will take care of deploying the [`cert-manager` by Jetstack](https://cert-manager.io/). For the local setup a dedicated cert-manager will be deployed. 
To sign each Certificate for the KymaCRs, we will use a self-signed RootCA Certificate. The Issuer and RootCACert setup can be found [here](https://github.com/kyma-project/lifecycle-manager/pull/340/files#diff-e7ef6c63ac89d132e24091b3c86fd6c287daa2ceb4457e52e18294ba6e9303a4).
We have a `ClusterIssuer` which will issue the RootCACert. When this certificate is created and signed, the Issuer `watcher-selfsigned-issuer` will use this certificate as CA to sign all certificates for the KymaCRs.
(This is an adoption of a popular cert-manager architecture, can be found [here](https://cert-manager.io/docs/configuration/selfsigned/) - This has been clarified by our Security experts and the SREs)

The Istio Gateway also needs a certificate to use strict mTLS. It will use the issued self-signed RootCACertificate. Due to this reason and for simplicity, we decided to have the whole cert setup in the istio-system namespace. Plus, the Issuer of cert-manager always issues the certificates in the same namespace as the Issuer recites. Thus, the Issuer setup and all kyma certificates would be in the Istio namespace.

**How does it work?**
In the reconciliation loop of a KymaCR, the loop will create a valid  [cert-manager `Certificate`](https://cert-manager.io/docs/concepts/certificate/) for the KymaCR and applies it using SSA. The deployed cert-manager will pick up the certificate, will create a `CertificateRequest`, and the deployed `Issuer` will create and sign the certificate using the `RootCACert`.

On the next step it will check if the secret of the Certifiacte is ready, if yes it will fetch the Secret, and copy it into the manifest of the watcher. Thus, it will directly be installed with the runtime-Watcher and thus get updated on each reconcile loop - Certificate Rotation automatically working. For the general certificate rotation the cert-manager will take care of. I used the default values for the cert-manager, but this can be discussed in future (default: expires after 90days, will be rotated after 60 days).


jeremyharisch:
Not ready fro review, just converted from draft to normal PR to trigger pipelines for testing

jakobmoellersap:
> The [docs](https://github.com/kyma-project/lifecycle-manager/pull/340/files#diff-34361f12f6e93aca001114ecaf58dcee6248694dd93fa68533f24d6a9f94dde9) have been updated and thus can be used for testing this PR. Please build your own docker image and use it for deploying the KLM, since the latest image does not include the changes from this PR. Or you can also use my build image to test it.
> 
> ```
> # Use this in Step 4
> make local-deploy-with-watcher-secured IMG=jaythedevil666/klm:latest
> ```

I believe we have an image build in our PR job which you can use instead: 
```
make local-deploy-with-watcher-secured eu.gcr.io/kyma-project/lifecycle-manager:PR-340
```

controllers/watcher_controller.go
```diff
@@ -69,6 +69,12 @@ func (r *WatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	logger := ctrlLog.FromContext(ctx).WithName(req.NamespacedName.String())
 	logger.Info("Reconciliation loop starting")
 
+	// sync CA Certificate to Istio Namespace
+	if err := r.IstioClient.SyncCertificateSecretToIstio(ctx, r.Client); err != nil {
```
jakobmoellersap, line 73: why do you need to pass a client into a method of a client? This looks a bit suspicious. Maybe you want to refactor the structure of these clients / methods

controllers/watcher_controller.go
```diff
@@ -69,6 +69,12 @@ func (r *WatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		logger.Error(err, "Could not synchronise CA Root Certificate to Istio Namespace")
+	}
+	logger.Info("Synchronised CA Root Certificate to Istio Namespace")
```
jakobmoellersap, line 76: maybe we should put this on debug level

flags.go
```diff
@@ -57,6 +57,10 @@ func defineFlagVar() *FlagVar {
 		"Enabling KCP Watcher to reconcile Watcher CRs created by KCP run operators")
 	flag.StringVar(&flagVar.skrWatcherPath, "skr-watcher-path", "./skr-webhook",
 		"The path to the skr watcher chart.")
+	flag.BoolVar(&flagVar.enableTLSCertSync, "enable-tls-cert-sync", true,
+		"Enabling synchronisation of TLS-Certificate-Secrets, created during the Watcher installation")
+	flag.StringVar(&flagVar.tlsCertSecretName, "tls-cert-secret-name", "skr-webhook-tls",
```
jakobmoellersap, line 62:
```suggestion
	flag.StringVar(&flagVar.tlsCertSyncSecretName, "tls-cert-sync-secret-name", "skr-webhook-tls",
```

main.go
```diff
@@ -50,10 +50,15 @@ import (
...
 
+	certManagerV1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
+
 	operatorv1alpha1 "github.com/kyma-project/lifecycle-manager/api/v1alpha1"
 	"github.com/kyma-project/lifecycle-manager/controllers"
 	moduleManagerV1alpha1 "github.com/kyma-project/module-manager/api/v1alpha1"
+
 	//+kubebuilder:scaffold:imports
+
+	aggregatorclientsetscheme "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/scheme"
```
jakobmoellersap, line 61: why is this necessary?

pkg/istio/client.go
```diff
@@ -382,3 +392,59 @@ func prepareIstioHTTPRouteForCR(obj *v1alpha1.Watcher) *istioapi.HTTPRoute {
...
+	certSecret := &corev1.Secret{}
+	err := kcpClient.Get(ctx,
+		client.ObjectKey{
+			Namespace: c.config.WatcherRootCertificateNamespace,
+			Name:      c.config.WatcherRootCertificateName,
+		}, certSecret)
+
+	if apierrors.IsNotFound(err) {
+		// if CA Certificate does not exist, check if it exists in istio Namespace, if yes remove it
+		c.logger.Info("CA Root Certificate does not exist, " +
```
jakobmoellersap, line 409: I believe this should be at Debug Level

pkg/istio/client.go
```diff
@@ -382,3 +392,59 @@ func prepareIstioHTTPRouteForCR(obj *v1alpha1.Watcher) *istioapi.HTTPRoute {
...
+func (c *Client) SyncCertificateSecretToIstio(ctx context.Context, kcpClient client.Client) error {
+	certSecret := &corev1.Secret{}
+	err := kcpClient.Get(ctx,
+		client.ObjectKey{
+			Namespace: c.config.WatcherRootCertificateNamespace,
+			Name:      c.config.WatcherRootCertificateName,
+		}, certSecret)
+
+	if apierrors.IsNotFound(err) {
+		// if CA Certificate does not exist, check if it exists in istio Namespace, if yes remove it
```
jakobmoellersap, line 408: Could you explain to me again why we don't directly sync it into the correct namespace always?

pkg/istio/client.go
```diff
@@ -382,3 +392,59 @@ func prepareIstioHTTPRouteForCR(obj *v1alpha1.Watcher) *istioapi.HTTPRoute {
...
+			Namespace:   c.config.IstioCertificateNamespace,
+			Labels:      certSecret.Labels,
+			Annotations: certSecret.Annotations,
+		},
+
+		Data: certSecret.Data,
+	}
+	// CA Certificate exists, copy it to istio namespace
+
+	if err := kcpClient.Update(ctx, istioCertSecret); apierrors.IsNotFound(err) {
```
jakobmoellersap, line 441: I believe this should be changed to a SSA with a correct agent instead of an Update Call

pkg/watcher/certificate.go
```diff
@@ -0,0 +1,263 @@
...
+type CertificateSecret struct {
+	CACrt           string
+	TLSCrt          string
+	TLSKey          string
+	ResourceVersion string
+}
+
+// NewCertificate returns a new Certificate, which can be used for creating a cert-manager Certificate.
+func NewCertificate(kcpClient client.Client, kyma *v1alpha1.Kyma) (*Certificate, error) {
+	if kcpClient == nil || kyma == nil {
```
jakobmoellersap, line 64: is this ever a valid scenario? Kyma and KCP Client are always present if you're in the reconcile loop

pkg/watcher/certificate.go
```diff
@@ -0,0 +1,263 @@
...
+	exists, err := c.exists(ctx)
+	if exists {
+		return nil
+	} else if err != nil {
+		return err
+	}
+	// fetch Subject-Alternative-Name from KymaCR
+	subjectAltNames, err := c.getSubjectAltNames(config)
+	if err != nil {
+		c.logger.Error(err, "Error get Subject Alternative Name from KymaCR")
```
jakobmoellersap, line 92: all of these errors imo should be logged one level higher and you should remove the logger as part of the Certificate struct.

pkg/watcher/certificate.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+type SubjectAltName struct {
+	DNSNames       []string
+	IPAddresses    []string
+	URIs           []string
+	EmailAddresses []string
+}
+
+type Certificate struct {
```
jakobmoellersap, line 47: This is imo not really well named since its not representing a certificate resource, its representing a struct containing pointers that are necessary to edit certificates. Maybe call it CertificateSync?

pkg/watcher/certificate.go
```diff
@@ -0,0 +1,263 @@
...
+
+func CreateSecretName(kymaObjKey client.ObjectKey) string {
+	return fmt.Sprintf("%s%s", kymaObjKey.Name, CertificateSuffix)
+}
+
+// Create creates a cert-manager Certificate with a sufficient set of Subject-Alternative-Names.
+func (c *Certificate) Create(ctx context.Context, config *SkrChartManagerConfig) error {
+	c.logger = logf.FromContext(ctx)
+	// Check if Certificate exists
+	exists, err := c.exists(ctx)
```
jakobmoellersap, line 83: not necessary if you change the entire logic to ServerSideApply. then you just need to call apply

pkg/watcher/certificate.go
jakobmoellersap: If you really need a logger, always use debug level and use log.FromContext(ctx) and pass the context of the reconciliation. Passing the logger is pretty cumbersome.

pkg/watcher/certificate.go
```diff
@@ -0,0 +1,263 @@
...
+func (c *Certificate) getIssuer(ctx context.Context) (*v1.Issuer, error) {
+	issuerList := &v1.IssuerList{}
+	err := c.kcpClient.List(ctx, issuerList, &client.ListOptions{
+		LabelSelector: k8slabels.SelectorFromSet(LabelSet),
+		Namespace:     c.kyma.Namespace,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("could not list cert-manager issuer %w", err)
+	}
+	if len(issuerList.Items) == 0 {
```
jakobmoellersap, line 253: If there is more than one issuer, you are missing a warning log or an error here.

jakobmoellersap (changes requested):
Did a superficial review now, need to test it in detail still

khlifi411 (commented):
The local testing was working fine for me 💯 
But I was getting the following error in the `lifecycle-manager` logs:
```log
E0206 09:32:53.5106651 reflector.go:140] pkg/mod/k8s.io/client-go@v0.26.1/tools/cache/reflector.go:169: Failed to watch *v1.Certificate: unknown (get certificates.cert-manager.io)
```

Makefile
```diff
@@ -125,10 +125,11 @@ lt-deploy: manifests kustomize ## Deploy controller to the K8s cluster specified
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/load_test | kubectl apply -f -
 
-.PHONY: local-deploy-with-watcher
-local-deploy-with-watcher: install manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+.PHONY: local-deploy-with-watcher-secured
```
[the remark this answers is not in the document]
jeremyharisch: This is a left over, originally I planned to have both (secured and unsecured), but that would make the Docs way harder to grasp, and since the actual setup should be secured, I just kept that one. I will rename it!

pkg/watcher/certificate.go
jeremyharisch: You are right, they should always be not nil.

flags.go
jeremyharisch: This is anyway a leftover from old commits - I will remove it completely

pkg/watcher/certificate.go
jeremyharisch: @jakobmoellersap Please have a look here: https://github.com/kyma-project/lifecycle-manager/pull/340/files/40eaf127431a197fe9fa4fd1fe00d728b87ab9fe#diff-e1231291a556a62c67cb33d06ebce147d6142aa744e71eeb98500915c780f66aR81
I did not pass the logger, I created it from the context. Or should I recreate in every function where it is needed?

jakobmoellersap: My problem is that you use a struct called `c.logger` instead of using a local reference variable. IMO you can and should get rid of c.logger and just create the logger on demand. WDYT?

api/v1alpha1/operator_labels.go
```diff
@@ -26,6 +26,9 @@ const (
 	// WatchedByLabel defines a redirect to a controller that should be getting a notification
 	// if this resource is changed.
 	WatchedByLabel = OperatorPrefix + Separator + "watched-by"
+	// PurposeLabel defines the purpose of the resource, i.e. Secrets which will be used to certificate management.
+	PurposeLabel = OperatorPrefix + Separator + "purpose"
```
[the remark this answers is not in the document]
jeremyharisch: This is used to filter the Issuer ([here](https://github.com/jeremyharisch/kyma-operator/blob/30eec89c4ce5654443612835a66b6f493c7d8a33/pkg/watcher/certificate.go#L247))

Since I did not find another label which can be reused for such purpose, I thought of introducing this new one.
Labels are set [here](https://github.com/jeremyharisch/kyma-operator/blob/ebccd8005142697a05494d62b64dd83d3ed69eaa/config/watcher/certificate_setup.yaml#L18)

Meaning, the Issuer will have the Operator Name label and a purpose label. We could also remove the purpose label and just keep the `managed-by`, but then we always need to make sure we will only have this single Issuer

pkg/watcher/certificate.go
jeremyharisch: I think `CertificateSync` is more misleading than `Certificate`, since there is not sync happening.

The `Certificate` is a struct which has all necessary information to create a single cert-manager certificate for one single Kyma instance. Meaning, this is not meant to be for multiple KymaCRs or to sync some sort of state.

I tried to come up with a different name, but for me the most sufficient is `Certificate` - What do you think? or do you may have a different proposal than `CertificateSync`?

api/v1alpha1/operator_labels.go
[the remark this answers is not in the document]
jeremyharisch: Furthermore, this label will be set to the secrets the cert-manager will create - just for a better understanding why such secrets are being created

pkg/watcher/certificate.go
jakobmoellersap: Understood the proposition now I think. Suggestion: A Certificate is a Certificate, the thing that creates/deletes a single cert-manager Certificate is a `CertificateManager` with a Method called `Create` and a method called `Delete` which deal with the certificate and only get passed the `context`.

You also don't need to pass `config *SkrChartManagerConfig` into Create if its already in the `CertificateManager`

WDYT?

khlifi411:
> > The local testing was working fine for me 💯 But I was getting the following error in the `lifecycle-manager` logs:
> > ```
> > E0206 09:32:53.5106651 reflector.go:140] pkg/mod/k8s.io/client-go@v0.26.1/tools/cache/reflector.go:169: Failed to watch *v1.Certificate: unknown (get certificates.cert-manager.io)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Also, the SKR watcher logs need to be updated to reflect that it's connecting via `https` to 9433 port instead of 9080.
> 
> I will create an issue for the KSR-Watcher to resolve this, since this is an issuer from the runtime-watcher repo
> 
> Regarding the log error in the KLM, I will double check why this happens

Thank you for double checking 👍 
About the logs, I meant the logs in the local testing guide not on the SKR-watcher side.

pkg/watcher/skr_webhook_manifest_manager.go
```diff
@@ -18,6 +20,11 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+const (
+	// always true since unsecured watcher setup will no longer be supported.
+	tlsEnabled = "true"
```
jakobmoellersap, line 25: I would remove this flag since its not needed

go.mod
```diff
@@ -4,10 +4,12 @@ go 1.19
 
 require (
 	github.com/Masterminds/semver/v3 v3.2.0
+	github.com/cert-manager/cert-manager v1.11.0
 	github.com/gardener/component-spec/bindings-go v0.0.78
 	github.com/go-logr/logr v1.2.3
 	github.com/go-logr/zapr v1.2.3
 	github.com/kyma-project/module-manager v0.0.0-20230203122329-275f0c18b284
+	github.com/kyma-project/module-manager/operator v0.0.0-20221125150159-5e84c4855249
```
jakobmoellersap, line 12: why is this import back? :O this should not be here, maybe a malformed import?

pkg/watcher/certificate_suite_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	operatorv1alpha1 "github.com/kyma-project/lifecycle-manager/api/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	certManagerV1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
+	. "github.com/kyma-project/lifecycle-manager/pkg/testutils"
+	moduleManagerV1alpha1 "github.com/kyma-project/module-manager/operator/api/v1alpha1"
```
jakobmoellersap, line 16: this one is a culprit!

config/watcher_local_test/kustomization.yaml
```diff
@@ -44,48 +44,53 @@ patches:
     target:
       kind: Deployment
 
+# Non-Native k8s resources
 patchesJson6902:
```
jakobmoellersap, line 48: can you run kustomize edit fix on all this?

PROJECT
```diff
@@ -27,6 +27,10 @@ resources:
     defaulting: true
     validation: true
     webhookVersion: v1
+- controller: true
```
jakobmoellersap, line 30: Why do we now have a secret controller?

jakobmoellersap (changes requested):
Still have to do functional review but overall Looks good! just having smaller comments right now

PROJECT
jeremyharisch: Weird leftover from first iteration, will remove

pkg/watcher/skr_webhook_manifest_manager.go
jeremyharisch: Moved it!

go.mod
jeremyharisch: Removed

docs/developer/watcher-local-test.md
```diff
@@ -19,165 +28,176 @@ sudo nano /etc/hosts
...
+    apiVersion: v1
+    kind: Secret
+    metadata:
+      name: kyma-sample
+      namespace: kcp-system
+      labels:
+        "operator.kyma-project.io/kyma-name": "kyma-sample"
+        "operator.kyma-project.io/managed-by": "lifecycle-manager"
+    data:
+      config: $(echo "${$(k3d kubeconfig get skr-local)/0.0.0.0/host.k3d.internal}" | base64)
```
khlifi411, line 103:
```suggestion
      config: $(k3d kubeconfig get skr-local | sed "s/0.0.0.0/host.k3d.internal/" | base64 | tr -d '\n')
```
for compatibility with linux and posix systems

docs/developer/watcher-local-test.md
```diff
@@ -19,165 +28,176 @@ sudo nano /etc/hosts
...
+    apiVersion: v1
+    kind: Secret
+    metadata:
+      name: kyma-sample
+      namespace: kcp-system
+      labels:
+        "operator.kyma-project.io/kyma-name": "kyma-sample"
+        "operator.kyma-project.io/managed-by": "lifecycle-manager"
+    data:
+      config: $(echo "$(k3d kubeconfig get skr-local)" | base64)
```
khlifi411, line 135:
```suggestion
      config: $(k3d kubeconfig get skr-local | base64 | tr -d '\n')
```
for compatibility with linux and posix systems

docs/developer/watcher-local-test.md
```diff
@@ -19,165 +28,176 @@ sudo nano /etc/hosts
...
-{"level":"INFO","date":"2023-01-05T09:21:51.011080512Z","caller":"controllers/kyma_controller.go:87","msg":"reconciling modules","context":{"controller":"kyma","controllerGroup":"operator.kyma-project.io","controllerKind":"Kyma","kyma":{"name":"kyma-sample","namespace":"default"},"namespace":"default","name":"kyma-sample","reconcileID":"f9b42382-dc68-41d2-96de-02b24e3ac2d6"}}
-{"level":"INFO","date":"2023-01-05T09:21:51.043800866Z","caller":"controllers/kyma_controller.go:206","msg":"syncing state","context":{"controller":"kyma","controllerGroup":"operator.kyma-project.io","controllerKind":"Kyma","kyma":{"name":"kyma-sample","namespace":"default"},"namespace":"default","name":"kyma-sample","reconcileID":"f9b42382-dc68-41d2-96de-02b24e3ac2d6","state":"Processing"}}
-```
-
-### Full-blown setup
+    ```log
+    1.6711877286771238e+09    INFO    skr-webhook    Kyma UPDATE validated from webhook 
+    1.6711879279507768e+09    INFO    skr-webhook    incoming admission review for: operator.kyma-project.io/v1alpha1, Kind=Kyma 
+    1.671187927950956e+09    INFO    skr-webhook    KCP    {"url": "http://host.k3d.internal:9443/v1/lifecycle-manager/event"}
```
khlifi411, line 187:
```suggestion
    1.671187927950956e+09    INFO    skr-webhook    KCP    {"url": "https://host.k3d.internal:9443/v1/lifecycle-manager/event"} 
```

Makefile
```diff
@@ -128,7 +128,9 @@ lt-deploy: manifests kustomize ## Deploy controller to the K8s cluster specified
 .PHONY: local-deploy-with-watcher
 local-deploy-with-watcher: install manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/watcher_local_test | kubectl apply -f -
+	$(KUSTOMIZE) build config/watcher_local_test_secured | kubectl apply -f -
+	# TODO PKI remove unsecrued and rename secured without secured
```
khlifi411, line 132:
```suggestion
	$(KUSTOMIZE) build config/watcher_local_test | kubectl apply -f -
```

config/watcher_local_test/kustomization.yaml
jeremyharisch: I did it, but this actuelly destroyed the kustomize and made it not working anymore

khlifi411 (approved):
Tested locally and it works! 🔥

jakobmoellersap (approved):

PR #342, opened 2023-01-04
remove watcher related resources in control-plane config

ruanxin:
only allow https hosts entry for watcher-gateway
remove watcher related resources from control plane

**Be aware:**
The PR in management plane config should be merged first: https://github.tools.sap/kyma/management-plane-config/pull/2705, otherwise the kustomize deployment will be failed due to can't found resources to patch


config/watcher/kustomization.yaml
```diff
@@ -3,7 +3,8 @@ kind: Component
 
 resources:
 - operator_v1alpha1_watcher.yaml
-
+- gateway.yaml
+- certificate_for_watcher_gateway.yaml
```
ruanxin, line 7: To be discussed:

I still have some concern about this gateway and certificate should belong to watcher, I mean istio gateway is just one option for traffic management of watcher, right? If we want to offer a customizable setup, I think it's not a good idea to build this dependencies between watcher and istio gateway (and in order for it to work, now even have to introduced one more dependeny - a cert manager Certificate). Maybe we should create an extra config folder, for those traffic management resources, as an option when enable watcher.

jakobmoellersap: My take: if people want to use a different traffic solution in control plane, they have to provide their own configuration. IMO this is not necessary to be considered right now. WDYT?

jakobmoellersap (approved):

PR #347, opened 2023-01-16
chore: concurrent Server Side Apply for catalog sync

jakobmoellersap:
Applies Catalog Data concurrently to allow for faster synchronisation of ModuleTemplates

### DIFF
--- a/pkg/catalog/remote_catalog.go
+++ b/pkg/catalog/remote_catalog.go
@@ -77,27 +77,10 @@ func (c *RemoteCatalog) CreateOrUpdate(
 
 	results := make(chan error, len(diffApply)+len(diffDelete))
 	for _, diff := range diffApply {
-		diff := diff
-		go func() {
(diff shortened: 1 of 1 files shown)


pkg/catalog/remote_catalog.go
```diff
@@ -77,28 +75,42 @@ func (c *RemoteCatalog) CreateOrUpdate(
...
+			}
+		}()
+	}
+	for _, diff := range diffDelete {
+		diff := diff
+		go func() {
+			if err := syncContext.RuntimeClient.Delete(ctx, diff); err != nil {
+				results <- fmt.Errorf("could not delete module template from diff: %w", err)
+			} else {
+				results <- nil
```
ruanxin, line 98: I suggest we can use [errgroup](https://bostonc.dev/blog/go-errgroup) to deal with the concurrent collection of errors

ruanxin (approved):

Issue #167, opened 2022-09-06
[Proposal] Introduce new Target for local deployment


ruanxin:
**Description:**
For local deployment, it make sense to add a new target value to kyma CR,  which make the kyma deployment target intuitive to understand.
See discussion: https://github.com/kyma-project/lifecycle-manager/pull/163#discussion_r963594259

**Acceptance Criteria:**
- [ ] Introduce a new Target value for the local (single-cluster) setup

PR #348, opened 2023-01-16
parse OCIRegistryCredLabel from descriptor resources.

ruanxin:
config load test
remove deprecated skr module
related issues: https://github.com/kyma-project/module-manager/issues/201


pkg/testutils/utils.go
```diff
@@ -160,14 +160,8 @@ func ModuleTemplateFactory(module v1alpha1.Module, data unstructured.Unstructure
 	return &moduleTemplate, nil
 }
 
-func readModuleTemplate(module v1alpha1.Module, moduleTemplate *v1alpha1.ModuleTemplate) error {
-	var template string
-	switch module.ControllerName {
-	case "manifest":
-		template = "operator_v1alpha1_moduletemplate_skr-module.yaml"
```
ruanxin, line 167: To reviewer:
At the moment, we only deploy manifest in kcp cluster, right? I think we can remove this skr-module example to avoid confusion.

jakobmoellersap: manifests are deployed only in kcp but the target of the manifest can still be skr or kcp. For 2 cluster mode this is relevant I believe.

api/v1alpha1/moduletemplate_types.go
```diff
@@ -66,7 +66,7 @@ type ModuleTemplateSpec struct {
 
 	// Target describes where the Module should later on be installed if parsed correctly. It is used as installation
 	// hint by downstream controllers to determine which client implementation to use for working with the Module
-	Target Target `json:"target"`
+	Target Target `json:"target,omitempty"`
```
jakobmoellersap, line 69: if this is optional, what is the default? Maybe add that to the description

config/samples/component-integration-installed/operator_v1alpha1_moduletemplate_skr-module.yaml
```diff
@@ -1,75 +0,0 @@
-apiVersion: operator.kyma-project.io/v1alpha1
```
jakobmoellersap, line 1: I believe that this should still be used in dual cluster setups / tests.

api/v1alpha1/moduletemplate_types.go
ruanxin: without mentioning this target, the default is local single cluster mode, or should we in progress this issue to introduce a new target https://github.com/kyma-project/lifecycle-manager/issues/167

config/samples/component-integration-installed/operator_v1alpha1_moduletemplate_skr-module.yaml
ruanxin: clarified, not needed

pkg/testutils/utils.go
ruanxin: clarified this change is correct.

adityabhatia (dismissed):

jakobmoellersap (changes requested):
I believe the optional field of target should stay mandatory, see discussion above.

api/v1alpha1/moduletemplate_types.go
ruanxin: But then how about the single cluster mode? should we immediately introduce new target https://github.com/kyma-project/lifecycle-manager/issues/167

jakobmoellersap: I am really not sure why we need an additional target right now. We have 2 targets already, control-plane and remote. Why is there a new target necessary?

ruanxin: after discussion, the decision is keep the current target, make this field as mandatory, for single cluster mode, use `control-plane` as target, if have further request, extend the target definition.

jakobmoellersap (approved):

PR #350, opened 2023-01-18
feat: introduce unified client and skr rate limit override

jakobmoellersap:
Introduces a new Client Abstraction that includes rest config. Also fixes Webhook lookup of SKR clients as it was skipping the cache for the chart installation and removal (was looking up the secret explicitly). Also overrides SKR Rate Limiting Config with KCP values to avoid rate limiting on catalog sync if configured properly.

### DIFF
--- a/controllers/kyma_controller.go
+++ b/controllers/kyma_controller.go
@@ -101,7 +101,8 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 
 	if kyma.Spec.Sync.Enabled {
 		var err error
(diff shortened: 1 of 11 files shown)


controllers/setup.go
```diff
@@ -54,7 +54,7 @@ func (r *KymaReconciler) SetupWithManager(mgr ctrl.Manager,
 	controllerBuilder = controllerBuilder.Watches(&source.Kind{Type: &moduleManagerV1alpha1.Manifest{}},
```
ruanxin, line 54: can we just append this `.Watches` directly after line 52, but not reassign this `controllerBuilder`?

jakobmoellersap: dropped the commit

ruanxin (approved):

PR #351, opened 2023-01-18
fix: catalog accidental delete of runtime templates due to legacy diff calculation

jakobmoellersap:
This fixes accidental deletes from the diff computation that is caused by SKR items not being present before the SSA and thus being deleted right after as potential diff.

This is a leftover regression as previously we used client side apply and the old diff logic was still in place to calculate deletes. However the delete calculation happened BEFORE the apply, so they were deleted right after

### DIFF
--- a/pkg/catalog/remote_catalog.go
+++ b/pkg/catalog/remote_catalog.go
@@ -2,6 +2,7 @@ package catalog
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/kyma-project/lifecycle-manager/api/v1alpha1"
@@ -13,6 +14,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+var ErrTemplateCRDNotReady = errors.New("module template crd for catalog sync is not ready")
 
 type Settings struct {
 	// this namespace flag can be used to override the namespace in which all ModuleTemplates should be applied.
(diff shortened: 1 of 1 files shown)


pkg/catalog/remote_catalog.go
```diff
@@ -130,48 +151,25 @@ func (c *RemoteCatalog) patchDiff(
...
-		// if the controlPlane Template does not exist in the remote, we already know we need to create it
-		// in the runtime
-		if _, exists := existingOnRemote[controlPlane.Namespace+controlPlane.Name]; !exists {
-			c.prepareForSSA(controlPlane)
-			diffToApply = append(diffToApply, controlPlane)
+) []*v1alpha1.ModuleTemplate {
+	kcp := controlPlaneList.Items
+	skr := runtimeList.Items
+	toDelete := make([]*v1alpha1.ModuleTemplate, 0, len(runtimeList.Items))
+	for skrIndex := range skr {
```
ruanxin, line 160: I'm fine with current implementation, but I think you can keep the existing way of calculating the `diffToDelete`, it should be fine, and on performance perspective, the previous one use hashmap is also better than current two for loop. And I think we should remove the previous function comments, it's still `CalculateDiffs` and the explanation is not up-to-date.

```
for i := range runtimeList.Items {
		remote := &runtimeList.Items[i]
		controlPlaneIndex, exists := existingOnControlPlane[remote.Namespace+remote.Name]

		// if the remote Template does not exist in the control plane, we already know we need to delete it
		if !exists {
			diffToDelete = append(diffToDelete, remote)
			continue
		}
	}
```

jakobmoellersap: I disagree on the performance I believe. The Hashmaps I do not need now since I am not iterating multiple times. Frequency Maps should only result in benefits If I plan to iterate multiple times over any of the 2 lists.

Regarding the comments, I fully agree, I will update these.

jakobmoellersap: I created a similar logic with frequency maps now but made it a bit more simple since we do not need to compute apply sets anymore (because we can simply always apply all templates, that way we don't have to compare them)

ruanxin (approved):

PR #355, opened 2023-01-20
Add removed files which is needed for two cluster setups

jeremyharisch:
Files was removed with https://github.com/kyma-project/lifecycle-manager/pull/348/files#diff-35bb8ee4aeb31abc92bd1be598b3b94833be5a5acdf7388d015e52f888f3507d

This file is needed for local two cluster setups.


ruanxin (approved):
we don't need two module template to do test, and there is no such skr module template concept exists any more, all moduletemplate should deployed under kcp, that's why the only leave the kcp moduletemplate.

If there is some spec you want to change

PR #361, opened 2023-01-24
Revert "Refactor SKR watcher installation to use native Kubernetes client (#313)

jakobmoellersap:
This reverts commit 4643b27eb755081affd6a250507432e70260e029.

It causes Test failures that were not discovered as well as regressions on the Cred Selector that we first have to identify.


adityabhatia:
should we close this PR, as this was not source of the failing tests in main?

PR #366, opened 2023-01-26
!feat: introduce FQDN and namespace/name module mappings

jakobmoellersap:
This feature introduces a new support for mapping modules to templates not only via label, but also via FQDN or Namespace/Name.

Two tests are failing now as they expect to install 2 modules with the same FQDN but the FQDN is now the way to differentiate a module. This is intended as there is no use case to enable the same module multiple times.

```yaml
kind: Kyma
...
spec:
  channel: alpha
  modules:
  - name: keda (can also be kyma.project.io/module/keda or kcp-system/moduletemplate-keda or moduletemplate-keda)
  sync:
    enabled: false
    moduleCatalog: true
    noModuleCopy: true
    strategy: secret
status:
  activeChannel: alpha
  conditions:
  - lastTransitionTime: "2023-01-26T17:12:55Z"
    message: not all modules are in ready state
    observedGeneration: 2
    reason: ModulesAreReady
    status: "False"
    type: Ready
  moduleStatus:
  - for: keda
    fqdn: kyma.project.io/module/keda
    generation: 1
    name: default-kyma-4191454207
    namespace: kcp-system
    state: Error
    templateInfo:
      channel: alpha
      generation: 1
      name: moduletemplate-keda
      namespace: kcp-system
      version: 0.0.1-6cd5086
  state: Processing
```


api/v1alpha1/kyma_types.go
```diff
@@ -173,11 +179,23 @@ const (
...
+	// GroupVersionKind is used to track the Kind that was created from the ModuleTemplate. This is dynamic to not bind
+	// ourselves to any kind of Kind in the code and allows us to work generic on deletion / cleanup of
+	// related resources to a Kyma Installation.
+	metav1.GroupVersionKind `json:"gvk"`
 
-	// ModuleName is the unique identifier of the module.
-	ModuleName string `json:"moduleName"`
+	// For defines the name of the Module in the Spec that the status is used for.
+	// It can be any kind of Reference format supported by Module.Name.
+	For string `json:"for"`
```
ruanxin, line 193: Use a preposition `for` as an attribute is a bit unusual here tbh, can we change to a noun word? why not still use the previously defined `ModuleName` for it?

jakobmoellersap: I think For is unfitting as well. 👍 Main Reason I do not want to use ModuleName is because we have `Name` already as the Resource and the status is in `.status.moduleStatus`, so the path would be `.status.moduleStatus.ModuleName` for me this kind of kills the purpose of nested keys. IMO we should try to refactor it to `.status.module.name` and then have name/namespace of the manifest in `.status.module.manifest.name/namespace`. WDYT?

api/v1alpha1/kyma_types.go
```diff
@@ -273,17 +287,16 @@ type moduleStatusExistsPair struct {
 }
 
 func (kyma *Kyma) GetNoLongerExistingModuleStatus() []*ModuleStatus {
-	moduleStatusMap := make(map[string]*moduleStatusExistsPair)
+	moduleStatusMap := make(map[int]*moduleStatusExistsPair)
 
 	for i := range kyma.Status.ModuleStatus {
 		moduleStatus := &kyma.Status.ModuleStatus[i]
-		moduleStatusMap[moduleStatus.ModuleName] = &moduleStatusExistsPair{exists: false, moduleStatus: moduleStatus}
+		moduleStatusMap[i] = &moduleStatusExistsPair{exists: false, moduleStatus: moduleStatus}
```
ruanxin, line 294: Use index here will not work I assume, we can't keep the `ModuleStatus` and `Modules` same order.

pkg/module/sync/runner_impl.go
```diff
@@ -159,38 +154,40 @@ func (r *RunnerImpl) updateModuleStatusFromExistingModules(modules common.Module
...
+			delete(moduleStatusMap, idx)
 		}
 	}
 	kyma.Status.ModuleStatus = convertToNewmoduleStatus(moduleStatusMap)
 	return updateRequired
 }
 
-func convertToNewmoduleStatus(moduleStatusMap map[string]*v1alpha1.ModuleStatus) []v1alpha1.ModuleStatus {
+func convertToNewmoduleStatus(moduleStatusMap map[int]*v1alpha1.ModuleStatus) []v1alpha1.ModuleStatus {
 	newModuleStatus := make([]v1alpha1.ModuleStatus, 0)
```
ruanxin, line 191: iterate through this map can't make sure original index keep same after it in the ModuleStatus array due to the go map is unordered.

ruanxin (changes requested):
the new moduleStatusMap needs to revisit

pkg/module/common/module.go
```diff
@@ -76,14 +75,15 @@ func (m *Module) ContainsExpectedOwnerReference(ownerName string) bool {
...
-}
+const maxModuleNameLength = 253
 
-func CreateModuleName(moduleName, kymaName string) string {
-	return fmt.Sprintf("%s-%s", kymaName, moduleName)
+func CreateModuleName(fqdn, kymaName string) string {
+	hash := fnv.New32()
+	_, _ = hash.Write([]byte(fqdn))
+	hashedFQDN := hash.Sum32()
+	name := fmt.Sprintf("%s-%v", kymaName, hashedFQDN)
```
ruanxin, line 84: I actually have some concern here for the naming, if we hash the module name, then by list the manifest CR with kubectl, basically don't know which module it is by first look, maybe get complained later for troubleshooting purpose.

pkg/module/sync/runner_impl.go
jakobmoellersap: I believe the index in this case is unimportant since we just need 0(1) access. We could even just use an array but I didn't want to touch the types too much

pkg/module/common/module.go
jakobmoellersap: I changed the manifests that are generated to have also an annotation for the module name. I would like to have used a label but unfortunately the FQDN is not a RFC compliant label.

All modules can still be received by introspecting the kyma CR.

api/v1alpha1/kyma_types.go
jakobmoellersap: Why can we not keep them in the same order?

pkg/module/sync/runner_impl.go
jakobmoellersap: changed to array indexing

api/v1alpha1/kyma_types.go
jakobmoellersap: changed to array indexing

api/v1alpha1/kyma_types.go
```diff
@@ -173,11 +179,23 @@ const (
 )
 
 type ModuleStatus struct {
-	// Name is the current deployed module name
+	// Name is the current manifest name
 	Name string `json:"name"`
```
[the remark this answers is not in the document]
jakobmoellersap: done

pkg/module/sync/runner_impl.go
```diff
@@ -66,134 +64,119 @@ func (r *RunnerImpl) Sync(ctx context.Context, kyma *v1alpha1.Kyma,
...
 			return fmt.Errorf("error setting owner reference on component CR of type: %s for resource %s %w",
-				name, kyma.Name, err)
+				module.GetName(), kyma.Name, err)
 		}
 	}
 
 	return nil
 }
 
 func (r *RunnerImpl) SyncModuleStatus(ctx context.Context, kyma *v1alpha1.Kyma, modules common.Modules) bool {
```
ruanxin, line 113: since currently we are not using the returned value to determine if need to do status update, we can simplify this logic without the return value.

pkg/module/sync/runner_impl.go
```diff
@@ -66,134 +64,119 @@ func (r *RunnerImpl) Sync(ctx context.Context, kyma *v1alpha1.Kyma,
...
+		return v1alpha1.StateError
 	}
-	return state
 }
 
-func (r *RunnerImpl) deleteNoLongerExistingModuleStatus(ctx context.Context,
-	moduleStatusMap map[string]*v1alpha1.ModuleStatus, kyma *v1alpha1.Kyma,
+func (r *RunnerImpl) deleteNoLongerExistingModuleStatus(ctx context.Context, kyma *v1alpha1.Kyma,
 ) bool {
 	updateRequired := false
```
ruanxin, line 164: we don't need this `updateRequired` as return value

pkg/module/sync/runner_impl.go
```diff
@@ -66,134 +64,119 @@ func (r *RunnerImpl) Sync(ctx context.Context, kyma *v1alpha1.Kyma,
...
+			Template: v1alpha1.TrackingObject{
+				PartialMeta: v1alpha1.PartialMetaFromObject(module.Template),
+				TypeMeta:    metav1.TypeMeta{Kind: templateKind, APIVersion: templateAPIVersion},
 			},
-			State: stateFromManifest(module.Manifest),
 		}
-		moduleStatus, exists := moduleStatusMap[module.Name]
-		if exists {
-			if moduleStatus.State != latestModuleStatus.State {
+		if len(kyma.Status.Modules) < idx+1 {
```
ruanxin, line 140: tbh, I haven't get the point of why needs to distinguish these two cases, but since we don't need to detect if update required, I think we can simplify to only append latestModuleStatus to `kyma.Status.Modules`

jakobmoellersap: if you do an append here instead of indexed access, we will not override the existing status. Thus we need it to make sure we do not simply add more elements to the slice every time

ruanxin (approved):

PR #370, opened 2023-02-01
Replace SKR webhook Helm Chart with raw Manifest

khlifi411:


pkg/deploy/skr_webhook_resources.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+	if err := unstructured.SetNestedStringMap(configMap.Object, configMapData, stringMapDataKey); err != nil {
+		return nil, err
+	}
+	return configMap, nil
+}
+
+func configureDeployment(cfg *unstructuredResourcesConfig, deployment *unstructured.Unstructured,
+) (*unstructured.Unstructured, error) {
+	err := unstructured.SetNestedField(deployment.Object, cfg.secretResVer, "spec", "template", "metadata",
```
jakobmoellersap, line 162: why do you not use the native APIs like v1.Deployment?

khlifi411: Can we unmarshal the yaml to native API structs instead of Unstructured?

jakobmoellersap: yes this is possible. you can also convert from unstructured to the typed one with runtime.DefaultUnstructuredConverter if you prefer this.

flags.go
```diff
@@ -53,10 +53,6 @@ func defineFlagVar() *FlagVar {
 		"Enabling KCP Watcher to reconcile Watcher CRs created by KCP run operators")
 	flag.StringVar(&flagVar.skrWatcherPath, "skr-watcher-path", "./skr-webhook",
 		"The path to the skr watcher chart.")
-	flag.StringVar(&flagVar.skrWebhookMemoryLimits, "skr-webhook-memory-limits", "200Mi",
```
jakobmoellersap, line 56: why were these flags removed? is it not relevant to configure anymore?

pkg/deploy/skr_webhook_resources.go
```diff
@@ -0,0 +1,316 @@
...
+	if resource.GetAPIVersion() == rbacV1.SchemeGroupVersion.String() && resource.GetKind() == "ClusterRoleBinding" {
+		return configureClusterRoleBinding(cfg, resource)
+	}
+	return resource, nil
+}
+
+func closeFileAndLogErr(closer io.Closer, logger logr.Logger, path string) {
+	err := closer.Close()
+	if err != nil {
+		logger.V(int(zap.DebugLevel)).Info("failed to close raw manifest file", "path", path)
```
jakobmoellersap, line 255: could you use the debug level constant that we introduced instead here?

pkg/deploy/skr_webhook_resources.go
```diff
@@ -0,0 +1,316 @@
...
+	if cfg == nil {
+		return nil, ErrExpectedNonNilConfig
+	}
+	manifestFilePath := fmt.Sprintf(rawManifestFilePathTpl, chartPath)
+	rawManifestFile, err := os.Open(manifestFilePath)
+	if err != nil {
+		return nil, err
+	}
+	defer closeFileAndLogErr(rawManifestFile, logger, rawManifestFile.Name())
+	decoder := yaml.NewYAMLOrJSONDecoder(rawManifestFile, defaultBufferSize)
```
jakobmoellersap, line 270: Why do you choose to use resources.yaml here instead of inlining the files as the proper API objects (e.g. a v1 deployment)? That way you don't have to do this yaml unstructuring and also don't have to use the runtime converter.

khlifi411: I am not sure if I fully understand the question. do you suggest to remove the resources.yaml and generate the API objects in the code?

jakobmoellersap: yes exactly, that's also what you're doing with some resources already I believe

flags.go
jakobmoellersap: I believe these flags are relevant for production and may need dynamic adjustments, so this is something that should be still settable without recompiling imo. Wdyt?

pkg/deploy/skr_webhook_resources.go
khlifi411: Yes. I chose to use the yaml manifest only to avoid having a lot of code written for the sole purpose of API object generation. Do you think it's a valid concern?

flags.go
khlifi411: I think you're right. I will reintroduce them then 👍

pkg/deploy/skr_webhook_resources.go
jakobmoellersap: I understand the concern. I believe the correct way to solve this is then to prepare the objects in advance instead of looking them up from the file system and parsing them on every reconcile. Other than that I think you can go ahead with that approach. I just want to avoid unnecessary effort in the reconcile loop

khlifi411: Makes sense. I will load them from the file system when the manager is created and configure them only when calling `Install()` 👍

controllers/kyma_controller.go
```diff
@@ -235,8 +235,8 @@ func (r *KymaReconciler) HandleProcessingState(ctx context.Context, kyma *v1alph
 	}
 	kyma.UpdateCondition(conditionReason, conditionStatus)
 
-	if kyma.Spec.Sync.Enabled && r.SKRWebhookChartManager != nil {
-		if _, err := r.SKRWebhookChartManager.Install(ctx, kyma); err != nil {
+	if kyma.Spec.Sync.Enabled && r.SKRWebhookManager != nil {
+		if err := r.SKRWebhookManager.Install(ctx, kyma); err != nil {
 			kyma.UpdateCondition(v1alpha1.ConditionReasonSKRWebhookIsReady, metav1.ConditionFalse)
 			return err
```
jakobmoellersap, line 241:
```suggestion
			return r.UpdateStatusWithEventFromErr(ctx, kyma, v1alpha1.StateError, err)
```

docs/developer/watcher-local-test.md
```diff
@@ -1,34 +1,34 @@
...
-   
+
 1. Create a local control-plane (KCP) cluster:
-    ```shell
-    k3d cluster create kcp-local --port 9080:80@loadbalancer \
-    --registry-create k3d-registry.localhost:0.0.0.0:5111 \
-    --k3s-arg '--no-deploy=traefik@server:0'
-    ```
-   
+   ```shell
```
jakobmoellersap, line 6: is it possible to replace this with a kyma provision k3d command?

docs/developer/watcher-local-test.md
```diff
@@ -37,28 +37,33 @@
...
 ### Create TLS secret
-Run the following commands from your local directory for the [kyma watcher repo](https://github.com/kyma-project/runtime-watcher) to generate PKI for the SKR webhook
-1. Run the following commands to generate the PKI for the watcher web-hook 
+
+Run the following commands from your local directory for
+the [kyma watcher repo](https://github.com/kyma-project/runtime-watcher) to generate PKI for the SKR webhook
+
+1. Run the following commands to generate the PKI for the watcher web-hook
     ```shell
     cd listener/example-mtls && \
```
jakobmoellersap, line 60:
```suggestion
    cd ../runtime-watcher/listener/example-mtls && \
```

docs/developer/watcher-local-test.md
```diff
@@ -69,27 +74,28 @@ Run the following commands from your local directory for the [kyma watcher repo]
...
       labels:
         "operator.kyma-project.io/managed-by": "lifecycle-manager"
     data:
-      ca.crt: $(cat ca.pem | base64)
-      tls.crt: $(cat signed-cert.pem | base64)
-      tls.key: $(cat signed-cert-key.pem | base64)
+      ca.crt: $(cat ca.pem | base64 | tr -d '\n')
+      tls.crt: $(cat signed-cert.pem | base64 | tr -d '\n')
+      tls.key: $(cat signed-cert-key.pem | base64 | tr -d '\n')
     EOF
```
jakobmoellersap, line 85: This EOF I think is somehow malindented, I had to manually adjust this script to make it run in my shell.

docs/developer/watcher-local-test.md
```diff
@@ -69,27 +74,28 @@ Run the following commands from your local directory for the [kyma watcher repo]
...
+```shell
+k3d cluster create skr-local && \
 kubectl create namespace kcp-system
 ```
 
-
 ### Create Kyma CR and remote secret
+
 1. Switch the context for using KCP cluster:
     ```shell
```
jakobmoellersap, line 100: This EOF I think is somehow malindented (and I could only apply the resources one after another), I had to manually adjust this script to make it run in my shell.

jakobmoellersap (changes requested):
Just some smaller changes, other than that LGTM!

jakobmoellersap (dismissed):

docs/developer/watcher-local-test.md
khlifi411: Reverted back to using k3d because `kyma provision k3d` has some bugs.
I created an issue on the cli repo https://github.com/kyma-project/cli/issues/1510

jakobmoellersap (approved):
Great work!

jakobmoellersap (approved):

PR #403, opened 2023-02-17
Istio smoke test setup

jakobmoellersap:
This heavily simplifies the smoke tests and adds a `control-plane` variety which also uses Istio and the service monitor CRD so that we can integrate into a managed simulated offering

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,9 +52,9 @@ jobs:
     strategy:
       matrix:
         cli-stability: ["unstable"]
-        smoke-run: [ "${{ github.repository }}" ]
+        target: ["default", "control-plane"]
     needs: [wait-for-img, kustomize]
-    name: "Kyma CLI - ${{ matrix.cli-stability }} - Integration (Tests of ${{ matrix.smoke-run }})"
+    name: "Kyma CLI - ${{ matrix.cli-stability }} - Integration (Tests of ${{ github.repository }})"
     runs-on: ubuntu-latest
     env:
       LIFECYCLE_MANAGER: ${{ github.repository }}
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -98,18 +91,29 @@ jobs:
...
-          kyma --ci alpha deploy \
-            -k $(go env GOPATH)/src/github.com/$LIFECYCLE_MANAGER/config/default
-      - name: Inject GOPATH into go Setup
-        run: echo "GOPATH=$(go env GOPATH)" >> $GITHUB_ENV
-      - name: Set up Go for ${{ matrix.smoke-run }}
+          curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+          chmod +x kubectl
+          mv kubectl /usr/local/bin
+          
+          curl -L https://istio.io/downloadIstio | TARGET_ARCH=x86_64 sh -
```
ruanxin, line 101: do we need to provide an ISTIO_VERSION here? as described in the this downloadIstio script
```
 curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.6.8 TARGET_ARCH=x86_64 sh -.
```

ruanxin: we should also deploy cert manager, or? The problem we faced in DEV is because the istio sidecar blocked the requestion coming from apiserver for the admission webhook, right?

jakobmoellersap: the cert manager is already deployed in kyma alpha deploy.
the ISTIO_VERSION is already provided from the environment variables so you can observe that in the logs

ruanxin (approved):

PR #412, opened 2023-02-21
Fix watcher setup

jeremyharisch:


jeremyharisch:
Will be fixed in https://github.com/kyma-project/lifecycle-manager/pull/429/files

PR #407, opened 2023-02-20
fix: remove bad naming scheme from `operator.kyma-project.io/owned-by`

jakobmoellersap:
The naming scheme of `operator.kyma-project.io/owned-by` causes errors during SSA as the naming can get too long:

```
Warning  ServerSideApply  4m1s  declarative.kyma-project.io/events  ServerSideApply failed (after 34.811534ms): patch for clusterrolebindings/istio-manager-rolebinding failed: ClusterRoleBinding.rbac.authorization.k8s.io "istio-manager-rolebinding" is invalid: metadata.labels: Invalid value: "kcp-system__226094f5-ac2a-4ffc-bfd2-1d7d41185286-istio-4034478325": must be no more than 63 characters
```

PR #413, opened 2023-02-21
Refactor labels pkg and add owned-by annotation

jeremyharisch:
**Description**

This [PR](https://github.com/kyma-project/lifecycle-manager/pull/407/files#diff-7b7b8e548274819827296add582c47d013f039f15b3502e0944dde5a9a0b326a) changed the format of the `OwnedBy` label to fix the problem of reaching the limit of label lengths. But this broke the functionality of the runtime-watcher, since it needs the ns and name of the corresponding resource.
Thus, a new format will be introduced with this PR and instead of putting it as a label to the resource, we will use annotations instead, since it it not used for any filtering and annotations have a bigger limit (256 kB) than labels.

Furthermore, I removed the labels pkg since those where duplicates of labels which already have been implemented in other packages (v1beta). Thus, instead of using the `labels` pkg `v1beta1` etc. will be used.


controllers/manifest_controller.go
```diff
@@ -42,7 +41,7 @@ func SetupWithManager(
 	}
 
 	runnableListener, eventChannel := listener.RegisterListenerComponent(
-		settings.ListenerAddr, strings.ToLower(labels.OperatorName), verifyFunc,
+		settings.ListenerAddr, strings.ToLower(declarative.OperatorName), verifyFunc,
```
jakobmoellersap, line 44: I think this is a different operator name than the one coming from labels.

jeremyharisch: As you can see [here](https://github.com/kyma-project/lifecycle-manager/pull/413/files#diff-a2c2307a5750317022e9cecac20fba60845972de6f2e6ec6d1f03492a722d018L9) it was set to `module-manager`. And it is now still set to `module-manager` ([here](https://github.com/kyma-project/lifecycle-manager/pull/413/files#diff-7b7b8e548274819827296add582c47d013f039f15b3502e0944dde5a9a0b326aR11))

This label has just usages in the declarative, since they are already included in the v1beta1 pkg. Thus, I moved the duplicates to the declarative pkg to have a better understanding what it is being used for and to avoid confusion. Unfortunately, we cannot use it directly from v1beta1 due to cyclic dependencies.

jakobmoellersap (approved):

api/v1beta1/operator_annotations.go
```diff
@@ -3,4 +3,10 @@ package v1beta1
 const (
 	LastSync = OperatorPrefix + Separator + "last-sync"
 	FQDN     = OperatorPrefix + Separator + "fqdn"
+
+	// OwnedByAnnotation defines the resource managing the resource. Differing from ManagedBy
+	// in that it does not reference controllers. Used by the runtime-watcher to determine the
```
Tomasz-Smelcerz-SAP, line 8:
```suggestion
	// OwnedByAnnotation defines the resource managing the resource. unlike the  'v1beta1.ManagedBy' label
	// it doesn't reference controllers. Used by the runtime-watcher to determine the
```

pkg/declarative/v2/default_transforms.go
```diff
@@ -2,17 +2,30 @@ package v2
...
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
 const (
-	DisclaimerAnnotation      = "reconciler.kyma-project.io/managed-by-reconciler-disclaimer"
+	OperatorName   = "module-manager"
+	OperatorPrefix = "operator.kyma-project.io"
+	Separator      = "/"
+
+	ManagedByLabel      = OperatorPrefix + Separator + "managed-by"
```
Tomasz-Smelcerz-SAP, line 15: This has changed. Any impact on the environment?

PR #414, opened 2023-02-21
Fix v1alpha1 validation webhook get called unintentionally

ruanxin:
If deployed v1alpha1 webhook and enabled admission webhook, even if controller runtime only handles v1beta1 CR, the validation webhook still get called, which is unnecessary.

This fix removed v1alpha1 webhook since v1alpha1 resource will convert to v1beta1 first, it's not necessary to validate v1alpha1 CR anymore.


config/manager/kustomization.yaml
```diff
@@ -15,5 +15,5 @@ configMapGenerator:
 
 images:
 - name: controller
-  newName: eu.gcr.io/kyma-project/lifecycle-manager
-  newTag: latest
+  newName: ghcr.io/ruanxin/lifecycle-manager
```
jakobmoellersap, line 18: this is an accidental commit I believe

config/samples/operator_v1alpha1_manifest.yaml
```diff
@@ -18,12 +18,12 @@ spec:
     spec:
       randomkey: samplevalue
   installs:
-#    - source:
-#        name: kyma.project.io/module/kyma-redis
-#        repo: europe-west3-docker.pkg.dev/sap-kyma-jellyfish-dev/operator-demo
-#        ref: sha256:cd98ce9b440fef2e1d11fe382ccf6fab971f516efa9ee174d73209777ed318b5
-#        type: oci-ref
-#      name: kyma-redis-operator
+    - source:
```
jakobmoellersap, line 21: this manifest is used in samples, are we sure this is okay to remove comments here?

config/manager/kustomization.yaml
ruanxin: for sure 😅

config/samples/operator_v1alpha1_manifest.yaml
ruanxin: committed unintentionally, it's used for me locally to test if manifest convert works or not. It indeed rejects this manifest cr due to `ErrSingleInstallOnly`

PR #417, opened 2023-02-22
OCM migration

jakobmoellersap:
Migrates all existing schemes to be compatible with the new OCM schemas. Uses the new OCM repository for compatibility. Now also supports v3 style schemas. Also removes internal memory struct by using runtime.RawExtensions internal Object instead. Still needs a way to call the convert method reliably so that the RawExtension is converted to a Descriptor instead.


config/samples/component-integration-installed/operator_v1beta1_moduletemplate_ocm.software.v3alpha1.yaml
```diff
@@ -0,0 +1,101 @@
...
+      name: kyma-project.io/my-module
+      provider:
+        labels:
+          - name: kyma-project.io/built-by
+            value: cli
+            version: v1
+        name: kyma-project.io
+      version: v0.0.1
+    repositoryContexts:
+      - baseUrl: ghcr.io/jakobmoellersap/testmodule
```
lindnerby, line 37: Can we find a reg location for samples or the team?

lindnerby (approved):

Issue #380, opened 2023-02-09
Set up monitoring for Kyma sync status


janmedrek:
**Description:**
Monitoring and metrics should be set up to reflect the sync status of Kyma resources that the Lifecycle Manager manages. This will provide an easily accessible overlook of the overall state of the Kyma Control Plane.

Alert rules will be set up as the follow-up ticket.

**Acceptance Criteria:**
- [x] Lifecycle Manager exposes metrics regarding the number of Kyma resources (gauge type)
- [x] Lifecycle Manager exposes metrics regarding the Kyma sync statuses (gauge type)
- [x] Lifecycle Manager exposes metrics representing how many Kyma CRs are in Error state and for how long - it would be nice to see which Kyma CRs are in error state for longer than X hours/days (grafana queries can provide the solution)
- [x] Metrics included in [the Grafana dashboard for the Lifecycle Manager](https://github.com/kyma-project/lifecycle-manager/tree/main/config/grafana) (propose the visualization)

ebensom:
The lifecycle-manager should expose for each Kyma CR (unique labels `kyma_name`, `instance_id`, and possibly `shoot` in all exposed metrics):

- The current value `.status.state` field. It could be exposed as a gauge mapping string states to numeric values, or reflecting it as boolean  0 or 1 metric values, where the state is a label, e.g. 
    ```
    lifecycle_mgr_kyma_state{kyma_name="xxxx", instance_id="yyyyy", state="Ready"} = 1 
    lifecycle_mgr_kyma_state{kyma_name="xxxx", instance_id="yyyyy", state="Progressing"} = 0
    lifecycle_mgr_kyma_state{kyma_name="xxxx", instance_id="yyyyy", state="Error"} = 0
    ```
- The current value of `.status.modules[*].state` fields for all modules (as gauge, or similarly as described above)

With the exposed metrics, we should be able to cover, and get notified about 4 scenarios:

1. When lifecycle-manager or module-manager fails to reconcile a change (during module enablement, or module manifest version update) of a module operator on the runtime
2. When lifecycle-manager or module-manager fails to properly delete Kyma module during module deactivation.
3. When lifecycle-manager or module-manager fails to properly delete Kyma module(s) during Kyma deletion (could also be triggered by KEB during deprovisioning) 
4. When module operator fails to reconcile the module resources locally on the runtime (e.g. because of customer caused issue).
[...]

PR #418, opened 2023-02-23
Introduce custom Kyma metrics

Tomasz-Smelcerz-SAP:
Introduce custome metrics for tracking Kyma status.
See: https://github.com/kyma-project/lifecycle-manager/issues/380

### DIFF
--- a/controllers/kyma_controller.go
+++ b/controllers/kyma_controller.go
@@ -20,9 +20,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/kyma-project/lifecycle-manager/pkg/log"
+	"github.com/kyma-project/lifecycle-manager/pkg/metrics"
 	"k8s.io/client-go/rest"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -428,4 +430,34 @@ func (r *KymaReconciler) deleteModule(ctx context.Context, moduleStatus *v1beta1
 	manifest.SetNamespace(moduleStatus.Manifest.GetNamespace())
 	manifest.SetName(moduleStatus.Manifest.GetName())
 	return r.Delete(ctx, &manifest, &client.DeleteOptions{})
+}
+
+// RecordKymaStatusMetrics updates prometheus metrics defined to track changes to the Kyma status.
+func (r *KymaReconciler) RecordKymaStatusMetrics(kyma *v1beta1.Kyma) {
+	shoot := ""
+	shootFQDN, keyExists := kyma.Annotations[v1beta1.SKRDomainAnnotation]
+	if keyExists {
+		parts := strings.Split(shootFQDN, ".")
+		// at least three labels in any real deployment.
+		if len(parts) > 2 { //nolint:gomnd
+			shoot = parts[0] // hostname
+		}
+	} else {
+		r.EventRecorder.Eventf(
+			kyma, "Warning", "AnnotationNotDefined",
+			"Expected annotation: %s not found.",
+			v1beta1.SKRDomainAnnotation,
+		)
+	}
+
+	instanceID, keyExists := kyma.Labels[v1beta1.InstanceIDLabel]
+	if !keyExists {
+		r.EventRecorder.Eventf(
+			kyma, "Warning", "LabelNotDefined",
+			"Expected label: %s not found.",
+			v1beta1.InstanceIDLabel,
+		)
+	}
+
+	metrics.RecordKymaStatus(kyma.Name, kyma.Status.State, shoot, instanceID)
 }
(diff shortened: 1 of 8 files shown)


controllers/kyma_controller.go
```diff
@@ -429,3 +430,12 @@ func (r *KymaReconciler) deleteModule(ctx context.Context, moduleStatus *v1beta1
 	manifest.SetName(moduleStatus.Manifest.GetName())
 	return r.Delete(ctx, &manifest, &client.DeleteOptions{})
 }
+
+// RecordKymaStatusMetrics updates prometheus metrics defined to track changes to the Kyma status
+func (r *KymaReconciler) RecordKymaStatusMetrics(kyma *v1beta1.Kyma) {
+	//TODO: where to take "instanceId" and "shoot" from?
```
jakobmoellersap, line 436: they come from labels in the runtime, please see our managed clusters and observe what KEB is introducing there.

controllers/kyma_controller.go
```diff
@@ -429,3 +430,12 @@ func (r *KymaReconciler) deleteModule(ctx context.Context, moduleStatus *v1beta1
...
+	err := metrics.RecordKymaStatus(kyma.Name, string(kyma.Status.State), "instanceId", "shoot")
```
lindnerby, line 437: In my opinion the only thing the whole metric setting code needs to boil down to is e.g. if Kyma is in state 'Ready':
```
statusMetric.With(prometheus.Labels{"kyma_name":kyma.Name, "state":"Ready"}).Set(1)
statusMetric.With(prometheus.Labels{"kyma_name":kyma.Name, "state":"Error"}).Set(0)
statusMetric.With(prometheus.Labels{"kyma_name":kyma.Name, "state":"Processing"}).Set(0)
statusMetric.With(prometheus.Labels{"kyma_name":kyma.Name, "state":"Deleting"}).Set(0)
```
so basically every time a Kyma gets reconciled you create new values, since it's important **when** a Kyma turned from Ready -> Processing -> Error for example.

Have you tried this or am I missing some requirement?

Tomasz-Smelcerz-SAP: @lindnerby I think this is exactly what I am doing. Take a look [here](https://github.com/kyma-project/lifecycle-manager/pull/418/files#diff-4bc91a74bc47b8467cdcd25a7e3fe1651dbe44907f478377eca31b982f444f23R33)

lindnerby: @Tomasz-Smelcerz-SAP Thanks for the sync. As discussed we:
* should only set labels `name` and `state`
* should probably align with controller-runtime's metric handling implementation and have an init for default values
* as well as may opt into using the previously mentioned method `With(Labels)` over `GetMetricWith(Labels)` to panic on dev errors to _fail-fast_ e.g. for mismatching label cardinality. But only after verifying that the errors are purely about label format, structure, etc. and can not stem from dynamic values from the CR, since controller-runtime is doing that as well. Don't know if we want to risk create only a warning event at runtime if a developer is making an error at some point in the future when e.g. adding new labels, new metrics, etc. Also no metrics would be updated anymore until someone notices that warning events are created.
* maybe for readability you could just iterate over all available statuses and set to 1 if Kyma status matches and to 0 if not

controllers/kyma_controller.go
Tomasz-Smelcerz-SAP: @lindnerby is on checking if these labels are automatically added to any metrics. In that case we don't have to do anything. For now I removed these from the code.

controllers/kyma_controller.go
Tomasz-Smelcerz-SAP: @lindnerby The errors are only related to number (and names) of labels, so they don't depend on runtime values. It means we can use a "fail-fast" approach, as the controller-runtime does. I've also introduced similar metrics for the modules statuses. Please take a look

Tomasz-Smelcerz-SAP:
Putting on-hold until we know for sure if `shoot` and `instanceId` labels are injected automatically into our metrics.

controllers/kyma_controller.go
```diff
@@ -429,3 +430,13 @@ func (r *KymaReconciler) deleteModule(ctx context.Context, moduleStatus *v1beta1
...
 	return r.Delete(ctx, &manifest, &client.DeleteOptions{})
 }
+
+// RecordKymaStatusMetrics updates prometheus metrics defined to track changes to the Kyma status
+func (r *KymaReconciler) RecordKymaStatusMetrics(kyma *v1beta1.Kyma) {
+
+	metrics.RecordKymaStatus(kyma.Name, kyma.Status.State)
+
+	for _, ms := range kyma.Status.Modules {
+		metrics.RecordKymaModuleStatus(kyma.Name, ms.Name, ms.State)
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Fine for me.

controllers/kyma_controller.go
lindnerby: As discussed, we can get these from the Kyma CR.

lindnerby (approved):

PR #421, opened 2023-02-23
Remove insecure flag

ruanxin:
At the moment, the module controller use a flag to determine if remote image repo is secured https or not, which means it can't change after component is initialized.

This PR changed the `pullLayer` function to first try with secure way, if failed, fallback into insecure, so that we can get rid of this global static flag to determine if remote repo is secured.


jakobmoellersap (commented):
The reason why this was put as a flag is because it is considered insecure to fallback to http connections, exposing a vulnerability in concept. Thus this flag is start-time and can never be changed once the controller is running. I believe this should stay that way, even though it might be slightly more inconvenient. WDYT?

ruanxin:
I underst

> The reason why this was put as a flag is because it is considered insecure to fallback to http connections, exposing a vulnerability in concept. Thus this flag is start-time and can never be changed once the controller is running. I believe this should stay that way, even though it might be slightly more inconvenient. WDYT?

I understood the concern, but with the current implementation, if by default the configuration is correct, it will not fallback to http, maybe I can continue to improve the error handling a bit more specific, for example, if the error is related to authentication error, then not do the fallback.

If this is really a problem, I still think we should provide an alternative solution that the controller can detect the repo dynamically, for example, `baseUrl` must provide `http://` as prefix to indicate it's not a secured host, so that by default the connection should always consider as https, but not statically defined during the start time. How about this? @jakobmoellersap

jakobmoellersap:
In general, this should still be too insecure as to even consider HTTP we would need to accept insecure connections which is not nice. I think it is totally okay to error out by default and use HTTPS only out of the box. everything else will result in us sometimes having to accept HTTP as transport which we do not want out of the box. So even if you use "http://" as a prefix, it would be the same as defaulting to HTTP for me.

jakobmoellersap (commented):
LGTM but has a conflict

PR #429, opened 2023-02-28
Fix local setup guide, update dependencies, set Watcher status as sub…

jeremyharisch:
**Description:**
Due to the merge of the KLm and KMM into on repository and binary, the local setup for the Watcher was broken. Thus, I took some time to fix it and enhance it with some features, plus removed a guide which was outdated and at this point a duplicate to the local setup guide with the watcher.

**Changes made:**
- Set the reference in the main README to the new local setup guide
- Deleted old setup guide (`creating-test-environment.md`)
- Wrote new local setup guide; used old Watcher setup guide as basis
- Added `status` in Watcher CRD as subresource
- Fixed prefixes in local kustomize setup
- Named the ports of the Watcher service, since after merging KLM and KMM we have two ports and thus they need to be named; otherwise Service cannot be created
- Use `default` kustomize as base for local setup, to have an easified kustomize setup with less patches
- Fixed patches to make local setup work again

**For Reviewer:**
When going through the local setup guide and trying it out by yourself, you will encounter two problems:
- In your local SKR cluster the pod of the Template Operator will not be able to pull the image, since it does not exists. This, is not a problem of this PR and can be ignored, since the module itself got deployed and that is the goal of this PR. To correct the image, a PR has been created: https://github.com/kyma-project/template-operator/pull/27 (it does not matter which one gets merged first)
- When checking your Kyma CR on your local k3d cluster, you will see that the template operator module is in `Processing`, this is due to the first bug - BUT. the Kyma status itself is already set to `Ready` - this should not be the case and needs to be resolved in a follow up issue: TODO LINK ISSUE


config/watcher_local_test/kustomization.yaml
```diff
@@ -3,24 +3,26 @@
...
-  - ../rbac
   - ../watcher
   - ../istio
 
-namespace: kcp-system
 
 generatorOptions:
   disableNameSuffixHash: true
 
+# TODO REMOVE this nameprefix and adapt whole watcher setup -> everything should have `lifecycle-manager-` as prefix
```
khlifi411, line 18: Can we apply this TODO in this PR?

khlifi411 (commented):
Tested the setup locally and it works flawlessly 🔥
Left one comment about the To-dos.

config/watcher_local_test/kustomization.yaml
jeremyharisch: Ups,...leftover TODO - as you can see this was already done

khlifi411 (approved):

PR #445, opened 2023-03-06
rework Conditions

jeremyharisch:


api/v1beta1/kyma_types.go
```diff
@@ -375,28 +366,29 @@ func init() {
 	SchemeBuilder.Register(&Kyma{}, &KymaList{})
 }
 
-func (kyma *Kyma) UpdateCondition(reason KymaConditionReason, status metav1.ConditionStatus) {
+func (kyma *Kyma) UpdateCondition(conditionType KymaConditionType, status metav1.ConditionStatus) {
```
jakobmoellersap, line 369: be aware: breaking change for existing kymas, so backwards check has to be done with v1beta1 from before and after! this also needs to be back ported to `v1alpha1`

jakobmoellersap (commented):
Overall LGTM, but v1alpha1 back port is missing

api/v1beta1/kyma_types.go
jeremyharisch: @jakobmoellersap I agree with the breaking change. Means we should remove all conditions which have Type `Ready`, since with the old KLM a module could have been in error state, then new KLM will be deployed, module gets ready but will have a new own conditions. Thus, an orphaned condition `Ready` = false will be left. Such condition should be removed.

But why do we need a backport? Since this does not change the api of Kyma v1alpha1, but also fixes a bug of it. The Kyma CRD did not change at all

jakobmoellersap: understood. yes that makes sense we should probably introduce a compat package for this and then note that it should be removed after the next apiversion

jeremyharisch: I will set the `Conditions` to nil, when converting from or to `v1alpha1`, since the Conditions will not be sued for STate Handling, but they will be updated in each reconcile loop. Thus, when converting from `v1alpha1` the one `Ready` condition will be removed, and replaced by the three current in the first reconcile loop

jakobmoellersap: But this only takes care of APIVersion upgrades. What if there is an existing v1beta1 CR that already has the wrong state?

api/v1beta1/condition_messages.go
```diff
@@ -4,31 +4,28 @@ import (
...
-	ConditionReasonSKRWebhookIsReady    KymaConditionReason = "SKRWebhookIsReady"
+	ConditionTypeModulesAreReady      KymaConditionType = "ModulesAreReady"
+	ConditionTypeModuleCatalogIsReady KymaConditionType = "ModuleCatalogIsReady"
+	ConditionTypeSKRWebhookIsReady    KymaConditionType = "SKRWebhookIsReady"
 )
 
-func GenerateMessage(reason KymaConditionReason, status metav1.ConditionStatus) string {
-	switch reason {
-	case ConditionReasonModulesAreReady:
+func GenerateMessage(conditionType KymaConditionType, status metav1.ConditionStatus) string {
```
ruanxin, line 26: Just left my proposal here for reference:
IMO, this `GenerateMessage` function is fine, we don't need to change it, the logic here is a bit complicate I agree, but it's designed solid, and it's not this generate messge cause problem.
The root cause of the current problem is we don't have different conditionType.
Why not just design three conditiontype one-to-one mapped with reason?

for example: ConditionTypeModules (`Modules`) <-> ConditionReasonModulesAreReady (`ModulesAreReady`)
Then in the existing `UpdateCondition`, we also don't need to change any signatures, still use `reason KymaConditionReason`, the only change need to do is just add a function to convert reason to type.
```
Type:               convertToType(reason)
```

jakobmoellersap: Honestly imo the entire file is pretty much over engineered and we would be fine with something like this:

```suggestion

type ConditionType string

const (
	ConditionTypeModules    ConditionType = "Modules"
)

type ConditionReason string

const (
	ConditionReasonModulesAvailable ConditionReason = "Available"
)

type ConditionMessage string

const (
       ConditionMessageModulesAvailable ConditionMessage = "modules are synchronised and available"
)


func newResourcesCondition(obj metav1.Object) metav1.Condition {
	return metav1.Condition{
		Type:               string(ConditionTypeModules),
		Reason:             string(ConditionReasonModulesAvailable),
		Message:            string(ConditionMessageModulesAvailable),
		Status:             metav1.ConditionFalse,
		ObservedGeneration: obj.GetGeneration(),
	}
}
```

jeremyharisch: I agree with @ruanxin at this point.
We had a couple of discussion and we agreed on the following format:
```
metav1.Condition{
	Type:               Can be `Modules`, `ModuleCatalogue`, `SKRWebhook`
	Status:             True, False or Unknown,
	Reason:             `Ready`,
	Message:            Message which actually represents the current state in human readable format,
	ObservedGeneration: kyma.GetGeneration(),
}
```

pkg/status/conditions.go
```diff
@@ -0,0 +1,32 @@
...
+// SetConditions checks if all required conditions do exists and removes deprecated Conditions.
+// If one or many Conditions are missing, they will be created and `true` will be returned
+// to determine that an update is needed.
+// If all required conditions exist, they will be set to `False`.
+func SetConditions(kyma *v1beta1.Kyma, watcherEnabled bool) bool {
+	conditionUpdateNeeded := false
+
+	existingConditions := kyma.Status.Conditions
+	// Remove deprecated `Ready` condition
+	if deprecatedCondition := apimeta.FindStatusCondition(existingConditions,
```
ruanxin, line 18: please add a TODO here to indicate remove this tech debt code later, and do we have a backlog issue related to this? If no, please create one.

ruanxin: consider rename this `DeprecatedConditionReady` to `DeprecatedConditionTypeReady`

controllers/kyma_controller.go
```diff
@@ -104,6 +102,13 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		return ctrl.Result{}, client.IgnoreNotFound(err) //nolint:wrapcheck
 	}
 
+	if status.SetConditions(kyma, r.WatcherEnabled(kyma)) {
```
ruanxin, line 105: Please consider renaming it to `InitConditions`, set method is misleading here, the function only initializes conditions, it's not for generic set condition purposes.

pkg/status/conditions.go
```diff
@@ -0,0 +1,32 @@
...
+		string(v1beta1.DeprecatedConditionReady)); deprecatedCondition != nil {
+		apimeta.RemoveStatusCondition(&existingConditions, string(v1beta1.DeprecatedConditionReady))
```
ruanxin, line 20: the `existingConditions` do not point to the original condition, when you do the assignment `existingConditions := kyma.Status.Conditions`,  the value gets copied, so the remove here will not remove actual `kyma.Status.Conditions`. Please also write a test for it after the fix.

controllers/kyma_controller.go
```diff
@@ -104,6 +102,13 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+		if err := r.UpdateStatus(ctx, kyma, kyma.Status.State, "Initialising Conditions"); err != nil {
```
ruanxin, line 106: Here you don't need to update the status immediately, just need to have an initialized conditions, in the `stateHandling`, update status will happen eventually anyway.

controllers/kyma_controller.go
```diff
@@ -232,25 +235,24 @@ func (r *KymaReconciler) HandleInitialState(ctx context.Context, kyma *v1beta1.K
...
 		if moduleStatus.State != v1beta1.StateReady {
-			conditionStatus = metav1.ConditionFalse
+			allModulesReady = false
 			break
 		}
 	}
-	kyma.UpdateCondition(conditionReason, conditionStatus)
 
-	if kyma.Spec.Sync.Enabled && r.SKRWebhookManager != nil {
+	if allModulesReady {
```
ruanxin, line 250: Please conside extract the `allModulesReady` related code into a function.

controllers/kyma_controller_test.go
```diff
@@ -262,3 +279,27 @@ func updateModuleTemplateSpecData(kymaName, valueUpdated string) func() error {
...
+func CheckKymaConditions(ctx context.Context, kcpClient client.Client, kymaName string,
+	requiredConditions []v1beta1.KymaConditionType) func() bool {
+	return func() bool {
+		kymaFromCluster, err := GetKyma(ctx, kcpClient, kymaName, "")
+		if err != nil || len(kymaFromCluster.Status.Conditions) != len(requiredConditions) {
+			return false
+		}
+
+		for _, conditionType := range requiredConditions {
+			exists := false
```
ruanxin, line 292: This logic has flaws, what you implement here is just check if all `requiredConditions` is a subset of `Status.Conditions`, but what you want to check is if requiredCondition is same as `Status.Conditions`, you need to prevent unwanted condition not exists.

Take your test as an example, you can't make sure if that `Deprecated` condition has been removed or not.

ruanxin (changes requested):
Please check my comments.

controllers/kyma_controller_test.go
ruanxin: I overlooked the precondition that the length of two conditions will also compare, thanks @LeelaChacha point it out. The logic is no problem then (however the logic can still be improved, for example, use hashmap to save requiredConditions type as key, then iterating `status.conditions` to check if all key exists, that's the normal way to compare if two unsorted arrays are equal)

The actual reason why that `Deprecated` condition can't be detected is due to the condition initialization having problem (this made-up condition never applied to kyma CR actually), this will be fixed in the follow-up commit.

PR #446, opened 2023-03-06
manifest reconciler can't handle different target module managed by same kyma

ruanxin:
At the moment, for the performance perspective, we cached remote cluster client, and the cache key is using kyma name, so each module defined under same kyma will shared same cluster client. 

However, if modules target are different, for example, some of them defined in control-plane, some in remote (although in practice, we don't have this scenario), then the client should be different.

The PR fixes this bug by providing a different cache key generation logic.

### DIFF
--- a/internal/manifest/v1beta1/client.go
+++ b/internal/manifest/v1beta1/client.go
@@ -66,31 +66,22 @@ func WithClientCacheKey() declarative.WithClientCacheKeyOption {
 	cacheKey := func(ctx context.Context, resource declarative.Object) (any, bool) {
 		logger := log.FromContext(ctx)
 
-		if resource == nil {
-			return nil, false
-		}
 		manifest := resource.(*manifestv1beta1.Manifest)
 		labelValue, err := internal.GetResourceLabel(resource, manifestv1beta1.KymaName)
 		objectKey := client.ObjectKeyFromObject(resource)
 		var labelErr *types.LabelNotFoundError
 		if errors.As(err, &labelErr) {
+			logger.V(internal.DebugLogLevel).Info(
+				"client can not been cached due to lack of expected label",
+				"resource", objectKey)
 			return nil, false
 		}
-		cacheKey := generateCacheKey(labelValue, strconv.FormatBool(manifest.Spec.Remote), manifest.GetNamespace())
-		logger.V(internal.DebugLogLevel).Info(
-			"resource will be cached",
-			"resource", objectKey,
-			"cachekey", cacheKey,
-		)
+		cacheKey := GenerateCacheKey(labelValue, strconv.FormatBool(manifest.Spec.Remote), manifest.GetNamespace())
 		return cacheKey, true
 	}
 	return declarative.WithClientCacheKeyOption{ClientCacheKeyFn: cacheKey}
 }
 
-func generateCacheKey(values ...string) string {
-	var sb strings.Builder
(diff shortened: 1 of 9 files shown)


internal/manifest/v1beta1/client.go
```diff
@@ -54,3 +61,36 @@ func (cc *ClusterClient) GetRESTConfig(
 	}
 	return restConfig, err
 }
+
+func WithClientCacheKey() declarative.WithClientCacheKeyOption {
+	cacheKey := func(ctx context.Context, resource declarative.Object) (any, bool) {
+		logger := log.FromContext(ctx)
+
+		if resource == nil {
```
jakobmoellersap, line 69: why is this case necessary?

internal/manifest/v1beta1/client.go
```diff
@@ -54,3 +61,36 @@ func (cc *ClusterClient) GetRESTConfig(
...
+			"resource will be cached",
+			"resource", objectKey,
+			"cachekey", cacheKey,
+		)
+		return cacheKey, true
+	}
+	return declarative.WithClientCacheKeyOption{ClientCacheKeyFn: cacheKey}
+}
+
+func generateCacheKey(values ...string) string {
```
jakobmoellersap, line 90: maybe use `strings.Join()` instead of a custom string builder

jakobmoellersap (commented):
Overall LGTM, just a few nits

internal/manifest/v1beta1/client.go
ruanxin: not necessary after a second look, just take from the previous code 😄

internal/manifest/v1beta1/client.go
```diff
@@ -54,3 +61,36 @@ func (cc *ClusterClient) GetRESTConfig(
...
+
+		if resource == nil {
+			return nil, false
+		}
+		manifest := resource.(*manifestv1beta1.Manifest)
+		labelValue, err := internal.GetResourceLabel(resource, manifestv1beta1.KymaName)
+		objectKey := client.ObjectKeyFromObject(resource)
+		var labelErr *types.LabelNotFoundError
+		if errors.As(err, &labelErr) {
+			return nil, false
```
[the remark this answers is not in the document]
ruanxin: done

internal/manifest/v1beta1/client.go
```diff
@@ -54,3 +61,36 @@ func (cc *ClusterClient) GetRESTConfig(
...
+		}
+		cacheKey := generateCacheKey(labelValue, strconv.FormatBool(manifest.Spec.Remote), manifest.GetNamespace())
+		logger.V(internal.DebugLogLevel).Info(
```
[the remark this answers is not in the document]
ruanxin: I think this log info can be removed, the information here to indicate client can be cached is redundant since it's default behavior, and knowing cachekey, I think it does not help too much for troubleshooting purposes. Basically, my point is if needs to downgrade to tracelevel, it's a sign that probably this log is put into an unexpected place.

internal/manifest/v1beta1/client.go
ruanxin: I use `strings.Join()`, it's a bit flexible if later needs to introduce more values as the source of cache key, although I believe will not happens.

jakobmoellersap (approved):

Issue #424, opened 2023-02-27
Documentation: Quick-Start for the Modularization


janmedrek:
**Description:**

This guide would include just a few simple steps describing how to set up the environment and deploy/enable the module in the cluster:
- setting up the environment (preferably with the `kyma alpha deploy` command of the CLI)
- deploying the ModuleTemplate in the cluster (can be a basic example such as [the one in the template operator repository](https://github.com/kyma-project/template-operator#deploying-a-moduletemplate-into-the-control-plane))
- enabling the module (kyma CLI enable command or by editing the resource)
- short description of what happens next, where and when the module is installed by which component

**Acceptance Criteria:**
- [ ] Quick-Start guide implemented
- [ ] The guide is available in the `/docs/user/quick-start` directory

PR #449, opened 2023-03-07
Quick start documentation

ruanxin:
related issue: https://github.com/kyma-project/lifecycle-manager/issues/424


docs/user/quick-start.md
```diff
@@ -0,0 +1,80 @@
...
+```
+kyma provision k3d
+```
+Since the services deployed in the control plane are managed by Istio, you need to install Istio into the local k3d cluster using the following command:
+```
+istioctl install -y
+```
+Lifecycle Manager also exposes metrics that are collected by Prometheus Operator in the control plane to provide better observability. To simplify the local setup, you only need to deploy the ServiceMonitor CRD using the following command:
+```
+kubectl apply -f hack/servicemonitors.yaml
```
jakobmoellersap, line 28: this hack file is not necessary, I think we could just use a URL for the servicemonitors crd

docs/user/quick-start.md
```diff
@@ -0,0 +1,80 @@
...
+```
+You can also follow official [quick start](https://prometheus-operator.dev/docs/prologue/quick-start/) guide to deploy a full set of prometheus operator into cluster as an alternative solution if you want to monitor the component performance.
+
+## Deploy Kyma lifecycle manager
+We recommend deploying Kyma Lifecycle Manager with the control plane kustomize profile, and `kcp-system`, `kyma-system` namespace must be configured upfront.
+```
+kubectl create ns kcp-system
+kubectl create ns kyma-system
+kyma alpha deploy -c alpha -k https://github.com/kyma-project/lifecycle-manager/config/control-plane
```
jakobmoellersap, line 37: why do we need the alpha channel here specifically?

docs/user/quick-start.md
```diff
@@ -0,0 +1,80 @@
...
+With Kyma CLI [enable module](https://github.com/kyma-project/cli/blob/main/docs/gen-docs/kyma_alpha_enable.md) and [disable module](https://github.com/kyma-project/cli/blob/main/docs/gen-docs/kyma_alpha_disable.md) command, you can manage kyma modules without directly modify Kyma CR.
+
+As an example, you can use following command to enable keda module in `alpha` channel for `default-kyma` under `kyma-system` namespace
+```
+kyma alpha enable module keda -c alpha -n kyma-system -k default-kyma
+```
+### Manage modules in remote cluster mode
+To allow Kyma Lifecycle Manager manages Kyma modules in remote cluster, two prerequisite must be fulfilled.
+
+1. For the module to be deployed remotely, the related Module Template must be configured as `remote` in `spec.target`.
```
jakobmoellersap, line 58: Maybe also add a kubectl patch command that allows you to do that in a one-liner

docs/user/quick-start.md
```diff
@@ -0,0 +1,80 @@
...
+2. A kubernetes secret resource which contains remote cluster kubeconfig access data must be deployed into control plane cluster upfront.
```
jakobmoellersap, line 59: I think this secret needs to be explained a little more (why its needed, not how its created as that's mentioned below)

docs/user/quick-start.md
```diff
@@ -0,0 +1,80 @@
...
+Deploy this secret under local control plane cluster:
+```
+kubectl config use-context k3d-kyma 
+kubectl apply -f default-kyma-secret.yaml
+```
+
+After this access secrete successfully deployed, you can start to use Kyma CLI manage modules for remote cluster.
+
+## Next Steps
+[TBD]
```
jakobmoellersap, line 80: What is TBD here?

docs/user/quick-start.md
```diff
@@ -0,0 +1,80 @@
+# Quick Start
```
jakobmoellersap, line 1: I think you should remove the quick start from the main README.md then as this here is much better imo. Additionally, I think the local start guide with the watcher has big overlaps so that should be unified.

docs/user/quick-start.md
ruanxin: unfortunately, there is no existing servicemonitors crd exists, current [prometheus-operator](https://github.com/prometheus-operator/prometheus-operator/) will generate all resources as a bundle.yaml to create with. Maybe search online can find this archived [crd](https://github.com/helm/charts/blob/master/stable/prometheus-operator/crds/crd-servicemonitor.yaml), but that's already outdated, still use `apiVersion: apiextensions.k8s.io/v1beta1`

docs/user/quick-start.md
ruanxin: to let the generated `default-kyma` use `alpha` channel (by default, it's `regular`), so that works with the existing moduletemplate.

docs/user/quick-start.md
ruanxin: I plan to put here some reference links to other user documents, but they are not created yet.

docs/user/quick-start.md
ruanxin: I think for a quick start, it's better not to introduce too much manual modification to interrupt the flow, the command I put here is basically all necessary for the setup.

In the following apply kyma-project modules, I highlighted there the target is configured as `remote` already. And above cli deploy command, the target is `control-plane` (haven't implemented yet), so the user doesn't have to worry about this spec value during quick start. If later needs to modify this field, the choice is left to the user. what do you think?

docs/user/quick-start.md
jakobmoellersap:
```suggestion
kubectl apply -f https://raw.githubusercontent.com/prometheus-community/helm-charts/main/charts/kube-prometheus-stack/crds/crd-servicemonitors.yaml
```
here you go :)

docs/user/quick-start.md
jakobmoellersap: Understood. I think in this case you should probably reference single/dual cluster mode then into a different guide but that's probably in the reference docs that Im building up. LGTM for me then

docs/user/quick-start.md
jakobmoellersap: Maybe write that in the TBD if you plan to commit it

docs/user/quick-start.md
jakobmoellersap: Can you align with @jeremyharisch when to officially enable watcher in our control plane? then we need to unify these guides asap

docs/user/quick-start.md
ruanxin: done

docs/user/quick-start.md
ruanxin: I remove this part, will add it in later refactoring

docs/user/quick-start.md
ruanxin: confirmed with @janmedrek , the watcher is not planed to introduced into current document, until we released it, the document can be revisited later.

PR #451, opened 2023-03-08
ignore IsNotFound and IsNoMatchError for preDelete

ruanxin:
During resource deletion, if the resources configured in preDelete have already been removed, it should not prevent the deletion process to be continue.

### DIFF
--- a/pkg/declarative/v2/reconciler.go
+++ b/pkg/declarative/v2/reconciler.go
@@ -7,6 +7,7 @@ import (
 
 	"helm.sh/helm/v3/pkg/kube"
 	v1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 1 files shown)


internal/manifest/v1beta1/custom_resource.go
```diff
@@ -73,7 +74,7 @@ func PreDeleteDeleteCR(
 		return ErrWaitingForAsyncCustomResourceDeletion
 	}
 
-	if !k8serrors.IsNotFound(err) {
+	if !k8serrors.IsNotFound(err) && !meta.IsNoMatchError(err) {
```
jakobmoellersap, line 77: if you ignore it here, it will be now thrown in line 86 on the get request, thus the error will not be fixed I believe. Could you confirm that?

ruanxin: no, in line 86, the err is related to get manifest CR, but this err here is for the CR which defined in manifest spec.data

jakobmoellersap (approved):

Issue #360, opened 2023-01-23
Remove Helm from the Module bundling


janmedrek:
**Description**
Since we have moved away from the Helm-based approach we need to get rid of Helm-related logic in the CLI and Managers.
One of the remaining functionalities was in the Module bundling process where the Helm OCI layer is created. This should be dropped as the raw Kubernetes YAMLs are now the preferred (and supported) option.

Since this will be removed, the Module Manager's interpretation of layers needs to be properly adapted.

**Acceptance Criteria**
- [x] Helm layers removed from the CLI `kyma alpha create module` command:
https://github.com/kyma-project/cli/pull/1581
- [x] Module Manager adapted to properly interpret the layers:
https://github.com/kyma-project/lifecycle-manager/pull/469
- [x] Sample Helm CR removed from template-operator [repository](https://github.com/kyma-project/template-operator):
https://github.com/kyma-project/template-operator/pull/47

PR #469, opened 2023-03-16
Remove helm chart packaging support for module installation

khlifi411:
Implementation of: #360


controllers/suite_test.go
```diff
@@ -141,7 +141,7 @@ var _ = BeforeSuite(func() {
 		RemoteClientCache: remoteClientCache,
 		KcpRestConfig:     k8sManager.GetConfig(),
 	}).SetupWithManager(k8sManager, controller.Options{},
-		controllers.SetupUpSetting{ListenerAddr: UseRandomPort})
+		controllers.SetupUpSetting{ListenerAddr: ":8080"})
```
ruanxin, line 144: why changed to static port?

khlifi411: I reverted it as it is working now with the random port.

lindnerby (approved):

internal/declarative/v2/renderer_raw_test.go
```diff
@@ -1,258 +0,0 @@
-package v2_test
```
ruanxin, line 1: Why this test get removed?

internal/declarative/v2/renderer_raw.go
```diff
@@ -2,31 +2,101 @@ package v2
...
+		meta.SetStatusCondition(&status.Conditions, r.prerequisiteCondition(obj))
+		obj.SetStatus(status)
+		return ErrConditionsNotYetRegistered
+	}
+
 	return nil
 }
 
-func (r *RawRenderer) EnsurePrerequisites(_ context.Context, _ Object) error {
+func (r *RawRenderer) EnsurePrerequisites(ctx context.Context, obj Object) error {
```
ruanxin, line 69: If it's a raw manifest, all the resources are already saved in that yaml, why we need to implement this ensure prerequisites? This file don't need to change.

internal/manifest/v1beta1/manifest_controller_test.go
```diff
@@ -134,41 +125,37 @@ var _ = Describe(
...
 					WithArguments(manifest).Should(Succeed())
 				Eventually(expectManifestState, standardTimeout, standardInterval).
 					WithArguments(manifest.GetName()).Should(Succeed())
-				Eventually(expectedHelmClientCache, standardTimeout, standardInterval).
-					WithArguments(internalV1beta1.GenerateCacheKey(manifest.GetLabels()[v1beta1.KymaName],
-						strconv.FormatBool(manifest.Spec.Remote), manifest.GetNamespace())).Should(BeTrue())
 				Eventually(deleteManifestAndVerify(manifest), standardTimeout, standardInterval).Should(Succeed())
 			},
 			Entry(
 				"When Manifest CR contains a valid install OCI image specification, "+
```
ruanxin, line 137: the expected condition should be changed, we are not testing any helm cache-related issues anymore.

internal/manifest/v1beta1/ready_check_test.go
```diff
@@ -16,76 +16,57 @@ import (
...
-			Expect(os.RemoveAll(filepath.Join(os.TempDir(), mainOciTempDir))).To(Succeed())
-		},
-	)
+const nginxControllerDeploymentSuffix = "nginx-ingress-controller"
 
-	It("Install OCI manifest and wait until it's ready", func() {
-		manifest := NewTestManifest("custom-check-oci")
+var _ = Describe("Custom Manifest consistency check, given Manifest CR with Helm specs", Ordered, func() {
+	setHelmEnv()
+	validHelmChartSpec := v1beta1.HelmChartSpec{
```
ruanxin, line 27: Why here test helm-chart?

internal/manifest/v1beta1/spec_resolver.go
```diff
@@ -83,7 +83,7 @@ func (m *ManifestSpecResolver) Spec(ctx context.Context, obj declarative.Object)
 	case v1beta1.HelmChartType:
 		mode = declarative.RenderModeHelm
```
ruanxin, line 84: Why we still support Helm and Kostomize? If the purpose for this issue is remove helm chart support?

internal/manifest/v1beta1/spec_resolver.go
```diff
@@ -83,7 +83,7 @@ func (m *ManifestSpecResolver) Spec(ctx context.Context, obj declarative.Object)
...
 	case v1beta1.OciRefType:
-		mode = declarative.RenderModeHelm
+		mode = declarative.RenderModeRaw
```
ruanxin, line 86: The previous oci layer was rendered as helm chart, but currently changed to Raw manifest, it will introduce breaking changes when we handle those helm chart oci layer.

internal/declarative/v2/renderer_raw_test.go
khlifi411: I removed this test because it was failing after I changed the raw renderer implementation. It was testing only the happy path and was too coupled to the original raw_renderer implementation. I think the manifest controller tests are sufficient.

internal/declarative/v2/renderer_raw.go
khlifi411: Yes. I agree in principle. But I faced problems related to missing CRDs when applying the raw manifest without ensuring that CRDs included in it are already installed. I couldn't find a different approach to that which fixes the CRD issue.

internal/manifest/v1beta1/spec_resolver.go
khlifi411: I am not sure either, TBH. As far as I understand, the last time we discussed the scope of the change, we agreed not to remove other renderers (as we are not live yet, and we can not be sure that some partial Kustomize or Helm support might be needed). But if you think we can drop it, that's also valid.

internal/manifest/v1beta1/spec_resolver.go
khlifi411: True. I have prepared [this](https://github.com/kyma-project/lifecycle-manager/pull/469) CLI change to stop using helm-chart bundling in OCI layers created for the module templates.

ruanxin: The thing is you need to communicate to other teams that we introduced these breaking changes because they already rendered module template with helm chart oci layer.

internal/declarative/v2/renderer_raw_test.go
ruanxin: This test is only for render raw manifest, that's the purpose it is designed for, if it fails, then it means either your change introduced problems, or you need to adapter the test, but not completely delete it.

internal/declarative/v2/renderer_raw.go
ruanxin: Please always keep in mind we are using reconcile pattern, if some resources fail in the first several try, but it can deploy successfully eventually, then it's okay. You don't have to make sure CRD get deployed first, raw manifest anyway include CRD, it's just in this reconcile round, CRD is not being fully deployed, but the manifest controller anyway always tries to deploy/check resources consistencies in each reconcile loop, eventually all resources can be deployed.

internal/manifest/v1beta1/ready_check_test.go
ruanxin: As we clarified, helm support should be completely removed.

internal/manifest/v1beta1/spec_resolver.go
ruanxin: As we clarified, helm and kostomize support should be completely removed.

PR #474, opened 2023-03-17
Replace warning events when setting metrics by info log

lindnerby:
Description:
* Whenever the expected `skr-domain` annotation or `kyma-project.io/instance-id` label on a Kyma are missing when setting kyma status metric it should not create a warning event rather than just log that


lindnerby:
Sorry about the imports reorder noise...

jakobmoellerdev (approved):

PR #484, opened 2023-03-22
kyma_controller: Decouple Handling of Processing State

### LINKED ISSUE #376: Move processes to the seperate go-routines and aggregate Status
**Description**
Currently, the Watcher-SKR-Webhook is being installed in the Kyma control-loop sequentially.  Thus, a work-around has been introduced with the PKI-Setup, if the certificate sync fails, that the module reconciliation is not affected.

A refactoring is needed to move the installation of the Webhook into a separate go-routine, to have the module reconciliation  and the webhook installation decoupled from each other.

Fruthermore, consider about introducing an own condition if the Certificate for a Kyma is ready. Thus, it can be easier to debug if installation of Watcher is stuck.

**ACs:**
- Move Webhook installation to separate go-routine
- Module installation moved to separate go-routine
- Module catalogue sync moved to separate go-routine
- Collect the state from the above three go-routines and create a Status out of that
- If installation of Webhooks fails, module reconciliation should not be affected
- Think of introducing an own condition if certificate is ready; continue Webhook installation even if Certificate Secret is not ready

LeelaChacha:
Closes #376


controllers/kyma_controller.go
```diff
@@ -229,23 +224,30 @@ func (r *KymaReconciler) HandleInitialState(ctx context.Context, kyma *v1beta1.K
 func (r *KymaReconciler) HandleProcessingState(ctx context.Context, kyma *v1beta1.Kyma) error {
```
lindnerby, line 224: Btw. this controller contains a bunch of methods that are exported, but since they don't belong to any interface implementation or are not used from other packages they can be made package-private. Can you check them please?

lindnerby (changes requested):

controllers/kyma_controller.go
```diff
@@ -26,13 +26,14 @@ import (
 	"github.com/kyma-project/lifecycle-manager/pkg/log"
 	"github.com/kyma-project/lifecycle-manager/pkg/metrics"
 	"k8s.io/client-go/rest"
+	"sync"
```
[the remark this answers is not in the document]
LeelaChacha: import of sync is not needed anymore

controllers/kyma_controller.go
LeelaChacha: "Find Usage" Tool shows no use outside the KymaReconciler and I know or can think of no other piece of code that might use these function.
Also making them package-private breaks no test and manual testing works.

lindnerby (approved):

Issue #470, opened 2023-03-16
Control-Loop for Force Finaliser Removal (Gardener Deprovisioning Compatibility)


jakobmoellerdev:
As a cluster user on Gardener that is using Kyma as managed offering, the (gardener) shoot is managed completely by a controller sitting outside the shoot. When deciding to kill my cluster, aka delete all of its nodes and resources within, gardener waits for any finalisers to be completely removed from resources and will get stuck during deprovisioning if this cannot be guaranteed. 

Thus we want to introduce a custom controller that is able to force-remove any finalisers set after a set period after the deletionTimestamp was set in order to forcefully remove all finalisers. Since this is gardener specific, we will do this only if a feature flag is active and a label is set on the kyma resource: `purge.finalisers.kyma-project.io`. 

If set together with a flag `purge-finaliser-timeout` set to a duration (e.g. `5m`), then the control-loop should actively remove all finalisers as soon as `deletionTimestamp+purge-finalizer-timeout` is reached.

For more reference on the root cause, see https://github.com/gardener/gardener/issues/3110

AC:

- Custom Control Loop is introduced to Lifecycle Manager
- Loop is triggered based on deletionTimestamp set on Kyma and still being present after a set timeout
- Control Loop only gets started based on `enable-purge-finaliser` feature-flag as `bool`
- Control Loop timeout is controlled with `purge-finaliser-timeout` feature-flag as `time.Duration`
- Control Loop continuously removes all finalisers in remote clusters only, and will not work with control-plane clusters as its effect can destroy lifecycle manager / kyma reconciliation as well. 
- Consider the amount requests to APIServer while removing all finalizers, avoid too many parallel requests to overload APIServer.

Tomasz-Smelcerz-SAP:
After thinking about the `Finaliser Removal` loop triggering, I think the correct approach must be discussed.
Currently, as I understand the AC, the proposal is: `control-loop should actively remove all finalisers as soon as deletionTimestamp+purge-finalizer-timeout is reached`
What if timeout is 5 minutes, but LM actually deletes the Kyma object in 1 minute? We're waiting 4 additional minutes to drop finalizers. It could have been done as soon the Kyma object is deleted.
If we can detect fast (or immediately) that Kyma is gone, I'd propose the following logic:
- drop finalizers as soon as Kyma is gone (object not found) in the control plane
[...]

PR #485, opened 2023-03-22
Cascade delete control loop

Tomasz-Smelcerz-SAP:
**Related Issues:**
*  https://github.com/kyma-project/lifecycle-manager/issues/470


controllers/kyma_controller.go
```diff
@@ -20,30 +20,32 @@ import (
 	"context"
```
lindnerby, line 20: Guess we don't need these import sorting changes in `kyma_controller`. Can you revert them?

Tomasz-Smelcerz-SAP:
There is a problem with reading secrets from the cluster. The controller-runtime client we use can't find them, though it can find other objects.
I am investigating the issue.

controllers/purge/purge_controller_helper_test.go
```diff
@@ -0,0 +1,30 @@
...
+	. "github.com/onsi/gomega"
+)
+
+func RegisterDefaultLifecycleForKyma(kyma *v1beta1.Kyma) {
+	BeforeAll(func() {
+		Expect(controlPlaneClient.Create(ctx, kyma)).Should(Succeed())
+	})
+
+	/*
+		AfterAll(func() {
```
lindnerby, line 16: You need these enabled?

controllers/purge/suite_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"context"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/kyma-project/lifecycle-manager/api"
+	"github.com/kyma-project/lifecycle-manager/pkg/log"
+	"go.uber.org/zap/zapcore"
+
+	//nolint:gci
```
lindnerby, line 29: If you enable this linter it will sort the imports according to the structure defined in `.golangci.yaml`
But you need to enable it in the config file and run it once to sort your imports

controllers/purge/suite_test.go
```diff
@@ -0,0 +1,148 @@
...
+	//+kubebuilder:scaffold:imports
+)
+
+// These tests use Ginkgo (BDD-style Go testing framework). Refer to
+// http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
+
+const UseRandomPort = "0"
+
+var (
+	purgeReconciler    *controllers.PurgeReconciler //nolint:gochecknoglobals
```
lindnerby, line 57: I think you can exclude `gochecknoglobals` on the pkg level

controllers/purge/suite_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	//+kubebuilder:scaffold:imports
+)
+
+// These tests use Ginkgo (BDD-style Go testing framework). Refer to
+// http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
+
+const UseRandomPort = "0"
```
lindnerby, line 54: Can be pkg private

controllers/purge/suite_test.go
```diff
@@ -0,0 +1,148 @@
...
+		Client:           k8sManager.GetClient(),
+		EventRecorder:    k8sManager.GetEventRecorderFor(operatorv1beta1.OperatorName),
+		RequeueIntervals: intervals,
+		VerificationSettings: signature.VerificationSettings{
+			EnableVerification: false,
+		},
+		RemoteClientCache:     remoteClientCache,
+		ResolveRemoteClient:   useLocalClient,
+		KcpRestConfig:         k8sManager.GetConfig(),
+		PurgeFinalizerTimeout: time.Second, //TODO: As short as possible?
```
lindnerby, line 126: Anything TODO left?

controllers/purge_controller.go
```diff
@@ -0,0 +1,209 @@
...
+	RequeueIntervals
+	signature.VerificationSettings
+	SKRWebhookManager     watcher.SKRWebhookManager
+	KcpRestConfig         *rest.Config
+	RemoteClientCache     *remote.ClientCache
+	ResolveRemoteClient   RemoteClientResolver
+	PurgeFinalizerTimeout time.Duration
+}
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
```
lindnerby, line 63: Obsolete comment

controllers/purge_controller.go
```diff
@@ -0,0 +1,209 @@
...
+				logger.Info(fmt.Sprintf("%s", msg))
+			} else {
+				logger.Info("Purging was not successful!")
+				logger.Info(fmt.Sprintf("%s", msg))
+			}
+			return ctrl.Result{}, nil
+		}
+
+		return ctrl.Result{
+			//	Requeueing exactly after 1s of the deadline
```
lindnerby, line 112: Obsolete comment. The code says the same thing.

controllers/purge_controller.go
```diff
@@ -0,0 +1,209 @@
...
+			RequeueAfter: deletionDeadline.Add(time.Second).Sub(time.Now()),
+		}, nil
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// Helper functions
```
lindnerby, line 120: Obsolete comment

controllers/purge_controller.go
```diff
@@ -0,0 +1,209 @@
...
+func (r *PurgeReconciler) UpdateStatus(
+	ctx context.Context, kyma *v1beta1.Kyma, state v1beta1.State, message string,
+) error {
+	if err := status.Helper(r).UpdateStatusForExistingModules(ctx, kyma, state, message); err != nil {
+		return fmt.Errorf("error while updating status to %s because of %s: %w", state, message, err)
+	}
+	return nil
+}
+
+// RecordKymaStatusMetrics updates prometheus metrics defined to track changes to the Kyma status.
```
lindnerby, line 206: Obsolete comment

go.mod
```diff
@@ -172,13 +184,27 @@ require (
...
 	github.com/tonglil/buflogr v1.0.1 // indirect
 	github.com/ulikunitz/xz v0.5.10 // indirect
 	github.com/vbatts/tar-split v0.11.2 // indirect
 	github.com/xeipuuv/gojsonpointer v0.0.0-20190905194746-02993c407bfb // indirect
 	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 // indirect
 	github.com/xlab/treeprint v1.1.0 // indirect
+	go.etcd.io/etcd/api/v3 v3.5.5 // indirect
+	go.etcd.io/etcd/client/pkg/v3 v3.5.5 // indirect
+	go.etcd.io/etcd/client/v3 v3.5.5 // indirect
+	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.35.0 // indirect
```
lindnerby, line 198: Where are all these otel dependencies coming from?

controllers/setup.go
```diff
@@ -145,3 +145,23 @@ func (r *WatcherReconciler) SetupWithManager(mgr ctrl.Manager, options controlle
...
 		Complete(r)
 }
+
+// SetupWithManager sets up the Kyma controller with the Manager.
+func (r *PurgeReconciler) SetupWithManager(mgr ctrl.Manager,
+	options controller.Options, settings SetupUpSetting,
+) error {
+	controllerBuilder := ctrl.NewControllerManagedBy(mgr).For(&v1beta1.Kyma{}).WithOptions(options)
+	// here we define a watch on secrets for the lifecycle-manager so that the cache is picking up changes
+	//TODO: Is it necessary here?
```
lindnerby, line 155: No, we don't need handling Secret changes in this reconciler

controllers/purge_controller.go
```diff
@@ -0,0 +1,209 @@
...
+			}
+
+			return ctrl.Result{}, nil
+		}
+
+		return ctrl.Result{}, err
+	}
+
+	// condition to check if deletionTimestamp is set, retry until it gets fully deleted
+	if !kyma.DeletionTimestamp.IsZero() || kyma.Status.State == v1beta1.StateDeleting {
```
lindnerby, line 94: We should only perform clean-up if `!kyma.DeletionTimestamp.IsZero()` and don't care about `v1beta1.StateDeleting`!

controllers/purge_controller.go
```diff
@@ -0,0 +1,209 @@
...
+	var crdList = apiextensions.CustomResourceDefinitionList{}
+
+	if err := remoteClient.List(ctx, &crdList); err != nil {
+		return false, err
+	}
+
+	for _, crdResource := range crdList.Items {
+		staleResources, err := getStaleResourcesFrom(ctx, remoteClient, crdResource)
+		//	Skipping Kyma Custom Resource deletion as it should be handled by the main reconciler loop
+		if crdResource.Spec.Group == "operator.kyma-project.io" && crdResource.Spec.Names.Kind == "Kyma" {
```
lindnerby, line 131: You wrap this condition check in a func `isKymaCR` and even do the skip before getting the stale resources

controllers/purge_controller.go
```diff
@@ -0,0 +1,209 @@
...
+
+	// check if kyma resource exists
+	kyma := &v1beta1.Kyma{}
+	if err := r.Get(ctx, req.NamespacedName, kyma); err != nil {
+		if apierrors.IsNotFound(err) {
+			remoteClient, err := r.ResolveRemoteClient(ctx, client.ObjectKey{Namespace: req.Namespace, Name: req.Name})
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			if done, msg := performCleanup(ctx, remoteClient, logger); done {
```
lindnerby, line 79: So if we can't get the Kyma with a NotFound, we are doing a clean-up anyways? That can't be correct, right?

flags.go
```diff
@@ -95,6 +96,11 @@ func defineFlagVar() *FlagVar {
 		&flagVar.logLevel, "log-level", defaultLogLevel,
 		"indicates the current log-level, enter negative values to increase verbosity (e.g. 9)",
 	)
+	flag.BoolVar(&flagVar.enablePurgeFinalizer, "enable-purge-finalizer", true,
```
lindnerby, line 99: I think default should be false and it should be explicitly enabled in the control-plane config

lindnerby (changes requested):

controllers/purge/suite_test.go
amritanshusikdar: We want the timeout to be as short as possible for the testing in the pipelines, else it can take unusually long to just build. 1 Second is what we came down to. But in case the tests fail due to timeout issues in the future, then this timeout can be increased.

controllers/purge_controller.go
amritanshusikdar: Yes, it is valid. Referring to the optimization suggested in this [comment](https://github.com/kyma-project/lifecycle-manager/issues/470#issuecomment-1486857998), we must drop the finalizers immediately once the Kyma object is gone instead of waiting till the timeout.

go.mod
amritanshusikdar: Removed the potentially unused dependencies

controllers/purge_controller.go
```diff
@@ -0,0 +1,223 @@
...
+
+		return ctrl.Result{
+			RequeueAfter: time.Until(deletionDeadline.Add(time.Second)),
+		}, nil
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func performCleanup(ctx context.Context, remoteClient client.Client, logger logr.Logger) (bool, interface{}) {
```
lindnerby, line 122: Let's return (bool, error) and let the caller unwrap the error msg

lindnerby (changes requested):

jeremyharisch (commented):
Just some small comments, but still in progress to check for improvments

Tomasz-Smelcerz-SAP:
I have found a reason for endless looping during Kyma deletion.
The main Kyma reconcile loop implementation is responsible for that.
The flow looks as follows:
1) Kyma deletion is triggered
2) Kyma status is changed to "Deleting"
3) Main kyma loop is executed on Kyma in "Deleting" state. It ensures Kyma finalizer (`operator.kyma-project.io/Kyma`) is there and then goes to `handleDeletingState()` function
4) In the `handleDeletingState()` function the logic drops the Kyma finalizer (`operator.kyma-project.io/Kyma`) and updates Kyma. This update is an event that will immediately re-trigger the reconcile loop.
5) Main kyma loop is executed on Kyma in "Deleting" state, without Kyma finalizer. The logic **adds** the Kyma finalizer again, because adding finalizer logic executes with every loop. The Kyma is updated. Then the logic continues to the `handleDe
letingState()`. Go to step 4)

Steps 4 and 5 execute in the loop without any delays.
Normally this loop is not observed because once all the finalizers are gone, the API server deletes the Kyma object from the cluster, and processing of this Kyma object is stopped.
But when we add another finalizer (in the purge reconciler loop), the Kyma object stays and is subject to the status-change-loop as described.

To reproduce:
a) Add a log statement that executes every time the loop enters `handleDeletingState()` function.
b) Start the controller (without the purge reconciler loop)
c) Create some Kyma object
d) Add a custom finalizer to the Kyma object alongside the existing one (`operator.kyma-project.io/Kyma`)
e) Delete the Kyma and observe endless stream of log messages - the one you added in step a)

To fix: It seems it's enough to skip the `kyma.CheckLabelsAndFinalizers()` logic if the Kyma is being deleted (kyma.DeletionTimestamp.IsZero() == false)

procedure.txt
jeremyharisch, line 1: This file should be removed! I guess this was just for local testing, but should not be on main branch :D

controllers/purge_controller.go
```diff
@@ -0,0 +1,279 @@
...
+		if err := r.DropPurgeFinalizer(ctx, kyma); err != nil {
+			logger.Error(err, "Couldn't remove Purge Finalizer from the Kyma object")
+			return ctrl.Result{}, err
+		}
+
+		return ctrl.Result{}, nil
+	}
+
+	return ctrl.Result{
+		RequeueAfter: time.Until(deletionDeadline.Add(time.Second)),
```
jeremyharisch, line 96: Why are her 0 Seconds added to the time? If you want to convert it into seconds, you could also cast it. WDYT?

controllers/purge_controller.go
```diff
@@ -0,0 +1,279 @@
...
+) error {
+	if err := status.Helper(r).UpdateStatusForExistingModules(ctx, kyma, state, message); err != nil {
+		return fmt.Errorf("error while updating status to %s because of %s: %w", state, message, err)
+	}
+	return nil
+}
+
+func (r *PurgeReconciler) RecordKymaStatusMetrics(_ context.Context, _ *v1beta1.Kyma) {}
+
+func (r *PurgeReconciler) IsKymaManaged() bool { return false }
```
jeremyharisch, line 207: Why is this always false? You could also use the flag bool, as in the kyma controller

controllers/purge_controller.go
amritanshusikdar: `time.Second` resolves to 1 second. So essentially we're just adding a second to the deletion deadline here.

lindnerby:
How are you making sure here that this loop removes only finalisers in remote clusters and it does not work with control-plane cluster?

Tomasz-Smelcerz-SAP:
> How are you making sure here that this loop removes only finalisers in remote clusters and it does not work with control-plane cluster?

@lindnerby Similarly to the main Kyma loop, we're using a separate client for these operations. Take a look at `performCleanup` method and how it takes the remote client as an argument. In a "production" configuration, this client is bootstrapped using a Kubeconfig taken from a secret existing in the control-plane. This Kubeconfig "points to" the target cluster.
It is possible to configure things so that this "remote client" is actually accessing the same cluster as the main "Reconcile()" loop client. But then also the main loop is affected and so it's not directly related to the code in this PR.

lindnerby (approved):

PR #489, opened 2023-03-23
Detect managed kyma

nesmabadr:
- Introducing the `is-kyma-managed` flag to indicate when the Kyma is managed.
- Having the flag as true for the control-plane configs.
- Changing the manager in the managed field depending on the `is-kyma-managed` flag.

This PR is part of the changes required for this [issue](https://github.com/kyma-project/cli/issues/1472)


controllers/kyma_controller_test.go
Tomasz-Smelcerz-SAP, line 1: A test that verifies that `lifecycle-manager` is set as a field manager for the `status` subresource would be nice

controllers/kyma_controller.go
```diff
@@ -152,7 +153,7 @@ func (r *KymaReconciler) deleteKyma(ctx context.Context, kyma *v1beta1.Kyma) (ct
 	// if the status is not yet set to deleting, also update the status of the control-plane
 	// in the next sync cycle
 	if err := status.Helper(r).UpdateStatusForExistingModules(
-		ctx, kyma, v1beta1.StateDeleting, "waiting for modules to be deleted",
+		ctx, kyma, v1beta1.StateDeleting, "waiting for modules to be deleted", r.IsKymaManaged,
```
Tomasz-Smelcerz-SAP, line 156: `r.IsKymaManaged` is not changing between calls to  `UpdateStatusForExistingModules`, the same value is always used - either `true` or `false`, because that is coming from the flag.
Extending the method signature with  a parameter that has always the same value for the entire life of a program is not a good design.
Because the Helper already accepts `r` - which contains information about `isKymaManaged` - perhaps this knowledge could be used in the implementation of `UpdateStatusForExistingModules`, instead of passing that value explicitly?

Tomasz-Smelcerz-SAP (approved):

PR #490, opened 2023-03-23
Fix cm patch for kustomize version 4.x

lindnerby:
Description:
* Previous version of the patch was tested with Kustomize 5.0.0, but we are pinned to 4.5.6
* The current patch doesn't work on argo apps, this should fix it
* Also fixed some linter issues
* Bump the golangci-lint version to the version that is used on the pipeline

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,7 +149,7 @@ GOLANG_CI_LINT = $(LOCALBIN)/golangci-lint
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v4.5.6
 CONTROLLER_TOOLS_VERSION ?= v0.11.2
-GOLANG_CI_LINT_VERSION ?= v1.51.1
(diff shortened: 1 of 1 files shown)


ruanxin (dismissed):

controllers/withwatcher/watcher_controller_test.go
```diff
@@ -106,10 +103,7 @@ func deleteWatcher(name string) error {
 	if apierrors.IsNotFound(err) {
 		return nil
 	}
-	if err := controlPlaneClient.Delete(suiteCtx, &watcher); err != nil {
-		return err
-	}
-	return nil
+	return controlPlaneClient.Delete(suiteCtx, &watcher)
```
ruanxin, line 106: can't believe we have so much unnecessary error handling 😄

ruanxin (approved):

PR #494, opened 2023-03-27
Fix LM overview dashboard and delete generated ones

lindnerby:
Description:
* Remove MM Panels
* Rename
* ref correct metrics service
* remove kubebuilder plugin generated dashboards, since they don't work and information is kind of redundant


config/grafana/dashboards/resources-metrics.json
```diff
@@ -0,0 +1,349 @@
+{
```
ruanxin, line 1: The file you removed was generated by this `make grafana-dashboard` command which is configured in the makefile. They are automatically generated by a kubebuilder plugin `grafana.kubebuilder.io/v1-alpha`

lindnerby: As discussed I deleted the make target and the kubebuilder plugin generated dashboards. They don't work and we keep our manually created dashboard only.

ruanxin (approved):

PR #499, opened 2023-03-29
Make tests more stable

ruanxin:
This PR address some flaky tests due to different design problems.

- removed the function to fetch external crds from remote url, get rid of the problem in case of network failure.
- Some gomega tests using `Eventually` polling function are not designed in a solid way, which brings flaky tests.


controllers/kyma_module_channel_test.go
```diff
@@ -163,14 +163,16 @@ var _ = Describe("Switching of a Channel with higher version leading to an Upgra
 	It(
 		"should create kyma with standard modules in default channel normally", func() {
 			Expect(controlPlaneClient.Create(ctx, kyma)).ToNot(HaveOccurred())
-			Eventually(GetKymaState(kyma.Name), 5*time.Second, Interval).
+			Eventually(GetKymaState, 20*time.Second, Interval).
+				WithArguments(kyma.GetName()).
```
Tomasz-Smelcerz-SAP, line 167: Is there any reason to do it this way instead of plain `GetKymaState(kyma.Name)` ?
I think it is less readable, since the function arguments are provided indirectly, in another line of code, for no obvious reasons.

ruanxin: it's purely because of gomega design.

```
Eventually can also accept functions that take arguments, however you must provide those arguments using .WithArguments().  For example, consider a function that takes a user-id and makes a network request to fetch a full name:

	func FetchFullName(userId int) (string, error)
```

ruanxin: If we go for the previous way, then we should rewrite the `GetKymaState` to return function (you can check that function before change), that's actually the thing I want to improve, especially regarding readability, for those functions used in `Eventually`, not return function, but just write like a normal function, but currently, most of them write in that way, I just take this as an example to change, we can refactor for other functions later and prevent to write like that for future tests.

Tomasz-Smelcerz-SAP (approved):

Tomasz-Smelcerz-SAP (approved):

PR #503, opened 2023-03-31
Accepting valid yaml files for the config layer

### LINKED ISSUE #502: Template Operator Verification Bug Hunt: config layers with yaml containing only comments should not error out and inform the user
When defining a config layer, any valid yaml can be used as the layer input. If the manifest operator currently observes a config layer, it tries to parse it into a map. This can lead to the following error:

```{r setup, include=FALSE}
1.661234725879773e+09   ERROR   Reconciler error        {"controller": "manifest", "controllerGroup": "component.kyma-project.io", "controllerKind": "Manifest", "manifest": {"name":"templatekyma-sample","namespace":"kyma-system"}, "namespace": "kyma-system", "name": "templatekyma-sample", "reconcileID": "0960e0eb-eb4d-40c1-804a-58ef541c3fce", "error": "reading install .spec.config resulted in an error for Manifest kyma-system/templatekyma-sample"}
```

It intead should tell the user that the configuration was malformed and give a better error message. Either we introduce a dedicated Config Struct for this or we find another way for better validation.

AC:

- [ ] Introduce a better error handling for this scenario
- [ ] Fail with an error message hinting at the content of the configuration of the module being broken
- [ ] Introduce the test for this case

nesmabadr:
- Accept any valid yaml file for the config layer (including yaml files which have comments only)
- Introduced a new struct to be able to decode the configs
- Added a unit test for parsing the install configs function

Fixes #502


internal/manifest/v1beta1/spec_resolver.go
```diff
@@ -174,27 +172,28 @@ func (m *ManifestSpecResolver) getValuesFromConfig(
...
+	ErrConfigObjectInvalid = errors.New(".spec.config is invalid")
 )
 
-func parseInstallConfigs(decodedConfig interface{}) ([]interface{}, error) {
+func ParseInstallConfigs(decodedConfig interface{}) ([]interface{}, error) {
 	var configs []interface{}
-	installConfigObj, decodeOk := decodedConfig.(map[string]interface{})
-	if !decodeOk {
+
+	type ConfigsYaml struct {
```
ruanxin, line 181: Define a struct inside function in general not a best practice, for example, you immediately lost the benefits to implement function for this struct, although in this case, you don't need it.

internal/manifest/v1beta1/spec_resolver.go
```diff
@@ -174,27 +172,28 @@ func (m *ManifestSpecResolver) getValuesFromConfig(
...
+		Configs []interface{}
+	}
+
+	var configsValues ConfigsYaml
+	err := mapstructure.Decode(decodedConfig, &configsValues)
```
ruanxin, line 186: I think we don't need to introduce this mapstructure library just for this purpose, when we introduce some new library, please always keep in mind it introduce dependencies to 3rd parties which is not under our control, especially more things for security scanner to bother us.

Regarding this issue, I think we can just return an empty `config` if the cast decoding not return `decodeOk`, right? or do you have specific concern here?

nesmabadr: No I was just using it for the decoding.

internal/manifest/v1beta1/spec_resolver.go
```diff
@@ -174,27 +173,31 @@ func (m *ManifestSpecResolver) getValuesFromConfig(
...
+type ConfigsYaml struct {
+	Configs []interface{}
+}
+
+func ParseInstallConfigs(decodedConfig interface{}) ([]interface{}, error) {
 	var configs []interface{}
-	installConfigObj, decodeOk := decodedConfig.(map[string]interface{})
-	if !decodeOk {
+	var configsValues ConfigsYaml
+	configsValuesYamlObj, err := json.Marshal(decodedConfig)
```
ruanxin, line 186: Do we have a specific reason to use `json.marshal` and `unmarshal`? Suppose you check the code how this `decodedConfig` get generated. In that case, it's actually from a yaml file, which already go through the yaml to json decoding, and be aware, the operation to do those transferring is expansive, will affect the reconcile performance.

```
err := yaml.NewYAMLOrJSONDecoder(blob, internal.YamlDecodeBufferSize).Decode(&decodedConfig)
```
Besides this, I suggest for this case, the test should test directly from reading an empty config.yaml to check the final parsed config []interface{}

internal/manifest/v1beta1/test-config-files/empty-configs.yaml
```diff
@@ -0,0 +1,5 @@
+# Samples Config
```
ruanxin, line 1: it's not a recommended approach to put test-related data together with production code, instead you can put it under `pkg/test_samples`

ruanxin (approved):

PR #504, opened 2023-04-03
kyma_controller: Add Channel Whitelisting Logic

### LINKED ISSUE #455: Limited modules availability per Kyma instance
**Description:**
Provide a way to limit the list modules available for the Kyma instance. It can be a label or annotation on the Kyma instance created in the control plane. The expected behaviour is that only module templates that match the Kyma resource label can be copied to the SKR and used to install modules.

In other words, there are groups of users in the same landscapes (dev-stage-prod) that should have different modules available to them. It is important to provide a mechanism that will be able to determine module templates should be synced to which SKRs.

Example:
- Kyma runtime A, used by an external customer, should not have an access to modules marked as internal
- Kyma runtime B, used for internal development, should have an access to internal modules

**Reasoning**
We need to support internal modules that should not be shared with other users. Those modules should be available to certain groups of clusters.

LeelaChacha:
### ♻️ Changes:
* Label-based channel whitelisting:
    * The following label must be set for a channel to whitelist it-> operator.kyma-project.io/whitelist-channel-{channelName}: "true"
    * This label is not synced in SKR Kyma copy.
    * Only a value of "true" whitelists a channel.
    * ModuleTemplates from non-whitelisted channels are not synced to the SKR.
    * Manifests are not created for modules with non-whitelisted channels as templates.
 
### ⚠️ Note:
* AFAIK, labels longer than 64 characters cause problems. This approach will limit the length of a channel name.

Closes #455

### DIFF
--- a/api/v1beta1/operator_labels.go
+++ b/api/v1beta1/operator_labels.go
@@ -3,6 +3,7 @@ package v1beta1
 import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"strings"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


LeelaChacha:
Everthing works as expected. Local testing with two-cluster setup demonstrates desired behaviour.
#### TODO: Write Env Tests

janmedrek:
@LeelaChacha let's align when the PR is ready, we will need to make sure that the communication is there and that others can adapt to the changes before we enable that in our environments. 🙂

ruanxin:
This PR will be closed due to the implementation is not aligned with current request.

Issue #302, opened 2022-11-07
Implement Watcher CR state handling


khlifi411:
**Description**
We are missing a state on Watcher CR if the Virtual Service has been created/updated correctly. 

**Reasons**
Central (in-KCP) Module developers would like to know what is the state of the resources managed by the Watcher Reconciler.

**ACs**
- [x] Propose a solution regarding the State object, we can do something similar to the one in Kyma CR
- [x] Implement state handling for watcher CRs in the watcher reconciler
- [x] Implement tests for state handling

**Related discussions**
https://github.com/kyma-project/lifecycle-manager/pull/277#discussion_r1014820500

jakobmoellersap:
Please add ACs @khlifi411

PR #516, opened 2023-04-06
Increase workqueue workers for KCP

### LINKED ISSUE #515: Increase workqueue workers in KCP
At the moment, the worker configured for lifecycle manager kyma-controller and manifest-controller are default one, which means that only one Kyma custom resource (CR) can be processed at a time. This leads to high latency when there are many Kyma CRs in the queue, as observed in the KCP DEV grafana dashboard, the latency increased to 3s for around 100 kyma CR.

The proposed solution is to increase the number of workers that can run in parallel for each controller, by adjusting the `MaxConcurrentReconciles` parameter. This parameter determines how many reconciliations can be performed simultaneously by a controller. By increasing this value, more Kyma CRs can be processed in parallel, reducing the queue size and latency.

AC:

- [x] Implement the logic to allow each controller (kyma, manifest, watcher) to configure its own `MaxConcurrentReconciles` value.
- [x] Expose each of these values as a flag that can be set during deployment, with a default value of 1.
- [x] Update the control-plane Kustomize config to set the `MaxConcurrentReconciles` value for kyma controller and manifest controller to 10.
- [x] Deploy the updated Lifecycle Manager in KCP and monitor the workqueue behavior, latency and resource consumption. If needed, adjust the values accordingly.

nesmabadr:
- Added `max-concurrent-kyma-reconciles` flag for the Kyma controller, with value for control-plane: 10
- Added `max-concurrent-manifest-reconciles` flag for the Manifest controller, with value for control-plane: 10
- Added `max-concurrent-watcher-reconciles` flag for the Manifest controller, with default value of 1

Closes #515

### DIFF
--- a/main.go
+++ b/main.go
@@ -203,8 +203,8 @@ func controllerOptionsFromFlagVar(flagVar *FlagVar) controller.Options {
 				Limiter: rate.NewLimiter(rate.Limit(flagVar.rateLimiterFrequency), flagVar.rateLimiterBurst),
 			},
 		),
-		MaxConcurrentReconciles: flagVar.maxConcurrentReconciles,
-		CacheSyncTimeout:        flagVar.cacheSyncTimeout,
+
+		CacheSyncTimeout: flagVar.cacheSyncTimeout,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -304,7 +308,7 @@ func setupKcpWatcherReconciler(mgr ctrl.Manager, options controller.Options, fla
 	// In total, we probably only have 20 watcher CRs, one worker can sufficiently handle it,
 	// and we don't have to deal with concurrent write to virtual service.
 	// although eventually the write operation will succeed.
-	options.MaxConcurrentReconciles = 1
```
ruanxin, line 307: The comment can be deleted. After this issue https://github.com/kyma-project/lifecycle-manager/issues/302 is resolved, we don't need to limit workers for watcher, and also with your change, this comment is irrelevant anyway.

ruanxin (approved):

PR #517, opened 2023-04-06
kyma_controller: Add Kyma CR Conditions Whitelist

### LINKED ISSUE #509: Refactor status.InitConditions function to handle with any future condition type will be deprecated.
At the moment, the `status.InitConditions` can only handle `DeprecatedConditionTypeReady` which is hardcoded, and can only handle with this specific condition. The better way is to write a function that is more generic, that we can handle with any condition in the future will be deprecated.
Idea: whitelist our current support condition type, and condition not in the whitelist should be removed during init condition.

AC:
- [ ] the InitConditions can remove any condition which consider being deprecated in the future.

LeelaChacha:
Closes #509

### DIFF
--- a/controllers/kyma_controller_test.go
+++ b/controllers/kyma_controller_test.go
@@ -1,17 +1,14 @@
 package controllers_test
 
 import (
-	"context"
 	"errors"
 	"time"
 
+	"github.com/kyma-project/lifecycle-manager/api/v1beta1"
+	. "github.com/kyma-project/lifecycle-manager/pkg/testutils"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	"github.com/kyma-project/lifecycle-manager/api/v1beta1"
-	. "github.com/kyma-project/lifecycle-manager/pkg/testutils"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


pkg/status/conditions.go
```diff
@@ -1,22 +1,27 @@
...
-		apimeta.RemoveStatusCondition(existingCondition, string(v1beta1.DeprecatedConditionTypeReady))
-	}
+	removeDeprecatedConditions(kyma)
 	// Add required Conditions
 	for _, cond := range v1beta1.GetRequiredConditionTypes(kyma.Spec.Sync.Enabled, watcherEnabled) {
 		kyma.UpdateCondition(cond, metav1.ConditionFalse)
 	}
 }
+
+func removeDeprecatedConditions(kyma *v1beta1.Kyma) {
```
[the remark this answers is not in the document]
LeelaChacha: Unit Test Implemented for InitConditions function. Test is passing

controllers/kyma_controller_test.go
```diff
@@ -43,13 +41,22 @@ var _ = Describe("Kyma with deprecated Condition", Ordered, func() {
...
 			LastTransitionTime: metav1.Time{Time: time.Now()},
 		})
-		kyma.ManagedFields = nil
-		Eventually(controlPlaneClient.Patch(ctx, kyma, client.Apply,
-			client.FieldOwner(v1beta1.UnmanagedKyma)), Timeout*2, Interval).Should(Succeed())
-		By("having transitioned the CR State to Ready as there are no modules")
-		Eventually(CheckKymaConditions(ctx, controlPlaneClient, kyma.GetName(),
-			[]v1beta1.KymaConditionType{v1beta1.ConditionTypeModules}),
-			Timeout, Interval).Should(BeTrue())
+		kyma.Status.Conditions = append(kyma.Status.Conditions, metav1.Condition{
```
ruanxin, line 44: This test is fine, but in this test, you only test the `InitConditions`, which can be written with a simple unit test, and doesn't need to be based on Ginkgo framework.

You can reference this [unit test example](https://github.com/kyma-project/lifecycle-manager/blob/main/internal/manifest/v1beta1/spec_resolver_test.go) to check how to write a unit test, and please also reference this [goland test guide](https://www.jetbrains.com/help/go/testing.html), it offers us a feature to automatically generate good unit test template to unit test the functions.

LeelaChacha: But Ginkgo is a general framework for testing in Golang. I think it is better to stick to one framework for consistency

ruanxin (approved):

PR #528, opened 2023-04-14
Feature/#457 support remote module templates

### LINKED ISSUE #457: Support remote module templates
Provide a way to install a module template that is not available in the control plane but is present in the target cluster.

**Reason**
Teams want to verify that their module template can be applied safely on real environment before they make it available to thousands of customers. Such test case should not require access to Kyma Control Plane (to apply module template there) as we want to keep number of people (teams) having write access to KCP limited. 

**Sample CR**
Here is a proposal how it could look like, but feel free to come with own ideas:
```
apiVersion: operator.kyma-project.io/v1beta1
kind: Kyma
metadata:
  name: default-kyma
  namespace: kyma-system
spec:
  channel: regular
  modules:
  - name: {moduleName}
    template: {namespace}/{name}    # reference to moduletemplate resource in my cluster
```

LeelaChacha:
### ♻️ Changes:
* Augment querying SKR cluster for custom ModuleTemplates
* Prevent deletion of custom ModuleTemplates

### ⚠️ Note:
* getTemplateWithoutIndexing() function was needed because we cannot pass ListOptions to the **runtimeClient**.List() function due to lack of indexing ([See Reference](https://github.com/kubernetes-sigs/controller-runtime/issues/1342#issuecomment-762020648)). We do not have access to the manager of on the runtime cluster hence we cannot add a FieldIndexer. So fetching all ModuleTemplates and filtering them manually is the only way, AFAIK.

Closes #457


LeelaChacha:
**TODO: Implement Env Test**

pkg/channel/lookup.go
```diff
@@ -28,14 +29,29 @@ type ModuleTemplate struct {
...
-func GetTemplates(
-	ctx context.Context, client client.Reader, kyma *operatorv1beta1.Kyma,
+func GetTemplates(ctx context.Context, kymaClient client.Reader, kyma *operatorv1beta1.Kyma,
 ) (ModuleTemplatesByModuleName, error) {
 	logger := ctrlLog.FromContext(ctx)
 	templates := make(ModuleTemplatesByModuleName)
 
+	var runtimeClient client.Reader
+	if kyma.Spec.Sync.Enabled {
+		runtimeClient = remote.SyncContextFromContext(ctx).RuntimeClient
```
ruanxin, line 39: If `kyma.Spec.Sync.Enabled` is not enabled, `runtimeClient` will be nil.

Here we should make a decision, if sync is not enabled, then this `RemoteModuleTemplateRef` should not be used, IMO.

pkg/channel/lookup.go
```diff
@@ -133,26 +149,46 @@ type Lookup interface {
...
 }
 
 func (c *TemplateLookup) WithContext(ctx context.Context) (*ModuleTemplate, error) {
 	desiredChannel := c.getDesiredChannel()
 
-	template, err := c.getTemplate(ctx, desiredChannel)
+	var template *operatorv1beta1.ModuleTemplate
+	var err error
+	if c.readerIndexed {
+		template, err = c.getTemplate(ctx, desiredChannel)
```
ruanxin, line 187: If I get your point currently, the concern here to treat differently is because this index field can't be used in remote module template? `err := c.reader.List(ctx, templateList, option, index.TemplateChannelField.WithValue(desiredChannel))`

Let's have a sync regarding this problem, and if it's true, then we should concern about using a unified way to get template, maybe you solution is more generic, introducing more different solutions means more potential divergence and more maintenance efforts, and we should definitely avoid it.

ruanxin (changes requested):
Please check my comments

pkg/remote/remote_catalog.go
```diff
@@ -180,13 +181,22 @@ func (c *RemoteCatalog) diffsToDelete(
 		presentInKCP[kcp[i].Namespace+kcp[i].Name] = struct{}{}
 	}
 	for i := range skr {
-		if _, inKCP := presentInKCP[skr[i].Namespace+skr[i].Name]; !inKCP {
+		if _, inKCP := presentInKCP[skr[i].Namespace+skr[i].Name]; !inKCP && isManagedByKcp(skr[i]) {
```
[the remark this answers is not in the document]
LeelaChacha: test has now been implemented

api/v1beta1/kyma_types.go
```diff
@@ -62,6 +62,11 @@ type Module struct {
 	// +kubebuilder:validation:MinLength:=3
 	Channel string `json:"channel,omitempty"`
 
+	// RemoteModuleTemplateRef is the reference (FQDN, Namespace/Name, Module Name Label)
+	// to the module template on the remote cluster.
+	// If specified, the module template will be fetched from the SKR and reconciled.
+	RemoteModuleTemplateRef string `json:"remoteModuleTemplateRef,omitempty"`
```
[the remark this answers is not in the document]
LeelaChacha: forgot to push. sorry :)

pkg/channel/lookup.go
LeelaChacha: good idea. I thought of this but it felt like it's an very unlikely edge case.
But let's just be sure and check for the sync when using the RemoteModuleTemplateRef.

jeremyharisch (commented):
Code Review Done: Looks good to me!
Still need to test the functionallity

ruanxin:
Since currently we introducing new API upgrade, to simplify implementation, migrate this code to next version.

ruanxin:
I merged the implementation to https://github.com/kyma-project/lifecycle-manager/pull/553 @LeelaChacha

PR #531, opened 2023-04-18
Use GCR registry instead of local k3d registry for module template used in tests

khlifi411:
**Proposed changes**

- replace local k3d registry with remote GCR registry for module template used in tests

### DIFF
--- a/config/samples/component-integration-installed/kcp-template-module.yaml
+++ b/config/samples/component-integration-installed/kcp-template-module.yaml
@@ -28,7 +28,7 @@ spec:
       name: kyma-project.io/template-operator
       provider: internal
       repositoryContexts:
-        - baseUrl: http://k3d-registry.localhost:5111
+        - baseUrl: europe-west3-docker.pkg.dev/sap-kyma-jellyfish-dev/template-operator
           componentNameMapping: urlPath
           type: ociRegistry
       resources:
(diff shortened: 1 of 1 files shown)


config/samples/component-integration-installed/kcp-template-module.yaml
```diff
@@ -28,7 +28,7 @@ spec:
       name: kyma-project.io/template-operator
       provider: internal
       repositoryContexts:
-        - baseUrl: http://k3d-registry.localhost:5111
+        - baseUrl: europe-west3-docker.pkg.dev/sap-kyma-jellyfish-dev/template-operator
```
ruanxin, line 31: Please don't create this new file to serve as testing module template, we previously use this [operator_v1beta1_moduletemplate_kcp-module.yaml](https://github.com/kyma-project/lifecycle-manager/blob/main/config/samples/component-integration-installed/operator_v1beta1_moduletemplate_kcp-module.yaml), which contains version, which is important indicate which version we are using for testing.

ruanxin: What I mean is this file should be changed: https://github.com/kyma-project/lifecycle-manager/blob/main/pkg/testutils/utils.go#L172

khlifi411: Hi, I see that this change is already in main. Is this PR still needed?

Issue #456, opened 2023-03-10
Security Considerations: Module Signature Configuration Enforcement through Signature Validation


jakobmoellerdev:
As a module user, I want to make sure I have the required provenance available to verify that all modules I want to install in my cluster can be trusted and are integral and not mutated. Thus I want to be able to verify the module data through an inbuilt mechanism.

For this to work, the existing module descriptor in the ModuleTemplate is an issue. Since the descriptor is accepted as [runtime.RawExtension](https://github.com/kyma-project/lifecycle-manager/blob/main/api/v1beta1/moduletemplate_types.go#L99) into the API-Server, we currently cannot avoid changing the data of the ModuleTemplate descriptor after it was accepted into the API-Server.

The only way to discover this tampering is by enforcing change denial in the admission webhook that already validates the ModuleTemplate.

For this to work we have to make use of the inbuilt signature support for ModuleTemplates.

Let's imagine a ModuleTemplate is signed by the key`private.pem` and can be verified with `public.pem`. (To sign a module, we can use `kyma alpha sign module` and to verify `kyma alpha verify module`, but there is no way to create a `template.yaml` with the signature included in CLI yet). Let us also assume that the key-pair above was created from an intermediary certificate that is trusted.

Now we need to assure the following for the `ModuleTemplate`:
1. Every time a change is made to the `ModuleTemplate` without a signature, we need to deny the request (as the signature always needs to be `enforced` or otherwise it could be skipped), this can also be termed `strict` validation
2. Every time a change is made to the `ModuleTemplate` with a signature, we need to verify the signature with a public-key that is either
  - inbuilt into the binary (e.g. for all kyma-modules we have an inbuilt public key)
  - accessible as a trusted secret in the cluster (e.g. for third-party signatures or modules)
3. When the Signature is accepted, only then the Update of the AdmissionWebhook will be accepted as well.
4. In case of denial, we will deny the request to change the ModuleTemplate.

Now that we setup the ModuleTemplate for verification, the next step is to safeguard the `Manifest` from changes. here we have a few Options:

1. The easy way: Rely on Lifecycle Manager to reconcile out bad changes inconsistent with the ModuleTemplate but accept side-effects between the discovered change and the correction.
[...]

PR #532, opened 2023-04-18
Allow each module template have it's own signature

ruanxin:
This PR refactored the existing signature verification process, the problem of the current implementation is, it expected all modules use a shared signature. But for principle of least privilege, it's not a sold concern to use a shared signature, using shared signatures can potentially create security risks, as it may grant unnecessary access or permissions to certain modules that do not require them.  it's better allow each module to have it's own signature.

Related issues: https://github.com/kyma-project/lifecycle-manager/issues/456


pkg/module/parse/template_to_module.go
```diff
@@ -32,26 +33,31 @@ type Parser struct {
...
 		InKCPMode:                inKCPMode,
 		remoteSyncNamespace:      remoteSyncNamespace,
+		EnableVerification:       enableVerification,
+		PublicKeyFilePath:        publicKeyFilePath,
 	}
 }
 
-func (p *Parser) GenerateModulesFromTemplates(ctx context.Context,
+func (parser *Parser) GenerateModulesFromTemplates(ctx context.Context,
 	kyma *v1beta2.Kyma,
```
medmes, line 59: I prefer to keep it as p instead of parser.

pkg/module/parse/template_to_module.go
```diff
@@ -117,14 +122,13 @@ func overwriteNameAndNamespace(template *channel.ModuleTemplateTO, name, namespa
 	}
 }
 
-func (p *Parser) newManifestFromTemplate(
+func (parser *Parser) newManifestFromTemplate(
 	ctx context.Context,
```
medmes, line 126: Here too!

pkg/module/parse/template_to_module.go
```diff
@@ -150,13 +154,22 @@ func (p *Parser) newManifestFromTemplate(
...
+		parser.Client,
+		parser.EnableVerification,
+		parser.PublicKeyFilePath,
+		module.Name)
+	if err != nil {
+		return nil, err
+	}
+
 	if err := signature.Verify(componentDescriptor, verification); err != nil {
 		return nil, fmt.Errorf("could not verify descriptor: %w", err)
```
medmes, line 174: Better to specify the message log as it's a signature verification than a general template verification. Maybe: "could not verify signature"

medmes, line 174: I am not sure if the parser package will be also responsible to verify module signature. For me what the parser responsibility is to parse, unmarshal and deserlize the template. What do you think?

medmes (commented):
All good just small questions/comments!

pkg/module/parse/template_to_module.go
ruanxin: To me using full name is always better for readability purposes, I actually haven't found strict rule defined by idiomatic Go for the naming of the receiver, and by this article https://blog.heroku.com/neither-self-nor-this-receivers-in-go#naming-the-receiver, actually there is no restriction for this, if you find use full name is an antipattern, I'm free to change it, otherwise I still prefer use fullname, don't like this short name, especially for long function. 
```
If the name of the struct is Server, we'll usually see s or srv or even server. 
All of these are fine—short is convenient, but it's more about uniquely identifying the variable in a consistent way.
```

pkg/module/parse/template_to_module.go
ruanxin: Depending on the use case and the scope of the `parse`, this parse package defined here is responsible for parsing moduletemplate to manifest cr, and verification is the necessary middle step to be involved, if can't pass verification, then it wouldn't be converted to manifest cr, tbh, I don't see the responsible problem in this case.

Maybe the package name is a bit misleading, instead of `parse`, calling it `converter` is more suitable. But this can be a follow up issue if needed.

pkg/module/parse/template_to_module.go
medmes: I got your point, just for the purpose of the consistency with other receivers I prefer it with `p` instead of `parser`.
Of course, I strongly disagree of use of `this` or `self` for example as its described on the previous link as it;s remind to oop language.
Please if you don't mind change it back to `p`!

pkg/module/parse/template_to_module.go
medmes: Yep, if so its a bit misleading. Maybe a refactoring of the package name will be better.

medmes (approved):

PR #537, opened 2023-04-20
Reference the module CR in the Kyma CR

### LINKED ISSUE #506: Provide a reference to the module CR after enabling module
**Description:**
After enabling the module it is unclear which resources were created and can be used to configure that module. To track it the user would need to go to the ModuleTemplate and check what the apiVersion is and what kind of a resource was created and then ask for a specific resource.

This can be simplified by providing a direct reference to the CR that was created in the Status field of the Kyma CR, i.e. for Keda it would be something like this:
```
modules:
    - name: keda
      state: Ready
      channel: alpha 
      ...
      resource:
        apiVersion: operator.kyma-project.io/v1beta1
        kind: Keda
        metadata:
          name: keda-sample
          namespace: kyma-system
```

**Acceptance Criteria:**
- [ ] Kyma CR contains a direct reference to the module CR that was created by enabling a module

nesmabadr:
**Description**

- Reference the module CR in the Kyma CR

**Related issue(s)**
Resolves #506

### DIFF
--- a/pkg/module/sync/runner_impl.go
+++ b/pkg/module/sync/runner_impl.go
@@ -124,11 +124,14 @@ func (r *RunnerImpl) updateModuleStatusFromExistingModules(
 		manifestObject := module.Object.(*v1beta1.Manifest)
 		manifestAPIVersion, manifestKind := manifestObject.GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
 		templateAPIVersion, templateKind := module.Template.GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
-		moduleCRAPIVersion := ""
-		moduleCRKind := ""
+		var moduleResource *v1beta1.TrackingObject
 		if manifestObject.Spec.Resource != nil {
-			moduleCRAPIVersion, moduleCRKind = manifestObject.Spec.Resource.
+			moduleCRAPIVersion, moduleCRKind := manifestObject.Spec.Resource.
 				GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
+			moduleResource = &v1beta1.TrackingObject{
+				PartialMeta: v1beta1.PartialMetaFromObject(manifestObject.Spec.Resource),
+				TypeMeta:    metav1.TypeMeta{Kind: moduleCRKind, APIVersion: moduleCRAPIVersion},
+			}
 		}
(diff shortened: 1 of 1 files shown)


pkg/module/sync/runner_impl.go
```diff
@@ -121,22 +121,29 @@ func (r *RunnerImpl) updateModuleStatusFromExistingModules(
 
 	for idx := range modules {
 		module := modules[idx]
-		manifestAPIVersion, manifestKind := module.Object.GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
+		manifestObject := module.Object.(*v1beta1.Manifest)
+		manifestAPIVersion, manifestKind := manifestObject.GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
 		templateAPIVersion, templateKind := module.Template.GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
+		moduleCRAPIVersion, moduleCRKind := manifestObject.Spec.Resource.
```
ruanxin, line 127: `manifestObject.Spec.Resource` is nullable, you need to check this.

pkg/module/sync/runner_impl.go
```diff
@@ -121,22 +121,33 @@ func (r *RunnerImpl) updateModuleStatusFromExistingModules(
 
 	for idx := range modules {
 		module := modules[idx]
-		manifestAPIVersion, manifestKind := module.Object.GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
+		manifestObject := module.Object.(*v1beta1.Manifest)
+		manifestAPIVersion, manifestKind := manifestObject.GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
 		templateAPIVersion, templateKind := module.Template.GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
+		moduleCRAPIVersion := ""
+		moduleCRKind := ""
+		if manifestObject.Spec.Resource != nil {
```
ruanxin, line 129: If the `manifestObject.Spec.Resource` is nil, then the ModuleStatus.Resource should also nil, you don't need to filling some empty object.

ruanxin: that's why design the `Resource *TrackingObject` as a pointer reference, so if it's not initiated, this filed will be empty.

ruanxin (approved):

Issue #544, opened 2023-04-21
Remove `spec.sync` from the Kyma CR


janmedrek:
**Description**

The `sync` section of the Kyma CR should be removed. This change will simplify the Kyma CR and eliminate confusion around the `sync` section.

**Reasons**

The `spec.sync` section is currently a bit confusing to the user, and its behaviour is not meaningful in the SKR. Additionally, recent changes to the requirements for module availability and channels have made the `sync` section less relevant. We would prefer to base the synchronisation logic on the labels/different solutions in the KCP, such as this proposal: https://github.com/kyma-project/lifecycle-manager/issues/542#issuecomment-1517370383

**Acceptance Criteria**
- [ ] `spec.sync` removed from the Kyma CR
- [ ] Documentation reflects the changes

**Attachments**

https://github.com/kyma-project/lifecycle-manager/issues/542
https://github.com/kyma-project/lifecycle-manager/issues/455

Issue #542, opened 2023-04-20
Provide a way to stop ModuleTemplate synchronization to the SKRs


janmedrek:
**Description**

Provide a way to stop specific ModuleTemplate synchronisation to the SKR clusters.

**Reasons**

Currently, there is no way to prevent a single ModuleTemplate from being synced to the SKR clusters. It would be useful to have a possibility to disable the sync for a specific module, i.e. during the migration process. Right now the only option to achieve similar result is to turn off Module Catalog sync, which affects other ModuleTemplates as well.

**Acceptance Criteria**
- [ ] Create a way to stop single ModuleTemplate synchronisation to the SKR clusters
- [ ] The marked ModuleTemplates should not be put into the SKR clusters
- [ ] The logic could be based on a label, there is already a similar [`skip-reconciliation` label](https://github.com/kyma-project/lifecycle-manager/tree/main/api#label-based-side-effects-from-operatorkyma-projectio) for Kyma CR
- [ ] Provide a documentation of the feature

**Attachments**

N/A

pbochynski:
If you base the implementation of https://github.com/kyma-project/lifecycle-manager/issues/455 on labels we can use it to fulfil this requirement. I can imagine that both Kyma and ModuleTemplate resources in the control plane can have labels such:
- operator.kyma-project.io/sync=true|false
- operator.kyma-project.io/internal=true|false
- operator.kyma-project.io/beta=true|false

Scenarions:
- Default: Kyma and ModuleTemplate have label operator.kyma-project.io/sync=true
- Disable module synchronisation in one cluster: remove label operator.kyma-project.io/sync from Kyma resource
- Disable synchronisation of single module: remove label operator.kyma-project.io/sync from ModuleTemplate resource
- Enable internal modules for cluster: add label operator.kyma-project.io/internal on Kyma
- Enable internal module for external customer (publish): remove label operator.kyma-project.io/internal from ModuleTemplate
- Disable synchronisation of beta modules in cluster: add label perator.kyma-project.io/beta=false on Kyma resource

Tomasz-Smelcerz-SAP:
A quick summary of the provided AC:

scenario name: |   | Disable sync | default | disable single module | publish beta modules | publish internal modules
-- | -- | -- | -- | -- | -- | --
  |   |   |   |   |   |  
labels on the Kyma instance: |   |   |   |   |   |  
  | sync: | FALSE | TRUE | TRUE | TRUE | TRUE
[...]

Issue #540, opened 2023-04-20
Remove `target` field from the ModuleTemplate


janmedrek:
**Description**

Currently, the `target` field in the ModuleTemplate CR is not needed anymore and only introduces confusion. This feature request is to drop the `target` field from the ModuleTemplate CR.

Additionally, we would like to adjust the Lifecycle Manager logic to always deploy the ModuleTemplates in the previously called `remote` mode, unless it is deployed in a local k3d cluster (it should be automatically detected).

**Reasons**

The `target` field in the ModuleTemplate CR is no longer used, and its presence only leads to confusion. Dropping it will make managing the ModuleTemplate CR simpler and more straightforward.

The proposed change to the Lifecycle Manager logic will ensure that ModuleTemplates are always deployed in the correct mode, regardless of the environment in which they are deployed. This will reduce the likelihood of errors and ensure consistent behaviour across environments.

**Acceptance Criteria**

- [ ] The `target` field should be removed from the ModuleTemplate CR.
- [ ] The Lifecycle Manager logic should be updated to always treat the ModuleTemplates as a `remote` mode unless it is deployed in a local k3d cluster.
- [ ] The above changes should be reflected in the documentation.

**Attachments**

None.

Issue #507, opened 2023-04-03
Status for faulty modules is missing in the Kyma CR


janmedrek:
**Description:**
Whenever an invalid module is enabled the Kyma goes into the `error` state (as expected). However, there will be no information on which module caused an error - this is because KLM cannot fetch the module and thus that module is not present in the module list in the status section of the CR.

The easiest way to reproduce such a situation is to enable a module that has no ModuleTemplate available for it. Then, the error will pop up in the `lastOperation` struct:
```
lastOperation:
      lastUpdateTime: "2023-04-03T17:19:34Z"
      operation: 'error while fetching modules during processing: templates could
        not be fetched: no unique template could be identified: no module template
        found for module: missing-module, attempted to lookup via [map[operator.kyma-project.io/module-name:missing-module]
        map[spec.descriptor.component.name:missing-module] map[metadata.name:missing-module]]'
```

There will be no module listed in the `modules` list:
```
modules:
    - name: keda
       ...
```

The expected status would be to have something like this:
```
modules:
    - name: keda
       ...
    - name: missing-module
       state: Error
       (+some info about the error)
```

**Acceptance Criteria:**
- [ ] Invalid modules pop up on the `status.modules` list
- [ ] The error state and short error message/explanation is visible

PR #553, opened 2023-04-25
provide v1beta2

ruanxin:
# Related Issue:
- https://github.com/kyma-project/lifecycle-manager/issues/544
  - Drop `Sync.Namespace` support for customized SKR Kyma CR namespace, based on the decision made with @janmedrek 
  - `Sync.SyncStrategy` replaced with an annotation on Kyma CR `SyncStrategyAnnotation`, by default, the sync strategy is always `SyncStrategyLocalSecret`, only when this annotation is configured with `SyncStrategyLocalClient` as value, it uses local client, and only for env test purpose.
  - `Sync.ModuleCatalog`, `Sync.Enabled` relies on https://github.com/kyma-project/lifecycle-manager/issues/542
  - `Sync.NoModuleCopy` replaced with the logic always not copy init modules to remote Kyma CR.

- https://github.com/kyma-project/lifecycle-manager/issues/540
  - new logic relies on `in-kcp-mode` flag to determine which mode lifecycle manager should be, if `true`, module always deploys to remote, by default `false`, deploy in the same cluster as single cluster mode.

- https://github.com/kyma-project/lifecycle-manager/issues/507

- Include the implementation from this PR https://github.com/kyma-project/lifecycle-manager/pull/528 @LeelaChacha 
# TODO
- revisit TODO and skipped test after #542 merged


lindnerby:
lgtm, as discussed after this version increment we should talk about, if we want to increase/align versions of all resources always of if we do a fine-grained version upgrade only for the type that is affected by changes

Tomasz-Smelcerz-SAP (approved):

lindnerby (approved):

api/README.md
```diff
@@ -106,55 +104,6 @@ In addition to this very flexible way of referencing modules, there is also anot
 While `CreateAndDelete` will cause the ModuleTemplate's `.spec.data` to be created and deleted to initialize a module with preconfigured defaults, `Ignore` can be used to only initialize the operator without initializing any default data. 
 This allows users to be fully flexible when and how to initialize their Module.
 
-### `.spec.sync`
```
ruanxin, line 109: The alternative to this spec.sync should be documented

api/README.md
```diff
@@ -271,24 +220,14 @@ spec:
             memory: "150Mi"
 ```
 
-The `namespace` of the resource mentioned in `.spec.data` will be defaulted to `.spec.sync.namespace` or `.metadata.namespace` of the referencing `Kyma` if not filled, otherwise it will be respected. All other attributes (including `.metadata.name` and `apiVersion` and `kind`) are taken over as is. Note that since it behaves similarly to a `template` any subresources, like `status` are ignored, even if specified in the field.
+The `namespace` of the resource mentioned in `.spec.data` will be defaulted to `.metadata.namespace` of the referencing `Kyma` if not filled, otherwise it will be respected. All other attributes (including `.metadata.name` and `apiVersion` and `kind`) are taken over as is. Note that since it behaves similarly to a `template` any subresources, like `status` are ignored, even if specified in the field.
```
ruanxin, line 223: This is not the current implementation, the namespace will be controlled by flag `sync-namespace` if  `.spec.data` not mentioned

PR #558, opened 2023-04-28
cleanup helm and kustomize renderers

khlifi411:
**Description**

- ...
- ...
- ...

**Related issue(s)**


api/v1beta2/manifest_types.go
```diff
@@ -27,7 +27,7 @@ const ManifestKind = "Manifest"
 
 // InstallInfo defines installation information.
 type InstallInfo struct {
-	// Source can either be described as ImageSpec, HelmChartSpec or KustomizeSpec
+	// Source convert to ImageSpec
```
Tomasz-Smelcerz-SAP, line 30:
```suggestion
	// Source in the ImageSpec format
```

config/crd/bases/operator.kyma-project.io_manifests.yaml
```diff
@@ -457,8 +457,7 @@ spec:
                     description: Name specifies a unique install name for Manifest
                     type: string
                   source:
-                    description: Source can either be described as ImageSpec, HelmChartSpec
-                      or KustomizeSpec
+                    description: Source convert to ImageSpec
```
Tomasz-Smelcerz-SAP, line 460:
```suggestion
                    description: Source in the ImageSpec format
```

config/crd/bases/operator.kyma-project.io_manifests.yaml
```diff
@@ -714,8 +713,7 @@ spec:
...
```
Tomasz-Smelcerz-SAP, line 716:
```suggestion
                    description: Source in the ImageSpec format
```

internal/declarative/v2/inmemory_rendered.go
```diff
@@ -30,10 +34,8 @@ type InMemoryManifestCache struct {
 func (c *InMemoryManifestCache) Parse(
 	ctx context.Context, renderer Renderer, obj Object, spec *Spec,
 ) (*internal.ManifestResources, error) {
-	file := filepath.Join(manifest, spec.Path, spec.ManifestName)
-	hashedValues, _ := internal.CalculateHash(spec.Values)
```
Tomasz-Smelcerz-SAP, line 34: This is removed, and from what I can see, the `spec.Values` are no longer used. Perhaps it should be alse removed then?

internal/declarative/v2/renderer_raw.go
```diff
@@ -112,80 +41,6 @@ func (r *RawRenderer) Render(_ context.Context, obj Object) ([]byte, error) {
...
-		crdGVK := apiextensionsv1.SchemeGroupVersion.WithKind("CustomResourceDefinition")
-		if resource.GroupVersionKind() == crdGVK {
-			r.crds = append(r.crds, resource)
-			continue
-		}
-	}
-	return nil
-}
-
-func (r *RawRenderer) installCRDs(ctx context.Context) error {
```
Tomasz-Smelcerz-SAP, line 150: Is it really obsolete?

internal/declarative/v2/inmemory_rendered.go
ruanxin: which should be removed? the `spec.Values` previously used as part of the cache key, current change is just avoid using it as the cache key, but the cache is still valid

internal/declarative/v2/renderer_raw.go
ruanxin: this renderer_raw just need to support `Render`, there is no need to implement EnsurePrerequisites, RemovePrerequisites, we previously already finish the implementation of all three rendering mode. And this additional is added by @khlifi411 since his previous implementation which is not needed.

Please check my comments [here](https://github.com/kyma-project/lifecycle-manager/pull/469#discussion_r1171675364)
What I did here is bring it back to the previous working version.

internal/declarative/v2/inmemory_rendered.go
Tomasz-Smelcerz-SAP: @ruanxin I meant `Spec.Values` from `internal/declarative/v2` package. After your changes I see it's no longer used in the project. Or I missed something? If the `Spec.Values` is no longer used, maybe it's good time to also remove it from the project?

internal/declarative/v2/renderer_raw.go
Tomasz-Smelcerz-SAP: :+1:

internal/declarative/v2/inmemory_rendered.go
ruanxin: good catch, `Spec.Values` removed

Tomasz-Smelcerz-SAP (approved):

PR #562, opened 2023-05-03
Refactor SynchronizeRemoteKyma; add Spec and Labels-AnnotationSync

jeremyharisch:
**Description**
We introduced a bug by removing an update call in the `SynchronizeRemoteKyma` function, thus the spec, labels and annotations have not been synced anymore. This PR introduced it again and refactors the sequence of the sync process.

- Renamed InsertWatcherLabelsAnnotations to SyncWatcherLabelsAnnotations
- Refactored the sequence of the sync calls
    - First check if the Kyma is being deleted, if yes nothing needs to be done
    - Sync Watcher Labels and annotations
    - Call update to update labels, annoations and sync spec
    - Update Status
- Introduce new testcase to check for Watcher Labels and Annotations

### DIFF
--- a/controllers/kyma_controller_helper_test.go
+++ b/controllers/kyma_controller_helper_test.go
@@ -22,8 +22,10 @@ import (
 )
 
 var (
-	ErrKymaNotFound          = errors.New("kyma not exists")
-	ErrExpectedLabelNotReset = errors.New("expected label not reset")
+	ErrKymaNotFound             = errors.New("kyma not exists")
+	ErrExpectedLabelNotReset    = errors.New("expected label not reset")
+	ErrWatcherLabelMissing      = errors.New("watcher label missing")
+	ErrWatcherAnnotationMissing = errors.New("watcher annotation missing")
 )
 
 func RegisterDefaultLifecycleForKyma(kyma *v1beta1.Kyma) {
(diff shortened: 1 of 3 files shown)


controllers/kyma_controller_helper_test.go
```diff
@@ -237,6 +239,23 @@ func ModuleTemplatesExist(clnt client.Client, kyma *v1beta1.Kyma, remote bool) f
 	}
 }
 
+func WatcherLabelsAnnotationsExist(clnt client.Client, kyma *v1beta1.Kyma) func() error {
```
ruanxin, line 242: Let's use the normal way to write the test function, not continue to introduce this `return func() error`, basically you can write this function to normal one to `return error`, like `func WatcherLabelsAnnotationsExist(clnt client.Client, kyma *v1beta1.Kyma) error`, so that it's more reusable. 

then in the Eventually condition, write like this:
```
Eventually(WatcherLabelsAnnotationsExist, Timeout, Interval).WithArguments(runtimeClient, kyma).Should(Succeed())
```
There are many existing test are refactored using this way, you can find some examples, like `updateModuleTemplateSpec`

jeremyharisch: Oke, I can do that. I know both ways. I just used this way since the other test above was written in the same way

controllers/kyma_controller_helper_test.go
```diff
@@ -237,6 +239,21 @@ func ModuleTemplatesExist(clnt client.Client, kyma *v1beta1.Kyma, remote bool) f
 	}
 }
 
+func WatcherLabelsAnnotationsExist(clnt client.Client, kyma *v1beta1.Kyma) error {
+	remoteKyma, err := GetKyma(ctx, clnt, kyma.GetName(), kyma.Spec.Sync.Namespace)
```
ruanxin, line 243:
```suggestion
	remoteKyma, err := GetKyma(ctx, clnt, kyma.GetName(), kyma.GetNamespace())
```
so that I don't have to refactor again because of this issue https://github.com/kyma-project/lifecycle-manager/issues/544

controllers/kyma_controller_helper_test.go
ruanxin: this test is failing

ruanxin (approved):

Issue #505, opened 2023-04-03
Protect the `kyma-system` namespace during module removal


janmedrek:
**Description:**
`kyma-system` namespace is the default for the modules to be installed in. This leads to the necessity of including the namespace in the module YAMLs for the stand-alone (without the KLM) deployment (this is also rendered by default by the kustomize as well).

Whenever a module with `kyma-system` namespace is enabled and later removed the `kyma-system` namespace is put into the `deleting` state. This should not happen as it's the KLM who creates and manages this namespace. Thus it should be protected and even if the namespace definition is among the module YAMLs it should not be removed along the module.

**Acceptance Criteria:**
- [ ] `kyma-system` namespace is protected in module removal process

PR #563, opened 2023-05-04
Filter out namespace kyma-system in declarative reconciler

lindnerby:
**Description**

- `kyma-system` namespace should not be considered when diffing manifest resources

**Related issue(s)**
* https://github.com/kyma-project/lifecycle-manager/issues/505

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -19,6 +19,7 @@ linters:
     - structcheck # deprecated (since v1.49.0) and replaced by 'unused'
     - sqlclosecheck # not needed for this project
     - varcheck # deprecated (since v1.49.0) and replaced by 'unused'
+    - varnamelen # too many false positives
     - wsl # too strict and mostly code is not more readable
     - wrapcheck # TODO enable and fix
 linters-settings:
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -19,7 +19,6 @@ linters:
     - structcheck # deprecated (since v1.49.0) and replaced by 'unused'
     - sqlclosecheck # not needed for this project
     - varcheck # deprecated (since v1.49.0) and replaced by 'unused'
-    - varnamelen # too many false positives
     - wsl # too strict and mostly code is not more readable
     - wrapcheck # TODO enable and fix
 linters-settings:
--- a/internal/declarative/v2/reconciler.go
+++ b/internal/declarative/v2/reconciler.go
@@ -25,7 +25,7 @@ var (
 )
 
 const (
-	kymaSystem = "kyma-system"
+	namespaceNotBeRemoved = "kyma-system"
 )
 
 func NewFromManager(mgr manager.Manager, prototype Object, options ...Option) *Reconciler {
@@ -378,7 +378,7 @@ func pruneKymaSystem(diff []*resource.Info) []*resource.Info {
 		if obj.GetObjectKind().GroupVersionKind().Kind != "Namespace" {
 			continue
 		}
-		if obj.GetName() != kymaSystem {
+		if obj.GetName() != namespaceNotBeRemoved {
 			continue
 		}
 		return append(diff[:i], diff[i+1:]...)
--- a/internal/declarative/v2/reconciler.go
+++ b/internal/declarative/v2/reconciler.go
@@ -357,7 +357,7 @@ func (r *Reconciler) renderTargetResources(
 func (r *Reconciler) pruneDiff(
 	ctx context.Context, clnt Client, obj Object, renderer Renderer, diff []*resource.Info,
 ) error {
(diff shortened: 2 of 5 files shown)


.golangci.yaml
```diff
@@ -19,6 +19,7 @@ linters:
     - structcheck # deprecated (since v1.49.0) and replaced by 'unused'
     - sqlclosecheck # not needed for this project
     - varcheck # deprecated (since v1.49.0) and replaced by 'unused'
+    - varnamelen # too many false positives
```
ruanxin, line 22: I suggest not ignoring this lint, a meaningful name is always helpful for readability purposes and should be consider as best practice, especially for multi-member teams with different experiences. And one bad habit of this go language I found is the short name which is widely used, and it influences newcomers to follow this antipattern to write code. We should prevent it to introduce into our codebase.

internal/declarative/v2/reconciler_test.go
```diff
@@ -0,0 +1,54 @@
...
+		infos := []*resource.Info{
+			obj1,
+			obj2,
+			obj3,
+			obj4,
+		}
+
+		result := pruneKymaSystem(infos)
+
+		require.Len(t, result, 3)
```
ruanxin, line 34: Can you explain this magic `3`?  If it's needed, then please replace it with a meaningful variable

internal/declarative/v2/reconciler_test.go
```diff
@@ -0,0 +1,54 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/cli-runtime/pkg/resource"
+)
+
+func TestPruneKymaSystem(t *testing.T) {
+	t.Parallel()
+	obj1 := &resource.Info{Object: &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "kube-system"}, TypeMeta: metav1.TypeMeta{Kind: "Namespace"}}}
```
ruanxin, line 17: this obj1, obj2.... random number name we should prevent.

internal/declarative/v2/reconciler.go
```diff
@@ -368,6 +368,20 @@ func (r *Reconciler) pruneDiff(
 	return renderer.RemovePrerequisites(ctx, obj)
 }
 
+func pruneKymaSystem(diff []*resource.Info) []*resource.Info {
+	for i, info := range diff {
+		obj := info.Object.(client.Object)
+		if obj.GetObjectKind().GroupVersionKind().Kind != "Namespace" {
+			continue
+		}
+		if obj.GetName() != "kyma-system" {
```
ruanxin, line 377: this magic `kyma-system` it's better to be extracted as a const, and referenced in the test, so that we get the purpose here more clearly

ruanxin (changes requested):
LGTM, mainly readability issues.

internal/declarative/v2/reconciler_test.go
lindnerby: Removing one object from 4 objects will result in 3. What should be more meaningful in this test? It's as simple as it gets :D

internal/declarative/v2/reconciler.go
lindnerby: Const is fine, but it should **not** be referenced in tests, because if it gets changed accidentally the test won't notice.

internal/declarative/v2/reconciler_test.go
lindnerby: Agreeing on this for production code. Will give the test data some names anyways.

internal/declarative/v2/reconciler_test.go
ruanxin: What I mean is this `3` can be replaced with some value to explain it.
```
var resourceAfterPruneKymaNs = 3
```

internal/declarative/v2/reconciler.go
```diff
@@ -24,6 +24,10 @@ var (
 	ErrObjectHasEmptyState                       = errors.New("object has an empty state")
 )
 
+const (
+	kymaSystem = "kyma-system"
```
ruanxin, line 28: this one can be renamed to something like `namespaceNotBeRemoved` so that it's clear why we need to ignore it.

internal/declarative/v2/reconciler.go
ruanxin: after reading your refactored comment, it's fine, what I originally meant for this reference is, to reference this const when you create your input for the test, but of cause not reference it as the expected value for the test. I think that should be clear.

ruanxin (approved):

PR #564, opened 2023-05-04
Detect differences between KCP and SKR CRDs if they have the same version

### LINKED ISSUE #534: Introduce a process to detect differences between KCP and SKR even if CRDs are in the same version
This issue addresses our CRD upgrade strategies. in order to handle **non-breaking** changes in the existing CRD version and still possible to sync with SKR CRD, we want to introduce a process to detect differences between KCP and SKR even if CRDs are in the same version and update SKR CRDs if necessary.

1. Doing time windowed research, find if there is an efficient and robust mechanism to directly compare the remote SKR CRDs spec to KCR CRDs spec, it also needs to run a performance test to approve it's not being bottlenecked.
2. More preferred way: use kyma CR annotation to record existing CRDs generation numbers, both SKR and KCR CRDs, check for differences in CRD generation numbers, and if there are any, apply CRD updates to SKR.

Be aware:
- We should prevent any CRD changes before this issue gets resolved.
- This approach is only for introducing non-breaking changes, but **for any breaking changes, we still need to upgrade the CRD version.**

nesmabadr:
**Description**

- Store Kyma and ModuleTemplate CRD generation as annotations in the Kyma CR for the SKR and KCP clusters
- Upgrade the CRD in SKR if the stored generation is not the same as the latest CRD generation

**Related issue(s)**
Resolves #534


pkg/remote/remote_catalog.go
```diff
@@ -250,8 +253,31 @@ func (c *RemoteCatalog) CreateModuleTemplateCRDInRuntime(ctx context.Context, pl
...
+			kyma.Annotations[v1beta2.SkrModuleTemplateCRDGenerationAnnotation], runtimeCRDGeneration) {
+		err = PatchCRD(ctx, syncContext.RuntimeClient, crd)
+		if err != nil {
+			return err
+		}
+
+		if kyma.Annotations == nil {
+			kyma.Annotations = make(map[string]string)
+		}
+		err = syncContext.RuntimeClient.Get(ctx, client.ObjectKey{
```
ruanxin, line 270: why need this Get request here?

pkg/remote/remote_catalog.go
```diff
@@ -129,8 +132,8 @@ func (c *RemoteCatalog) createOrUpdateCatalog(ctx context.Context,
 	}
 
 	// it can happen that the ModuleTemplate CRD is not existing in the Remote Cluster when we apply it and retry
-	if containsMetaIsNoMatchErr(errs) {
-		if err := c.CreateModuleTemplateCRDInRuntime(ctx, v1beta2.ModuleTemplateKind.Plural()); err != nil {
+	if containsMetaIsNoMatchErr(errs) || len(errs) == 0 {
```
ruanxin, line 135: now I get it, by this, you change the way the `CreateModuleTemplateCRDInRuntime` was originally designed for. It should only designed for create module template crd in remote skr, as the described, it should be really cautious when you change the existing function behavior, because it can introduce unexpected result, for example, if we have test for this function, then probably it will break.

Two options here:
1. rename this function to something like createRemoteCRDAndUpdateKymaAnnotation, but as you can see, this function did two much job, which violates the single responsibility principle. It's also hard for you to write a unit test for it. Which is not suggest way.
2. Put you update Kyma annotation in a different function, just for updating kyma annotation, and also please write a unit test for it.

pkg/remote/kyma_synchronization_context.go
```diff
@@ -172,10 +198,11 @@ func (c *KymaSynchronizationContext) CreateOrFetchRemoteKyma(
...
 
-	if meta.IsNoMatchError(err) {
-		recorder.Event(kyma, "Normal", err.Error(), "CRDs are missing in SKR and will be installed")
-
-		if err := c.CreateOrUpdateCRD(ctx, v1beta2.KymaKind.Plural()); err != nil {
+	if meta.IsNoMatchError(err) || err == nil {
+		if meta.IsNoMatchError(err) {
+			recorder.Event(kyma, "Normal", err.Error(), "CRDs are missing in SKR and will be installed")
+		}
+		if err = c.CreateOrUpdateCRD(ctx, v1beta2.KymaKind.Plural(), kyma); err != nil {
```
ruanxin, line 205: can you use this `CreateOrUpdateCRD` to replace with `CreateModuleTemplateCRDInRuntime`, they basically do same thing

ruanxin (changes requested):
please check my comment

pkg/remote/kyma_synchronization_context.go
```diff
@@ -141,24 +164,52 @@ func (c *KymaSynchronizationContext) CreateOrUpdateCRD(ctx context.Context, plur
...
+	skrAnnotation := v1beta2.SkrKymaCRDGenerationAnnotation
+
+	if plural == v1beta2.ModuleTemplateKind.Plural() {
+		kcpAnnotation = v1beta2.KcpModuleTemplateCRDGenerationAnnotation
+		skrAnnotation = v1beta2.SkrModuleTemplateCRDGenerationAnnotation
+	}
+
+	latestGeneration := strconv.FormatInt(crd.Generation, 10)
+	runtimeCRDGeneration := strconv.FormatInt(crdFromRuntime.Generation, 10)
+	if k8serrors.IsNotFound(err) || !ContainsLatestVersion(crdFromRuntime, v1beta2.GroupVersion.Version) ||
```
ruanxin, line 186: This if condition you can extract to a function, and design a unit test for it, basically this is the most important part for this feature request. You test in what kind of situation, we are patching remote CRD.

pkg/remote/kyma_synchronization_context_test.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	tests := []struct {
+		name string
+		args args
+	}{
+		{
+			name: "Update ModuleTemplate Annotations",
+			args: args{
+				kyma: testutils.NewTestKyma("test-kyma"),
+				skrCRD: &v1extensions.CustomResourceDefinition{
```
ruanxin, line 31: 1. regarding how to design a unit test

these two crd, you don't have to offer as args, you can create them in the test body.

The template will automatically create that args struct, but most time, you'd better adjust them and don't have to use the original one, the args should be easy to read, the idea is to design the input value and expected value for test, the input value is basically for your function, and expect value is the output of the function.

2. this `UpdateKymaAnnotations` function is just for updating annotation, it does not bring much value to writing a test for it, and it's not critical for this feature request.

pkg/remote/kyma_synchronization_context.go
```diff
@@ -128,11 +129,33 @@ func (c *KymaSynchronizationContext) ensureRemoteNamespaceExists(ctx context.Con
...
+	if kcpCRD.Spec.Names.Kind == string(v1beta2.ModuleTemplateKind) {
+		kcpAnnotation = v1beta2.KcpModuleTemplateCRDGenerationAnnotation
+		skrAnnotation = v1beta2.SkrModuleTemplateCRDGenerationAnnotation
+	}
+
+	kyma.Annotations[kcpAnnotation] = strconv.FormatInt(kcpCRD.Generation, 10)
+	kyma.Annotations[skrAnnotation] = strconv.FormatInt(skrCRD.Generation, 10)
+}
+
+func CreateOrUpdateCRD(
```
ruanxin, line 152: since it's a shared function, you need to consider put it in to a better place, in this cse, `crd_upgrade.go` is a good candidate.

pkg/remote/kyma_synchronization_context.go
```diff
@@ -174,11 +225,24 @@ func (c *KymaSynchronizationContext) CreateOrFetchRemoteKyma(
 
 	if meta.IsNoMatchError(err) {
 		recorder.Event(kyma, "Normal", err.Error(), "CRDs are missing in SKR and will be installed")
+	}
 
-		if err := c.CreateOrUpdateCRD(ctx, v1beta2.KymaKind.Plural()); err != nil {
+	if meta.IsNoMatchError(err) || err == nil {
+		var kcpCrd, skrCrd *v1extensions.CustomResourceDefinition
```
ruanxin, line 231: The follow logic you already used in both remote catalog sync and remote kyma sync, then you should consider extract them in to a shared function, and used in both place.

The logic I mean CreateOrUpdateCRD -> check annotation with the crd generation -> patch remote crd

pkg/remote/kyma_synchronization_context.go
```diff
@@ -174,11 +225,24 @@ func (c *KymaSynchronizationContext) CreateOrFetchRemoteKyma(
...
+			ctx, v1beta2.KymaKind.Plural(), kyma, c.RuntimeClient, c.ControlPlaneClient); err != nil {
 			return nil, err
 		}
 
+		if !ContainsLatestCRDGeneration(kyma.Annotations[v1beta2.KcpKymaCRDGenerationAnnotation],
+			strconv.FormatInt(kcpCrd.Generation, 10)) ||
+			!ContainsLatestCRDGeneration(kyma.Annotations[v1beta2.SkrKymaCRDGenerationAnnotation],
+				strconv.FormatInt(skrCrd.Generation, 10)) {
+			UpdateKymaAnnotations(kyma, kcpCrd, skrCrd)
+			if err = c.ControlPlaneClient.Update(ctx, kyma); err != nil {
```
ruanxin, line 242: don't do update kyma in the middle of reconciler loop, always update kyma in the end of reconciliation, mainly two reason:

1. save request to api server, each update is a remote http call, it's expensive regarding performance. 
2. after you update, the kyma in memory is different than the actual kyma in cluster, if you use this in memory kyma update later, it will failed, because the optimistic lock designed by apiserver to prevent old resource get write into etcd.

pkg/remote/kyma_synchronization_context.go
```diff
@@ -128,11 +129,33 @@ func (c *KymaSynchronizationContext) ensureRemoteNamespaceExists(ctx context.Con
 	return nil
 }
 
-func (c *KymaSynchronizationContext) CreateOrUpdateCRD(ctx context.Context, plural string) error {
+func UpdateKymaAnnotations(
```
ruanxin, line 132: this function I suggest you design it more generic, you can create a string type attribute `CrdType` (either `KCP` or `SKR`) to indicate which crd annotation to update with,
```
UpdateKymaAnnotations(kyma *v1beta2.Kyma, crd *v1extensions.CustomResourceDefinition, crdType CrdType) {
     if kyma.Annotations == nil {
		kyma.Annotations = make(map[string]string)
     }
     kyma.Annotations[getAnotation(crd, crdType)] = strconv.FormatInt(crd.Generation, 10)
}
```
for that `getAnotation` function, you determine which annotation based on `crd.Spec.Names.Kind ` and `crdType`

ruanxin (changes requested):
please check my comment

pkg/remote/kyma_synchronization_context_test.go
ruanxin: for point 1, you can have a look at this test as a reference: https://github.com/kyma-project/lifecycle-manager/blob/main/api/v1beta2/moduletemplate_webhook_test.go

nesmabadr: Yes this test now is not of value. I had it earlier because I did the check here, so it served as a test for the conditions as well.

controllers/kyma_controller.go
```diff
@@ -147,12 +145,32 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		if err := r.syncRemoteKymaSpecAndStatus(ctx, kyma); err != nil {
 			return r.CtrlErr(ctx, kyma, fmt.Errorf("could not synchronize remote kyma: %w", err))
 		}
+		if err := r.syncCrdsAndUpdateKymaAnnotations(ctx, kyma); err != nil {
```
ruanxin, line 148: this condition is wrong, you still did update() but did not immediately quit the reconciling, if update success, there is no error returned, you will still continue to state handling, 

you should do something like this, meaning when you detected it needs to update, then update and return immediately. 
```
if syncCrdsAndUpdateKymaAnnotations() {
    if err := r.Update(ctx, kyma); err != nil {
			return err
    }
    return ctrl.Result{}, nil
}
```

pkg/remote/crd_upgrade.go
```diff
@@ -21,6 +29,142 @@ func PatchCRD(ctx context.Context, clnt client.Client, crd *v1extensions.CustomR
...
+	}
+	annotation, err := getAnnotation(crd, crdType)
+	if err != nil {
+		return err
+	}
+	kyma.Annotations[annotation] = strconv.FormatInt(crd.Generation, 10)
+	return nil
+}
+
+func getAnnotation(crd *v1extensions.CustomResourceDefinition, crdType CrdType) (string, error) {
```
ruanxin, line 115: Actually, you can create the annotation dynamically and don't need to return the error, something like:
```
func getAnnotation(crd *v1extensions.CustomResourceDefinition, crdType CrdType) (string) {
     return fmt.sprintf("%s-%s-crd-generation", crd.Spec.Names.Kind, crdType)
}
```

And since you have this function, don't need to declare a cons value in the `operator_annotations.go`

pkg/remote/crd_upgrade.go
```diff
@@ -21,6 +29,142 @@ func PatchCRD(ctx context.Context, clnt client.Client, crd *v1extensions.CustomR
...
+		return false, nil, nil, err
+	}
+
+	return false, nil, nil, nil
+}
+
+func ShouldPatchRemoteCRD(
+	runtimeCrd *v1extensions.CustomResourceDefinition, kcpCrd *v1extensions.CustomResourceDefinition,
+	kyma *v1beta2.Kyma, err error) bool {
+	if k8serrors.IsNotFound(err) {
```
ruanxin, line 91: don't put err as input

pkg/remote/crd_upgrade.go
```diff
@@ -29,3 +173,7 @@ func ContainsLatestVersion(crdFromRuntime *v1extensions.CustomResourceDefinition
 	}
 	return false
 }
+
+func containsLatestCRDGeneration(storedGeneration string, latestGeneration string) bool {
```
ruanxin, line 177: it's just a simple compare, doesn't have to be a function

pkg/remote/crd_upgrade.go
```diff
@@ -21,6 +29,142 @@ func PatchCRD(ctx context.Context, clnt client.Client, crd *v1extensions.CustomR
...
+	SKR CrdType = "SKR"
+)
+
+func updateRemoteCRD(
+	ctx context.Context, plural string, kyma *v1beta2.Kyma, runtimeClient Client, controlPlaneClient Client) (
+	bool, *v1extensions.CustomResourceDefinition, *v1extensions.CustomResourceDefinition, error) {
+	crd := &v1extensions.CustomResourceDefinition{}
+	crdFromRuntime := &v1extensions.CustomResourceDefinition{}
+	var err error
+	err = controlPlaneClient.Get(
```
ruanxin, line 45: Since CRD in KCP can be only updated when we deploy LM, for performance purposes, we can cache the CRD generation, because during the runtime of LM, CRD wouldn't get updated, we can save many remote calls in this case. But this can be done in the future issue. https://github.com/kyma-project/lifecycle-manager/issues/598

pkg/remote/crd_upgrade.go
nesmabadr: We no longer need this check since we do the generation check separately

controllers/kyma_controller.go
```diff
@@ -143,16 +143,37 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
 
 	// update the remote kyma with the state of the control plane
-	if kyma.SyncEnabled() {
+	if kyma.SyncEnabled() { //nolint:nestif
 		if err := r.syncRemoteKymaSpecAndStatus(ctx, kyma); err != nil {
 			return r.CtrlErr(ctx, kyma, fmt.Errorf("could not synchronize remote kyma: %w", err))
 		}
+		updateKymaRequired, err := r.syncCrdsAndUpdateKymaAnnotations(ctx, kyma)
+		if err != nil {
+			return r.CtrlErr(ctx, kyma, fmt.Errorf("could not sync CRDS: %w", err))
```
ruanxin, line 152:
```suggestion
			return r.CtrlErr(ctx, kyma, fmt.Errorf("could not sync CRDs: %w", err))
```

pkg/remote/crd_upgrade.go
```diff
@@ -21,6 +25,136 @@ func PatchCRD(ctx context.Context, clnt client.Client, crd *v1extensions.CustomR
...
+	kyma.Annotations[annotation] = strconv.FormatInt(crd.Generation, 10)
+}
+
+func getAnnotation(crd *v1extensions.CustomResourceDefinition, crdType CrdType) string {
+	return fmt.Sprintf("%s-%s-crd-generation", strings.ToLower(crd.Spec.Names.Kind), strings.ToLower(string(crdType)))
+}
+
+func SyncCrdsAndUpdateKymaAnnotations(ctx context.Context, kyma *v1beta2.Kyma,
+	runtimeClient Client, controlPlaneClient Client) (bool, error) {
+	kcpKymaCrd, skrKymaCrd, err := fetchCrds(ctx, controlPlaneClient, runtimeClient, v1beta2.KymaKind.Plural())
```
ruanxin, line 78: you can extract line 78 to 97 as a function, reuse it to fetchCrds and updateKymAnnotation for kyma and moduletemplate.

line 99 to 119 is basically just copy-paste which you should avoid.

pkg/remote/crd_upgrade.go
```diff
@@ -21,6 +25,136 @@ func PatchCRD(ctx context.Context, clnt client.Client, crd *v1extensions.CustomR
...
+	}
+
+	err = runtimeClient.Get(
+		ctx, client.ObjectKey{
+			Name: fmt.Sprintf("%s.%s", plural, v1beta2.GroupVersion.Group),
+		}, crdFromRuntime,
+	)
+
+	if k8serrors.IsNotFound(err) {
+		err = nil
```
ruanxin, line 148: why not found, you assign nil to err?

pkg/remote/crd_upgrade_test.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	tests := []struct {
+		name string
+		args args
+		want bool
+	}{
+		{
+			name: "Different Skr Generation",
+			args: args{
+				err:                       nil,
```
ruanxin, line 31: If you follow the order of each field in your args struct definition, you don't need to redeclare each fields name here.

pkg/remote/crd_upgrade_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+//nolint:funlen
+func TestShouldPatchRemoteCRD(t *testing.T) {
+	t.Parallel()
+	type args struct {
+		runtimeCrdGeneration      int64
+		kcpCrdGeneration          int64
+		kymaKcpCrdAnnotationValue string
+		kymaSkrCrdAnnotationValue string
+		err                       error
```
ruanxin, line 21: what was the usage of this err here? If have a specific purpose, you need to test this err, but in your test case, all given value is nil.

pkg/remote/crd_upgrade_test.go
```diff
@@ -0,0 +1,100 @@
...
+			},
+			want: true,
+		},
+		{
+			name: "Different Kcp Generation",
+			args: args{
+				err:                       nil,
+				runtimeCrdGeneration:      1,
+				kcpCrdGeneration:          1,
+				kymaKcpCrdAnnotationValue: "0",
```
ruanxin, line 45: there is no such case generation value will be `0`

nesmabadr: it's just a dummy value :D but will change it

pkg/remote/crd_upgrade.go
ruanxin: that's the problem, if crd not created, then the err is `nil`, it will return `crd, crdFromRuntime, nil`, but the `crdFromRuntime` is empty.
and in general, you should not mutate err, err should be treated as a readonly value.

nesmabadr: if the crd is not created, the error Is not nil. it is crd "x" not found.

ruanxin (approved):

Issue #565, opened 2023-05-04
[Bug] remote Kyma CR can't deleted


ruanxin:
When KCP Kyma is under deleting, the remote Kyma CR synced will only sync as deleting status, however, the deletion timestamp is not set, which means the resource itself is not under deleting.

Be aware, in kyma-controller, we have the logic `DeleteRemotelySyncedKyma` implemented, it seems it's not handled properly. 
```
func (r *KymaReconciler) TriggerKymaDeletion(ctx context.Context, kyma *v1beta1.Kyma) error {
	logger := ctrlLog.FromContext(ctx).V(log.InfoLevel)

	if kyma.Spec.Sync.Enabled {
		if err := remote.DeleteRemotelySyncedKyma(ctx, kyma); client.IgnoreNotFound(err) != nil {
			logger.Error(err, "Failed to be deleted remotely!")
			return fmt.Errorf("error occurred while trying to delete remotely synced kyma: %w", err)
		}
		logger.Info("Successfully deleted remotely!")
	}
	return nil
}

```
AC:
- [ ] After KCP Kyma under deleting, the remote Kyma CR should be deleted.

PR #577, opened 2023-05-05
Fix remote kyma creating again after deleted

ruanxin:
**Description**
During Kyma Deleting, the remote synced Kyma actually deleted, however due to the remote sync, the remote Kyma created again when LM detected that it does not exist during sync, this PR fixed this by preventing remote sync if kcp kyma is under deleting.

When a KCP Kyma instance was deleted, the remote-synced Kyma would also get deleted. However, due to the remote sync feature, the remote Kyma would get created again when LM detected that it does not exist during sync.

To fix this issue, this PR prevent remote sync if the Kyma instance was under deletion. This would ensure that the remote synced Kyma does not get recreated when the KCP Kyma is deleted.

**Related issue(s)**
https://github.com/kyma-project/lifecycle-manager/issues/565


jeremyharisch (approved):
Works like a charm 👍🏻

PR #585, opened 2023-05-09
Refactor reconciliation error handling and state handling

lindnerby:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/controllers/kyma_controller.go
+++ b/controllers/kyma_controller.go
@@ -49,13 +49,17 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-type EventErrorType string
+type EventReason string
 
 const (
-	ModuleReconciliationError  EventErrorType = "ModuleReconciliationError"
-	SyncContextError           EventErrorType = "SyncContextError"
-	DeletionError              EventErrorType = "DeletionError"
-	DefaultRemoteSyncNamespace string         = "kyma-system"
+	moduleReconciliationError  EventReason = "ModuleReconciliationError"
+	syncContextError           EventReason = "SyncContextError"
+	deletionError              EventReason = "DeletionError"
+	DefaultRemoteSyncNamespace string      = "kyma-system"
+)
+
+var (
+	errInitSyncCtxFailed = errors.New("initializing sync context failed")
 )
 
 type RequeueIntervals struct {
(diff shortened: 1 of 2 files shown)


pkg/status/conditions_test.go
```diff
@@ -70,7 +75,7 @@ func TestInitConditions(t *testing.T) {
 				Reason:             "Deprecated",
 			})
 
-			status.InitConditions(kyma, tcase.args.syncEnabled, tcase.args.watcherEnabled)
+			status.InitConditions(kyma, tcase.args.watcherEnabled)
```
ruanxin, line 78: if you change to this way, it lacks one test case, which is kyma don't have sync label.

controllers/kyma_controller.go
```diff
@@ -90,95 +91,95 @@ type KymaReconciler struct {
...
-	}
-	return ctrl.Result{}, nil
+func (r *KymaReconciler) requeueWithError(ctx context.Context, kyma *v1beta2.Kyma, err error) (ctrl.Result, error) {
+	return ctrl.Result{Requeue: true}, r.updateStatusWithError(ctx, kyma, err)
 }
 
-func (r *KymaReconciler) CtrlErr(ctx context.Context, kyma *v1beta2.Kyma, err error) (ctrl.Result, error) {
-	return ctrl.Result{Requeue: true},
-		r.UpdateStatusWithEventFromErr(ctx, kyma, v1beta2.StateError, err)
+func (r *KymaReconciler) enqueueWarningEvent(kyma *v1beta2.Kyma, reason EventReason, err error) {
```
ruanxin, line 174: I think it make sense to design two types of EventReason, to distinguish warning and normal event, and avoid misuse.

controllers/kyma_controller.go
```diff
@@ -90,95 +91,95 @@ type KymaReconciler struct {
...
+		}
+		if err := r.updateStatus(ctx, kyma, v1beta2.StateDeleting, "waiting for modules to be deleted"); err != nil {
+			return r.requeueWithError(ctx, kyma, fmt.Errorf("could not update kyma status after triggering deletion: %w", err))
+		}
+		return ctrl.Result{}, nil
 	}
 
-	// check finalizer
-	if kyma.CheckLabelsAndFinalizers() {
+	if needsUpdate := kyma.EnsureLabelsAndFinalizers(); needsUpdate {
```
ruanxin, line 142: what this reason to have this `needsUpdate`?

lindnerby: Readability in terms of why this method returns a bool.

controllers/kyma_controller.go
lindnerby: There are two functions for that now. Reason is just an informative string. What do you mean by can be misused here?

pkg/status/conditions_test.go
lindnerby: Will add a testcase with missing labels

ruanxin (approved):

Issue #550, opened 2023-04-25
Documentation improvements for Lifecycle Manager


janmedrek:
**Description**

The documentation for Lifecycle Manager needs to be improved so that it is more transparent and guides the user through the whole component and Module contribution journey.

**Acceptance Criteria**

- [x] The documentation includes an overview of Lifecycle Manager and its architecture.
- [x] The documentation provides a clear explanation of the different components of Lifecycle Manager.
- [ ] The documentation includes a step-by-step guide for contributing a new Module to Lifecycle Manager.
- [x] The documentation provides examples and use cases to illustrate the functionality of Lifecycle Manager.
- [ ] The documentation describes the channel concept.
- [ ] The documentation is up-to-date with the latest changes and features of Lifecycle Manager.

**Reasons**

Lifecycle Manager is a key component of Kyma, and it is important that users have a clear understanding of how to use it and contribute to it. Improving the documentation will make it easier for users to get started with Lifecycle Manager and reduce the number of support requests and improve the overall user experience.

**Attachments**

N/A

mmitoraj:
We can think about the order of the developer tutorials - from basic to more complex ones. Tutorials' titles could also be changed to imperatives to describe their goal more precisely.

PR #586, opened 2023-05-09
Update structure

mmitoraj:
**Description**

- Updated structure of the Lifecycle Manager documents

**Related issue(s)**
#550


mmitoraj:
In the PR, I made further changes to the documentation structure. I have doubts regarding the `/user-tutorials` dir if we need it now. But maybe I misplaced some of the tutorials and they should be moved from the`/developer-tutorials` folder. @janmedrek have a look :)

janmedrek:
Hey @mmitoraj, I would move the `Managing module initialization with the CustomResourcePolicy` tutorial to the user tutorials, besides that I have no comments. 🙂

I would keep the user tutorials directory for now, when we are done with the content refactor then we can think if it is indeed needed or can be merged to another dir. 🙂

janmedrek (approved):

Tomasz-Smelcerz-SAP (approved):

PR #588, opened 2023-05-09
fix module template update check

ruanxin:
**Description**
This PR mainly fixes the bug that prevents downgrading version during the channel switch. With this change, if the user switch channel and the module version in the channel is lower, in the following case, the module deployed under fast channel with version `v3.1.0`, then switched by user to `regular` channel with version `v3.2.0`.

The expected result:
1. the already deployed manifest will not get updated, stay in `v3.2.0`
2. Kyma.status report this module in ready state with message
3. Kyma CR still result in ready state
4. Kyma.spec.modules changed to `regular` channel

```
apiVersion: operator.kyma-project.io/v1beta2
kind: Kyma
spec:
  channel: alpha
  modules:
  - channel: regular
    name: template-operator
status:
  activeChannel: alpha
 ...
  modules:
  - channel: fast
    fqdn: kyma-project.io/template-operator
    ...
    message: 'module template update not allowed: ignore channel skew, as a higher
      version of the module was previously installed'
    name: template-operator
    state: Ready
    template:
      apiVersion: operator.kyma-project.io/v1beta2
      kind: ModuleTemplate
      metadata:
        generation: 1
        name: template-operator-3.2
        namespace: kcp-system
    version: v3.2.0
  state: Ready

``` 

**Related issue(s)**


pkg/channel/lookup.go
```diff
@@ -82,28 +82,33 @@ func DetermineTemplatesVisibility(kyma *v1beta2.Kyma, templates ModuleTemplatesB
...
 			}
 		}
+		templates[moduleName] = moduleTemplate
 	}
 }
 
-// CheckForOutdatedTemplate verifies if the given ModuleTemplate is outdated and sets their Outdated Flag based on
-// provided Modules, provided by the Cluster as a status of the last known module state.
+func moduleMatch(moduleStatus *v1beta2.ModuleStatus, moduleName string) bool {
+	return moduleStatus.FQDN == moduleName || moduleStatus.Name == moduleName
```
lindnerby, line 101: Not sure if this is correct and we should rather compare FQDN with FQDN and name with name, but rest lgtm.

lindnerby (approved):

PR #590, opened 2023-05-10
Remove v1alpha1 from stored versions

### LINKED ISSUE #574: clean up v1alpha1
- [x] Write a function in the lifecycle manager to remove `v1alpha1` in all CRDs 
golang script reference https://github.com/elastic/cloud-on-k8s/issues/2196#issuecomment-560318404 
- [x] Make sure `v1alpha1` removed in `status.storedVersions` in all CRDs (Kyma, Manifest, ModuleTemplate, Watcher)
- [x] Remove all `v1alpha1` related code in LM, deploy CRDs without `v1alpha1`

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version

nesmabadr:
**Description**

- Remove v1alpha1 from all CRDs storedVersions

**Related issue(s)**
Part of #574


ruanxin (changes requested):
v1alpha1 can't be dropped in crd atm, the script in main.go should be released first, so that it can remove the CRD with v1alpha1 in status.storedVersion, before that happens, v1alpha1 must keep at CRD definition.

main.go
```diff
@@ -357,3 +361,36 @@ func setupKcpWatcherReconciler(mgr ctrl.Manager, options controller.Options, fla
...
+	}
+
+	for _, crdItem := range crdList.Items {
+		if crdItem.Spec.Group != "operator.kyma-project.io" {
+			continue
+		}
+		oldStoredVersions := crdItem.Status.StoredVersions
+		newStoredVersions := make([]string, 0, len(oldStoredVersions))
+		for _, stored := range oldStoredVersions {
+			if stored != "v1beta2" {
```
ruanxin, line 384: you should remove v1alpha1, but not v1beta2

nesmabadr: @ruanxin Yes I had this pushed by mistake, thanks!

main.go
```diff
@@ -357,3 +361,36 @@ func setupKcpWatcherReconciler(mgr ctrl.Manager, options controller.Options, fla
...
+		setupLog.V(log.DebugLevel).Error(err, "unable to initialize client to remove v1alpha1")
+	}
+	ctx := context.TODO()
+	var crdList *v1extensions.CustomResourceDefinitionList
+	if crdList, err = kcpClient.ApiextensionsV1().CustomResourceDefinitions().List(ctx, v1.ListOptions{}); err != nil {
+		setupLog.V(log.DebugLevel).Error(err, "unable to list CRDs")
+	}
+
+	for _, crdItem := range crdList.Items {
+		if crdItem.Spec.Group != "operator.kyma-project.io" {
```
ruanxin, line 378: all module crd are also using this group, we should only limited to our crd, kyma, moduletemplate, watcher, manifest

main.go
```diff
@@ -357,3 +363,39 @@ func setupKcpWatcherReconciler(mgr ctrl.Manager, options controller.Options, fla
 		os.Exit(1)
 	}
 }
+
+func dropV1alpha1FromStoredVersions(mgr manager.Manager) {
```
ruanxin, line 367: we'd better put it in a separate go routing to avoid blocking the main thread.

refactor this function to accept the version to be dropped, so that we can reuse it in the later deletion.

ruanxin (approved):

PR #596, opened 2023-05-11
change ReplaceWithVirtualKyma to only accept one remote Kyma

ruanxin:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/pkg/remote/kyma_synchronization_context.go
+++ b/pkg/remote/kyma_synchronization_context.go
@@ -233,23 +233,20 @@ func (c *KymaSynchronizationContext) SynchronizeRemoteKyma(
 	return nil
 }
 
-// ReplaceWithVirtualKyma creates a virtual kyma instance from a control plane Kyma and N Remote Kymas,
+// ReplaceWithVirtualKyma creates a virtual kyma instance from a control plane Kyma and one Remote Kymas,
 // merging the module specification in the process.
-func (c *KymaSynchronizationContext) ReplaceWithVirtualKyma(
+func ReplaceWithVirtualKyma(
 	kyma *v1beta2.Kyma,
-	remotes ...*v1beta2.Kyma,
+	remoteKyma *v1beta2.Kyma,
 ) {
 	totalModuleAmount := len(kyma.Spec.Modules)
-	for _, remote := range remotes {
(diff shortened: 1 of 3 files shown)


pkg/remote/kyma_synchronization_context.go
```diff
@@ -258,6 +255,7 @@ func (c *KymaSynchronizationContext) ReplaceWithVirtualKyma(
 	for _, m := range modules {
 		kyma.Spec.Modules = append(kyma.Spec.Modules, m)
 	}
+	kyma.Spec.Channel = remoteKyma.Spec.Channel
```
jeremyharisch, line 258: With this overwrite we are always using the global channel configured in the SKR KymaCR. Thus, the SREs are not able to change the global channel for a specific Kyma Cluster in the kcp-cluster.

Example:
We have a pre-configured module in the kcp-cluster with no specific channel mentioned in the spec.
```yaml
spec:
  channel: regular
  modules:
  - customResourcePolicy: CreateAndDelete
    name: istio
```
This module will not be reflected in the KymaCR on SKR, which is fine and a wanted behaviour &#8594; User cannot change "prerequisite" components. 
If the SREs are now changing the global channel in the kcp-cluster for a Kyma, this change will **not** be taken over. &#8594; complete ignore of kcp global channel after initialisation.


This should be clarified with @janmedrek if we want to have a behaviour like this, or we need to refine the wanted behaviour more.

jeremyharisch (commented):
Besides the one comment with need to be resolved with @janmedrek , it looks good and works.

pkg/remote/kyma_synchronization_context.go
janmedrek: Having the KCP channel set only for initialisation is a good thing. We want to let the users do all the stuff with their runtimes as long as they do not (accidentally) break anything.

I also do not imagine a scenario where we want to override the default channel for a specific runtime in KCP, even if we needed that then we can just execute some script to iterate over SKRs and change that.

So if you ask me - I would leave it as it is, it is completely fine and in line with the idea behind the modularization flow. 🙂 

_(it would be good to write down the behaviour for this though)_

jeremyharisch (approved):

PR #607, opened 2023-05-16
Ignore sync label for single cluster mode

### LINKED ISSUE #576: single cluster mode should ignore sync label
Since v1beta2, the mode of LM has been fixed after deployment, which means when LM running under single cluster mode, there is no need to support kyma and moduletemplate sync, and this can be detected using [`inKCPMode`](https://github.com/kyma-project/lifecycle-manager/blob/main/flags.go#L103-L104) flag.

AC:
- [ ] LM deployed with `inKCPMode=false`, disable sync for kyma and moduletemplate, ignore the sync label presented in the CR.

nesmabadr:
**Description**

- Ignore `sync` label if there is no `in-kcp-mode` flag

**Related issue(s)**
Resolves #576


controllers/control-plane/kcp_controller_helper_test.go
```diff
@@ -0,0 +1,160 @@
...
+)
+
+var (
+	ErrNotFound                 = errors.New("resource not exists")
+	ErrExpectedLabelNotReset    = errors.New("expected label not reset")
+	ErrWatcherLabelMissing      = errors.New("watcher label missing")
+	ErrWatcherAnnotationMissing = errors.New("watcher annotation missing")
+)
+
+func RegisterControlPlaneLifecycleForKyma(kyma *v1beta2.Kyma) {
```
Tomasz-Smelcerz-SAP, line 24: The function shouldn't be exported

controllers/control-plane/kcp_controller_helper_test.go
Tomasz-Smelcerz-SAP, line 1: All the helper functions defined here should be unexported.

nesmabadr: There is already the testutils pkg, so I added the common functions there instead of having a new package for them.

PR #611, opened 2023-05-17
Validate unique modules in kyma spec

### LINKED ISSUE #610: CEL Validation for Remote Kyma CR
**Description**
At the moment it is possible to edit the Kyma CR on SKR to have multiple same modules. In the KCP it is already prevented using a webhook. We should also introduce same behaviour on our remote KymaCRs.

A possible and easy solution would be to use the new [CEL](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation) Validation feature from Kubernetes.

**Reasons**
It should not be possible to have the same module more than once in a single remote Kyma CR.

**ACs**
- [ ] Prevent having module more than once in a single Remote Kyma CR
- [ ] Use [CEL](https://github.com/google/cel-spec) if possible

nesmabadr:
**Description**

- Add a validation to have unique modules based on their name

**Related issue(s)**
Resolves #610

### DIFF
--- a/api/v1beta2/kyma_types.go
+++ b/api/v1beta2/kyma_types.go
@@ -46,6 +46,8 @@ type KymaSpec struct {
 	Channel string `json:"channel"`
 
 	// Modules specifies the list of modules to be installed
+	// +listType=map
+	// +listMapKey=name
 	Modules []Module `json:"modules,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


medmes (dismissed):
Smart solution, Thanks! 👍

jeremyharisch (dismissed):

ruanxin (approved):

Issue #518, opened 2023-04-09
Remove CPU limit for lifecycle manager


ruanxin:
Referenced article: [CPU limits on Kubernetes are an antipattern](https://home.robusta.dev/blog/stop-using-cpu-limits)

Main takeaway from this article, it's not a best practice to setup CPU limit, because it will potentially introduce CPU throttling if application has burst during high CPU usage.

AC:
- [x] lifecycle manager deployment should not set CPU limit
- [x] consult SRE regarding this change

amritanshusikdar:
The SKR Webhook CPU Limits are preserved, aka, not removed.

ruanxin:
let's postpone this until we resolve current high CPU usage issue:  https://github.com/kyma-project/lifecycle-manager/pull/651

ruanxin:
done https://github.com/kyma-project/lifecycle-manager/pull/732

PR #620, opened 2023-05-19
remove cpu limits

amritanshusikdar:
- Remove CPU Limits from the Lifecycle Manager

**Related issue(s)**
Resolves: #518

### DIFF
--- a/docs/technical-reference/api/moduleTemplate-cr.md
+++ b/docs/technical-reference/api/moduleTemplate-cr.md
@@ -36,7 +36,6 @@ spec:
       resources:
         operator:
           limits:
-            cpu: "800m"
             memory: "200Mi"
           requests:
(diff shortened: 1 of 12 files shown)


ruanxin:
let's postpone this until we resolve current high CPU usage issue: https://github.com/kyma-project/lifecycle-manager/pull/651

docs/technical-reference/api/moduleTemplate-cr.md
```diff
@@ -36,7 +36,6 @@ spec:
       resources:
         operator:
           limits:
-            cpu: "800m"
```
ruanxin, line 39: This is related to module CR, we do not care if cpu limit should be set or not. In certain scenario, limit CPU is needed.

Issue #618, opened 2023-05-19
Invalid `status.conditions` for a remote Kyma


Tomasz-Smelcerz-SAP:
**Description**

During synchronization of a Kyma CR the Module Catalog (ModuleTemplate CRs) is synced correctly.
The Kyma object with an empty module list goes immediately to a `ready` state, and it's synced to the SKR.
The `status.conditions[]` in the SKR copy are wrong though.
Conditions for `Modules` and `ModuleCatalog` are set to "Ready == False" and the message indicates an error has occured. This is different from the values in the original Kyma (the control-plane Kyma CR), which reports all conditions as "Ready == True"

**Expected result**

Conditions for `Modules` and `ModuleCatalog` are correct

**Actual result**
```
  spec:
    channel: regular
  status:
    activeChannel: regular
    conditions:
    - lastTransitionTime: "2023-05-19T07:24:42Z"
      message: not all modules are in ready state
      observedGeneration: 1
      reason: Ready
      status: "False"
      type: Modules
    - lastTransitionTime: "2023-05-19T07:24:42Z"
      message: module catalog is out of sync and needs to be resynchronized
      observedGeneration: 1
      reason: Ready
      status: "False"
      type: ModuleCatalog
    - lastTransitionTime: "2023-05-19T07:24:42Z"
      message: skrwebhook is out of sync and needs to be resynchronized
      observedGeneration: 1
      reason: Ready
      status: "False"
      type: SKRWebhook
    lastOperation:
      lastUpdateTime: "2023-05-19T07:24:42Z"
      operation: kyma is ready
    state: Ready
```
**Steps to reproduce**

Follow the [developer tutorial](https://github.com/kyma-project/lifecycle-manager/blob/main/docs/developer-tutorials/local-test-setup.md) for local testing.
Create a Kyma with an empty module list and observe it's `status.conditions`

**Troubleshooting**
n/a

Tomasz-Smelcerz-SAP:
Note for reviewer: This PR contains also a [fix](#624) for a critical bug discovered during testing the local setup. It's included here because otherwise this PR is harder to test.

PR #621, opened 2023-05-22
Fix invalid remote kyma status sync

Tomasz-Smelcerz-SAP:
**Description**

Fix invalid sync of the remote Kyma `status.conditions`

- Fix invalid sync of the remote Kyma `status.conditions`
- minor renaming to more readable identifiers

**Related issue(s)**
#618

### DIFF
--- a/config/watcher_local_test/kustomization.yaml
+++ b/config/watcher_local_test/kustomization.yaml
@@ -28,6 +28,9 @@ patches:
       value: --in-kcp-mode=true
     - op: add
       path: /spec/template/spec/containers/0/args/-
+      value: --in-kcp-mode=true
+    - op: add
+      path: /spec/template/spec/containers/0/args/-
       value: --listener-http-local-mapping=9443
     - op: add
       path: /spec/template/spec/containers/0/args/-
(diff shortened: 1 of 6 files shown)


config/watcher_local_test/kustomization.yaml
```diff
@@ -23,6 +23,7 @@ patches:
     - op: add
       path: /spec/template/spec/containers/0/args/-
       value: --enable-watcher-local-testing
+      value: --in-kcp-mode=true
```
jeremyharisch, line 26:
```suggestion
    - op: add
      path: /spec/template/spec/containers/0/args/-
      value: --in-kcp-mode=true
```

Lets stick to the normal yaml pattern and avoid having duplicate keys, otherwise we will get the warning `Key 'value' is duplicated `.

If you would like to change it to use `value` multiple times in one `op`, then I would suggest to change to everywhere. But your IDE will flood you with warnings 😓

jeremyharisch (dismissed):

jeremyharisch (dismissed):

jeremyharisch (approved):

Issue #623, opened 2023-05-22
panic: the given value is not a pointer to a kyma synchronization context


Tomasz-Smelcerz-SAP:
**Description**

The Lifecycle Manager build from current `main` panics during Kyma reconciliation.

**Expected result**

Reconciliation works as expected

**Actual result**

```
panic: the given value is not a pointer to a kyma synchronization context

goroutine 960 [running]:
github.com/kyma-project/lifecycle-manager/pkg/remote.SyncContextFromContext({0x2e2bde0?, 0xc0014bf440?})
        /workspace/pkg/remote/context.go:30 +0x71
github.com/kyma-project/lifecycle-manager/pkg/watcher.(*SKRWebhookManifestManager).Install(0xc0012633e0, {0x2e2bde0, 0xc0014bf440}, 0xc000805100?)
        /workspace/pkg/watcher/skr_webhook_manifest_manager.go:73 +0xbc
github.com/kyma-project/lifecycle-manager/controllers.(*KymaReconciler).handleProcessingState.func3()
        /workspace/controllers/kyma_controller.go:259 +0x3b
golang.org/x/sync/errgroup.(*Group).Go.func1()
        /go/pkg/mod/golang.org/x/sync@v0.1.0/errgroup/errgroup.go:75 +0x64
created by golang.org/x/sync/errgroup.(*Group).Go
        /go/pkg/mod/golang.org/x/sync@v0.1.0/errgroup/errgroup.go:72 +0xa5
```
**Steps to reproduce**

follow the tutorial from: `docs/developer-tutorials/local-test-setup.md`
**Troubleshooting**
n/a

PR #624, opened 2023-05-22
fix missing synchronization context for local testing

Tomasz-Smelcerz-SAP:
**Description**

Fixes panic in local testing introduced in #607 

- Watcher is enabled only if Kyma Sync is enabled
- add missing "in-kcp-mode" flag to local testing Makefile target

**Related issue(s)**
#623

### DIFF
--- a/config/watcher_local_test/kustomization.yaml
+++ b/config/watcher_local_test/kustomization.yaml
@@ -23,6 +23,7 @@ patches:
     - op: add
       path: /spec/template/spec/containers/0/args/-
       value: --enable-watcher-local-testing
+      value: --in-kcp-mode=true
     - op: add
       path: /spec/template/spec/containers/0/args/-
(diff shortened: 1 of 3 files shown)


config/watcher_local_test/kustomization.yaml
```diff
@@ -23,6 +23,7 @@ patches:
     - op: add
       path: /spec/template/spec/containers/0/args/-
       value: --enable-watcher-local-testing
+      value: --in-kcp-mode=true
```
jeremyharisch, line 26:
```suggestion
    - op: add
      path: /spec/template/spec/containers/0/args/-
      value: --in-kcp-mode=true
```

Lets stick to the normal yaml pattern and avoid having duplicate keys, otherwise we will get the warning `Key 'value' is duplicated `.

PR #625, opened 2023-05-23
Normalize remote Kyma CR name

### LINKED ISSUE #599: Normalize the synced Kyma CR name
**Description**

Currently, the name of the Kyma custom resource synced to the runtime cluster is based on its uid, which is not user-friendly and may cause confusion. Therefore, we should consider normalizing the name to make it more transparent and comfortable to use.

**Reasons**

Normalizing the name of the Kyma custom resource would improve the user experience and increase transparency.

**Acceptance Criteria**
- [ ] Kyma name in runtime clusters is normalized (i.e `default`)

nesmabadr:
**Description**

- Normalize the name of the remote Kyma CR to `default`

**Related issue(s)**
Resolves #599


lindnerby (approved):

pkg/remote/kyma_synchronization_context.go
```diff
@@ -167,7 +167,7 @@ func (c *KymaSynchronizationContext) CreateOrFetchRemoteKyma(
 	recorder := adapter.RecorderFromContext(ctx)
 	remoteKyma := &v1beta2.Kyma{}
 
-	remoteKyma.Name = kyma.Name
+	remoteKyma.Name = v1beta2.DefaultRemoteKymaName
```
ruanxin, line 170: if fetch with defaultname -> found -> 
                                          -> not found ->  fallback to fetch with kyma.Name
                                                                                       found ->  update with defaultname
                                                                                       notfound ->  created with defaultname

pkg/remote/kyma_synchronization_context.go
```diff
@@ -186,24 +187,47 @@ func (c *KymaSynchronizationContext) CreateOrFetchRemoteKyma(
...
+	return remoteKyma, nil
+}
+
+func (c *KymaSynchronizationContext) copyOldKymaCRIfExisting(
+	ctx context.Context, kyma *v1beta2.Kyma, remoteSyncNamespace string,
+	remoteKyma *v1beta2.Kyma, recorder record.EventRecorder) error {
+	oldRemoteKyma := &v1beta2.Kyma{}
+	oldRemoteKyma.Name = kyma.Name
+	oldRemoteKyma.Namespace = remoteSyncNamespace
+	err := c.RuntimeClient.Get(ctx, client.ObjectKeyFromObject(oldRemoteKyma), oldRemoteKyma)
```
ruanxin, line 219: here we need to consider the error differently, if the error is not found, we can return, but if it's other error, it might be temporary can't reach the oldRemoteKyma, but it still exists, which means we should not create new kyma, because after the new kyma get created, there is no chance to delete this old kyma anymore.

lindnerby (approved):

PR #642, opened 2023-05-26
Remove Kyma and Manifest finalizers if no access secret is found and they are being deleted

### LINKED ISSUE #640: If kyma is under deleting and no secret found, remove finalizer
The current decision when SKR is under deprovisioning, there is 1h waiting period,  if Kyma is still not been deleted successfully, it will have a process to dropping whole SKR cluster, and kyma access secret. However, this brings the problem that Kyma CR still exists, but secret are gone.

To resolve this problem, LM detect if current Kyma is under deleting (`!kyma.DeletionTimestamp.IsZero()`), but can't find access secret, LM should remove Kyma CR finalizer so that it can be removed by apiserver.

nesmabadr:
**Description**

- Remove Kyma and manifests finalizers if no secret is found and they are being deleted

**Related issue(s)**
Resolves #640


controllers/kyma_controller.go
```diff
@@ -122,13 +122,19 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
-//nolint:cyclop
+//nolint:funlen,cyclop,gocognit
 func (r *KymaReconciler) reconcile(ctx context.Context, kyma *v1beta2.Kyma) (ctrl.Result, error) {
-	if r.SyncKymaEnabled(kyma) {
+	if r.SyncKymaEnabled(kyma) { //nolint:nestif
 		var err error
 		remoteClient := remote.NewClientWithConfig(r.Client, r.KcpRestConfig)
 		if ctx, err = remote.InitializeSyncContext(ctx, kyma,
 			r.RemoteSyncNamespace, remoteClient, r.RemoteClientCache); err != nil {
+			if !kyma.DeletionTimestamp.IsZero() && apierrors.IsNotFound(err) {
```
medmes, line 132: I am thinking about reducing the nested if-statement. Did you think somehow about switch case for this scenario?

nesmabadr: @medmes  How would I use a switch for this case?

medmes:
As we discussed that we need either to not use the **cyclop** golang linting or refactor the code in order to solve the code complexity.
I don't want to approve it until we pick one solution described above.

nesmabadr:
@medmes  the extra cyclop linting was not introduced in this PR, that's why I believe we should address this in a separate PR since it's not within this PR's scope

medmes (approved):
The logic is all fine, but a following Github issue should be created regarding the re visit all suppressed go linting especially the code complexity.

PR #655, opened 2023-06-02
Introduce cache for KCP CRDs

### LINKED ISSUE #598: Introduce cache to fetch KCP CRD
In the current CRD generation detect, each reconciliation loop, LM need to fetch the KCP and SKR CRD to do the comparison. But KCP CRD basically can be only updated during LM deployment, which means each request to KCP will return the same results. To save the remote call and for performance purposes, we can cache this result.

AC:
- [ ] The KCP CRD of Kyma and ModuleTemplate generation should be cached.

nesmabadr:
**Description**

- Added a cache to store the KCP ModuleTemplate and Kyma CRDs

**Related issue(s)**
Resolves #598


internal/crd_cache.go
```diff
@@ -0,0 +1,29 @@
...
+
+type CustomResourceDefinitionCache struct {
+	storage *sync.Map
+}
+
+func NewCustomResourceDefinitionCache() *CustomResourceDefinitionCache {
+	return &CustomResourceDefinitionCache{storage: &sync.Map{}}
+}
+
+func (cache *CustomResourceDefinitionCache) Get(key client.ObjectKey) *v1.CustomResourceDefinition {
```
ruanxin, line 18: `client.ObjectKey` is not suitable here, you don't need Name and Namespace to distinguish with the CRD (CRD is also not namespaced). The Key here can be CRD name only, a pure string is enough.

pkg/remote/crd_upgrade.go
```diff
@@ -116,23 +117,30 @@ func fetchCrdsAndUpdateKymaAnnotations(ctx context.Context, controlPlaneClient C
...
-			Name: fmt.Sprintf("%s.%s", plural, v1beta2.GroupVersion.Group),
-		}, crd,
-	)
-	if err != nil {
-		return nil, nil, err
+
+	kcpNamespacedNameCrd := client.ObjectKey{
+		// this object name is derived from the plural and is the default kustomize value for crd namings, if the CRD
+		// name changes, this also has to be adjusted here. We can think of making this configurable later
+		Name: fmt.Sprintf("%s.%s", plural, v1beta2.GroupVersion.Group)}
```
ruanxin, line 129: You should not use self made name for `client.ObjectKey`, ObjectKey should be only used for go client to fetch k8s resources with its name and namespace. That's another reason why you should not use it as cache key.

pkg/remote/kyma_synchronization_context.go
```diff
@@ -129,20 +130,26 @@ func (c *KymaSynchronizationContext) ensureRemoteNamespaceExists(ctx context.Con
 	return nil
 }
 
-func (c *KymaSynchronizationContext) CreateOrUpdateCRD(ctx context.Context, plural string) error {
-	crd := &v1extensions.CustomResourceDefinition{}
+func (c *KymaSynchronizationContext) CreateOrUpdateCRD(ctx context.Context, plural string,
```
ruanxin, line 133: Don't introduce cache to this function, CreateOrUpdateCRD only get called when there is no CRD found, we don't need to cache it.

pkg/remote/remote_catalog.go
```diff
@@ -130,7 +133,8 @@ func (c *RemoteCatalog) createOrUpdateCatalog(ctx context.Context,
 
 	// it can happen that the ModuleTemplate CRD is not existing in the Remote Cluster when we apply it and retry
 	if containsMetaIsNoMatchErr(errs) {
-		if err := c.CreateModuleTemplateCRDInRuntime(ctx, v1beta2.ModuleTemplateKind.Plural()); err != nil {
+		if err := c.CreateModuleTemplateCRDInRuntime(ctx, v1beta2.ModuleTemplateKind.Plural(),
```
ruanxin, line 136: Same here, it's only called when CRD is not found, we don't need cache

ruanxin (changes requested):
please see my comments

pkg/remote/kyma_synchronization_context.go
nesmabadr: This function creates or updates the CRD for the runtime, but the cache is only for the KCP

pkg/remote/remote_catalog.go
nesmabadr: Same here, the cache is only for the KCP so it doesn't guarantee that it's there for the runtime

internal/crd_cache.go
```diff
@@ -0,0 +1,28 @@
...
+
+type CustomResourceDefinitionCache struct {
+	storage *sync.Map
+}
+
+func NewCustomResourceDefinitionCache() *CustomResourceDefinitionCache {
+	return &CustomResourceDefinitionCache{storage: &sync.Map{}}
+}
+
+func (cache *CustomResourceDefinitionCache) Get(key string) *v1.CustomResourceDefinition {
```
ruanxin, line 17: It's better we return back a copied value, but not a memory reference, so that CRD saved in the cache can't be changed.

controllers/control-plane/kcp_controller_helper_test.go
```diff
@@ -210,8 +211,12 @@ func updateKymaCRD(clnt client.Client) (*v1extensions.CustomResourceDefinition,
 		client.FieldOwner(v1beta2.OperatorName)); err != nil {
 		return nil, err
 	}
-
 	crd, err = fetchCrd(clnt, v1beta2.KymaKind)
+	kymaCrdName := fmt.Sprintf("%s.%s", v1beta2.KymaKind.Plural(), v1beta2.GroupVersion.Group)
```
ruanxin, line 215: why we need this cache in test?

nesmabadr: because here I am updating the KCP crd for the test so I need to update it manually

ruanxin (approved):

Issue #614, opened 2023-05-17
refactor oci registry authentication using Workload Identity


ruanxin:
The final decision for the signed module template is hosted in Google Artifact Registry and in private mode, which requires an authentication process to access. Currently, we are using credential secrets as a solution, but this approach is not recommended due to credential rotation challenges.

Fortunately, we have deployed Lifecycle Manager under KCP based on GKE, and Google provides Workload Identity, which allows applications in GKE clusters to impersonate IAM service accounts for accessing Google Cloud services. Additionally, `go-containerregistry` supports handling authentication internally.

AC:
- [ ] Ensure that the existing solution using credential secrets is still supported to accommodate other private registries if needed.
- [ ] Make sure LM can access Google Artifact Registry private registries without offering credential secrets, but using Workload Identity

PR #666, opened 2023-06-06
support Workload Identity for GKE cluster

ruanxin:
related issue: https://github.com/kyma-project/lifecycle-manager/issues/614


pkg/module/parse/template_to_module.go
```diff
@@ -140,22 +135,6 @@ func (p *Parser) newManifestFromTemplate(
 	if err != nil {
 		return nil, err
 	}
-	var componentDescriptor *compdesc.ComponentDescriptor
```
Tomasz-Smelcerz-SAP, line 143: What is the reasoning behind taking the component descriptor from the module template instead of a "true" remote one? Is it just a temporary solution?

Tomasz-Smelcerz-SAP (approved):

Issue #663, opened 2023-06-06
Unit test manifest yaml parsing logic


ruanxin:
We recently got an internal incident, the root cause is that rendered resources being cached are inconsistent with the resources in raw yaml file. It might be due to parsing logic is not robust enough.

AC:
- [ ] refactoring current `ParseManifestStringToObjects` if necessary
- [ ] write unit test for it, the test should use real case raw manifest yaml, some candidate: btp-operator, keda manager

PR #670, opened 2023-06-07
Refactor parsing logic

ruanxin:
- Replace spaghetti parsing function with resource builder to provide solid parsing logic.

- Cleanup unused code
 
https://github.com/kyma-project/lifecycle-manager/issues/663


internal/declarative/v2/inmemory_rendered.go
```diff
@@ -31,46 +28,24 @@ type InMemoryManifestCache struct {
...
+func (c *InMemoryManifestCache) Parse(spec *Spec,
+) (internal.ManifestResources, error) {
 	file := filepath.Join(ManifestFilePrefix, spec.Path, spec.ManifestName)
 	key := fmt.Sprintf("%s-%s", file, spec.Mode)
 
 	item := c.Cache.Get(key)
 	if item != nil {
 		resources := item.Value()
 
-		copied := &internal.ManifestResources{
```
Tomasz-Smelcerz-SAP, line 44: I think this defensive copy was here to ensure that no code can change the cached object (due to a bug or misuse).
Just an example of what is possible: `resources.Items[0].SetDeletionTimestamp(someValue)`.
Now, because you're returning a cache item directly, this protection is disabled. Maybe I am missing something?

internal/declarative/v2/ssa.go
```diff
@@ -99,7 +99,7 @@ func (c *ConcurrentDefaultSSA) serverSideApplyResourceInfo(
 			"%s is not a valid client-go object: %w", info.ObjectName(), ErrClientObjectConversionFailed,
 		)
 	}
-
+	obj.SetManagedFields(nil)
```
Tomasz-Smelcerz-SAP, line 102: What does this do?

internal/declarative/v2/inmemory_rendered.go
```diff
@@ -31,46 +28,24 @@ type InMemoryManifestCache struct {
...
-			Blobs: resources.Blobs,
-		}
-		for _, res := range resources.Items {
-			copied.Items = append(copied.Items, res.DeepCopy())
-		}
-
-		return copied, nil
-	}
-
-	rendered, err := renderer.Render(ctx, obj)
```
Tomasz-Smelcerz-SAP, line 55: After removing this line, the `Render` method from the `Renderer` interface is unused. Is it OK?

internal/declarative/v2/inmemory_rendered.go
ruanxin: you are right 👍

Tomasz-Smelcerz-SAP (approved):

PR #673, opened 2023-06-12
Remove CRD after other resources

### LINKED ISSUE #668: For module deleting, remove CRD only after all other resources get deleted.
the current deleting logic will delete all resources all at once, in most cases, it's fine, however, if module CR is in delete pending state, then the CRD should not been deleted. Otherwise, CRD will under terminating and prevent LM to recreate the CR. This can happen when use disable module and enable it again.

AC:
- [ ] refactor module deleting logic, make sure CRD deleting happens at the end, and only after all other resources get deleted.

nesmabadr:
**Description**

- Prevent deletion of the CRD before CR

**Related issue(s)**
Fixes #668 
#663

### DIFF
--- a/controllers/manifest_controller.go
+++ b/controllers/manifest_controller.go
@@ -92,5 +92,6 @@ func ManifestReconciler(
 		declarative.WithPostRun{internalv1beta1.PostRunCreateCR},
 		declarative.WithPreDelete{internalv1beta1.PreDeleteDeleteCR},
 		declarative.WithPeriodicConsistencyCheck(checkInterval),
+		declarative.WithCRDName(internalv1beta1.GetCRDName),
 	)
 }
--- a/controllers/manifest_controller.go
+++ b/controllers/manifest_controller.go
@@ -92,6 +92,6 @@ func ManifestReconciler(
 		declarative.WithPostRun{internalv1beta1.PostRunCreateCR},
 		declarative.WithPreDelete{internalv1beta1.PreDeleteDeleteCR},
 		declarative.WithPeriodicConsistencyCheck(checkInterval),
-		declarative.WithCRDName(internalv1beta1.GetCRDName),
+		declarative.WithModuleCRDName(internalv1beta1.GetModuleCRDName),
 	)
 }
--- a/internal/declarative/v2/options.go
+++ b/internal/declarative/v2/options.go
@@ -62,6 +62,8 @@ type Options struct {
 	Namespace       string
 	CreateNamespace bool
 
+	CRDName func(obj Object) string
+
 	Finalizer string
 
 	ServerSideApply bool
@@ -117,6 +119,12 @@ type WithFinalizer string
 
 func (o WithFinalizer) Apply(options *Options) {
 	options.Finalizer = string(o)
+}
+
+type WithCRDName func(obj Object) string
+
+func (o WithCRDName) Apply(options *Options) {
+	options.CRDName = o
 }
 
 type WithManagerOption struct {
@@ -239,8 +247,6 @@ func (o WithDeleteCRDs) Apply(options *Options) {
 }
 
 type ManifestCache string
-
-const NoManifestCache ManifestCache = "no-cache"
 
 type WithManifestCache ManifestCache
 
--- a/internal/declarative/v2/options.go
+++ b/internal/declarative/v2/options.go
@@ -62,7 +62,7 @@ type Options struct {
 	Namespace       string
 	CreateNamespace bool
 
-	CRDName func(obj Object) string
+	ModuleCRDName func(obj Object) string
 
 	Finalizer string
 
@@ -121,10 +121,10 @@ func (o WithFinalizer) Apply(options *Options) {
 	options.Finalizer = string(o)
 }
 
-type WithCRDName func(obj Object) string
+type WithModuleCRDName func(obj Object) string
 
-func (o WithCRDName) Apply(options *Options) {
-	options.CRDName = o
+func (o WithModuleCRDName) Apply(options *Options) {
+	options.ModuleCRDName = o
 }
 
 type WithManagerOption struct {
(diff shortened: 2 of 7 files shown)


controllers/manifest_controller.go
```diff
@@ -92,5 +92,6 @@ func ManifestReconciler(
 		declarative.WithPostRun{internalv1beta1.PostRunCreateCR},
 		declarative.WithPreDelete{internalv1beta1.PreDeleteDeleteCR},
 		declarative.WithPeriodicConsistencyCheck(checkInterval),
+		declarative.WithCRDName(internalv1beta1.GetCRDName),
```
ruanxin, line 95: This `WithCRDName` is not clear enough, which CRD? Maybe rename to something like `WithModuleCRDName`

internal/declarative/v2/options.go
```diff
@@ -62,6 +62,8 @@ type Options struct {
 	Namespace       string
 	CreateNamespace bool
 
+	CRDName func(obj Object) string
```
ruanxin, line 65: also please rename this, it's too generic.

internal/declarative/v2/reconciler_test.go
```diff
@@ -27,14 +29,32 @@ func TestPruneKymaSystem(t *testing.T) {
 			service,
 			kymaNs,
 			deployment,
+			crd,
```
ruanxin, line 32: for this test, you don't need to add your crd

internal/declarative/v2/reconciler_test.go
```diff
@@ -27,14 +29,32 @@ func TestPruneKymaSystem(t *testing.T) {
...
+
+		infos := []*resource.Info{
+			kubeNs,
+			service,
+			kymaNs,
+			deployment,
+			crd,
+		}
+
+		result := pruneResource(infos, "CustomResourceDefinition", "some-crd")
```
ruanxin, line 52: please replace those two `"some-crd"` with a variable so that others know why you can prune this crd

internal/manifest/v1beta1/custom_resource.go
```diff
@@ -96,3 +120,14 @@ func PreDeleteDeleteCR(
 	}
 	return nil
 }
+
+func GetCRDName(obj declarative.Object) string {
+	manifest := obj.(*v1beta2.Manifest)
+	if manifest.Spec.Resource != nil {
+		group := manifest.Spec.Resource.GroupVersionKind().Group
+		name := manifest.Spec.Resource.GroupVersionKind().Kind
+		return fmt.Sprintf("%ss.%s", strings.ToLower(name), group)
```
ruanxin, line 129: consider making that plural `s` more visible for readability purposes, e.g: `fmt.Sprintf("%s.%s", getPlural(name), group)`

internal/declarative/v2/reconciler.go
```diff
@@ -366,7 +366,12 @@ func (r *Reconciler) renderTargetResources(
 func (r *Reconciler) pruneDiff(
 	ctx context.Context, clnt Client, obj Object, renderer Renderer, diff []*resource.Info,
 ) error {
-	if err := r.deleteResources(ctx, clnt, obj, pruneKymaSystem(diff)); err != nil {
+	diff = pruneResource(diff, "Namespace", namespaceNotBeRemoved)
+	resourceName := r.ModuleCRDName(obj)
+	if resourceName != "" {
```
ruanxin, line 371: I think you don't need this additional check if resourceName is empty, `pruneResource` anyway will skip this case.

internal/manifest/v1beta1/custom_resource.go
```diff
@@ -79,6 +86,23 @@ func PreDeleteDeleteCR(
 		return err
 	}
 
+	var crd unstructured.Unstructured
```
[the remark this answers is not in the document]
ruanxin: the test can be an enhancement in this PR

internal/declarative/v2/reconciler.go
```diff
@@ -377,17 +381,18 @@ func (r *Reconciler) pruneDiff(
...
 		obj := info.Object.(client.Object)
-		if obj.GetObjectKind().GroupVersionKind().Kind != "Namespace" {
+		if obj.GetObjectKind().GroupVersionKind().Kind != resourceType {
 			continue
 		}
-		if obj.GetName() != namespaceNotBeRemoved {
+		if obj.GetName() != resourceName {
 			continue
 		}
 		return append(diff[:i], diff[i+1:]...)
```
Tomasz-Smelcerz-SAP, line 393: This line is executed only if: `obj.GetObjectKind().GroupVersionKind().Kind == resourceType && obj.GetName() == resourceName`
Why not replace the lines above that use negated conditions and loop skipping with such an explicit check?

internal/manifest/v1beta1/custom_resource.go
```diff
@@ -79,6 +86,23 @@ func PreDeleteDeleteCR(
...
+	crd.SetName(GetModuleCRDName(obj))
+	crd.SetGroupVersionKind(schema.GroupVersionKind{
+		Version: "v1",
+		Group:   "apiextensions.k8s.io",
+		Kind:    "CustomResourceDefinition",
+	})
+	crdCopy := crd.DeepCopy()
```
Tomasz-Smelcerz-SAP, line 96: I don't think you need to `DeepCopy()` - it's not the object read using the client, it was just created in line 89?

nesmabadr: the reason I did this is that skr here is declarative.Client not Client.client so the skr.Delete needs an *Unstructured

Tomasz-Smelcerz-SAP: :+1

Tomasz-Smelcerz-SAP (approved):

PR #674, opened 2023-06-12
Prevent diff resources to be deleted if oci ref not changed

ruanxin:
This PR is a defensive protection related to the current resources cache inconsistency issue. If the resources read from the cache are still incomplete, this implementation prevents synced resources to be deleted, and will update the manifest CR with Warning state, also it evicts the cache so that the resource cache can be recreated.


Tomasz-Smelcerz-SAP:
I've manually reproduced the "invalid cache" condition, but it ends up in an invalid state.
I've used a two-cluster development setup described in our docs.
steps:
- when provisioning `kcp` k3d cluster, mount a local directory to the cluster so that it becomes the TMPDIR directory for the Lifecycle-Manager process
- reduce the TTL cache time to a couple of seconds (I used 24 seconds instead of 24 hours)
- setup both clusters as usual
- install a sample kyma with a sample moduletemplate. I used `keda-regular`
- wait until all is `ready`
- navigate to your local directory mounted into k3d Pod. You should see there files created by the Lifecycle Manager.
- find a cached file for `keda` manifests. Delete something from it, save.
- change something in the `keda` moduletemplate to enforce reconciliation. I changed some annotation.

Tomasz-Smelcerz-SAP:
I've created an issue to improve the file handling logic: https://github.com/kyma-project/lifecycle-manager/issues/684

PR #675, opened 2023-06-13
Fix parallel file cache access

Tomasz-Smelcerz-SAP:
**Description**

This PR introduces an in-memory locking mechanism to avoid parallel writes to the same file when handling raw manifests.
This also ensures there are no `reads` of the file when the `write` is still pending.
All of this is to ensure the read/write consistency of the data.

For any specific file:

1) only a single goroutine may check if the file exists, all other writers/readers must wait
2) only a single goroutine may create the file and write data to it, all other writers/readers must wait

Operation 1) is very fast so though it's performed many times, the additional locking is not expected to impact performance (to be verified).
Operation 2) is slow, but it's executed only once per module version in the Lifecycle-Manager lifetime. It creates a local file with raw manifest data for a specific module template hash.

The file operations on different files (paths) are not mutually locked because these can be safely executed in parallel.

**Related issue(s)**
https://github.com/kyma-project/lifecycle-manager/issues/663

### DIFF
--- a/internal/manifest/v1beta1/parse.go
+++ b/internal/manifest/v1beta1/parse.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path"
 	"regexp"
+	"sync"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
@@ -20,6 +21,8 @@ import (
 
 const manifestFileName = "raw-manifest.yaml"
 
+var fileMutexMap = sync.Map{}
+
 func GetPathFromRawManifest(ctx context.Context,
 	imageSpec v1beta2.ImageSpec,
 	keyChain authn.Keychain,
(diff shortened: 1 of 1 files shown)


internal/manifest/v1beta1/parse.go
```diff
@@ -76,3 +87,14 @@ func pullLayer(ctx context.Context, imageRef string, keyChain authn.Keychain) (v
 	}
 	return crane.PullLayer(noSchemeImageRef, crane.WithAuthFromKeychain(keyChain), crane.WithContext(ctx))
 }
+
+// getLockerForPath always returns the same sync.Locker instance for given path argument.
+func getLockerForPath(path string) sync.Locker {
+	val, ok := fileMutexMap.Load(path)
```
ruanxin, line 93: Why not only use `LoadOrStore`

Tomasz-Smelcerz-SAP: To avoid creating `sync.Mutex{}` objects with every single call, which would not be used - because after some time all the necessary entries in the Map will already have some Mutex assigned.
In this implementation (assume 10 workers), in the worst case, 10 goroutines will invoke this method, 10 `sync.Mutex{}` objects are created, and only one "wins" and becomes a Map entry. The other 9 are garbage collected. But it's just 9 objects instead of one object per every call to the method (essentially an unbounded number).
I know it's not much for GC to "eat" these objects, but if we can easily spare few CPU cycles, why not?

ruanxin: tbh, considering readability and clean code purpose, this tiny performance increase is not relevant so much, there are lots of temporary objects created for each reconciling which are left for GC to be cleaned up, you can run with pprof to verify, this additional amount of empty `sync.Mutex{}` are nothing compare to that. I suggest simplifying this.

ruanxin: I just realized another problem, the lifetime of this `fileMutexMap` is not permanent, it only lives in each reconciliation loop, which means for each concurrent worker, everytime when it do `fileMutexMap.Load(path)`, they get no result, then all of them will enter into `LoadOrStore`, so basically, the performance gain you imagined also not true.

Tomasz-Smelcerz-SAP: Can you help me to fix the problem you described? How should I change the implementation then so that `fileMutexMap` is permanent?

nesmabadr: In my opinion, using Load would have a slightly better performance as Tomasz mentioned. I believe the if condition should be met only if the path is not in the map or when a worker is just spawned not with each reconciliation loop right?

nesmabadr (approved):

PR #686, opened 2023-06-19
Improve integration tests

ruanxin:
This PR try to rename and reorgenize current integration tests as a base for futher improvement.

Reference:
https://wiki.one.int.sap/wiki/display/kyma/Lifecycle+Manager+Integration+Test+Coverage


controllers/control-plane/kyma_remote_sync_test.go
```diff
@@ -24,102 +24,8 @@ var (
 	ErrAnnotationNotUpdated          = errors.New("kyma CR annotation not updated")
 )
 
-var _ = Describe("Kyma with multiple module CRs in remote sync mode", Ordered, func() {
```
ruanxin, line 27: This test are merged into `Kyma sync into Remote Cluster`

nesmabadr (approved):

PR #705, opened 2023-07-05
Enqueue owner instead of watched Resource

jeremyharisch:
**Description**
Enqueue `owner` instead of watched. Since we changed the default name of the KymaCR in SKR to not match the name and namespace of the corresponding KymaCR in KCP. Thus, we need to enqueue the owner of the watcher CR.

(Owner = corresponding KymaCR in KCP)

**Related issue(s)**


controllers/setup.go
```diff
@@ -95,29 +95,34 @@ func (r *KymaReconciler) SetupWithManager(mgr ctrl.Manager,
...
 			logger := ctrl.Log.WithName("listener")
 			unstructWatcherEvt, conversionOk := event.Object.(*unstructured.Unstructured)
 			if !conversionOk {
 				logger.Error(errConvertingWatcherEvent, fmt.Sprintf("event: %v", event.Object))
 				return
 			}
-			watched, ok := unstructWatcherEvt.Object["watched"]
+
+			// get owner object from unstructured event, owner = KymaCR object reference in KCP
+			unstructuredOwner, ok := unstructWatcherEvt.Object["owner"]
```
ruanxin, line 107: what's `watched` now? why previously choose watched instead of owner.

ruanxin: is this `watched` information still useful?

jeremyharisch: `watched` is not used anymore as described to enqueue a KymaCR.

The "old" behaviour: KymaCR was created in kcp-system namespace with i.e. "kyma1234" as name, thus the KymaCRs namespace and name on SKR was kcp-system/kyma1234
How it is now: KymaCR in SKR will always be deployed in kyma-system namespace and will have the name "default".

The Event which is send from SKR to KCP now includes the watched object (namespace and name), its GVK, and the owner of it (corresponding KymaCR in KCP).

For the Kyma reconcile we just need the owner, since this is the object we want to reconcile.

ruanxin: ok, then it means we should clean up the runtime watcher for this `watched` field, right? can you create a followup issue for it?

jeremyharisch: it will be used for manifest watcher and for the other teams which want to leverage the watcher.

ruanxin: but if still used for manifest watcher, where is the place to watch and enqueue it? this is the only place to handle this  "watched" right?

ruanxin (approved):

PR #713, opened 2023-07-08
Refactoring ready check state handling and integeration test package

ruanxin:
This PR address several issues:

- refactered ready check state handling, it creates a transfer data object `StateInfo` which carries non error state with additional info, so that make manifest error state only based on returned error, which distingushed with other state handling.

- it refactered manifest cr test which use remote host that brings unnessary dependencies, replaced it with mocked oci host.

- renamed integration test package name by it's controller name to make it more obverse which controller is involved in the tests


jeremyharisch (approved):

.github/workflows/e2e-test.yml
```diff
@@ -117,18 +116,20 @@ jobs:
...
-          else
-            echo "Deploy encountered some error, will retry after 1 minute"
-            sleep 60
+          maxRetry=5
+          for i in $(seq 1 $maxRetry)
+          do
             if make local-deploy-with-watcher IMG=europe-docker.pkg.dev/kyma-project/$KLM_IMAGE_REPO/lifecycle-manager:$KLM_VERSION_TAG; then
-                echo "KLM deployed successfully"
+              echo "KLM deployed successfully"
+              exit 0
```
jeremyharisch, line 124: `exit 0` will end the pipeline here, or am I wrong? Since this is not the last step in the pipeline, we should not use `exit` here.

Please use `break` instead. See [here](https://www.cyberciti.biz/faq/bash-for-loop/#Conditional_exit_with_break)

ruanxin: no, it ends this step only

jeremyharisch: Can you show me a reference to your described behaviour? Since in official docs, if ends the actions. https://docs.github.com/en/actions/creating-actions/setting-exit-codes-for-actions

And anyway, `exit 1` ends the action fully, so I would consider anyway using break, so it is more clear that this does not ends the actions, and continues on the next step.

controllers/control-plane/kyma_remote_sync_test.go
```diff
@@ -272,6 +183,50 @@ var _ = Describe("Kyma sync into Remote Cluster", Ordered, func() {
 			Should(Succeed())
 	})
 
+	It("Enable Custom ModuleTemplate in SKR", func() {
```
ruanxin, line 186: To reviewer:

This test combines "Kyma with remote module templates" test and is refactored according to real scenario, in the original test, it enabled `moduleInKcp` in KCP kyma, which did not meet the actual case, although the test works.

```
kyma.Spec.Modules = []v1beta2.Module{moduleInSkr, moduleInKcp}
```

controllers/kyma_controller/manifest_test.go
```diff
@@ -316,7 +316,7 @@ func validateManifestSpecInstallSourceRepo(manifestImageSpec *v1beta2.ImageSpec,
 //nolint:goerr113
 func validateManifestSpecInstallSourceType(manifestImageSpec *v1beta2.ImageSpec) error {
 	actualSourceType := string(manifestImageSpec.Type)
-	expectedSourceType := "oci-ref" //no corresponding value in the ModuleTemplate?
+	expectedSourceType := "oci-ref" // no corresponding value in the ModuleTemplate?
```
Tomasz-Smelcerz-SAP, line 319:
```suggestion
	expectedSourceType := "string(v1beta2.OciRefType)" // no corresponding value in the ModuleTemplate?
```

internal/manifest/v1beta1/manifest_controller_test.go
```diff
@@ -1,151 +0,0 @@
...
-			func() {
-				PushToRemoteOCIRegistry(installName)
-			},
-		)
-		BeforeEach(
-			func() {
-				Expect(os.RemoveAll(filepath.Join(os.TempDir(), mainOciTempDir))).To(Succeed())
-			},
-		)
-		It(
```
Tomasz-Smelcerz-SAP, line 90: This test case is not present in the PR. Is it redundant/obsolete?

internal/declarative/v2/ready_check.go
```diff
@@ -24,16 +25,20 @@ func NewExistsReadyCheck() ReadyCheck {
 
 type ExistsReadyCheck struct{}
 
-func (c *ExistsReadyCheck) Run(ctx context.Context, clnt Client, _ Object, resources []*resource.Info) (State, error) {
+func (c *ExistsReadyCheck) Run(
```
Tomasz-Smelcerz-SAP, line 28: The `ExistsReadyCheck` type is actually only used in tests. Maybe it's worth to move it to some *_test package?

internal/manifest/ready_check.go
```diff
@@ -0,0 +1,102 @@
...
+	deploy := &appsv1.Deployment{}
+	found := false
+	for _, res := range resources {
+		err := clt.Scheme().Convert(res.Object, deploy, nil)
+		if err == nil {
+			found = true
+			break
+		}
+	}
+	// not every module operator use Deployment by default, e.g: StatefulSet also a valid approach
```
Tomasz-Smelcerz-SAP, line 90: The comment is valid. But previously the function returned an error, now it returns "true". I think it changes the contract. Is it tested?

internal/manifest/ready_check.go
```diff
@@ -0,0 +1,102 @@
...
+		return declarative.StateInfo{State: declarative.StateReady}, nil
+	}
+	res := manifest.Spec.Resource.DeepCopy()
+	if err := clnt.Get(ctx, client.ObjectKeyFromObject(res), res); err != nil {
+		return declarative.StateInfo{State: declarative.StateError}, err
+	}
+	stateFromCR, stateExists, err := unstructured.NestedString(res.Object,
+		strings.Split(customResourceStatePath, ".")...)
+	// CR state might not been initialized, put manifest state into processing
+	if !stateExists {
```
Tomasz-Smelcerz-SAP, line 54: is it ok to verify that before error check?

internal/manifest/v1beta1/manifest_controller_test.go
ruanxin: glad you mention it! this test to me is meaningless, because:
1. it just tests installing a manifest layer with the same content twice, but not verifying anything. `withValidInstallImageSpec` only creates manifest CR, not checking for its state.
2. I believe the test trying to verify the situation that same module deployed into different kyma, so it can have multiple manifest CR with same OCI image (belongs to that module), but this situation can only happens LM in control-plane mode, it's not a setup can be easily configured in this integration test. And in single cluster mode, there is no such situation that multiple manifest CR have same OCI image, because each manifest CR belongs to a dedicated module.

internal/declarative/v2/reconciler.go
```diff
@@ -316,40 +316,52 @@ func hasDiff(oldResources []Resource, newResources []Resource) bool {
...
-	state, err := resourceReadyCheck.Run(ctx, clnt, obj, target)
+	crStateInfo, err := resourceReadyCheck.Run(ctx, clnt, manifest, target)
 
-	if errors.Is(err, ErrResourcesNotReady) || errors.Is(err, ErrCustomResourceStateNotFound) ||
-		errors.Is(err, ErrDeploymentNotReady) {
-		waitingMsg := fmt.Sprintf("waiting for resources to become ready: %s", err.Error())
-		r.Event(obj, "Normal", "ResourceReadyCheck", waitingMsg)
-		obj.SetStatus(status.WithState(StateProcessing).WithOperation(waitingMsg))
-		return err
+	if crStateInfo.State == StateProcessing {
```
Tomasz-Smelcerz-SAP, line 327: is it ok to check the state before error check?

internal/manifest/ready_check.go
ruanxin: Ok, this one needs to clarify a bit. So the purpose here is to extend the LM to support not only kyma module operator, but other 3rd party operators which not use deployment to provision operator, as I mentioned, some operators still use StatefulSet. That's why here just want to exclude those operators which not contain deployment.  

That's why here we should not return error, because deployment resource is not a hard requirement, we should not through errors if there are no deployment resources exist.

internal/manifest/ready_check.go
ruanxin: yes, because if the CR state is not initialized, which means the `unstructured.NestedString` will return error, if we deal with error before this exist check, the manifest CR will always be in `error` first, then into `processing`

internal/declarative/v2/reconciler.go
ruanxin: The position is irrelevant, if you look at those returned `crStateInfo`, if the `State` is `StateProcessing`, the returned `error` is always `nil`, but same as previous comments, let's deal with error first to keep consistent.

controllers/kyma_controller/manifest_test.go
ruanxin: you should not quote it 😄

ruanxin: and actually it's your code, self-review is good practice. 😄

internal/declarative/v2/reconciler.go
Tomasz-Smelcerz-SAP: :+1: You are right, that by the **current** implementation it doesn't matter. But implementation will be for sure changed in the future. The `ReadyCheck` interface doesn't declare any specific error-handling policy for the `Run` method. That's why we should stick to the default, idiomatic golang error handling: If the returned error != nil, we ignore any returned values.
If  it's necessary for the `Run` method to have a custom error handling policy, it should be specified somehow (in the godoc?)

Tomasz-Smelcerz-SAP (approved):

Tomasz-Smelcerz-SAP (approved):

PR #717, opened 2023-07-12
manifest parser cache key should not use manifest name

ruanxin:
**Description**

This PR changed the cache key of the manifest parser, at the moment, we use manifest cr name as part of the key, but for each manifest with the same module template, the content (resources) rendered will be no different, have the different key will introduce unnecessary memory occupation, by this change, we use rendered raw manifest yaml path as the key, so that each module with same module template can share with same key.

**Related issue(s)**


Tomasz-Smelcerz-SAP:
This PR should be merged, as it simplifies the code, but I am afraid it will not increase the performance.

I debugged this, and these are the results:
- new cache key (path): `/tmp/kyma-project.io/foo-sha256:3eee012347c87061550a44a13dc91f4fd051a63fa510771d21fdaef3d4af93c5/raw-manifest.yaml`
- old cache key (path + name combined): `manifest/tmp/kyma-project.io/foo-sha256:3eee012347c87061550a44a13dc91f4fd051a63fa510771d21fdaef3d4af93c5/raw-manifest.yaml/raw-manifest-raw`

This new and old one are very similar. Some redundant prefix and suffix are added but this doesn't matter as far as uniqueness is concerned.
The reason for that is following: The manifest `name` (`raw-manifest`) comes from the ModuleTemplate. And in the ModuleTemplate that is built with the recent Kyma-CLI versions, the manifest name is hardcoded to the `raw-manifest` string value.
See:
```
pkg/module/layer.go
11:     rawManifestLayerName = "raw-manifest"
```

PR #728, opened 2023-07-14
Unregister histogram metrics with high cardinality

lindnerby:
**Description**

- Remove the generation of unused metrics since the cardinality is too high on productive environments

**Related issue(s)**

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -33,7 +33,6 @@ var (
 )
 
 func Initialize() {
-
 	requestLatency := prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Subsystem: "rest_client",
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -32,6 +32,7 @@ var (
 	}, []string{moduleNameLabel, kymaNameLabel, stateLabel, shootIDLabel, instanceIDLabel})
 )
 
+//nolint:gomnd
 func Initialize() {
 	requestLatency := prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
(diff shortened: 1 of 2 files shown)


pkg/metrics/metrics.go
```diff
@@ -32,7 +32,47 @@ var (
 	}, []string{moduleNameLabel, kymaNameLabel, stateLabel, shootIDLabel, instanceIDLabel})
 )
 
+//nolint:gomnd
 func Initialize() {
+	requestLatency := prometheus.NewHistogramVec(
```
ruanxin, line 37: Please mention it's temporary solution, and should be removed after controller-runtime upgraded to 0.15.0

pkg/metrics/metrics.go
```diff
@@ -32,7 +32,47 @@ var (
...
+			Name:      "request_latency_seconds",
+			Help:      "Request latency in seconds. Broken down by verb and URL.",
+			Buckets:   prometheus.ExponentialBuckets(0.001, 2, 10),
+		},
+		[]string{"verb", "url"},
+	)
+	requestDuration := prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "rest_client_request_duration_seconds",
+			Help:    "Request latency in seconds. Broken down by verb, and host.",
```
ruanxin, line 49: can we just offer required options? I think `Name` is enough

lindnerby: No sure we would need to test what's the minimal data to pass. From Documentation:
`Unregister` unregisters the Collector that equals the Collector passed in as an argument, and two Collectors are considered equal if their **Describe** method yields the same set of descriptors.
According to the Desc of metrics:
Descriptors registered with the same registry have to fulfill certain consistency and uniqueness criteria if they share the same fully-qualified name: They must have the same help string and the same label names (aka label dimensions) in each, constLabels and variableLabels, but they must differ in the values of the constLabels.

So since it works like this atm and since it's temporary anyways, I don't want to spend time trying out which of the fields can be omitted right now.

PR #730, opened 2023-07-14
enhance smoke test

ruanxin:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/tests/smoke_test/module_upgrade/moduletemplate_template_operator_regular.yaml
+++ b/tests/smoke_test/module_upgrade/moduletemplate_template_operator_regular.yaml
@@ -1,0 +1,48 @@
+apiVersion: operator.kyma-project.io/v1beta2
+kind: ModuleTemplate
+metadata:
+  name: template-operator-regular
+  namespace: kcp-system
+  labels:
+    "operator.kyma-project.io/module-name": "template-operator"
+spec:
+  channel: regular
+  data:
+    apiVersion: operator.kyma-project.io/v1alpha1
+    kind: Sample
+    metadata:
+      name: sample-yaml
+    spec:
+      resourceFilePath: "./base/config.yaml"
+  descriptor:
+    component:
+      componentReferences: []
+      name: kyma-project.io/template-operator
+      provider: internal
+      repositoryContexts:
+        - baseUrl: europe-west3-docker.pkg.dev/sap-kyma-jellyfish-dev/template-operator
+          componentNameMapping: urlPath
+          type: ociRegistry
+      resources:
+        - access:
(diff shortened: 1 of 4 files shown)


internal/declarative/v2/reconciler.go
```diff
@@ -296,6 +294,27 @@ func (r *Reconciler) syncResources(
...
+		if countMap[item.ID()] {
+			countMap[item.ID()] = false
+		}
+	}
+	for _, exists := range countMap {
+		if exists {
+			return true
+		}
+	}
+	return false
```
Tomasz-Smelcerz-SAP, line 315: A bit more efficient implementation.
Implicitly assumes that `item.ID()` is unique in both collections, but your code also requires that.
Should it return true if a new resource is detected, but the length of the two collections is the same? I assumed so.

```suggestion
	var diffs = 0
	countMap := map[string]bool{}
	for _, item := range oldResources {
		countMap[item.ID()] = true
		diffs += 1
	}
	for _, item := range newResources {
		if countMap[item.ID()] {
			diffs -= 1
		} else {
			return true //a new resource detected
		}
	}
	
	return diffs != 0
```

ruanxin: this solution have some edge case problem, it will return true if `newResources` list contains duplicate resources but the list size is the same as `oldResources`, check the updated [unit test case](https://github.com/kyma-project/lifecycle-manager/pull/730/files#diff-608c0e7ac0ff9d6c47c49efcc9c990d99b696100b56632bb72d961ec37d7e3a5R113-R116) as example

tests/smoke_test/module_upgrade/moduletemplate_template_operator_regular.yaml
```diff
@@ -0,0 +1,48 @@
...
+            digest: sha256:903eda97406482c42b7aa34ae7da1dac68fdacaccd2a70158ef13a09c2230915
+            type: localOciBlob
+          name: raw-manifest
+          relation: local
+          type: yaml
+          version: v1.1.0-smoke-test
+      sources:
+        - access:
+            commit: d9ddb3ddd0e4579bba24b7f56027a939fb559e1f
+            repoUrl: https://github.com/jakobmoellersap/cli
```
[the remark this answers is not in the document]
ruanxin: I think it's because the cli binary I generated is from a local release branch, updated it with main branch.

internal/declarative/v2/reconciler.go
ruanxin: I enhanced the parsing logic to remove duplication, please [check](https://github.com/kyma-project/lifecycle-manager/pull/730/files#diff-4a98c02a04a96038d6875b39e833779fc588b08dcf9ec098b46b572d5bd3583eR38-R51)

Tomasz-Smelcerz-SAP (approved):

Issue #694, opened 2023-06-27
Enable forcetypeassert and fix all related linter check


ruanxin:
type assertion without check will panic program if unexpected error happens, it's a bad practice without this check.

AC
- [x] double-check whether this is handled by the errcheck (see if this is enabled) - [link](https://gist.github.com/maratori/47a4d00457a92aa426dbd48a18776322)
- [x] enable `forcetypeassert` and fix all unchecked type assertion

lindnerby:
Enabled on both LM and CLI.

PR #735, opened 2023-07-19
Enable type assertion linter

lindnerby:
**Description**

- Bump golangci-lint version
- Disabled `depguard` that comes with this version bump and we have no package whitelist
- Introduce errors for failed type assertions
- Check type assertions

**Related issue(s)**
* https://github.com/kyma-project/lifecycle-manager/issues/694


api/v1beta2/error.go
```diff
@@ -0,0 +1,11 @@
+package v1beta2
+
+import "errors"
+
+var (
+	ErrTypeAssertKyma           = errors.New("raw can not be cast to v1beta2.Kyma")
```
Tomasz-Smelcerz-SAP, line 6: I don't understand what `raw` means in the generic context. Perhaps the `value can not be cast to v1beta2.Manifest` would be better?

pkg/common/error.go
```diff
@@ -0,0 +1,5 @@
+package common
+
+import "errors"
+
+var ErrTypeAssert = errors.New("generic TypeAssertion error")
```
Tomasz-Smelcerz-SAP, line 5: Perhaps just: "TypeAssertion failed" ?

Tomasz-Smelcerz-SAP:
I have found a not-detected type assertion:

```
controllers/control-plane/kcp_controller_helper_test.go
110:    value, found := initKey.(map[string]any)["initKey"]
```
Why it wasn't detected by the linter?

internal/declarative/v2/reconciler.go
```diff
@@ -463,8 +468,12 @@ func updateSyncedOCIRefAnnotation(obj Object, ref string) {
 }
 
 func pruneResource(diff []*resource.Info, resourceType string, resourceName string) []*resource.Info {
+	//nolint:varnamelen
 	for i, info := range diff {
-		obj := info.Object.(client.Object)
+		obj, ok := info.Object.(client.Object)
+		if !ok {
```
Tomasz-Smelcerz-SAP, line 474: Is hiding the conversion problem here OK? Perhaps an error should be returned instead?

lindnerby:
> I have found a not-detected type assertion:
> 
> ```
> controllers/control-plane/kcp_controller_helper_test.go
> 110:    value, found := initKey.(map[string]any)["initKey"]
> ```
> 
> Why it wasn't detected by the linter?
Guess because the map key exists check is covering it?  No idea, I separated it.

PR #742, opened 2023-07-24
Disable Manifest Watcher

jeremyharisch:
**Description**

Since we currently want to focus on the KymaWatch mechanism, we decided to disable the Manifest Watcher for the current time.

In future it may will be enabled again.

### DIFF
--- a/config/watcher/operator_v1beta2_watcher.yaml
+++ b/config/watcher/operator_v1beta2_watcher.yaml
@@ -22,29 +22,29 @@ spec:
       matchLabels:
         "operator.kyma-project.io/watcher-gateway": "default"
 ---
-apiVersion: operator.kyma-project.io/v1beta2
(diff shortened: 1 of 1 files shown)


config/watcher/operator_v1beta2_watcher.yaml
```diff
@@ -22,29 +22,29 @@ spec:
...
+#  field: "status"
+#  serviceInfo:
+#    name: klm-event-service
+#    port: 8083
+#    namespace: kcp-system
+#  gateway:
+#    selector:
+#      matchLabels:
+#        "operator.kyma-project.io/watcher-gateway": "default"
 ---
```
lindnerby, line 48:
```suggestion
# ---
```

lindnerby (changes requested):

Issue #733, opened 2023-07-18
Watcher issues on DEV


jeremyharisch:
**Description**
The Watcher has been enabled on Dev. Due to our istio setup on the landscape, we are encountering several issues in the mTLS communication inside of the SKR and between the SKR and KCP.

ACs:
- TLS issues to discuss with Goats (@Janmedrek will clarify)
    - Can we disable sidecar for ValidationWebhook inside of SKR?
    - If yes, include annotation to disable sidecar in [runtime-watcher deployment](https://github.com/kyma-project/lifecycle-manager/blob/8f2a55ddb9f3fb38ad6c6558bdcab115857dd5bd/skr-webhook/resources.yaml#L58)
- Investigate why requests from SKR to KCP are failing
    - Check if deployed Certificates contain the sufficient SANs
    - Gateway wrongly configured?
    - ...

PR #743, opened 2023-07-24
Make KCP-Gateway-HTTPS port configurable; incl minor changes

jeremyharisch:
**Description**

- Disable sidecar Proxy for Runtime-Watcher on Rmeote-Cluster
- Change default Port for local k3d Testing from 9080 to 9443
- Remove flag overwrite from localtest-setup kustomize, since default is now already set to 9443
- Introduce new flag `listenerGatewayPortName` to be able to configure the Gateway Port Name; default set to `https`-> should fix the DEV landscape bug
- Introduce new flag `defaultIstioIngressServiceName` to be able to configure the Istio Ingress Name; default set to `istio-ingressgateway`-
- Introduce new flag `defaultIstioNamespace` to be able to configure the Istio Namespace; default set to `istio-system`

**Related issue(s)**

Should fix https://github.com/kyma-project/lifecycle-manager/issues/733 , since the runtime-watcher requests have been send to `https://<dev-ip>:80/v1/lifecycle-manager/event` and got denied with the following error `"server gave HTTP response to HTTPS client"`. This happened, since in the KLM it was hardcoded that the runtime watcher should send requests to the http2 port of the Istio Loadbalancer, which is incorrect - It should send requests to the port called `https`. To be able to configure this on the fly, I introduced a new flag for it. -> Default is set to `https`, and can be changed using the flag `listener-gateway-port-name`.

This error was not catched in local testing or in the pipelines, since they are running with two k3d cluster on the same machine and thus, the request address is overwritten with `host.k3d.internal:9443`.

### DIFF
--- a/flags.go
+++ b/flags.go
@@ -29,6 +29,8 @@ const (
 	defaultMaxConcurrentManifestReconciles = 25
 	defaultMaxConcurrentKymaReconciles     = 25
 	defaultMaxConcurrentWatcherReconciles  = 1
+	defaultIstioIngressServiceName         = "istio-ingressgateway"
+	defaultIstioNamespace                  = "istio-system"
 )
 
 //nolint:funlen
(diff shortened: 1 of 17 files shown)


pkg/watcher/webhook_chart.go
```diff
@@ -75,7 +76,7 @@ func resolveKcpAddr(kcpConfig *rest.Config, managerConfig *SkrWebhookManagerConf
 	externalIP := loadBalancerService.Status.LoadBalancer.Ingress[0].IP
 	var port int32
 	for _, loadBalancerPort := range loadBalancerService.Spec.Ports {
-		if loadBalancerPort.Name == "http2" {
+		if loadBalancerPort.Name == managerConfig.ListenerGatewayPortName {
```
ruanxin, line 79: The `IngressServiceName` and `IstioSystemNs` better also make them configurable as flags, but not hardcoded.

flags.go
```diff
@@ -89,6 +90,8 @@ func defineFlagVar() *FlagVar {
 		"Enabling KCP Watcher two-cluster setup to be tested locally using k3d")
 	flag.StringVar(&flagVar.istioNamespace, "istio-namespace", "istio-system",
 		"CLuster Resource Namespace of Istio")
+	flag.StringVar(&flagVar.listenerGatewayPortName, "listener-gateway-port-name", defaultListenerGatewayPortName,
+		"Port name that is specified in the cluster's gateway used for incoming watcher requests")
 	flag.IntVar(&flagVar.listenerHTTPPortLocalMapping, "listener-http-local-mapping", defaultListenerPort,
```
ruanxin, line 95: The thing is ports `443` and `80` have special meaning, `443` is default for `https`, if you change the local port to `9443`, then it indicates it's https protocol, so the `listenerHTTPPortLocalMapping` better change to `listenerHTTPSPortLocalMapping`. But the current local k3d setup is HTTP only, right? why we use `9443`?

ruanxin (approved):

PR #744, opened 2023-07-24
template_to_module.go: Don't Include Config Layer in Manifest

### LINKED ISSUE #543: Remove config layer
Since now we only support raw manifest rendering, the config layer which was previously provided as helm chart, kustomize overwrite becomes useless. We should remove any config.yaml related code in lifecycle manager.

AC:
- [x] CLI don't need to handle `config.yaml`, and generated module template should not contains config layer.
- [x] Drop the code from the CLI related to the config layer
- [x] For module template which contains config, the config layer will not rendered into manifest CR.
- [x] All code in lifecycle manager related handling this config should be removed
- [x] Update the description in the ModuleTemplate CRD to indicate that the config field is no longer used

**Be aware, eventually, manifest spec should not contain this config field, but this might introduce breaking changes, should plan it in next API version. In this issue, don't remove this Config field.** 
```
type ManifestSpec struct {
	// Config specifies OCI image configuration for Manifest
	Config ImageSpec `json:"config,omitempty"`

```

LeelaChacha:
closes #543


pkg/module/parse/template_to_module.go
```diff
@@ -180,9 +180,8 @@ func insertLayerIntoManifest(
 	manifest *v1beta2.Manifest, layer img.Layer,
 ) error {
 	switch layer.LayerName {
+	case img.ConfigLayer: // Skipping Config Layer. Deprecated since removal of Helm and Kustomize from ModuleTemplates
```
ruanxin, line 183: Just skip is not enough, you need to remove the usage of `manifest.Spec.Config` code completely.

ruanxin (changes requested):
please fully cleanup of `manifest.Spec.Config`

Issue #702, opened 2023-07-04
[Regression test] For Manifest CR


ruanxin:
Reference the internal integration test wiki page, add missing integration test for Manifest CR.

AC:

- [ ] add missing test case for existing test
- [ ] create new test for missing test
- [ ] update wiki page accordingly

Tomasz-Smelcerz-SAP:
Reopening because of #866

PR #748, opened 2023-07-26
Extend manifest cr tests

Tomasz-Smelcerz-SAP:
**Description**

Extend existing `envtest`-based test cases for the `Manifest` CR.

- extends tests in the `controllers/manifest_test.go` file
- extend test that verifies that the `Manifest` CR is updated after `ModuleTemplate` is updated (change to the Module Descriptor)
- add test for correct rendering of the `Manifest` CR
- add test that verifies that the `Manifest` CR is reset according to the values in the `ModuleTemplate` when updated manually.

**Related issue(s)**
See: #702

### DIFF
--- a/controllers/manifest_test.go
+++ b/controllers/manifest_test.go
@@ -2,7 +2,6 @@ package controllers_test
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"reflect"
 	"strings"
@@ -110,7 +109,7 @@ var _ = Describe("Update Manifest CR", Ordered, func() {
 			hasDummyRepositoryURL := func(manifest *v1beta2.Manifest) error {
 				manifestImageSpec := extractInstallImageSpec(manifest.Spec.Install)
 				if !strings.HasPrefix(manifestImageSpec.Repo, updateRepositoryURL) {
-					return errors.New(fmt.Sprintf("Invalid manifest spec.install.repo: %s, expected prefix: %s", manifestImageSpec.Repo, updateRepositoryURL))
+					return fmt.Errorf("Invalid manifest spec.install.repo: %s, expected prefix: %s", manifestImageSpec.Repo, updateRepositoryURL)
 				}
 				return nil
 			}
@@ -168,7 +167,7 @@ var _ = Describe("Manifest.Spec is reset after manual update", Ordered, func() {
 	RegisterDefaultLifecycleForKyma(kyma)
 
 	It("update Manifest", func() {
-		//Await for the manifest to be created
+		// await for the manifest to be created
 		Eventually(GetManifestSpecRemote, Timeout, Interval).
 			WithArguments(ctx, controlPlaneClient, kyma, module).
 			Should(Equal(false))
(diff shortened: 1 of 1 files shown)


controllers/manifest_test.go
```diff
@@ -61,5 +77,299 @@ var _ = Describe("Update Manifest CR", Ordered, func() {
...
+	module := v1beta2.Module{
+		ControllerName: "manifest",
+		Name:           NewUniqModuleName(),
+		Channel:        v1beta2.DefaultChannel,
+	}
+	kyma.Spec.Modules = append(kyma.Spec.Modules, module)
+	RegisterDefaultLifecycleForKyma(kyma)
+
+	It("validate Manifest", func() {
+		moduleTemplate, err := GetModuleTemplate(ctx, controlPlaneClient, module.Name, "default")
```
nesmabadr, line 136: Please verify that the `sync-oci-ref` annotation also exists in this test

controllers/manifest_test.go
```diff
@@ -61,5 +77,299 @@ var _ = Describe("Update Manifest CR", Ordered, func() {
...
+	)
+
+	manifestImageSpec = extractInstallImageSpec(manifestInstall)
+	moduleTemplateDescriptor, err = moduleTemplate.GetDescriptor()
+	if err != nil {
+		return err
+	}
+
+	//compares the actual manifest spec.install.source.name with the corresponding values from the ModuleTemplate
+	compareManifestSourceName := func() error {
```
nesmabadr, line 234: I suggest having smaller functions to validate each of the spec.install.source components for better readability

controllers/manifest_test.go
```diff
@@ -61,5 +77,299 @@ var _ = Describe("Update Manifest CR", Ordered, func() {
...
+
+	It("validate Manifest", func() {
+		moduleTemplate, err := GetModuleTemplate(ctx, controlPlaneClient, module.Name, "default")
+		Expect(err).NotTo(HaveOccurred())
+
+		expectManifest := expectManifestFor(kyma)
+
+		By("checking Spec.Install")
+		hasValidSpecInstall := func(manifest *v1beta2.Manifest) error {
+			return validateManifestSpecInstall(manifest.Spec.Install, moduleTemplate)
```
nesmabadr, line 143: why don't you just do 
`Eventually(validateManifestSpecInstall, Timeout, Interval).WithArguments(manifest.Spec.Install, moduleTemplate).Should(Succeed())`?

and then there would be no need for `expectManifest` function

nesmabadr:
Also, please don't forget to update the wiki with the newly added test cases

controllers/manifest_test.go
Tomasz-Smelcerz-SAP: This annotation can't be tested here, because it's set by the Manifest controller, which is disabled in this test.
I'll test it's presence in the subsequent PR for tests in the `internal/manifest` package.

controllers/manifest_test.go
Tomasz-Smelcerz-SAP: The short answer is: I don't do that, because `validateManifestSpecInstall` is a pure function - it has no side effects.
Longer explanation:
Being "pure" means function's invocation always yields the same results given the function arguments are the same. If so, there's no reason to test it using `Eventually` assertion, which is intended to re-try the function in the hope it will `eventually` pass. For pure functions, like `validateManifestSpecInstall` there's no such hope :)
The `expectManifest` plays a very important role here. 
`validateManifestSpecInstall` is a pure function that just validates things and can be used both with `Eventually` assertion and with a standard one like `Expect`
`expectManifest` - returns an anonymous function that is actually used as a target for `Eventually` assertion. This anonymous function fetches the "fresh" `Manifest` object each time it's invoked and then passes that to the provided validation function - in that case it's wired to `validateManifestSpecInstall`.
Thanks to this function composition the `Eventually` assertion works as expected. In fact it can be observed that in some cases the assertion fails initially, and on second or third "try" it succeeds - it's because it takes time for the Kyma controller to update the Manifest's CR Spec.

nesmabadr (approved):

PR #749, opened 2023-07-26
linter fix: increase timeout and remove concurrency

lindnerby:
**Description**

- Set linter timeout to 15min
- Remove concurrency and use default no of cores

**Related issue(s)**

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -48,7 +48,5 @@ issues:
 output:
   sort-results: true
 run:
-  concurrency: 8
-  timeout: 3m
   skip-files:
     - zz_generated.deepcopy.go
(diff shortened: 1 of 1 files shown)


.golangci.yaml
```diff
@@ -48,7 +48,6 @@ issues:
 output:
   sort-results: true
 run:
-  concurrency: 8
```
ruanxin, line 51: high concurrency can faster process time, the default is 4, if you want, we can test this on this PR, but need a `.go` file change to trigger the job

lindnerby: docu says:   `# The default concurrency value is the number of available CPU.`

ruanxin (approved):

PR #764, opened 2023-07-31
Add Github Actions lint workflow

lindnerby:
**Description**

- Add pull-lifecycle-mgr-lint workflow
- Pin version to 1.53.3
- Disabled `depguard` that comes with the version bump, we have no package whitelist
- Increase max cyclop a bit to 20 (recommended for go is 30)
- Add timeout: 15m
- Remove autofix for issues, this swallows issues on PRs
- Fix new issues appearing since version bump and autofix removal

**Related issue(s)**

### DIFF
--- a/.github/workflows/pull-lifecycle-mgr-lint.yml
+++ b/.github/workflows/pull-lifecycle-mgr-lint.yml
@@ -12,7 +12,6 @@ jobs:
       - uses: actions/setup-go@v4
         with:
           go-version: '1.20'
-          cache: false
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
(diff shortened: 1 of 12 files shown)


.github/workflows/pull-lifecycle-mgr-lint.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.20'
+          cache: false
```
ruanxin, line 15: why not cache enabled?

lindnerby: No thoughts given to that, just taken from the golangci-lint action sample.
Just read a bit about it, for the cache to work, we need to make sure that the checkout step is executed before, which is the case here.
Will try with enable cache (default for v4 setup-go)

ruanxin (approved):

PR #768, opened 2023-08-01
Fix Watcher Landscape Setup

jeremyharisch:
**Description**

- Remove Manifest Watcher fully
- Remove duplicated install of istioctl and cmctl
- Change Gateway host to correct format; it anyway needs an overwrite in kustomize for each landscape, but it is nicer to already have the correct format
- Fix selector of KLM service

**Related issue(s)**


.github/workflows/e2e-test.yml
```diff
@@ -92,15 +92,9 @@ jobs:
         run: kubectl config use-context k3d-kcp
       - name: Deploy Istio on KCP Cluster
         run: |
-          curl -L https://istio.io/downloadIstio | TARGET_ARCH=x86_64 sh -
-          chmod +x istio-$ISTIO_VERSION/bin/istioctl
-          mv istio-$ISTIO_VERSION/bin/istioctl /usr/local/bin
           istioctl install --set profile=demo -y
```
ruanxin, line 95: why here remove download `istioctl` but still use istioctl?

.github/workflows/e2e-test.yml
```diff
@@ -92,15 +92,9 @@ jobs:
...
       - name: Deploy Cert Manager on KCP Cluster
         run: |
-          OS=$(go env GOOS); ARCH=$(go env GOARCH); curl -fsSL -o cmctl.tar.gz https://github.com/cert-manager/cert-manager/releases/latest/download/cmctl-$OS-$ARCH.tar.gz
-          tar xzf cmctl.tar.gz
-          sudo mv cmctl /usr/local/bin
           kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/$CM_VERSION/cert-manager.yaml
           cmctl check api --wait=2m
```
ruanxin, line 99: same here for `cmctl`

pkg/watcher/webhook_chart.go
```diff
@@ -53,31 +56,27 @@ func resolveKcpAddr(kcpConfig *rest.Config, managerConfig *SkrWebhookManagerConf
...
 		return "", err
 	}
-	ctx := context.TODO()
-	loadBalancerService := &corev1.Service{}
-	if err := kcpClient.Get(ctx, client.ObjectKey{
-		Name:      managerConfig.IstioIngressServiceName,
-		Namespace: managerConfig.IstioNamespace,
-	}, loadBalancerService); err != nil {
+
+	gateway, err := kcpIstioClient.NetworkingV1beta1().Gateways(managerConfig.IstioGatewayNamespace).
```
ruanxin, line 67: I'm not recommend to import `kcpIstioClient`, instead, what you can do is, add istio schema in main.go `init()`, then you can use `kcpClient` to get this istio gateway.

```
	utilruntime.Must(istiov1beta1.AddToScheme(scheme))
```

.github/workflows/e2e-test.yml
jeremyharisch: we do not need to download it twice

.github/workflows/e2e-test.yml
jeremyharisch: we do not need to download it twice

ruanxin (approved):

PR #770, opened 2023-08-02
Remove flaky module upgrade integration test

ruanxin:
This module upgrade implemented as integration test is not a wise decision, should be covered in smoke test (already have) and e2e test later.


lindnerby (approved):
@ruanxin We have an issue for the e2e extension of this?

ruanxin:
@lindnerby the e2e test issue will create later

PR #786, opened 2023-08-08
Remove Config Layer Code for Manifest

### LINKED ISSUE #543: Remove config layer

LeelaChacha:
Closes #543

### DIFF
--- a/controllers/kyma_controller/moduletemplate_test.go
+++ b/controllers/kyma_controller/moduletemplate_test.go
@@ -51,9 +51,6 @@ func expectManifestSpecNotContainsCredSecretSelector(kymaName string) func() err
 			if err != nil {
 				return err
 			}
-			if moduleInCluster.Spec.Config.CredSecretSelector != nil {
-				return ErrContainsUnexpectedCredSecretSelector
-			}
 			installImageSpec := extractInstallImageSpec(moduleInCluster.Spec.Install)
 
 			if installImageSpec.CredSecretSelector != nil {
@@ -74,12 +71,6 @@ func expectManifestSpecContainsCredSecretSelector(kymaName string) func() error 
 			moduleInCluster, err := GetManifest(ctx, controlPlaneClient, createdKyma, module)
 			if err != nil {
 				return err
-			}
-			var emptyImageSpec v1beta2.ImageSpec
-			if moduleInCluster.Spec.Config != emptyImageSpec {
-				if err := expectCredSecretSelectorCorrect(&moduleInCluster.Spec.Config); err != nil {
-					return fmt.Errorf("config %v is invalid: %w", &moduleInCluster.Spec.Config, err)
-				}
 			}
 
 			installImageSpec := extractInstallImageSpec(moduleInCluster.Spec.Install)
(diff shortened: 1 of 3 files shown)


controllers/kyma_controller/moduletemplate_test.go
```diff
@@ -40,108 +25,9 @@ func expectManifestSpecDataEquals(kymaName, value string) func() error {
...
-			installImageSpec, ErrNotContainsExpectedCredSecretSelector)
-	}
-
-	value, found := installImageSpec.CredSecretSelector.MatchLabels[credSecretLabel]
-	Expect(found).To(BeTrue())
-	Expect(value).To(Equal(credSecretValue))
-	return nil
-}
-
-var _ = Describe("ModuleTemplate.Spec.descriptor not contains RegistryCred label", Ordered, func() {
```
ruanxin, line 113: Why you deleted this two test?

ruanxin: They are valid testing

LeelaChacha: Since they were reading the Manifest.Spec.Config. Now that we won't write anything to the config, the tests don"t make sense.

LeelaChacha: No wait. You're right.

LeelaChacha: We don't have the credselector in the raw install

ruanxin: no, that's not true, the `CredSecretSelector` used in `manifest.Spec.Install` you can find here: https://github.com/kyma-project/lifecycle-manager/blob/main/internal/manifest/spec_resolver.go#L139

LeelaChacha: ok fair point.

ruanxin (changes requested):
but now you still have leftover in the test which uses `.Spec.Config` in `moduletemplate_test.go`, previously you completely removed the test, but `.Spec.Install` is still in use, but it doesn't means you should not care about that test,  because in the test case,  the deprecated  `.Spec.Config` still being test, you should do the cleanup accordingly.

ruanxin (approved):

PR #787, opened 2023-08-08
Fix ignored error and improper panic

ruanxin:
This PR fix ignored error and improper panic which complained by checkmarx


medmes (approved):
LGTM!

Issue #693, opened 2023-06-27
Enable wrapcheck and fix all unwrapped error in the KLM


ruanxin:
For better troubleshooting, we need to make sure that errors from external packages must attach extra information so that we can track down to the place where the error happens easily.

At the moment, we have 143 errors need to be fixed after enabled wrapcheck
```
make lint | grep "(wrapcheck)" | wc -l
```

Note: remove wrapcheck from diable list in `.golangci.yaml`

AC:
- [x] `wrapcheck` should be enabled by default
- [x] fix all existing unwrapped errors

PR #788, opened 2023-08-09
Enable wrapcheck linter & code refactor if needed

medmes:
**Description**
This PR purpose is for enabling `wrapcheck`, fix and refactor the unwrapped errors

- Wrap errors

**Related issue(s)**
#693


.golangci.yaml
```diff
@@ -41,7 +40,15 @@ linters-settings:
       - Client
       - record.EventRecorder
       - client.Object
+  wrapcheck:
+    ignorePackageGlobs:
+      - github.com/kyma-project/lifecycle-manager/pkg/*
```
ruanxin, line 45: If you ignore `pkg`, basically the meaning of this refactoring task is lost, right? most of the unwrapped errors come from pkg.

medmes: @lindnerby is working in parallel to tackle the `pkg/` package. He will commit the new changes soon!

lindnerby: this will ignore not wrapped errors that come from github.com/kyma-project/lifecycle-manager/pkg/*
since these are our pkgs, the consumer of this package (also us, controller most likely) doesn't need to always wrap them. but it is still analysing every error within pkg.

ruanxin (approved):

controllers/manifest_controller.go
```diff
@@ -69,7 +71,12 @@ func SetupWithManager(
 					queue.Add(ctrl.Request{NamespacedName: client.ObjectKeyFromObject(event.Object)})
 				},
 			},
-		).WithOptions(options).Complete(ManifestReconciler(mgr, codec, checkInterval))
+		).WithOptions(options)
+
+	if controllerManagedByManager.Complete(ManifestReconciler(mgr, codec, checkInterval)) != nil {
+		return fmt.Errorf("failed to initialize manifest controller by manager: %w", err)
+	}
+	return nil // never end up here :)
```
ruanxin, line 79: no, it should always end up here, this comment is misleading.

PR #806, opened 2023-08-14
Client cache invalidation when remote cluster is unreachable

### LINKED ISSUE #281: Client cache invalidation mechanism for changing secrets
**Description**
The remote client cache `(operator/pkg/remote/client_cache.go)` implements a client caching mechanism, which returns and saves cache entries. However, a mechanism is required to delete cache entries or invalidate them in scenarios like - deletion of Kyma CR or kubeconfig rotation stored in a secret. 

Consider including both the TTL and forced eviction on errors.

**Acceptance Criteria**

- [x] Implement invalidation (or deletion) if the remote cluster is not reachable - the kubeconfig should be removed from the cache and a new one should be fetched from the secret
- [ ] Add E2E test for that scenario

nesmabadr:
**Description**

- Implement client cache invalidation when remote cluster is unreachable
- Add E2E test

**Related issue(s)**
Resolves #281


tests/e2e_test/kyma_test.go
```diff
@@ -0,0 +1,106 @@
+//go:build kyma_e2e
```
ruanxin, line 1: I suggest integrating this e2e scenario into `deletion_e2e`, the scenario of this test is more suitable for "deleting"

controllers/kyma_controller.go
```diff
@@ -125,15 +125,16 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	return r.reconcile(ctx, kyma)
 }
 
-func (r *KymaReconciler) handleRemoteClusterConnectionErrorOnDeletion(
+func (r *KymaReconciler) handleRemoteClusterConnectionError(
 	ctx context.Context, kyma *v1beta2.Kyma, err error) (
 	ctrl.Result, error,
 ) {
+	if util.IsConnectionRefused(err) {
```
ruanxin, line 132: Code style suggestion:
For this function `handleRemoteClusterConnectionError`, it's better not to return `ctrl.Result`, but just `error` because `ctrl.Result` should be only handled in `reconcile`, which is a layer above this function, and only return `error` it can simplify your return logic for this function, and make this function more generic, not depends with controller-runtime items(`ctrl.Result`).

ruanxin (changes requested):
Please see my comments

controllers/kyma_controller.go
```diff
@@ -152,15 +149,17 @@ func (r *KymaReconciler) reconcile(ctx context.Context, kyma *v1beta2.Kyma) (ctr
 		remoteClient := remote.NewClientWithConfig(r.Client, r.KcpRestConfig)
 		if ctx, err = remote.InitializeSyncContext(ctx, kyma,
 			r.RemoteSyncNamespace, remoteClient, r.RemoteClientCache); err != nil {
-			return r.handleRemoteClusterConnectionErrorOnDeletion(ctx, kyma, err)
+			if err = r.handleRemoteClusterConnectionError(ctx, kyma, err); err != nil {
```
ruanxin, line 152: here you can write like this:
```
if ctx, err = remote.InitializeSyncContext(ctx, kyma,
			r.RemoteSyncNamespace, remoteClient, r.RemoteClientCache); r.handleRemoteClusterConnectionError(ctx, kyma, err) != nil {

}
```

ruanxin (approved):

PR #808, opened 2023-08-16
fix/enhance error handling

ruanxin:
This PR fixed some cases return nil error

### DIFF
--- a/controllers/kyma_controller.go
+++ b/controllers/kyma_controller.go
@@ -104,15 +104,11 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 
 	kyma := &v1beta2.Kyma{}
 	if err := r.Get(ctx, req.NamespacedName, kyma); err != nil {
-		if util.IsNotFound(err) {
-			// TODO: revisit after runtime-controller gets upgraded
-			// Related issue: https://github.com/kyma-project/lifecycle-manager/issues/579
+		if !util.IsNotFound(err) {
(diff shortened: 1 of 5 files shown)


controllers/manifest_controller.go
```diff
@@ -76,7 +76,7 @@ func SetupWithManager(
 	if controllerManagedByManager.Complete(ManifestReconciler(mgr, codec, checkInterval)) != nil {
 		return fmt.Errorf("failed to initialize manifest controller by manager: %w", err)
 	}
-	return nil // never end up here :)
```
ruanxin, line 79: To reviewer:
a misleading (it should be expected to return nil) and unnecessary comments.

medmes: True, it was an unnecessary comment on line 79.

medmes (approved):

controllers/kyma_controller.go
```diff
@@ -104,15 +104,11 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 
 	kyma := &v1beta2.Kyma{}
 	if err := r.Get(ctx, req.NamespacedName, kyma); err != nil {
-		if util.IsNotFound(err) {
-			// TODO: revisit after runtime-controller gets upgraded
-			// Related issue: https://github.com/kyma-project/lifecycle-manager/issues/579
+		if !util.IsNotFound(err) {
```
nesmabadr, line 107: why is this condition negated? (Same for the rest)

Issue #794, opened 2023-08-10
Force skr webhook start with TLS mode


lindnerby:
At the moment, runtime skr webhook service is possible to start with plaintext HTTP mode. https://github.com/kyma-project/runtime-watcher/blob/main/skr/main.go#L131-L135, which is also a vulnerable case [reported by Checkmarx](https://github.tools.sap/kyma/security-scans-modular/issues/16246).

AC:
- Remove the code to start with HTTP mode `server.ListenAndServe()`.
- Remove using `params.TLSServer` to control server start with `server.ListenAndServeTLS`, which should be the default setup.
- Referenced [ConigMap in LM](https://github.com/kyma-project/lifecycle-manager/blob/b0dab11e4ab8bf93a7b14ea00f53b71cc94cf9de/skr-webhook/resources.yaml#L29C4-L29C4) needs to be discussed at some other time

PR #817, opened 2023-08-17
Remove TLS config from deployed skr-webhook resources

lindnerby:
**Description**

- Remove TLS flag from watcher config 

**Related issue(s)**
* https://github.com/kyma-project/lifecycle-manager/issues/794


skr-webhook/resources.yaml
```diff
@@ -107,7 +95,7 @@ spec:
                   key: contractVersion
             - name: WEBHOOK_SIDE_CAR
               value: "true"
-          image: "europe-docker.pkg.dev/kyma-project/prod/runtime-watcher-skr:latest"
+          image: "europe-docker.pkg.dev/kyma-project/prod/runtime-watcher-skr:v20230824-f4459bad"
```
nesmabadr, line 98: is this intended?

lindnerby: yes, i would rather like to pin it to a specific version, than use latest, because any merge to runtime-watcher main will trigger the post build and create a new latest. Just to make it deliberate to bump the img version of the runtime-watcher and since the coverage of tests for that webhook is not so good, it would be easy to push bugs and have a faulty img

lindnerby: when the runtime-watcher is more stable, has more tests and is not changed frequently anymore, we can consider setting it to latest again :)

nesmabadr (approved):

Issue #836, opened 2023-08-29
LM should not block user enable module during deleting


ruanxin:
As reported by this [bug](https://github.com/kyma-project/lifecycle-manager/issues/833), LM will prevent the module from being upgraded if module CR is stuck under deleting. Which brings side effects.

The generic problem here is how we deal with the situation user disabled module, and module is under deletion, meanwhile, the user enabled it again.

The request here is not to block this user behaviour, and LM is still possible to hand status correctly.

AC:
- [ ] If module is under deletion, but user enables it during this time, module CR status should not be affected, other synced resources should be installed again.
- [ ] More specifically, if LM detects a new version to be applied (by channel switch, moduletemplate update), the upgrade process should not be blocked, which means all new resources should be installed.
- [ ] Manifest CR should marked for deleting (contains DeletionTimestamp) after all resources get deleted.

jakobmoellerdev:
Interesting issue... From what I can sense there are a few things that are relevant here:
1. Module Enabled means a Spec, e.g. "desired State" of the module in the Kyma CR changed.
2. If the Module is enabled after it is deleted, the entire deletion needs to proceed as normal and the old Module needs to be removed entirely (check deletionTimestamp on Manifest for this, if timestamp is set then that means we have to wait), and after this we need to recreate everything again from scratch based on the ModuleTemplate.
3. If the LifecycleManager detects a new version while the deletion is progressing, it is simply not possible to change the image of the module deployment, simply because the deletionTimestamp stops specs upgrade completely. It is only possible to force remove the module by dropping finalizers and possibly killing state.
4. IMO it is bad practice to mark for deletion after resources are deleted. This is against kubernetes best practices and also does not respect deletionTimestampGracePeriod in k8s.

Basically this is a case of delete Module, wait until gone / suceeded or force removal, then recreate new module from spec.

Everything else is a reimagination of the kubernetes spec and should be discouraged.

ruanxin:
In the ideal case, that's true, but in practice, we will have some challenges preventing module operator upgrades during deletion. For example, those security issues mentioned [here](https://github.com/kyma-project/lifecycle-manager/issues/833#issuecomment-1697009603) will be a problem if the module operator is not up-to-date for a long time.

Regarding the idea of postponing deleting manifest CR, it's basically trying to keep manifest CR not violating the k8s principle. If we immediately delete manifest CR (set deletionTimestamp), then there is no chance of bringing it to a normal state before deleting it completely.

[...]

PR #839, opened 2023-09-01
chore: Remove unnecessary ctx pointer

jakobmoellerdev:
**Description**

- removes unnecessary ctx pointer introduced in https://github.com/kyma-project/lifecycle-manager/pull/837

#alwayswatching

**Related issue(s)**


controllers/kyma_controller.go
```diff
@@ -125,7 +125,7 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 func (r *KymaReconciler) reconcile(ctx context.Context, kyma *v1beta2.Kyma) (ctrl.Result, error) {
 	if r.SyncKymaEnabled(kyma) {
 		remoteClient := remote.NewClientWithConfig(r.Client, r.KcpRestConfig)
-		if err := remote.InitializeSyncContext(&ctx, kyma,
+		if err := remote.InitializeSyncContext(ctx, kyma,
```
ruanxin, line 128: wow, @jakobmoellerdev, it's so glad to see you continue work on this project 😄 

This ctx pointer is actually needed, with this pointer reference, the ctx altered with additional syncContext can be used in further steps, otherwise, when it needs to use `SyncContextFromContext`, the ctx used there will not with syncContext, which leads to `ErrIsNoSyncContext`.

jakobmoellerdev: I just changed it now so that the ineffectual assignment is gone. I believe this method is a bit strange but now at least the ctx switch is fine again

jakobmoellerdev: btw is there any way I can look into the backlog you guys currently have? Would like to contribute from time to time still?

ruanxin: ok, now you changed back before this function gets reverted https://github.com/kyma-project/lifecycle-manager/pull/837/files#diff-f3f736c5fa557c8d1171caedf277f1364262cc378e79bcd147ed2d3d29659b17L145

That ineffectual assignment is because the `ctx` initiated in the last step is not used, the scope of that `ctx` without pointer is limited to the function body. 

But then let's change back to return this context specifically so that with the least surprise.

I tried to invite you as a reader role into the backlog project, but seems not possible, the backlog is configured as "Only those with access to this project can view it." This setting is unfortunately configured by the project owner, but I'll clarify if it's possible to have some exceptions.

jakobmoellerdev: Understood. Yeah I agree both options are kind of hacky, i think for me its simply a matter of dont have a pointer on interface. But you're right, the pointer in this instance would also work. I guess its just style, can also close the PR if you want.

Regarding the collaboration, are there any features you are willing to outsource and that external contributors can work on?

ruanxin: Of course, regarding our current progress, some topics in my mind which worth your attention:

We are enabling watcher atm, it's working but not been introduced to PROD yet, I am a bit concerned about its performance, it's just a single embedded server and can't scale, atm, it's probably fine to handle those requests, but will be problem when the coming request increased, btw, we introduced metric to monitoring it.

Another thing is regarding this [issue](https://github.com/kyma-project/lifecycle-manager/issues/836), it's also very valuable to have your opinion, basically, it needs to change current reconciliation logic, but I don't want to introduce much special logic to handle this situation, it should be generic enough.

One more thing appreciate your knowledge is regarding that ocm library, recently, I did an upgrade for go mod in this [PR](https://github.com/kyma-project/cli/pull/1758/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L34), as you can see, the ocm still not updated to the latest [0.3.0](https://github.com/open-component-model/ocm/releases/tag/v0.3.0), the reason is between 0.3.0-rc1 and 0.3.0, lots of breaking changes introduced, they completely rewrite some code, for example, to register repository, in [0.3.0](https://github.com/open-component-model/ocm/blob/v0.3.0/pkg/runtime/scheme.go#L440), it changes method signature to generic type, it also have other breaking changes introduced. I actually plan to have a concrete issue for this ocm upgrade, both in CLI and LM. If you can provide some insights for this upgrade or work on it, it will be great.

controllers/kyma_controller.go
```diff
@@ -125,8 +125,9 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 func (r *KymaReconciler) reconcile(ctx context.Context, kyma *v1beta2.Kyma) (ctrl.Result, error) {
 	if r.SyncKymaEnabled(kyma) {
 		remoteClient := remote.NewClientWithConfig(r.Client, r.KcpRestConfig)
-		if err := remote.InitializeSyncContext(&ctx, kyma,
-			r.RemoteSyncNamespace, remoteClient, r.RemoteClientCache); err != nil {
+		ctxWithSync, err := remote.InitializeSyncContext(ctx, kyma,
```
ruanxin, line 128: I'm thinking maybe it's better to relocate this `InitializeSyncContext` (renaming if needed) as a common one before passing into `reconcile(ctx context.Context, ...)`. And inside the function body, depends on if `SyncKymaEnabled`, return back context with syncContext or without, so that in this step, we don't need to deal with context anymore.

controllers/kyma_controller.go
ruanxin: I saw you already have a PR for the ocm upgrade 👍 
And thanks for the comments on the reconciliation issue, we can continue to discuss the issue there.
btw, I mentioned your PR within team on the daily(we are currently only half of the team size, @janmedrek is on vacation until 18.09), everyone is happy about you continuing to contribute to this project😄

ruanxin: and which open source Kyma slack channel?

jakobmoellerdev: > and which open source Kyma slack channel?

http://slack.kyma-project.io/ as referenced in https://github.com/kyma-project/community#slack

controllers/kyma_controller.go
jakobmoellerdev: Refactored to external method and added docs on special cases on the method.

janmedrek:
@jakobmoellerdev great to hear from you! I will try to set up the projects and the backlog in a more open-source-friendly way. 😄 

AFAIK you should still be a member of the organization, I will do the double-checks on that as well.

controllers/kyma_controller.go
```diff
@@ -148,16 +155,12 @@ func (r *KymaReconciler) getSyncedContext(ctx context.Context, kyma *v1beta2.Kym
...
-			logger.Info("access secret not found for kyma, assuming already deleted cluster")
-			return nil, r.removeAllFinalizersAndUpdateKyma(ctx, kyma)
-		}
 		if util.IsConnectionRefused(err) {
 			logger.Info("connection refused, assuming connection is invalid and resetting cache-entry for kyma")
 			r.RemoteClientCache.Del(client.ObjectKeyFromObject(kyma))
 		}
 		r.enqueueWarningEvent(kyma, syncContextError, err)
-		return nil, err
+		return ctx, err
```
Tomasz-Smelcerz-SAP, line 163: This doesn't follow the golang convention: In case of on non-nil error the other returned values should be considered unreliable by the caller. As such it should be documented. But I'd rather avoid it and just return `nil. err` here.

controllers/kyma_controller.go
```diff
@@ -148,16 +155,12 @@ func (r *KymaReconciler) getSyncedContext(ctx context.Context, kyma *v1beta2.Kym
 	ctxWithSync, err := remote.InitializeSyncContext(ctx, kyma,
 		r.RemoteSyncNamespace, remoteClient, r.RemoteClientCache)
 	if err != nil {
-		if !kyma.DeletionTimestamp.IsZero() && errors.Is(err, remote.ErrAccessSecretNotFound) {
```
Tomasz-Smelcerz-SAP: After moving this block, the function's docs must change alse

controllers/kyma_controller.go
```diff
@@ -148,16 +155,12 @@ func (r *KymaReconciler) getSyncedContext(ctx context.Context, kyma *v1beta2.Kym
...
-			logger.Info("access secret not found for kyma, assuming already deleted cluster")
-			return nil, r.removeAllFinalizersAndUpdateKyma(ctx, kyma)
-		}
 		if util.IsConnectionRefused(err) {
```
Tomasz-Smelcerz-SAP, line 158: I think this code goes agains "Single responsibility principle". The "getSyncedContext" function should be focused only on getting synced context, not on handling special cleanup cases, removing cache entries etc.
I think moving the `kyma.DeletionTimestamp.IsZero` away from this function is a good thing, and I'd also move **this** block out of it. Otherwise our logic is scattered across different functions and it's really hard to answer the question:
`What is controller logic in case of a remote connection problem?`

ruanxin: make sense, changed accordingly.

controllers/kyma_controller.go
ruanxin: updated func comment

controllers/kyma_controller.go
ruanxin: Yes, it's intended, if an error happens, the function should **return back original context**, otherwise, follow-up process will get nil pointer when it uses context.

I updated the function comment to emphasise this in case later we accidentally change it.

Tomasz-Smelcerz-SAP (approved):

Issue #841, opened 2023-09-04
Upgrade ocm dependencies in CLI


ruanxin:
At the moment, ocm dependencies in LM stays at [0.2.0](https://github.com/kyma-project/lifecycle-manager/blob/main/go.mod#L18) while in CLI are [0.3.0-rc1](https://github.com/kyma-project/cli/blob/main/go.mod#L34), but the latest version currently is already at [0.4.0](https://github.com/open-component-model/ocm/releases/tag/v0.4.0)

AC:
- [ ] update dependencies to the latest for LM and CLI

# Remark:
It's not a trivial task, there are breaking changes between those versions.

jakobmoellerdev:
Would like to tackle this change if possible. @janmedrek is that an option in your current planning?

ruanxin:
# DevNotes 

For CLI, to upgrade 0.3.0-rc1 to 0.3.0 will be complicated.

0.3.0 changed the way how [ComponentArchive](https://github.com/open-component-model/ocm/blob/v0.3.0/pkg/contexts/ocm/repositories/comparch/componentarchive.go#L37) initiated, it expected `ComponentDescriptor` data already presented in the local file system, and after initialized, the`NewComponentVersionAccessImpl` is immutable, which means `ComponentArchive` should be generated at last before push to remote.

However, CLI implementation [initialized](https://github.com/open-component-model/ocm/blob/v0.3.0-rc.1/pkg/contexts/ocm/repositories/comparch/componentarchive.go#L33) ComponentArchive first, then update ComponentDescriptor, this is impossible in 0.3.0

janmedrek:
@jakobmoellerdev feel free to pick it up. 🙂 

We just need to make sure that it either does not affect the Module teams or that we have a proper guide on what they need to do in order to migrate.

PR #843, opened 2023-09-04
chore!: Upgrade to ocm v0.3.0

jakobmoellerdev:
**Description**

- removes the legacy oci types
- upgrades to ocm v0.3.0
- introduces better context registrations for resolving types from moduleTemplate descriptor

WARNING: Introduces lots of dependencies in the graph that are probably not good, e.g. the previously removed helm is reintroduced in OCM. I believe OCM has tons of unnecessary deps for us which could be invoked as false positives in sec scannings. Still TBD how to get these down. Also had to pin docker version otherwise we would have a breaking go mod graph, still experimenting

WARNING: Should not be merged until all Module Templates are generated with OCM v3alpha1 scheme, as the v2 scheme still generates the old localOciBlob access type in the CLI. aligned with @ruanxin 

**Related issue(s)**
#841


ruanxin:
This PR was blocked until we fully migrated ModuleTemplate component descriptor from v2 to v3alpha1

jeremyharisch (approved):

PR #848, opened 2023-09-05
test: Extend manifest CR tests

Tomasz-Smelcerz-SAP:
**Description**

This PR extends existing tests for Manifest CR reconciliation. 
It adds one new suite that uses `CustomResourceReadyCheck` for state checking instead of a simplified implementation `ExistsReadyCheck`. Usage of `CustomResourceReadyCheck` should be preferred, because this is the one the production code is using.
Some improvement in timing configuration of the test suites is proposed. This should be fine-tuned in subsequent PRs, but a significant improvement in execution times is already observed.

- add checks for `oci-sync-ref` annotation
- add checks for `status.synced` resource list
- add checks if all resources are deleted when the Manifest CR is deleted
- add suite for Manifest CR following state transition in the Resource CR: `Ready` -> `Warning` -> `Ready`
- configure more aggresive timing for test suites (faster execution)

**Related issue(s)**
See: #702


controllers/manifest_controller/helper/helper.go
```diff
@@ -21,11 +25,22 @@ import (
...
 )
 
-const CredSecretLabelKeyForTest = "operator.kyma-project.io/oci-registry-cred" //nolint:gosec
+var (
+	Cancel                   context.CancelFunc
+	Ctx                      context.Context
+	K8sClient                client.Client
+	Server                   *httptest.Server
+	ErrManifestStateMisMatch = errors.New("ManifestState mismatch")
+	ManifestFilePath         = "../../pkg/test_samples/oci/rendered.yaml"
```
jeremyharisch, line 38: I think its rather an anti-pattern to have ctx, client, etc. as shared global variables between multiple test suits.
I would prefer to not to have it as global variables, but the suits them self should each have there own clients etc.

```suggestion
	ErrManifestStateMisMatch = errors.New("ManifestState mismatch")
	ManifestFilePath         = "../../pkg/test_samples/oci/rendered.yaml"
```

controllers/manifest_controller/helper/helper.go
jeremyharisch, line 1: I would keep the `test` somehow in the name of the file/pkg, since those are just utility functions used by test suits.

To have a seperated pkg for it makes sense, since those functions are used in multiple test suits - But preferably we should name it in a way that it makes clear it should only be used for testing, eventho it now need to go into the production code. (Sad but thats how Go works)

Please rename to something like `manifesttest`.


Examples how other projects solve this situation:
- https://golang.org/pkg/net/http/httptest/
-  https://godoc.org/go.uber.org/zap/zaptest

jeremyharisch (commented):
Test `Update Module Template version` is missing 🤔 

Left some other comments, rest looks fine

controllers/manifest_controller/helper/helper.go
```diff
@@ -21,11 +25,22 @@ import (
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/google/go-containerregistry/pkg/v1/types"
-	. "github.com/onsi/gomega"
+	gmg "github.com/onsi/gomega"
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Yes, the reason is it's not a `_test` package - it's the "normal" one - and the linter complains about dot imports. I could suppress the linter or use explicit package naming.

controllers/manifest_controller/helper/helper.go
Tomasz-Smelcerz-SAP: You are right. At the same time they are global variables in existing suites. This variables are necessary for the functions defined here. We have three options:
- duplicate the entire code between the suites, removing this file (I don't like it)
- have the common functions defined here, but without global variables - make them parameters to the functions (pollutes nice function API with low-level stuff like k8sClient)
- wrap all these params in some struct and make all the functions defined here a methods of this struct. Acceptable, but initially I considered it an over-engineering.

What's your opinion?

controllers/manifest_controller/helper/helper.go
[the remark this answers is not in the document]
jeremyharisch: Ok, got the point - but since this is anyway just for pure testing, lets suppress it I would say.
Please also check my other comment regards the file name

controllers/manifest_controller/helper/helper.go
jeremyharisch: I would go for the second option 👍🏻 So we could possibly use it in any combination of suits, without the need to think about race conditions

controllers/manifest_controller/helper/helper.go
Tomasz-Smelcerz-SAP: :+1:

jeremyharisch (approved):

PR #859, opened 2023-09-12
chore: Use GH-Action instead prow job for unit testing

jeremyharisch:
**Description**

- Use GH-Actions instead prow to run `make test`
- Prow job still needs to be cleaned up: https://github.com/kyma-project/test-infra/pull/8896
    - Way faster than prow (currently, test takes ~5min + waiting time for free runner)
- Use go 1.21 in GH-Actions, since KLM is written in 1.21

**Related issue(s)**

### DIFF
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -1,4 +1,4 @@
-name: TestSuite EnvTest and UnitTest
+name: TestSuite integration and unit tests
 
 on:
   pull_request:
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: '1.20'
+          go-version: '1.21'
       - name: Run 'make test'
         working-directory: ./lifecycle-manager
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/test-unit.yml
```diff
@@ -0,0 +1,22 @@
+name: TestSuite EnvTest and UnitTest
```
ruanxin, line 1:
```suggestion
name: TestSuite integration and unit tests
```

.github/workflows/test-unit.yml
```diff
@@ -0,0 +1,27 @@
...
+      - '**.go'
+jobs:
+  envtest-and-unittest:
+    name: "Run 'make test'"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout lifecycle-manager
+        uses: actions/checkout@v3
+        with:
+          path: ./lifecycle-manager/
```
lindnerby, line 19: Are these paths in `actions/checkout@v3` and `Run 'make test'` context should be this root already

jeremyharisch: For me personal it is more readable if we always mention the path, but that is personal preference :D I removed it

lindnerby (approved):

Issue #858, opened 2023-09-12
Too many restarts of skr-webhook


lindnerby:
Due to the `request-sidecar` container the `skr-webhook` pods on remotes are restarting very frequently, although not needed:
```
Containers:
  server:
    State:          Running
      Started:      Mon, 04 Sep 2023 12:27:28 +0200
    Ready:          True
    Restart Count:  0
  request-sidecar:
    State:          Running
      Started:      Tue, 12 Sep 2023 08:31:15 +0200
    Last State:     Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Tue, 12 Sep 2023 07:31:14 +0200
      Finished:     Tue, 12 Sep 2023 08:31:14 +0200
    Ready:          True
    Restart Count:  188
```
We need to investigate, it's roundabout restarting every hour and the sidecar has initial sleep of 60m.

ACs;
- [x] Double-check if side-car is necessary, remove if not needed
- [x] skr-webhook should not restart that often

ruanxin:
```
│ NAME↑            PF IMAGE                                                                           READY STATE    INIT   RESTARTS PROBES(L:R) CPU MEM CPU/R:L MEM/R:L %CPU/R %CPU/L MEM/R │
│ request-sidecar  ●  europe-docker.pkg.dev/kyma-project/prod/external/alpine:3.17.4                  true  Running  false        24 off:off       0   0  10:100  20:200      0      0     1 │
│ server           ●  europe-docker.pkg.dev/kyma-project/prod/runtime-watcher-skr:v20230824-f4459bad  true  Running  false         0 off:off       1  18  10:100  20:200     10      1    92 │
│                                                                                                                                                                                            │
│                                                                                                                                                                                            │

```
the restart happens only in `request-sidecar` container

lindnerby:
> ```
> │ NAME↑            PF IMAGE                                                                           READY STATE    INIT   RESTARTS PROBES(L:R) CPU MEM CPU/R:L MEM/R:L %CPU/R %CPU/L MEM/R │
> │ request-sidecar  ●  europe-docker.pkg.dev/kyma-project/prod/external/alpine:3.17.4                  true  Running  false        24 off:off       0   0  10:100  20:200      0      0     1 │
> │ server           ●  europe-docker.pkg.dev/kyma-project/prod/runtime-watcher-skr:v20230824-f4459bad  true  Running  false         0 off:off       1  18  10:100  20:200     10      1    92 │
[...]

PR #861, opened 2023-09-14
feat: Drop request-sidecar container

ruanxin:
This request-sidecar container do nothing but just sleep 60m then get killed and restart.

Related issue: https://github.com/kyma-project/lifecycle-manager/issues/858


lindnerby:
Let's also check if https://github.com/kyma-project/runtime-watcher/blob/main/runtime-watcher/internal/request.go and https://github.com/kyma-project/runtime-watcher/blob/16145a7dc6bfbbfb921662f239b901244995e790/runtime-watcher/internal/handler.go#L139 can be dropped

skr-webhook/resources.yaml
```diff
@@ -98,23 +96,6 @@ spec:
...
             - containerPort: 8443
               name: watcher-port
-        - name: request-sidecar
-          image: europe-docker.pkg.dev/kyma-project/prod/external/alpine:3.17.4
-          command:
-            - /bin/sh
-            - "-c"
-            - "sleep 60m"
-          imagePullPolicy: Always
-          volumeMounts:
```
lindnerby, line 108: You can also remove the volumeMount for the tmp path for the server container and the emptyDir volume itself

lindnerby (changes requested):

skr-webhook/resources.yaml
ruanxin: good catch 👍

lindnerby (approved):

Issue #734, opened 2023-07-18
Basic monitoring for PurgeReconciler in grafana dashboard


ruanxin:
Now we enabled PurgeReconciler in KCP, however, we don't have any monitoring dashboard configured for this reconcile.

AC
- [x] Expose metrics for Purge reconciler implemented by controler-runtime
- [x] Extend the KLM performance dashboard with the PurgeReconciler metrics (no need for new panels/diagrams)
- [x] All the default metrics visible on the dashboard
- [x] If it is not possible to distinguish the Purge Controller metrics to Kyma Controller metrics -> Implement custom critical metrics (i.e. Process time of purge controller)

janmedrek:
Reopening this one - we need to:
- check implementation for the `purge count` metric (it produces big numbers of purge counts for a small number of clusters)
- introduce improvements to the presentation of the `purge count` metric

PR #875, opened 2023-09-22
feat: Add purge-controller metrics

lindnerby:
**Description**

- Add purge-controller metrics
- Small refactorings
- Metrics visualization in dashboard as follow-up

**Related issue(s)**
* https://github.com/kyma-project/lifecycle-manager/issues/734


controllers/purge_controller.go
```diff
@@ -19,180 +19,195 @@ package controllers
...
 	logger := ctrlLog.FromContext(ctx)
 	logger.V(log.InfoLevel).Info("Purge Reconciliation started")
 
 	ctx = adapter.ContextWithRecorder(ctx, r.EventRecorder)
 
-	// check if kyma resource exists
 	kyma := &v1beta2.Kyma{}
 	if err := r.Get(ctx, req.NamespacedName, kyma); err != nil {
-		if !util.IsNotFound(err) {
+		if util.IsNotFound(err) {
```
ruanxin, line 62: Why change this logic? The condition here is if the err is not found, then this item should enqueue again, but if it's indeed not found, then stop enqueue.

controllers/purge_controller.go
```diff
@@ -19,180 +19,195 @@ package controllers
...
-			return ctrl.Result{Requeue: true}, nil
-		}
-
-		if err != nil {
 			return ctrl.Result{}, err
 		}
+		return ctrl.Result{Requeue: true}, nil
+	}
+	if err != nil {
+		return ctrl.Result{Requeue: true}, err
```
ruanxin, line 99:
```suggestion
		return ctrl.Result{}, err
```
If has err, it will requeue anyway

controllers/purge_controller.go
ruanxin, line 62:
```suggestion
		if !util.IsNotFound(err) {
```

lindnerby: Sure yes, reverted it. Thanks!

PR #876, opened 2023-09-22
chore(infra): Configure dependabot to adhere to the Conventional Commits spec

ameteiko:

- Configure Dependabot's dependency bumps to adhere to the Conventional Commits spec

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,8 +10,7 @@ updates:
     schedule:
       interval: "daily"
     commit-message:
-      prefix: "gomod"
-      include: "scope"
+      prefix: "chore(dependabot)"
   - package-ecosystem: "docker"
     directory: "/"
     labels:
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -10,8 +10,7 @@ updates:
     schedule:
       interval: "daily"
     commit-message:
-      prefix: "gomod"
-      include: "scope"
+      prefix: "chore(dependabot)"
```
nesmabadr, line 13: I'm confused now :sweat_smile: shouldn't this be `deps` not `chore`

ameteiko: Good remark.

My intent behind putting dependabot's bumps to `chore`, is that nothing from this category goes to the changelog (everything in this category is irrelevant for the consumers of our module).

`deps` on the other hand is the manifestation for all the changes in our external dependencies that can be informative for the consumers of our module. I think that minor dependabot's bumps should not go there.

What do you think?

nesmabadr: My argument was that in order to be consistent across all upgrades, but I get your point, makes sense :+1:

PR #879, opened 2023-09-25
fix: Delete Metrics Before Removing Finalizer

### LINKED ISSUE #700: Have a dedicated schedule process to check and remove non exists kyma metrics
**Description**
When Kyma CR is under deleting, LM will try to remove it in `handleDeletingState`, however, this just happens once, if `RemoveKymaStateMetrics` got error, there is no chance to remove the metrics again.

The idea here is, to create a dedicated schedule process to regularly check all existing kymas, and compared with kyma metrics, remove those not existing ones.

**Acceptance Criteria**
- [x] As proposed [here](https://github.com/kyma-project/lifecycle-manager/issues/700#issuecomment-1624345847) switch the statements to have a different order 
- [x] Write e2e test (use deletion e2e test)
    - [x] Create Kyma
    - [x] Check metrics Kyma count
    - [x] Remove Kyma
    - [x] Kyma count in metrics should go down again

# Reopen on 25.10.2023

The quick fix seems not working, we still got report the false positive alert gets triggered. We need to implement the proposal solution, to create a dedicated schedule process to regularly check all existing kymas, and compared with kyma metrics, remove those not existing ones.

LeelaChacha:
Closes #700

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -72,7 +72,7 @@ jobs:
           k3d cluster create skr -p 10080:80@loadbalancer -p 10443:443@loadbalancer --k3s-arg '--disable=traefik@server:0'
       - name: Provision KCP cluster
         run: |
-          kyma provision k3d --name=kcp -p 9080:80@loadbalancer -p 9443:443@loadbalancer --ci --registry-port 5111
+          kyma provision k3d --name=kcp -p 9080:80@loadbalancer -p 9443:443@loadbalancer -p 9081:8080@loadbalancer --ci --registry-port 5111
       - name: Update Kubeconfigs
         run: k3d kubeconfig merge -a -d
       - name: Export required Kubeconfig Env vars
@@ -150,6 +150,11 @@ jobs:
           kubectl config use-context k3d-kcp
           kubectl get crds
           kubectl apply -f template.yaml
+      - name: Expose Metrics Endpoint
+        working-directory: ./lifecycle-manager
+        if: ${{ matrix.e2e-test == 'test-kyma-deletion-e2e' }}
+        run: |
+          kubectl patch svc klm-metrics-service -p '{"spec": {"type": "LoadBalancer"}}' -n kcp-system
       - name: Run ${{ matrix.e2e-test }}
         working-directory: ./lifecycle-manager
         run: |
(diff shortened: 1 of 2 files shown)


tests/e2e_test/kyma_deletion_test.go
```diff
@@ -119,3 +123,93 @@ var _ = Describe("KCP Kyma CR should be deleted successfully when SKR cluster ge
...
+
+			By("should decrease the metric count")
+			metricCountAfterDeletion, err := getKymaStateMetricCount()
+			Expect(err).Should(Not(HaveOccurred()))
+			GinkgoWriter.Printf("Kyma State Metric count after CR deletion: %d", metricCountAfterDeletion)
+			Expect(metricCountAfterDeletion).To(Equal(metricCountBeforeDeletion - 1))
+		})
+	})
+
+func getKymaStateMetricCount() (int, error) {
```
ruanxin, line 187: It's better have this function to get each state count, basically this test should verify:
1. before kyma CR removed, the `Ready` state equals 1, 
2. after kyma CR removed, the `Deleting` state should not exists, equals 0

Then this function can be useful for other tests, to have a summary count is too generic.

ruanxin (changes requested):
good test case, just needs some enhancement.

.github/workflows/test-e2e.yml
```diff
@@ -72,7 +72,7 @@ jobs:
           k3d cluster create skr -p 10080:80@loadbalancer -p 10443:443@loadbalancer --k3s-arg '--disable=traefik@server:0'
       - name: Provision KCP cluster
         run: |
-          kyma provision k3d --name=kcp -p 9080:80@loadbalancer -p 9443:443@loadbalancer --ci --registry-port 5111
+          kyma provision k3d --name=kcp -p 9080:80@loadbalancer -p 9443:443@loadbalancer -p 9081:8080@loadbalancer --ci --registry-port 5111
```
ruanxin, line 75: can you just change this `9080:80` to `9080:8080`? I don't think expose `80` port is needed here.

tests/e2e_test/kyma_deletion_test.go
LeelaChacha: good thinking

.github/workflows/test-e2e.yml
LeelaChacha: The port 80 expose is needed for the communication between kcp and skr. It's advised as well in the developer docs. I"m not sure why though, since the watcher communicates over tls. but some e2e tests fail if this port is not open

ruanxin (approved):

Issue #391, opened 2023-02-14
Migrate kustomize setup to kustomize v5


jeremyharisch:
**Description**
The KLM is currently setup with kustomize v4. `5.0.0` was [released](https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv5.0.0) which introduces a lot of breaking changes. To stick to the latest changes, features and fixes we should think of upgrading to latest kustomize version.

Example breaking changes which at least affect this project:
- Dropped `patchesStrategicMerge`. Please use [patches](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/patches/) instead
- Dropped `patchesJson6902`. Please use [patches](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/patches/) instead
- Dropped `vars`. Please use [replacements](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/) instead.

Unfortunately, the new `patches` do not work as `patchesJson6902` did before. I.e. in kubebuilder v4 the render order of the kustomizations have been the following:
1. patches
2. global patches (i.e. `namespace: kcp-system`)
3. json patches

The following kustomization:
```yaml
namespace: GLOBAL-system

patches:
  - patch: |-
      - op: replace
        path: /metadata/namespace
        value: OLD-namespace
    target:
      group: networking.istio.io
      version: v1beta1
      kind: Gateway
      name: watcher-gateway

patchesJson6902:
  - patch: |-
      - op: replace
        path: /metadata/namespace
        value: JSON-namespace
    target:
      group: networking.istio.io
      version: v1beta1
      kind: Gateway
      name: watcher-gateway
```

Would have the following output:
```yaml
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: klm-watcher-serving-cert
  namespace: JSON-system
spec:
  <...>
```

Since JSON patches have been deprecated, the render order is the following:
1. patches
2. global patches (i.e. `namespace: kcp-system`)

Means the kustomization above, would render to the following with v5 using only `patches`:
```yaml
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: klm-watcher-serving-cert
  namespace: GLOBAL-system
spec:
  <...>
```

[...]

PR #887, opened 2023-09-26
feat: Upgrade kustomize to v5

Tomasz-Smelcerz-SAP:
**Description**

Upgrade to kustomize v5 and prepare for migration to Kustomization API version `v1`

Note: the `vars` configuration is also deprecated, but it is not migrated in this PR - on purpose. I wanted to avoid too many changes to the already complicated kustomize configurations we have. Once this PR is merged, `vars` will be migrated in the following PR.

- upgrade `kustomize` to `v5.1.1` and `controller-gen` to `0.13.0`
- migrate deprecated customizations: `patchesJson6902` and `patchesStrategicMerge` are moved to the `patches` configuration.
- ensure the files render in the same way as in the current `main` branch.

**Related issue(s)**
https://github.com/kyma-project/lifecycle-manager/issues/391


Makefile
```diff
@@ -116,7 +116,7 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
 
 .PHONY: deploy
-deploy: install manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
```
lindnerby, line 119: Can you check but I think `install` can also be removed from other targets, like `local-deploy-with-watcher`?

config/manager/kustomization.yaml
```diff
@@ -15,5 +15,4 @@ configMapGenerator:
 
 images:
 - name: controller
-  newName: europe-docker.pkg.dev/kyma-project/prod/lifecycle-manager
```
lindnerby, line 18: I don't understand this change, can you please explain how this still works?

lindnerby (commented):
Just left two questions. In general the changes look good of course.

config/manager/kustomization.yaml
Tomasz-Smelcerz-SAP: Good catch! This sneaked in from this Makefile command: `cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}`

Tomasz-Smelcerz-SAP:
Holding the PR until the changes are verified against control-plane kustomizations.

lindnerby (approved):

PR #911, opened 2023-10-04
fix: Make data as pointer to support module without default CR

ruanxin:
- configure moduletemplate `spec.data` as pointer so that generates empty field in case no default CR provided
- add e2e test to verify moduletemplate without default cr works
- upgrade go version in dockerfile
- refactor the existing test to better organize test utility code

bug reported in this issue: https://github.com/kyma-project/lifecycle-manager/issues/917

PR #912, opened 2023-10-04
feat(newReconcileOption): Introduce ModuleCRDeletionCheck + Non-Blocking Deletion for Modules

jeremyharisch:
**Description**

- Introduce new Option for a Custom deletion Check
- Deletion check for default CR in manifest reconciler
- New e2e test to check new custom deletion check of manifest reconciler

**Related issue(s)**


internal/manifest/cr_deletion_check.go
```diff
@@ -0,0 +1,48 @@
...
+	resourceCR := &unstructured.Unstructured{}
+	if err := clnt.Get(ctx,
+		client.ObjectKey{Name: name, Namespace: namespace}, resourceCR); err != nil {
+		if util.IsNotFound(err) {
+			return true, nil
+		}
+		return false, errors.Wrap(err, "failed to fetch default resource CR")
+	}
+
+	return false, clnt.Delete(ctx, resourceCR)
```
ruanxin, line 47: delete resource cr should not happens here, LM delete it in https://github.com/kyma-project/lifecycle-manager/blob/cc7c6e49a9da791f6d90188a63b6184abbf4b819/internal/manifest/custom_resource.go#L70

tests/e2e_test/module_deletion_test.go
```diff
@@ -0,0 +1,157 @@
...
+				"Deployment", runtimeClient).
+			Should(Succeed())
+		By("Checking if it is possible to update Sample CR")
+		Eventually(UpdateSampleCRSpec).
+			WithContext(ctx).
+			WithArguments("sample-yaml", "kyma-system", "newValue", runtimeClient).
+			Should(Succeed())
+	})
+
+	It("Should be able to re-enable module under deletion", func() {
```
ruanxin, line 92: In this senario, after re-enable, we need to verify the manifest CR is still in deleting state, and after sample CR finalizer removed, a new manifest CR created.

tests/e2e_test/module_deletion_test.go
```diff
@@ -0,0 +1,157 @@
...
+			WithArguments(defaultRemoteKymaName, remoteNamespace, "template-operator", "regular", runtimeClient).
+			Should(Succeed())
+		By("Removing all finalizers from sample CR")
+		Eventually(SetFinalizer).
+			WithContext(ctx).
+			WithArguments("sample-yaml", "kyma-system", "operator.kyma-project.io/v1alpha1", "Sample",
+				[]string{}, runtimeClient).
+			Should(Succeed())
+		By("New Sample CR should exists")
+		Eventually(CheckIfExists).
```
ruanxin, line 105: By this check, you can't assume this Sample CR is new one, so additionally, we need to check the deletiontimestamp not exists.

tests/e2e_test/module_deletion_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+import (
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+	"github.com/kyma-project/lifecycle-manager/pkg/testutils"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/api/meta"
+)
+
+var _ = Describe("Non Blocking Module Deletion", Ordered, func() {
```
ruanxin, line 11: I suggest we can already apply some BDD test patterns with this, I saw on CLI [PR](https://github.com/kyma-project/cli/pull/1792/files#diff-dae65c740f15475708a1e4a92f0d63e65dbd56ea50108646cf4194be51758a76R11), we introduced decent example already, maybe @ameteiko can help in this case also.

internal/declarative/v2/reconciler.go
```diff
@@ -403,15 +404,23 @@ func (r *Reconciler) doPreDelete(ctx context.Context, clnt Client, obj Object) e
...
 func (r *Reconciler) renderTargetResources(
 	ctx context.Context,
+	clnt client.Client,
 	converter ResourceToInfoConverter,
 	obj Object,
 	spec *Spec,
 ) ([]*resource.Info, error) {
-	if !obj.GetDeletionTimestamp().IsZero() {
-		// if we are deleting the resources,
+	if !obj.GetDeletionTimestamp().IsZero() && r.DeletionCheck != nil {
```
ruanxin, line 412: The initialization of this `DeletionCheck` is controlled by us, we just need to make sure it's getting init in declarative.WithCustomDeletionCheck, then we don't need to check this `nil` case, and will make the logic simplified.

internal/declarative/v2/reconciler.go
```diff
@@ -403,15 +404,23 @@ func (r *Reconciler) doPreDelete(ctx context.Context, clnt Client, obj Object) e
...
+		deleted, err := r.DeletionCheck.Run(ctx, clnt, obj)
+		if err != nil {
+			return nil, err
+		}
+		if deleted {
+			return ResourceList{}, nil
+		}
+	} else if !obj.GetDeletionTimestamp().IsZero() {
```
ruanxin, line 420: so we don't need this else if case

internal/manifest/cr_deletion_check.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// NewCustomResourceDeletionCheck creates a check that verifies that the Resource CR in the remote cluster is deleted.
+func NewCustomResourceDeletionCheck() *CRDeletionCheck {
+	return &CRDeletionCheck{}
+}
+
+type CRDeletionCheck struct{}
+
+func (c *CRDeletionCheck) Run(
```
ruanxin, line 22: Please think about some consistent name, now it has `NewCustomResourceDeletionCheck` `CRDeletionCheck`, `WithCustomDeletionCheck`, `DeletionCheck` ... we can use one name.

ruanxin: Among them, `CustomDeletionCheck` is misleading to me, basically, it's check if moduleCR gets deleted, please mention moduleCR at least.

internal/manifest/cr_deletion_check.go
```diff
@@ -0,0 +1,51 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// NewCustomResourceDeletionCheck creates a check that verifies that the Resource CR in the remote cluster is deleted.
+func NewCustomResourceDeletionCheck() *CRDeletionCheck {
+	return &CRDeletionCheck{}
+}
+
+type CRDeletionCheck struct{}
```
ruanxin, line 20: You can consider put `context`, `client` as struct field, so that not just an empty struct, and then `Run` just need 
 to pass one `Object` param

internal/manifest/cr_deletion_check.go
```diff
@@ -0,0 +1,51 @@
...
+	resourceCR := &unstructured.Unstructured{Object: map[string]interface{}{
+		"apiVersion": gvk.GroupVersion().String(),
+		"kind":       gvk.Kind,
+	}}
+	if err := clnt.Get(ctx,
+		client.ObjectKey{Name: name, Namespace: namespace}, resourceCR); err != nil {
+		if util.IsNotFound(err) {
+			return true, nil
+		}
+		return false, errors.Wrap(err, "failed to fetch default resource CR")
```
ruanxin, line 48: please don't use errors.Wrap, instead use `fmt.Errorf`

go.mod
```diff
@@ -32,6 +32,8 @@ require (
 )
 
 require (
+	github.com/kyma-project/template-operator v0.0.0-20230925081606-4ae25406be03
+	github.com/pkg/errors v0.9.1
```
ruanxin, line 36: also please don't introduce this errors as direct dependencies, it's a deprecated library since go lang natively introduced fmt.Errorf

ruanxin (dismissed):
Please check the comments

internal/manifest/cr_deletion_check.go
jeremyharisch: I disagree with mentioning at least module CR. Since this is an option in the declarative library and thus should be adaptable to some custom logic and not specific to only modules.

If we would remove/refactor the declarative library, we could get rid of all the options, because we anyway will only use it for the manifest reconciler.

But for now we are still sticking to it and I would not include module CR in the option name.
You can have a look at the other options, they are also named like this and the function passed into the options have the specific naming.

But I agree to have a consistent naming at least.

go.mod
```diff
@@ -32,6 +32,8 @@ require (
 )
 
 require (
+	github.com/kyma-project/template-operator v0.0.0-20230925081606-4ae25406be03
```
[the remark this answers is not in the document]
jeremyharisch: I use it in e2e testing to keep the code as simple as possible, is there a way I can exclude this fomr go.mod since it is just used in test files?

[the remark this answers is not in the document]
ruanxin: no good way to exclude it, otherwise you need to design this test as a dedicate go module, and then you can create new go.mod for this testing module, but it's not a practical way.

Anyway, template-operator is not needed, the only benefit you can get is to directly reference some API code, like the typed Sample CR, but you can also created it with `Unstructured{}`.

Instead, introducing it brings lots of dependencies problems, for example, both of them are operator, we may introduce different controller-runtime versions, think about later controller-runtime have breaking changes, you may blocked LM to be adapted smoothly because the template-operator still using old version.

We need to really be care to avoid introducing unnecessary dependencies.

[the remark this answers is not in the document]
ruanxin: But additionally, if we can refactor template operator API as a dedicated go module, then we can add it with introduce fewer dependencies. That's also one [idea](https://github.com/orgs/kyma-project/projects/30/views/13?pane=issue&itemId=38579080) I'm plan to do it for LM api.

internal/manifest/cr_deletion_check.go
ruanxin: As we discussed, we have a dedicated session for the declarative library.

And for the naming issue, I understand your concern, but still, if you look at the usage of this function `CustomDeletionCheck` - `options.DeletionCheck`, it's used in the `renderTargetResources` resources only, and the obj passing to it is manifest CR, in this case, it's not an option can be used anywhere. so when you decide the naming of this option, this should be aligned.

internal/manifest/cr_deletion_check.go
jeremyharisch: The problem here is when we init the options here, we do not have the runtime client nor the context, thus I would leave it as it is - Otherwise we would need to do a bigger refactoring - but since we plan to get rid of declarative in general, this refactoring is obsolite.

.github/workflows/test-e2e.yml
```diff
@@ -144,11 +144,20 @@ jobs:
...
-        if: ${{ matrix.e2e-test == 'module-status-propagation' }}
+        if: ${{ matrix.e2e-test == 'module-status-propagation' || matrix.e2e-test == 'non-blocking-deletion' }}
         with:
           repository: kyma-project/template-operator
           path: ./template-operator/
       - name: Create Template Operator Module
+        working-directory: ./template-operator
+        if: ${{ matrix.e2e-test == 'non-blocking-deletion' }}
+        run: |
+          kyma alpha create module --kubebuilder-project  --channel=regular --name kyma.project.io/module/template-operator --version 1.1.1 --path . --registry k3d-kcp-registry:5111 --insecure --module-archive-version-overwrite /
```
ruanxin, line 155: please change the create module to module config file instead of the deprecated kubebuilder-project. 

you can have a reference how it's doing in this [PR](https://github.com/kyma-project/lifecycle-manager/pull/911/files#diff-976614657a9537dd6476a2977a4a11c0ac912e262006a4ca2d4fe0b4d8cf0becR179)
or in [CLI](https://github.com/kyma-project/cli/blob/main/.github/workflows/test-e2e-create-module.yml#L74-L81)

tests/e2e_test/module_deletion_test.go
```diff
@@ -0,0 +1,187 @@
...
+	kyma := testutils.NewKymaWithSyncLabel("kyma-sample", "kcp-system", "regular",
+		v1beta2.SyncStrategyLocalSecret)
+	GinkgoWriter.Printf("kyma before create %v\n", kyma)
+
+	BeforeAll(func() {
```
ruanxin, line 17: This `beforeall` is not need, if KymaCRD is not installed, if it happens, in later `Eventually` when you create kyma CR it will fail and continue trying to create.

I already refactoring the shared BeforeAll and AfterAll statements in this [PR](https://github.com/kyma-project/lifecycle-manager/pull/911/files#diff-996d9d42109a1ee29b73f16aaea9f398b2a2a374f62f47b6ba881f845be21c57R16-R17), so please do not introduce more in.

internal/manifest/cr_deletion_check.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	if manifest.Spec.Resource == nil {
+		return true, nil
+	}
+
+	name := manifest.Spec.Resource.GetName()
+	namespace := manifest.Spec.Resource.GetNamespace()
+	gvk := manifest.Spec.Resource.GroupVersionKind()
+
+	resourceCR := &unstructured.Unstructured{Object: map[string]interface{}{
```
ruanxin, line 38: you can use `resourceCR.SetGroupVersionKind(gvk)` afterwards, instead of defining it as an untyped Object.

ruanxin: if you find other places in code base also define gvk like this, you can also refactor in this PR.

tests/e2e_test/module_deletion_test.go
```diff
@@ -0,0 +1,187 @@
...
+				WithContext(ctx).
+				WithArguments(kyma.GetName(), kyma.GetNamespace(), controlPlaneClient, v1beta2.StateProcessing).
+				Should(Succeed())
+			By("And Manifest CR is in state `Deleting`")
+			Eventually(CheckManifestIsInState).
+				WithContext(ctx).
+				WithArguments(kyma.GetName(), "template-operator", controlPlaneClient, v2.StateDeleting).
+				Should(Succeed())
+			By("And Sample CR is not removed")
+			Eventually(CheckIfExists).
```
ruanxin, line 83:
```suggestion
			Consistently(CheckIfExists).
```

tests/e2e_test/module_deletion_test.go
```diff
@@ -0,0 +1,187 @@
...
+				WithContext(ctx).
+				WithArguments("sample-yaml", "kyma-system", "operator.kyma-project.io/v1alpha1", "Sample",
+					runtimeClient).
+				Should(Succeed())
+
+			By("And the template-operator manager Deployment is not removed on runtime cluster")
+			Eventually(CheckIfExists).
```
ruanxin, line 90:
```suggestion
			Consistently(CheckIfExists).
```

tests/e2e_test/module_deletion_test.go
ruanxin: https://github.com/kyma-project/lifecycle-manager/pull/912#discussion_r1350506044
the second point I mentioned there is not covered, SampleCR should verify in `Deleting` state and deletiontimestamp not zero

tests/e2e_test/module_deletion_test.go
```diff
@@ -0,0 +1,187 @@
...
+				WithContext(ctx).
+				WithArguments("template-operator-controller-manager", "template-operator-system", "apps/v1",
+					"Deployment", runtimeClient).
+				Should(Succeed())
+			By("And it is possible to update the Sample CR in the runtime cluster")
```
ruanxin, line 95: This check to be honest, is not necessary.

If a resource is under delete, normally the user will not update spec, and the operator also should not design to react to any spec change when resource is under delete.

tests/e2e_test/module_deletion_test.go
```diff
@@ -0,0 +1,187 @@
...
+				WithContext(ctx).
+				WithArguments(defaultRemoteKymaName, remoteNamespace, "template-operator", "regular", runtimeClient).
+				Should(Succeed())
+			By("Then the Manifest CR is still in Deleting state")
+			Eventually(CheckManifestIsInState).
+				WithContext(ctx).
+				WithArguments(kyma.GetName(), "template-operator", controlPlaneClient, v2.StateDeleting).
+				Should(Succeed())
+			By("And the Sample CR is still in Deleting state")
+			Eventually(CheckSampleCRIsInState).
```
ruanxin, line 113:
```suggestion
			Consistently(CheckSampleCRIsInState).
```

internal/manifest/cr_deletion_check.go
ruanxin: make sense

tests/e2e_test/utils_test.go
```diff
@@ -202,3 +291,67 @@ func GetKymaStateMetricCount(ctx context.Context, kymaName, state string) (int,
...
+	}
+
+	if stateFromCR != expectedState {
+		return fmt.Errorf("%w: expect %s, but in %s",
+			errSampleCrNotInExpectedState, expectedState, stateFromCR)
+	}
+	return nil
+}
+
+func SampleCRNoDeletionTimeStampSet(ctx context.Context, name, namespace string, clnt client.Client) error {
```
ruanxin, line 337: Please consider 
1. refactor this as a more generic function to something like `CRExistsWithoutDeletionTimestamp`, this can be reused in other places.
2. or extract internal function body as a generic one and put it under `testutil` pkg, which accepts all object, not only SampleCR.

`CheckIfExists` can be reuse here, then only need to additionally check if deletionTimestamp exists.

tests/e2e_test/utils_test.go
```diff
@@ -202,3 +291,67 @@ func GetKymaStateMetricCount(ctx context.Context, kymaName, state string) (int,
...
+	}
+
+	if err := unstructured.SetNestedField(sampleCR.Object, resourceFilePath, "spec", "ResourceFilePath"); err != nil {
+		return err
+	}
+
+	return clnt.Update(ctx, sampleCR)
+}
+
+func CheckSampleCRIsInState(ctx context.Context, name, namespace string, clnt client.Client,
```
ruanxin, line 312: the internal check state logic you can reuse for other resources, including those typed one, like manifestCR, not only SampleCR.

Then you can reuse it for `CheckManifestIsInState`, `CheckKymaIsInState`

ruanxin: please also refactor and align this [IsKymaInState](https://github.com/kyma-project/lifecycle-manager/blob/174363eac89b6558c3393908fe2baaefff8bc17f/pkg/testutils/kyma.go#L111) with your `CheckKymaIsInState`

tests/e2e_test/utils_test.go
ruanxin: `ManifestNoDeletionTimeStampSet` can also share this code, what you need to do is convert typed manifest to unstructured use `runtime.DefaultUnstructuredConverter.ToUnstructured`

internal/manifest/custom_resource.go
```diff
@@ -85,10 +85,6 @@ func PreDeleteDeleteCR(
 		return ErrWaitingForAsyncCustomResourceDeletion
```
ruanxin, line 85: This function needs to be refactored further, the original description in the issue is not correct.

The block you need to remove is not `if !util.IsNotFound(err)` but this `if err == nil`

This function should: 
1. delete cr
2. if cr not found, further delete related CRD.
3. if CRD not found, remove manifest CR finalizer

So we can remove this wired `ErrWaitingForAsyncCustomResourceDeletion` when `err == nil`, as well as `ErrWaitingForAsyncCustomResourceDefinitionDeletion`

With the current change, basically this `PreDeleteDeleteCR` still block diff resource from being deleted. But due to you haven't covered this scenario in the test, it's not detected.

What you missing test is basically a module upgrade test when module CR is disabled, say we have v1.0.0 module contains resources A, B, C, and in upgrade v1.0.1 contains resources A, B, D. in this case, C will be deleted in this [step](https://github.com/kyma-project/lifecycle-manager/blob/c16f84e586922238616079ae78b58fc034393a01/internal/declarative/v2/reconciler.go#L482), but due to this module CR always return `ErrWaitingForAsyncCustomResourceDeletion`, it will prevent that upgrade happens.

So the test case should be designed like that
1. After module disabled.
2. apply a new moduletemplate with new version
3. verify all the resources in newer version get deployed.

For this test, we can use the moduletemplate within smoke test for module upgrade test [here](https://github.com/kyma-project/lifecycle-manager/blob/main/.github/workflows/test-smoke.yml#L150-L151)

ruanxin (changes requested):
some code refactoring, more important is the comment mentioned here

tests/e2e_test/utils_test.go
jeremyharisch: Refactored it, but I disagree with `ManifestNoDeletionTimeStampSet`, since we never know the exact name of the Manifest CR. The `ManifestNoDeletionTimeStampSet ` list all manifest and gets the correct one based on the naming schema (contains kymaName and moduleName). Then we could fetch the manifest again, becaise then we have the name and namespace or we just check the DeletionTimeStamp on the already fetched object from the manifestlist, since this includes the deletiontimestamp.

tests/e2e_test/utils_test.go
jeremyharisch: Check Kyma will be refactored

tests/e2e_test/module_deletion_test.go
ruanxin: We clarified the description of the original AC is misleading, Scenario 1: Test module CR still possible to be updated during deleting state, it's not mean test module CR update, but test module resources update.

tests/e2e_test/utils_test.go
ruanxin: First, we can get manifest CR name from `kyma.status.modules.manifests`, you can reuse this [`GetManifest`](https://github.com/kyma-project/lifecycle-manager/pull/911/files#diff-fe4282c0e1f2cefc83a319710b3e8db61f6f40a3b927d946357e786eed6c2e6eR32) after it merged.

Second, this does not prevent you from reusing `CRExistsWithoutDeletionTimestamp`, after you get a manifest CR, no matter which way mentioned above, what you need to do is just convert it to unstructured, then you use shared code.

pseudo code:
```
ManifestNoDeletionTimeStampSet() {
 manifest, err := getmanifest();
 manifestUnstructured := runtime.DefaultUnstructuredConverter.ToUnstructured(manifest)
 CRExistsWithoutDeletionTimestamp(manifestUnstructured)
}
```

tests/e2e_test/utils_test.go
ruanxin: as mentioned [here](https://github.com/kyma-project/lifecycle-manager/pull/912#discussion_r1357156505), no matter which way you get manifest CR, then you already have manifest name, namespace, gvk, what prevents you reuse code?

basically, for this case, you can also reuse `CRIsInState` in `CheckManifestIsInState`

internal/manifest/custom_resource.go
```diff
@@ -81,12 +81,9 @@ func PreDeleteDeleteCR(
 	resource := manifest.Spec.Resource.DeepCopy()
 	propagation := v1.DeletePropagationBackground
 	err := skr.Delete(ctx, resource, &client.DeleteOptions{PropagationPolicy: &propagation})
-	if err == nil {
-		return ErrWaitingForAsyncCustomResourceDeletion
-	}
 
 	if !util.IsNotFound(err) {
-		return fmt.Errorf("PreDeleteDeleteCR: %w", err)
+		return nil
```
ruanxin, line 86: in later step, CRD delete should also refactored like this

.github/workflows/test-e2e.yml
```diff
@@ -144,11 +144,20 @@ jobs:
...
-        if: ${{ matrix.e2e-test == 'module-status-propagation' }}
+        if: ${{ matrix.e2e-test == 'module-status-propagation' || matrix.e2e-test == 'non-blocking-deletion' }}
         with:
           repository: kyma-project/template-operator
           path: ./template-operator/
       - name: Create Template Operator Module
+        working-directory: ./template-operator
+        if: ${{ matrix.e2e-test == 'non-blocking-deletion' }}
+        run: |
+          kyma alpha create module --module-config-file ./module-config-no-cr.yaml --path . --registry k3d-kcp-registry:5111 --insecure --name kyma-project.io/module/template-operator
```
ameteiko, line 155: I wonder if we should create a subdirectory `/tests/e2e_test/scripts` in our project, create files with shell scripts there and then reference those scripts in here. For example:

```yml
- name: Create Template Operator Module
        working-directory: ./template-operator
        if: ${{ matrix.e2e-test == 'non-blocking-deletion' }}
        run: |
            $PATH_TO_SCRIPTS/deploy_module.sh
```

@ruanxin what do you think?

pkg/testutils/kyma.go
```diff
@@ -83,12 +83,13 @@ func GetKyma(ctx context.Context, testClient client.Client, name, namespace stri
 	return kymaInCluster, nil
```
ameteiko, line 83: @ruanxin, I suggest moving this package to the `/tests/testkit` or `/tests/testutils` (the former one is the convention used by Huskies, so if we adhere to it, it can become a standard in Kyma).

ameteiko (changes requested):

.github/workflows/test-e2e.yml
ruanxin: make sense, and especially if the script can be reused, for example, here this `deploy_module.sh` should be possible to accept some params, so that it can be reused in followup "Create Template Operator Module with final state `Warning`" step.

We can either progress this refactoring here or create a followup issue to address this, it is up to you @jeremyharisch .

pkg/testutils/kyma.go
ruanxin: For me, putting those utility functions under `pkg` or `tests` is both valid, if you reference to this golang-standards layout, you can find some examples.

for example:
under `pkg` https://github.com/containerd/containerd/tree/main/pkg/testutil
under `test` https://github.com/openshift/origin/tree/master/test/extended/util

but maybe we should rename `/tests` to `/test`, the latter one is the standard name defined in that layout.

regarding the name of `testkit`, IMO, it's not a good name, the functionality and coverage of `kit` are much heavier than `utility` to me, maybe Kuskies need to define some complicated logic to fulfill some test cases, but for us, the function in current package just some lightweight create test data, operator resources in cluster. Another concern, you can also observe in other good structured repo mentioned in that standard layout, most of them use `util` package. 

We can bring this topic to the team, after having decision, create an issue to address this specifically, it's not relevant to this PR, and will make the structure hard to review.

ameteiko: The reasoning behind the `testkit` name is the readability (subjective) of the test files:

```go

testkit.CRIsInState(...) /* vs */ utility.CRIsInState(...)

// or 

testkit.CreateNamespace(...) /* vs */ utility.CreateNamespace(...)
```

`testkit` here clearly manifests that a test-related package is involved, and in the case of the `utility`, it is ambiguous if the utility is a part of the application domain or a test one.

So, we need to brainstorm the API for this package and find a proper name and location.

.github/workflows/test-e2e.yml
jeremyharisch: I like the idea, I will create an issue for that - I do not want to blow  this PR up even up

jeremyharisch: Follow-up issue https://github.com/kyma-project/lifecycle-manager/issues/945

internal/manifest/custom_resource.go
jeremyharisch: I will resolve the missing testcase in a follow-up PR

tests/e2e_test/module_deletion_test.go
```diff
@@ -0,0 +1,180 @@
...
+		It("When a KCP Kyma CR is created on the KCP cluster", func() {
+			Eventually(CreateKymaSecret).
+				WithContext(ctx).
+				WithArguments(kyma.GetName(), kyma.GetNamespace(), controlPlaneClient).
+				Should(Succeed())
+			Eventually(controlPlaneClient.Create).
+				WithContext(ctx).
+				WithArguments(kyma).
+				Should(Succeed())
+			By("Then the Kyma CR is a \"Ready\" state on the KCP cluster ")
```
ameteiko, line 26:
```suggestion
			By("Then the Kyma CR is in a \"Ready\" State on the KCP cluster ")
```

tests/e2e_test/module_deletion_test.go
```diff
@@ -0,0 +1,180 @@
...
+			Eventually(testutils.IsKymaInState).
+				WithContext(ctx).
+				WithArguments(kyma.GetName(), kyma.GetNamespace(), controlPlaneClient, v1beta2.StateReady).
+				Should(Succeed())
+			By("And the Kyma CR is in \"Ready\" in the SKR cluster")
```
ameteiko, line 31:
```suggestion
			By("And the Kyma CR is in \"Ready\" State on the SKR cluster")
```

tests/e2e_test/module_deletion_test.go
```diff
@@ -0,0 +1,180 @@
...
+			Eventually(SetFinalizer).
+				WithContext(ctx).
+				WithArguments("sample-yaml", "kyma-system", "operator.kyma-project.io", "v1alpha1", "Sample",
+					[]string{"sample.kyma-project.io/finalizer", "blocking-finalizer"}, runtimeClient).
+				Should(Succeed())
+			Eventually(DisableModule).
+				WithContext(ctx).
+				WithArguments(defaultRemoteKymaName, remoteNamespace, "template-operator", runtimeClient).
+				Should(Succeed())
+			By(" Then the KCP Kyma CR is in a \"Processing\" State")
```
ameteiko, line 65:
```suggestion
			By("Then the KCP Kyma CR is in a \"Processing\" State")
```

tests/e2e_test/module_deletion_test.go
```diff
@@ -0,0 +1,180 @@
...
+				Should(Succeed())
+		})
+
+		It("Given the Kyma Module is under deletion", func() {
+			By("When the Module is re-enabled")
+			Eventually(EnableModule).
+				WithContext(ctx).
+				WithArguments(defaultRemoteKymaName, remoteNamespace, "template-operator", "fast", runtimeClient).
+				Should(Succeed())
+			By("Then the Module Manifest CR is still in a and in state \"Deleting\" State")
```
ameteiko, line 105:
```suggestion
			By("Then the Module Manifest CR is still in a \"Deleting\" State")
```

tests/e2e_test/module_deletion_test.go
```diff
@@ -0,0 +1,180 @@
...
+			Consistently(CheckManifestIsInState).
+				WithContext(ctx).
+				WithArguments(kyma.GetName(), kyma.GetNamespace(), "template-operator", controlPlaneClient, v2.StateDeleting).
+				Should(Succeed())
+			By("And the Module Default CR is in Deleting state")
```
ameteiko, line 110:
```suggestion
			By("And the Module Default CR is in a \"Deleting\" State")
```

tests/e2e_test/module_deletion_test.go
```diff
@@ -0,0 +1,180 @@
...
+			Eventually(CheckIfExists).
+				WithContext(ctx).
+				WithArguments("sample-yaml", "kyma-system", "operator.kyma-project.io", "v1alpha1", "Sample",
+					runtimeClient).
+				Should(Succeed())
+			Eventually(SampleCRNoDeletionTimeStampSet).
+				WithContext(ctx).
+				WithArguments("sample-yaml", "kyma-system", runtimeClient).
+				Should(Succeed())
+			By("And a new Manifest CR is created.")
```
ameteiko, line 133:
```suggestion
			By("And a new Manifest CR is created")
```

tests/e2e_test/module_deletion_test.go
```diff
@@ -0,0 +1,180 @@
...
+			Eventually(CheckManifestIsInState).
+				WithContext(ctx).
+				WithArguments(kyma.GetName(), kyma.GetNamespace(), "template-operator", controlPlaneClient, v2.StateReady).
+				Should(Succeed())
+			Eventually(ManifestNoDeletionTimeStampSet).
+				WithContext(ctx).
+				WithArguments(kyma.GetName(), kyma.GetNamespace(), "template-operator", controlPlaneClient).
+				Should(Succeed())
+			By("And KCP Kyma CR is in a Ready State")
```
ameteiko, line 142:
```suggestion
			By("And KCP Kyma CR is in a \"Ready\" State")
```

ameteiko (changes requested):

pkg/testutils/kyma.go
ruanxin: regarding this concern, actually from my perperctive, there is no much sense have a name when import test package, that's why I choose this way to [implicit](https://github.com/kyma-project/lifecycle-manager/blob/main/tests/e2e_test/kyma_deletion_test.go#L7) import test package.

And yes, regarding the rest, let's have a dedicate session discuss about it, if you can manage this discussion, it will be great! @ameteiko

ameteiko (approved):

tests/e2e_test/utils_test.go
```diff
@@ -378,3 +407,40 @@ func CheckKymaModuleIsInState(ctx context.Context,
...
+	return CRIsInState(ctx,
+		"operator.kyma-project.io", "v1alpha1", "Sample",
+		name, namespace,
+		[]string{"status", "state"},
+		clnt,
+		expectedState)
+}
+
+func SampleCRNoDeletionTimeStampSet(ctx context.Context, name, namespace string, clnt client.Client) error {
+	_, exists, err := GetDeletionTimeStamp(ctx, "operator.kyma-project.io", "v1alpha1",
```
ruanxin, line 423: Is there any purpose return this deletiontimestamp? no real usage

.github/workflows/test-e2e.yml
```diff
@@ -143,7 +143,7 @@ jobs:
           done
       - name: Checkout template-operator
         uses: actions/checkout@v3
-        if: ${{ matrix.e2e-test == 'module-status-propagation' || matrix.e2e-test == 'module-without-default-cr' || matrix.e2e-test == 'non-blocking-deletion'}}
+        if: ${{ matrix.e2e-test == 'module-status-propagation' || 'module-without-default-cr' || 'non-blocking-deletion' }}
```
ruanxin, line 146: this if condition is not working, you need that matrix.e2e-test == 'module-without-default-cr', otherwise, it just means `if 'module-without-default-cr'` which will always `true`, so basically means this step will run in each e2e test

tests/e2e_test/module_deletion_test.go
```diff
@@ -0,0 +1,180 @@
+package e2e_test
+
+import (
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+	v2 "github.com/kyma-project/lifecycle-manager/internal/declarative/v2"
+	"github.com/kyma-project/lifecycle-manager/pkg/testutils"
```
ruanxin, line 6: you can import this `testutils` implicitly, to avoid use function with the package name, anyway, the testutils are used for testing.
```suggestion
	. "github.com/kyma-project/lifecycle-manager/pkg/testutils"
```

tests/e2e_test/utils_test.go
jeremyharisch: Because I implemented it in the utils packge adn can be reused in future, if you think that is not neccessary, then I can remove it

.github/workflows/test-e2e.yml
jeremyharisch: Ups, leftover from merge conflict

tests/e2e_test/utils_test.go
```diff
@@ -378,3 +407,40 @@ func CheckKymaModuleIsInState(ctx context.Context,
...
+		return err
+	}
+
+	if exists {
+		return errSampleCRDeletionTimestampSet
+	}
+	return nil
+}
+
+func SampleCRDeletionTimeStampSet(ctx context.Context, name, namespace string, clnt client.Client) error {
```
ruanxin, line 435: you can consider put this function under `testutils/modulecr.go`

tests/e2e_test/utils_test.go
ruanxin: let's remove it, two reasons:
1. to check deletiontimestamp is not so usual, normally just need to verify if it exits.
2. avoid overrengineer, when we need, we can refactor this later.

ruanxin (approved):

ameteiko (approved):

PR #916, opened 2023-10-05
fix: Not caching descriptor for remote modules

### LINKED ISSUE #831: Do not cache descriptors for remote modules
**Description**

For performance, we cached component descriptors to avoid repeatedly unmarshal oci information. 

This is also the case for remote ModuleTemplates that are deployed in the SKRs and not present in the KCP, it causes potential problems and exposes the module to other clusters (which is not the desired state).

**Acceptance Criteria**

- [x] ModuleTemplate created in SKR cluster should not be cached (ones referenced with `remoteModuleTemplateRef`)
- [x] Implement envtest to check cache-key is not related to remote modules

**Implementation Hint**
https://github.com/kyma-project/lifecycle-manager/blob/main/api/v1beta2/moduletemplate_types.go#L186

nesmabadr:
**Description**

- Not using the descriptor cache for remote ModuleTemplates (based on the remoteModuleTemplateRef)

**Related issue(s)**
Fixes #831


api/v1beta2/moduletemplate_types.go
```diff
@@ -118,7 +118,7 @@ type CustomStateCheck struct {
 	MappedState State `json:"mappedState" yaml:"mappedState"`
 }
 
-func (m *ModuleTemplate) GetDescriptor() (*Descriptor, error) {
+func (m *ModuleTemplate) GetDescriptor(useCache bool) (*Descriptor, error) {
```
ruanxin, line 121: I suggest change this function to read only mode but no `SetDescToCache`, then we can git rid of this `useCache`, by your change, it makes the logic no so clear.

So for this function:
1. read from cache
2. if no result from cache, fallback to reading from `Descriptor.Raw`

ruanxin: Then you only need to set cache in this [condition](https://github.com/kyma-project/lifecycle-manager/pull/916/files#diff-fee2b3ec33be5b4b5a1396a28a16914df3bd0c257297d091d3174c900976e02cR48) for non remote module template

ruanxin: By this refactoring, I believe you can further clean up that  `IsRemoteModuleTemplate` field you introduced in `ModuleTemplateTO`

ruanxin (changes requested):
Please check my comment

api/v1beta2/moduletemplate_types.go
nesmabadr: I need to save the descriptor to cache in `CheckValidTemplateUpdate` but I don't have access to the module so either I call this function directly from the `GetTemplates` and pass if the module template is remote as an argument or keep it in the `ModuleTemplateTO` (current implementation after I made the GetDescriptor as Read-only). WDYT?

pkg/channel/lookup.go
```diff
@@ -328,3 +329,18 @@ func NewMoreThanOneTemplateCandidateErr(component v1beta2.Module,
...
 		ErrTemplateNotIdentified, component.Name, candidates)
 }
+
+func saveDescriptorToCache(template ModuleTemplateTO) {
+	if template.ModuleTemplate != nil {
+		descriptor := template.GetDescFromCache()
+		if descriptor == nil {
+			var err error
+			descriptor, err = template.GetDescriptor()
+			if err != nil {
```
ruanxin, line 339: please do not handle `template.Err` in this function, instead, return it, and handle it outside, then you minimum the responsibility of this function to only deal with read and save cache, nothing more.

```
if err := saveDescriptorToCache(); err != nil {
  template.Err = fmt.Errorf("failed to get descriptor: %w", err)
  continue
}
```

pkg/channel/lookup.go
```diff
@@ -328,3 +329,18 @@ func NewMoreThanOneTemplateCandidateErr(component v1beta2.Module,
 	return fmt.Errorf("%w: more than one module template found for module: %s, candidates: %v",
 		ErrTemplateNotIdentified, component.Name, candidates)
 }
+
+func saveDescriptorToCache(template ModuleTemplateTO) {
+	if template.ModuleTemplate != nil {
```
ruanxin, line 334: Write this condition with `if template.ModuleTemplate == nil`, then you can return error immediately and have less nested if block.

pkg/channel/lookup.go
```diff
@@ -45,6 +45,7 @@ func GetTemplates(
 		switch {
 		case module.RemoteModuleTemplateRef == "":
 			template = NewTemplateLookup(kymaClient, module, kyma.Spec.Channel).WithContext(ctx)
+			saveDescriptorToCache(template)
```
ruanxin, line 48: it's better handle `template.Err != nil` before `saveDescriptorToCache` in case you deal with some err template, also it prevents `template.Err` to be overwritten later in case it has error in  `saveDescriptorToCache`

pkg/channel/lookup.go
```diff
@@ -328,3 +334,20 @@ func NewMoreThanOneTemplateCandidateErr(component v1beta2.Module,
 	return fmt.Errorf("%w: more than one module template found for module: %s, candidates: %v",
 		ErrTemplateNotIdentified, component.Name, candidates)
 }
+
+func saveDescriptorToCache(template ModuleTemplateTO) error {
+	if template.ModuleTemplate == nil {
+		return ErrModuleTemplateIsNil
+	}
+	descriptor := template.GetDescFromCache()
+	if descriptor == nil {
```
ruanxin, line 343: still the coding style, it's better to always deal with unexpected behavior first, so that you can avoid nested if condition

```
if descriptor != nil {
  return nil
}

//the rest code 

```

pkg/channel/lookup.go
```diff
@@ -45,6 +46,11 @@ func GetTemplates(
 		switch {
 		case module.RemoteModuleTemplateRef == "":
 			template = NewTemplateLookup(kymaClient, module, kyma.Spec.Channel).WithContext(ctx)
+			if template.Err == nil {
```
ruanxin, line 49: coding style issue:
```
if template.Err != nil {
  continue
}
if err := saveDescriptorToCache(template); err != nil {
  template.Err = fmt.Errorf("failed to get descriptor: %w", err)
}
```

nesmabadr: I still need to reach this line `templates[module.Name] = &template` not to continue in this case. That's why I have it like this

ruanxin (approved):

PR #921, opened 2023-10-05
chore: Refactor unit/integration test to improve test resource creation

ruanxin:
This PR includes:

- extract fat test util functions in `utils.go` into different files scoped by the resources `kyma.go`, `manifest.go` ...
- introduce more builder pattern for `moduletemplate` and `defaultCR`
   - replaced ModuleTemplateFactory so that we can gid rid of long control params in module template creation
- decoupled `gomega` from `testutils` package
- adjust related tests


pkg/testutils/kyma.go
```diff
@@ -0,0 +1,134 @@
...
+	//nolint:wrapcheck
+	return client.IgnoreNotFound(err)
+}
+
+func KymaExists(ctx context.Context, clnt client.Client, name, namespace string) error {
+	kyma, err := GetKyma(ctx, clnt, name, namespace)
+	return CRExists(kyma, err)
+}
+
+func KymaDeleted(ctx context.Context,
```
lindnerby, line 58: For these helper function can the return type just be bool? What we need to error for in assertions?

pkg/testutils/kyma.go
```diff
@@ -0,0 +1,134 @@
...
+var (
+	ErrStatusModuleStateMismatch = errors.New("status.modules.state not match")
+	ErrKymaNotDeleted            = errors.New("kyma CR not deleted")
+)
+
+func NewTestKyma(name string) *v1beta2.Kyma {
+	return newKCPKymaWithNamespace(name, v1.NamespaceDefault, v1beta2.DefaultChannel, v1beta2.SyncStrategyLocalClient)
+}
+
+func NewKymaForE2E(name, namespace, channel string) *v1beta2.Kyma {
```
lindnerby, line 25: Can we move this to the e2e tests when it's only used by e2e tests?

pkg/testutils/kyma.go
lindnerby: Same for `KymaExists`

pkg/testutils/manifest.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/kyma-project/lifecycle-manager/pkg/module/common"
+	"github.com/kyma-project/lifecycle-manager/pkg/testutils/builder"
+	compdesc2 "github.com/open-component-model/ocm/pkg/contexts/ocm/compdesc/versions/v2"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/uuid"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func NewTestManifest(prefix string) *v1beta2.Manifest {
+	return &v1beta2.Manifest{
```
lindnerby, line 18: Maybe establish a ManifestBuilder as well?

pkg/testutils/builder/defaultcr.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"fmt"
+
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+type DefaultCRBuilder struct {
```
lindnerby, line 11: Struggling a bit with this name. Maybe CustomResourceBuilder? SampleCRBuilder? UnstructuredCRBuilder?

pkg/testutils/builder/defaultcr.go
```diff
@@ -0,0 +1,64 @@
...
+	defaultCR *unstructured.Unstructured
+}
+
+const (
+	InitSpecKey   = "initKey"
+	InitSpecValue = "initValue"
```
lindnerby, line 17: These are actually not used here and have no relation to the builder itself. Can you move them to the testcases that use them and do assertions with them?

lindnerby: Or the `NewDefaultCRBuilder()` should initialize the internal `defaultCR *unstructured.Unstructured` Spec already with an init kvp?

lindnerby (changes requested):

pkg/testutils/kyma.go
ruanxin: I actually encourage reuse of these util functions for e2e test, many of them can be shared, like check state, conditions, we don't need to strictly separate them.

pkg/testutils/manifest.go
ruanxin: We can do it later if needed, atm, I do not see the need to have a builder for manifest, manifest doesn't need so many configurations during creation, it's basically converted from module template.

pkg/testutils/builder/defaultcr.go
ruanxin: maybe `ModuleCRBuilder`?

pkg/testutils/kyma.go
ruanxin: The thing is return error can get more information for analysis if the test failed, but I will remove `ErrKymaNotDeleted`, instead just return `err`

pkg/testutils/builder/defaultcr.go
ruanxin: This is actually a problem of the existing test, this field you can see is defined in `operator_v1beta2_moduletemplate_kcp-module.yaml`, and we are using it across different test suites (kyma-controller, control-plane), I also have no good idea where to put it.

IMO the `NewDefaultCRBuilder` should not provide this spec field by default, it prevents us from having a test case which requires different `spec`.

ruanxin: ok, split finally.

pkg/testutils/builder/defaultcr.go
lindnerby: Anything than default :)

pkg/testutils/kyma.go
lindnerby: Of course, but they can be part of e2e_test/utils then right?

ruanxin: My idea is to use `testutils` as a shared package between all tests, if not necessary, do not need to differentiate with e2e test and rest.

Of course, if some specific function is only suitable for e2e, then putting it under the `e2e_test` package makes sense. But it's a bit overkill for kyma initialization. We only have kyma different between e2e and unit test, because of `SyncStrategy`. But for rest resources, the initialization of test resources can be shared.

And follow the naming pattern, actually, it's better to put it under `e2e_test/kyma.go`, it's the only use case, I just think it's not necessary to have this level of separation, at least at the moment. If you want, we can have a call.

lindnerby: Just not a big fan of offering functions by pkg for other tests if the specific test scenarios need a Kyma with sync label, where they can then just use the builder and create a Kyma with that label explicitly, explaining via code what is so special about this "e2e Kyma", because one can notice then "ah it is just a Kyma with a sync label" without needing to jump to a definition in another file. And if a convenience function is **really** needed to save that one line of code, then it should in my opinion reside as close to the usage as possible and not in some generic testutils pkg. 
But at this point I don't really care anymore. So maybe either you can rename it to `NewKymaWithSyncLabel` to explain that this is e2e agnostic, remove it or leave it. 👍

ruanxin: renamed to `NewKymaWithSyncLabel`

lindnerby (approved):

PR #947, opened 2023-10-16
feat: Improve XFCC header parsing in SAN Verification code (Checkmarx)

Tomasz-Smelcerz-SAP:
**Description**

Checkmarx reported potential DDoS via resource-exhaustion around the XFCC header parsing in `pkg/security/san_pinning.go:81` (line number as in the current PR)
This PR does not "fix" the Checkmarx alert - it looks like Checkmarx is complaining about "slice expression" which, on its own, is not leading to any resource exhaustion.
Nevertheless we should protect the code from incjection of big and potentially malicious Certificate data and introduce some limits.

Note to the reviewer: The limits proposed in the PR are arbitrary, AFAIK there are no "hard" values defined for certificate size or a number of SAN of the certificate.
The proposed values seem to be big enough to handle any real scenario considering the Lifecycle-Manager environment.

- Limit the size of the certificate data to 32 KiB
- Limit the number of SAN items to 100 (one hundred)

### DIFF
--- a/pkg/security/san_pinning.go
+++ b/pkg/security/san_pinning.go
@@ -24,6 +24,7 @@ const (
 	certificateKey = "Cert="
 	shootDomainKey = "skr-domain"
 	limit32KiB     = 32 * 1024
+	limitSANValues = 100
 )
 
 var (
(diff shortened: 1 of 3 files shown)


pkg/security/san_pinning.go
```diff
@@ -127,14 +131,25 @@ func (v *RequestVerifier) getDomain(request *http.Request, watcherEvtObject *typ
...
-		contains(certificate.IPAddresses, kymaDomain) {
+func (v *RequestVerifier) VerifySAN(certificate *x509.Certificate, kymaDomain string) (bool, error) {
+	uris := certificate.URIs
+	dnsNames := certificate.DNSNames
+	IPAddresses := certificate.IPAddresses
+
+	if len(uris) > limitSANValues ||
+		len(dnsNames) > limitSANValues ||
+		len(IPAddresses) > limitSANValues {
+		return false, errTooManySANValues
```
jeremyharisch, line 142:
```suggestion
	if (len(uris) + len(dnsNames) + len(IPAddresses)) > limitSANValues {
		return false, errTooManySANValues
```

jeremyharisch (approved):

PR #948, opened 2023-10-16
test: Purge Controller E2E tests

### LINKED ISSUE #936: Add e2e test for purge-controller
**Description**

The purge controller is not tested e2e. We can add a new e2e-test to the workflow at `jobs.e2e-integration.strategy.matrix` called `"purge-controller"`. 
There needs to be extra patches for the LM deployment, so before `Deploy LM local testing kustomize`:
```
- op: add
   value: --enable-purge-finalizer=true
- op: add
   value: --purge-finalizer-timeout=5s
```
So the purge-controller is actually kicking in after 5 seconds and starts to delete remote finalizers for deleted Kymas. ([Implementation hint](https://github.com/kyma-project/lifecycle-manager/blob/main/.github/workflows/test-e2e.yml#L155C10-L162C24))

**Acceptance Criteria**
* Add e2e test that verifies purge loop is triggered after timeout value and starts deleting finalizers from remote cluster CRs
    * Use template-operator
    * Add finalizer to SampleCR to block the deletion
    * Purge reconciler should remove all finalizers after Kyma deletion + timeout
    * Figure out if 5seconds of purge timeout is sufficient
* Add e2e test that checks if metrics endpoint exposes purge related metrics `lifecycle_mgr_purgectrl_time`, `lifecycle_mgr_purgectrl_requests_total` and `lifecycle_mgr_purgectrl_error`
    * [Implementation Inspiration](https://github.com/kyma-project/lifecycle-manager/blob/main/tests/e2e_test/kyma_metrics_test.go)

nesmabadr:
**Description**

- Purge controllers E2E test

**Related issue(s)**
Resolves #936


pkg/testutils/kyma.go
```diff
@@ -80,6 +80,34 @@ func DeleteKymaByForceRemovePurgeFinalizer(ctx context.Context, clnt client.Clie
 	return DeleteCR(ctx, clnt, kyma)
 }
 
+func DeleteKyma(ctx context.Context,
+	clnt client.Client,
+	kyma *v1beta2.Kyma,
+) error {
+	if err := SyncKyma(ctx, clnt, kyma); err != nil {
```
jeremyharisch, line 87: Why is this needed? Since if Kyma has already been deleted, then `SyncKyma` will return nil thus will continue and you will get an error when calling `clnt.Delete`.

I suggest to rewrite something like the following:

```
func DeleteKyma(ctx context.Context,
	clnt client.Client,
	kyma *v1beta2.Kyma,
) error {
    err := clnt.Delete(ctx, kyma)
    if client.IgnoreNotFound(err) != nil {
        return fmt.Errorf("updating kyma %w", err)
    }
    return nil
}
```

pkg/testutils/modulecr.go
```diff
@@ -58,3 +62,47 @@ func SampleCRDeletionTimeStampSet(ctx context.Context, name, namespace string, c
 	}
 	return nil
 }
+
+func AddFinalizerToModuleCR(ctx context.Context, clnt client.Client, moduleCR *unstructured.Unstructured,
+	finalizer string,
+) error {
+	err := clnt.Get(ctx, client.ObjectKey{
```
jeremyharisch, line 69: Please check the error, otherwise you just overwrite it in line 74

jeremyharisch (commented):
Just some small comments, rest looks fine 👍🏻

jeremyharisch (approved):

Issue #868, opened 2023-09-19
Module state is not propagated correctly to Kyma CR


shorim:
**Description**
After the changes done [here](https://github.com/kyma-project/telemetry-manager/pull/405), if the user attempts to delete the Telemetry module while some custom resources related to Telemetry still exist in the cluster, the module deletion is blocked and the state of the Telemetry module will change to `Warning` . However, the Kyma CR shows that the Telemetry module is in `Deleting` state instead of the actual `Warning` state.

Also, the `Processing` state was removed [here](https://github.com/kyma-project/telemetry-manager/pull/333) and initially the Telemetry module state is empty. However, the Kyma CR shows that the Telemetry module is in `Processing` state.

**Expected result**
When Telemetry module is in `Warning` state, it is expected that the Telemetry state in Kyma CR is also `Warning`
```
  modules:
  - channel: fast
    fqdn: kyma-project.io/module/telemetry
    manifest:
      apiVersion: operator.kyma-project.io/v1beta2
      kind: Manifest
      metadata:
        generation: 1
        name: default-kyma-telemetry-812080235
        namespace: kyma-system
    name: telemetry
    resource:
      apiVersion: operator.kyma-project.io/v1alpha1
      kind: Telemetry
      metadata:
        name: default
        namespace: kyma-system
    state: Warning
```

When the state of the Telemetry module is empty, it is expected that the Telemetry state in Kyma CR is also empty
```
  modules:
  - channel: fast
    fqdn: kyma-project.io/module/telemetry
    manifest:
      apiVersion: operator.kyma-project.io/v1beta2
      kind: Manifest
      metadata:
        generation: 1
        name: default-kyma-telemetry-812080235
        namespace: kyma-system
    name: telemetry
    resource:
      apiVersion: operator.kyma-project.io/v1alpha1
      kind: Telemetry
      metadata:
        name: default
        namespace: kyma-system
    state: ""
```
**Actual result**
When Telemetry module is in `Warning` state, the Telemetry state in Kyma CR is incorrectly set as `Deleting`
```
kubectl get telemetry -n kyma-system
NAME      GENERATION   AGE   STATE
default   3            47m   Warning
```

```
  modules:
  - channel: fast
    fqdn: kyma-project.io/module/telemetry
    manifest:
      apiVersion: operator.kyma-project.io/v1beta2
      kind: Manifest
      metadata:
        generation: 1
        name: default-kyma-telemetry-812080235
        namespace: kyma-system
[...]

PR #956, opened 2023-10-18
fix: State propagation during module CR deletion

ruanxin:
related issue: https://github.com/kyma-project/lifecycle-manager/issues/868

- the test of normal state propagation during module delete has been covered [here](https://github.com/kyma-project/lifecycle-manager/pull/956/files#diff-ea86e84ebe177150874aafaa958558196cdd8bf2461a02a11b00b5b7b03ad44bR55-R75)
- the test of warning state propagation during module delete has been covered [here](https://github.com/kyma-project/lifecycle-manager/pull/956/files#diff-02f29766ed71850fe0275f8cea90bd331073eba09d1770d43b9d2b0472c1b872R102-R131)
- refactored `helper_tests.go`, relocated more shared utility functions into `testutils`


pkg/testutils/kyma.go
```diff
@@ -117,10 +198,11 @@ func IsKymaInState(ctx context.Context, name, namespace string, clnt client.Clie
 		string(state))
 }
 
-func ExpectKymaManagerField(
-	ctx context.Context, controlPlaneClient client.Client, kymaName string, managerName string,
+func ContainsExpectKymaManagerField(
```
Tomasz-Smelcerz-SAP, line 201: Consider renaming it to `ContainsKymaManagerField` or `StatusContainsManagerField`

pkg/testutils/kyma.go
```diff
@@ -117,10 +198,11 @@ func IsKymaInState(ctx context.Context, name, namespace string, clnt client.Clie
...
+	ctx context.Context, clnt client.Client,
+	kymaName, kymaNamespace, managerName string,
 ) (bool, error) {
-	createdKyma, err := GetKyma(ctx, controlPlaneClient, kymaName, "")
+	createdKyma, err := GetKyma(ctx, clnt, kymaName, kymaNamespace)
```
Tomasz-Smelcerz-SAP, line 205: How about renaming this to just `kyma`?

internal/controller/kyma_controller/kyma_test.go
```diff
@@ -79,12 +90,23 @@ var _ = Describe("Kyma enable one Module", Ordered, func() {
...
-			Should(Equal(string(v1beta2.StateProcessing)))
+		Eventually(IsKymaInState, Timeout, Interval).
+			WithContext(ctx).
+			WithArguments(kyma.GetName(), kyma.GetNamespace(), controlPlaneClient, v1beta2.StateProcessing).
+			Should(Succeed())
 
 		By("having created new conditions in its status")
-		Eventually(GetKymaConditions, Timeout, Interval).WithArguments(kyma.GetName()).ShouldNot(BeEmpty())
+		Eventually(func() error {
+			createdKyma, err := GetKyma(ctx, controlPlaneClient, kyma.GetName(), kyma.GetNamespace())
```
Tomasz-Smelcerz-SAP, line 100: I think this code is less readable than the original one, which was a one-liner. Generally I think we should keep the test code as short as possible and possibly reflecting the "given/when/then" BDD style. IMO this code, although simple, is long enough to deserve it's own function.

pkg/testutils/kyma.go
ruanxin: changed to ContainsKymaManagerField

internal/controller/kyma_controller/kyma_test.go
ruanxin: the problem of [original one](https://github.com/kyma-project/lifecycle-manager/pull/956/files#diff-886762db4cf11d4f54f47a446185f94f18c2777d10def986b78b5509c0707982L101)is it expected return empty condition when get kyma get error, which is not solid enough, but I agree it deserve a function.

Tomasz-Smelcerz-SAP (approved):

PR #962, opened 2023-10-18
chore(linters): Add "gci" and "importas" linters

ameteiko:
**Description**

- adds a [gci](https://golangci-lint.run/usage/linters/#gci) linter
- adds an [importas](https://golangci-lint.run/usage/linters/#importas) linter
- refactors the codebase to conform to the new linting rules


.golangci.yaml
```diff
@@ -69,6 +143,17 @@ issues:
     - path: "_test\\.go"
       linters:
         - wrapcheck
+    - path: tests/e2e_test/
+      linters: [ gci ] # Disable gci due to the test utilities dot import.
+    - path: pkg/watcher/certificate_suite_test.go
+      linters: [ gci ] # Disable gci due to the test utilities dot import.
+    - path: internal/declarative/v2/test/v1/declarative_test.go
```
ruanxin, line 150: is it possible use `text:` to only exclude the dot import, but allow other gci rules? 
https://golangci-lint.run/usage/configuration/#issues-configuration

.golangci.yaml
```diff
@@ -6,12 +6,12 @@ linters:
     - depguard # checks if package imports are whitelisted
     - exhaustivestruct # deprecated (since v1.46.0) and replaced by 'exhaustruct'
     - exhaustruct # TODO enable and fix, use regex to exclude certain cases
-    - gci # keep disabled for now since it will autofix all imports and result in a git dirty state
     - godox # TODO enable and fix
     - golint # deprecated (since v1.41.0) and replaced by 'revive'
     - ifshort # deprecated (since v1.48.0)
     - interfacer # deprecated (since v1.38.0)
     - ireturn # TODO enable and fix
+    - lll
```
ruanxin, line 14: why we exclude `lll` ?

.golangci.yaml
ameteiko: Unfortunately, the `gci` linting error message is to generic: 

```
File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(github.com/kyma-project/lifecycle-manager) -s blank -s dot --custom-order
```

It just complains that the imports are not properly sorted without informing the particular section. So it makes it impossible to filter-out only the dot-imports, using the `text:` filter you suggested.

ruanxin (approved):

PR #974, opened 2023-10-20
test: Enhance the deletion E2E test

### LINKED ISSUE #752: Fix kyma CR can't be deleted due to cached runtime client
Currently, we are facing a problem during the deletion of Kyma CR instances. In some cases, the remote SKR cluster might have been removed before the Kyma CR deletion process. As a result, the lifecycle-manager tries to connect to a non-existent SKR cluster, causing it to return a "no such host" error. This prevents successful deletion of the Kyma CR.

```
dial tcp: lookup xxx on xxx : no such host'
```

**ACs:**

- When the Kyma CR deletion timestamp is set and the LM cannot reach the remote host (SKR cluster), the cached SKR client entry should be deleted. This will force the LM to obtain access secret from cluster.

- If the access secret has been removed, the LM will return a "not found" error. This action should trigger the removal of the finalizer function for the Kyma CR, allowing it to be successfully deleted. [related code](https://github.com/kyma-project/lifecycle-manager/blob/main/controllers/kyma_controller.go#L131-L136)
- E2E-Test
    - Re-Use already existing cluster; delete SKR after other e2e tests did run, then run that testcase
    - Test case: After SKR-Cluster has been deleted, KymaCR should also be deleted successfully

nesmabadr:
**Description**

- Adjusted the deletion E2E test

**Related issue(s)**
Resolves #752


pkg/testutils/kyma.go
```diff
@@ -88,7 +88,8 @@ func DeleteKyma(ctx context.Context,
 	clnt client.Client,
 	kyma *v1beta2.Kyma,
 ) error {
-	err := clnt.Delete(ctx, kyma)
+	propagation := v1.DeletePropagationForeground
```
ruanxin, line 91: It's better add a comment why delete kyma use foreground

.github/workflows/test-e2e.yml
```diff
@@ -155,13 +155,13 @@ jobs:
           done
       - name: Checkout template-operator
         uses: actions/checkout@v3
-        if: ${{ matrix.e2e-test == 'module-status-propagation' ||  matrix.e2e-test == 'module-without-default-cr' ||  matrix.e2e-test == 'non-blocking-deletion' ||  matrix.e2e-test == 'purge-controller' ||  matrix.e2e-test == 'purge-metrics' }}
+        if: ${{ matrix.e2e-test == 'module-status-propagation' ||  matrix.e2e-test == 'module-without-default-cr' ||  matrix.e2e-test == 'non-blocking-deletion' ||  matrix.e2e-test == 'purge-controller' ||  matrix.e2e-test == 'purge-metrics' || matrix.e2e-test == 'kyma-deletion'}}
```
ruanxin, line 158: which one doesn't need to check out template-operator? We can change to if not equals, this if condition is too long. Or even remove this if, actually doesn't matter have this checkout step if not needed.

pkg/testutils/kyma.go
```diff
@@ -88,7 +88,9 @@ func DeleteKyma(ctx context.Context,
 	clnt client.Client,
 	kyma *v1beta2.Kyma,
 ) error {
-	err := clnt.Delete(ctx, kyma)
+	// Foreground deletion is used to make sure all the dependents get deleted first before Kyma is deleted
```
ruanxin, line 91: The dependents here is only manifest CR, no other resources.

tests/e2e_test/kyma_deletion_test.go
```diff
@@ -12,71 +12,84 @@ import (
 var _ = Describe("KCP Kyma CR Deletion", Ordered, func() {
 	kyma := NewKymaWithSyncLabel("kyma-sample", "kcp-system", "regular",
 		v1beta2.SyncStrategyLocalSecret)
+	module := NewTestModuleWithFixName(moduleName, v1beta2.DefaultChannel)
```
ruanxin, line 15: Please use `NewTemplateOperator` to create a template-operator module, it's in `testutils`

tests/e2e_test/utils_test.go
```diff
@@ -50,6 +50,8 @@ const (
 	interval              = 1 * time.Second
 	remoteNamespace       = "kyma-system"
 	controlPlaneNamespace = "kcp-system"
+	moduleName            = "template-operator"
```
ruanxin, line 53: if you create all test module with `NewTemplateOperator`, we don't need to define this `moduleName` specifically.

ruanxin (commented):
lgtm, some refactoring comments.

PR #976, opened 2023-10-20
fix: share code between declaritive and api

ruanxin:
Since currently, the declarative library is only for LM internal usage, the dependencies between API and declarative should be designed carefully. declarative can import API, but not verse vice. 

This PR extract some common types which can be used by API and declarative into API `shared` package, to make sure the import only happens in one direction, to avoid cyclic dependencies.


api/shared/last_operation.go
```diff
@@ -0,0 +1,10 @@
+package shared
```
lindnerby, line 1: I like the changes, but can we talk about the package name?
From usage perspective `shared.State` or `shared.Resource` seems a bit weird, wdyt? The intention is to share that's right, but the actual property of the package is that it contains common types to be used from various consumers esp. `api`.
So what do you think about naming it `common` or `general`? The import reads nice and explains a bit:
`github.com/kyma-project/lifecycle-manager/api/common`
Just a thought, there maybe better names

lindnerby (commented):
Like the changes! Just a though about the pkg name below. Feel free to ignore or leave a thought as well. LGTM

api/shared/last_operation.go
ruanxin: I use `common` in the first try, but the name `common` is too common even in the LM repo, we already have two package named `common`, I don't want to introduce one more. I know this name is not perfect, but compared to `general`, it's better, general is way too general.

The intention of `shared` is sharing this type, that's right. For example, `shared.State` is a state shared by kyma, manifest, module CR, `shared.LastOperations` is same purpose. `shared.Resource` is only for manifest CR atm, but it can be used in future types if needed.

lindnerby: Alright if `common` is already taken too much, let's keep it 👍

lindnerby (approved):

PR #986, opened 2023-10-27
feat: Module Upgrade during blocking deletion

jeremyharisch:
**Description**

Implements https://github.com/kyma-project/lifecycle-manager/pull/986/files plus fixes KLM behaviour.

- Enhances E2E Module Deletion test by upgrade scenario
- Fixed the following:
    - If a module was stuck under deletion, got reenabled under different channel or new version, then got removed completely, the Module Default CR as well as Manifest CR was flickering between the states (`Processing'`, `Error`, `Ready` and `Deleting`). The end result is the same: Everything got deleted, but it should not flicker between the states before.
    - The Describes behaviour was due to racing conditions.
- Make sure to set the Manifest state to `Deleting` if the deletion timestamp has been set and if the Module default CR is under a deleting state or already has been removed.
- Enhance e2e scenario to cover corner cases
- Only add finalizer if Manifest is not under deletion in `PostRunCreateCR`
- Fixed racing condition when Module is under deleting
- Removed `WithModuleCRDName` Option, not used anymore
- Removed `WithCRDDelete` Option; was set to false, only used in one condition with always was resulting in true, thus also remvoed this condition
- Removed `renderer` interface incl. implementations, only nil implementations -> not used
- Moved `NewResourceToInfoConverter` to `renderResources`: Simplify `Reconcile` func


internal/declarative/v2/reconciler.go
```diff
@@ -81,7 +83,7 @@ func newResourcesCondition(obj Object) metav1.Condition {
 	}
 }
 
-//nolint:funlen,cyclop
+//nolint:funlen,cyclop, gocognit
```
lindnerby, line 86: Any chance we can fix the issue reported by this new linter or would the changeset be too big?

internal/manifest/custom_resource.go
```diff
@@ -45,16 +45,19 @@ func PostRunCreateCR(
...
-	oMeta.SetNamespace(obj.GetNamespace())
-	oMeta.SetFinalizers(obj.GetFinalizers())
-	if added := controllerutil.AddFinalizer(oMeta, declarative.CustomResourceManager); added {
-		if err := kcp.Patch(
-			ctx, oMeta, client.Apply, client.ForceOwnership, client.FieldOwner(declarative.CustomResourceManager),
-		); err != nil {
-			return fmt.Errorf("failed to patch resource: %w", err)
+	// only add finalizer if Manifest is not under deletion
+	if obj.GetDeletionTimestamp().IsZero() {
+		oMeta := &v1.PartialObjectMetadata{}
```
lindnerby, line 50: Can you extract this into a separate function where its name is explaining, could make the comment obsolete

internal/declarative/v2/reconciler.go
```diff
@@ -155,10 +157,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return r.ssaStatus(ctx, obj)
 	}
 
-	if !obj.GetDeletionTimestamp().IsZero() {
+	resourceCRDeleted, err := r.DeletionCheck.Run(ctx, clnt, obj)
```
ruanxin, line 160: why we need deletioncheck here? to verify resourceCR get deleted before remove manifest CR finalizer, it should and already detect in `doPreDelete`

ruanxin: ok, we refactored the `doPreDelete` to not wait for CR deletion.

ruanxin: This refactoring is not needed, the existing code just remove default finalizer `"declarative.kyma-project.io/finalizer"`, and in `doPreDelete` (`PreDeleteDeleteCR`), it will remove `"resource.kyma-project.io/finalizer"` after resourceCR and CRD get deleted. Which means manifest CR anyway will not deleted if that `"resource.kyma-project.io/finalizer"` still exists.

you current refactoring makes the manifest CR deletion logic complicated.

internal/manifest/ready_check.go
```diff
@@ -60,7 +60,7 @@ func (c *CustomResourceReadyCheck) Run(ctx context.Context,
 	}
 	moduleCR := manifest.Spec.Resource.DeepCopy()
 	if err := clnt.Get(ctx, client.ObjectKeyFromObject(moduleCR), moduleCR); err != nil {
-		return declarative.StateInfo{State: shared.StateError}, fmt.Errorf("failed to fetch resource: %w", err)
+		return declarative.StateInfo{State: shared.StateError}, declarative.ErrCustomResourceDoesNotExist
```
ruanxin, line 63: you can't assume all err here equals `DoesNotExist`

internal/declarative/v2/reconciler.go
```diff
@@ -215,7 +229,9 @@ func (r *Reconciler) initialize(obj Object) error {
 		status.Synced = []shared.Resource{}
 	}
 
-	if status.State == "" {
+	if !obj.GetDeletionTimestamp().IsZero() {
+		status.State = shared.StateDeleting
```
ruanxin, line 233: why we need to set deleting state manually here in the  `initialize` step? What happens if manifest CR is under deleting, but have warning or error state before? In the initialize step, we should not change original state, only add if there is no state.

internal/declarative/v2/reconciler.go
```diff
@@ -286,8 +302,19 @@ func (r *Reconciler) syncResources(ctx context.Context, clnt Client, obj Object,
 	status.Synced = newSynced
 
 	if hasDiff(oldSynced, newSynced) {
-		obj.SetStatus(status.WithState(shared.StateProcessing).WithOperation(ErrWarningResourceSyncStateDiff.Error()))
+		if obj.GetDeletionTimestamp().IsZero() {
+			obj.SetStatus(status.WithState(shared.StateProcessing).WithOperation(ErrWarningResourceSyncStateDiff.Error()))
+		}
 		return ErrWarningResourceSyncStateDiff
+	} else if !obj.GetDeletionTimestamp().IsZero() {
```
ruanxin, line 309: this `else if` is not the else condition for `if hasDiff`

ruanxin: again, here you introduce some logic I assume make things complicated, basically the logic here should be very simple, if hasDiff, we change the state of manifest CR.

now we have two cases:
1. manifest CR not under delete, then change it to `Processing`
2. manifest CR is under delete, then change it to `Deleting`

What's the reason here to introduce ready check? here no matter if module CR exists or not, it doesn't affect the state of manifest CR.

The ready check should also just do once in each reconciliation to minimum the request to cluster.

ruanxin (changes requested):
In general, this PR should not adjust so many existing logic.

internal/manifest/custom_resource.go
```diff
@@ -45,16 +45,19 @@ func PostRunCreateCR(
...
-	oMeta.SetGroupVersionKind(obj.GetObjectKind().GroupVersionKind())
-	oMeta.SetNamespace(obj.GetNamespace())
-	oMeta.SetFinalizers(obj.GetFinalizers())
-	if added := controllerutil.AddFinalizer(oMeta, declarative.CustomResourceManager); added {
-		if err := kcp.Patch(
-			ctx, oMeta, client.Apply, client.ForceOwnership, client.FieldOwner(declarative.CustomResourceManager),
-		); err != nil {
-			return fmt.Errorf("failed to patch resource: %w", err)
+	// only add finalizer if Manifest is not under deletion
+	if obj.GetDeletionTimestamp().IsZero() {
```
ruanxin, line 49: if manifest is under deletion, we should not create cr, so this check should put in the beginning.

jeremyharisch:
> In general, this PR should not adjust so many existing logic.

Took changes from https://github.com/kyma-project/lifecycle-manager/pull/998 to resolve commetns

internal/declarative/v2/reconciler.go
ruanxin: This `gocognit` basically means in this reconciliation loop, we have too much if else conditions and make the reconciliation flow complicated. But it can be easily reduced by extracting some logic into meaningful functions.

Just want to mention, with remove declarative library task, we will not change current reconciliation logic too much, maybe just rename some existing functions to more expressive compare to current genearic function.

For software refactoring, when we meet this kind of situation that linter is complain the complexities, it's always good to not ignore it and refactoring later, and in practice, in most cases, it will be neglected later.

I would say it worth take efforts to think avoid this `gocognit` complain, one hint, if you look at this `renderer`, it's actually only used in `pruneDiff`, and it can be encapsulated into that function, then you can reduce one if condition, and if you look further, this `renderer` only used in `renderer.RemovePrerequisites(ctx, obj)`, and currently, there is no need to remove any prerequisites, the implementation of `RemovePrerequisites` returns `nil`
```
renderer, err := InitializeRenderer(ctx, obj, spec, r.Options)
```
What I want to say is, take the newly introduced linter complain as a chance to revisit the code structure, it will make you start to think and even observe some tech debts in earlier stage.

But of course, regarding how to work with those tech debts, it depends, in most cases, it may not relavent to the current issue you working with, but then, we can create a follow-up issue, and plan to resolve later.

internal/declarative/v2/reconciler.go
```diff
@@ -450,11 +466,6 @@ func (r *Reconciler) pruneDiff(
 	if err != nil {
 		return err
 	}
-	resourceName := r.ModuleCRDName(obj)
```
ruanxin, line 453: do we still have usage for this function after removing it here? If not, we can cleanup completely.

internal/declarative/v2/reconciler.go
```diff
@@ -332,7 +346,9 @@ func (r *Reconciler) checkTargetReadiness(
 	crStateInfo, err := resourceReadyCheck.Run(ctx, clnt, manifest, target)
 	if err != nil {
 		r.Event(manifest, "Warning", "ResourceReadyCheck", err.Error())
-		manifest.SetStatus(status.WithState(shared.StateError).WithErr(err))
+		if manifest.GetDeletionTimestamp().IsZero() {
+			manifest.SetStatus(status.WithState(shared.StateError).WithErr(err))
```
ruanxin, line 350: do we really need this condition? What I mean is, if this `resourceReadyCheck.Run` return error, it always put `crStateInfo.State` in `shared.StateError`, no matter manifest cr is under delete or not.

And if manifest CR is under delete, it may happen module CR is already deleted during ready check, but for this case, you already have this condition to make sure not return error if is not found.
```
	err := clnt.Get(ctx, client.ObjectKeyFromObject(moduleCR), moduleCR)
	if err != nil {
		if util.IsNotFound(err) && !manifest.DeletionTimestamp.IsZero() {
			return declarative.StateInfo{State: shared.StateDeleting}, nil
		}
		return declarative.StateInfo{State: shared.StateError}, fmt.Errorf("failed to fetch resource: %w", err)
	}
```

internal/declarative/v2/reconciler.go
jeremyharisch: Good catch! Will clean it up, not needed anymore :D

internal/declarative/v2/reconciler.go
jeremyharisch: In theory you are correct, I will give it a shot and remove if everything works as expected

internal/declarative/v2/reconciler.go
jeremyharisch: I removed the renderer fully, since it has a nil implementation and the corresponding interface has only this one implementation.

Furthermore, removed the RemoveCRDs option, since this was always set to false, and thus I could also remove one more condition to simplify the code

internal/declarative/v2/reconciler.go
jeremyharisch: Removed it

internal/declarative/v2/renderer.go
```diff
@@ -1,36 +0,0 @@
...
-	"context"
-
-	"k8s.io/cli-runtime/pkg/resource"
-)
-
-type Prerequisites []*resource.Info
-
-type ConditionsNeedUpdate bool
-
-type Renderer interface {
```
ruanxin, line 13: it's nice we can remove this render completely.

internal/declarative/v2/spec.go
```diff
@@ -10,6 +10,12 @@ type SpecResolver interface {
 	Spec(ctx context.Context, object Object, remoteClient client.Client) (*Spec, error)
 }
 
+type RenderMode string
```
ruanxin, line 13: I believe we can completely drop this `RenderMode` also since currently there is only one render mode anyway, but let's create a follow up issue to address this tech debts.

tests/e2e_test/module_deletion_test.go
```diff
@@ -97,21 +97,32 @@ var _ = Describe("Non Blocking Kyma Module Deletion", Ordered, func() {
...
 				WithArguments("sample-yaml", "kyma-system", runtimeClient, "Deleting").
 				Should(Succeed())
+			By("And the Module Manifest CR is still in a \"Deleting\" State")
+			Consistently(CheckManifestIsInState).
+				WithContext(ctx).
+				WithArguments(kyma.GetName(), kyma.GetNamespace(), "template-operator", controlPlaneClient, shared.StateDeleting).
+				Should(Succeed())
 		})
 
 		It("When the blocking finalizers from the Default Module CR get removed", func() {
```
ruanxin, line 128: here in this step, can you add two more checks to verify  Module Default CR is in `Ready` state, and `Consistently (DeploymentIsReady)`, just to be bit solid.

tests/e2e_test/module_deletion_test.go
```diff
@@ -104,7 +104,7 @@ var _ = Describe("Non Blocking Kyma Module Deletion", Ordered, func() {
 				WithArguments(runtimeClient, defaultRemoteKymaName, remoteNamespace, module).
 				Should(Succeed())
 			By("Then the Module will be updated on SKR")
-			Eventually(DeploymentIsReady).
+			Consistently(DeploymentIsReady).
```
ruanxin, line 107: I believe this step will fail, it's a new deploy, can't be consistently in ready

jeremyharisch: You are correct, that was a stupid change.

ruanxin (approved):

PR #1002, opened 2023-11-03
feat(declarative_test.go): Add Skip Reconcile Label Test

### LINKED ISSUE #965: Add a test for the skip-reconciliation for the Manifest Controller
**Description:**

As a followup to this [issue](https://github.com/kyma-project/lifecycle-manager/issues/937), we need to test the functionality of the skip-reconciliation for the Manifest Controller

**ACs:**
- [x] Add an E2E test for the Manifest Controller [here](https://github.com/kyma-project/lifecycle-manager/blob/main/internal/declarative/v2/test/v1/declarative_test.go) to validate that the reconciliation is skipped if the `operator.kyma-project.io/skip-reconciliation` label exists.

Given kyma CR deployed in KCP Cluster
When enabling Template Operator
Then the Sample CR is deployed on the SKR cluster in Ready state

Given Manifest CR with skip label
When deleting Template Operator deployment
Then the deployment should not been recreated.
When deleting Sample CR
Then the Sample CR should not been recreated.

When removing Manifest CR skip label
Then the deployment should be created.
Then the Sample CR should be created.

LeelaChacha:
Closes #965


internal/declarative/v2/test/v1/declarative_test.go
```diff
@@ -225,6 +225,101 @@ var _ = Describe("Test Manifest Reconciliation for module deletion", Ordered, fu
...
+		func(context.Context, Object) (*ClusterInfo, error) {
+			return &ClusterInfo{
+				Config: cfg,
+			}, nil
+		},
+	)}
+	source := WithSpecResolver(DefaultSpec(filepath.Join(testSamplesDir, "raw-manifest.yaml"), ocirefSynced,
+		RenderModeRaw))
+	oldDeployedResources, err := internal.ParseManifestToObjects(path.Join(testSamplesDir, "raw-manifest.yaml"))
+	Expect(err).NotTo(HaveOccurred())
```
nesmabadr, line 255: please assert here that the oldDeployedResources list is not empty

internal/declarative/v2/test/v1/declarative_test.go
```diff
@@ -225,6 +225,101 @@ var _ = Describe("Test Manifest Reconciliation for module deletion", Ordered, fu
...
+
+	BeforeAll(func() {
+		env, cfg = StartEnv()
+		testClient = GetTestClient(cfg)
+		ctx, cancel = context.WithCancel(context.TODO())
+		reconciler = StartDeclarativeReconcilerForRun(ctx, runID, cfg, append(opts, WithSpecResolver(source))...)
+	})
+
+	It("Should create manifest resources", func() {
```
nesmabadr, line 264: could be renamed to `When creating manifest resources`

internal/declarative/v2/test/v1/declarative_test.go
```diff
@@ -225,6 +225,101 @@ var _ = Describe("Test Manifest Reconciliation for module deletion", Ordered, fu
...
+			BeInState(shared.StateReady),
+			HaveConditionWithStatus(ConditionTypeResources, metav1.ConditionTrue),
+			HaveConditionWithStatus(ConditionTypeInstallation, metav1.ConditionTrue),
+		)
+
+		Expect(testClient.Get(ctx, key, obj)).To(Succeed())
+		Expect(obj.GetStatus()).To(HaveAllSyncedResourcesExistingInCluster(ctx, testClient))
+	})
+
+	It("Should skip reconciliation of deployed module resources", func() {
```
nesmabadr, line 278: May be split this It to: 
`And skip-reconciliation label is added to the manifest` and
`Then the module resources do not get reconciled`

internal/declarative/v2/test/v1/declarative_test.go
```diff
@@ -225,6 +225,101 @@ var _ = Describe("Test Manifest Reconciliation for module deletion", Ordered, fu
...
+		obj.SetLabels(labels)
+		Expect(testClient.Update(ctx, obj)).To(Succeed())
+
+		source := WithSpecResolver(DefaultSpec(filepath.Join(testSamplesDir, "empty-file.yaml"),
+			"", RenderModeRaw))
+		reconciler.SpecResolver = source
+		Consistently(validateOldResourcesNotLongerDeployed, Timeout, Interval).
+			WithContext(ctx).
+			WithArguments(ctx, oldDeployedResources, testClient).
+			Should(Not(Succeed()))
```
nesmabadr, line 290: Instead of Not(Succeed) assert that the error equals `ErrOldResourcesStillDeployed` 
Also assert that they are still there in the status as well

internal/declarative/v2/test/v1/declarative_test.go
```diff
@@ -225,6 +225,101 @@ var _ = Describe("Test Manifest Reconciliation for module deletion", Ordered, fu
...
+	})
+
+	It("Should reconcile after skip label removed", func() {
```
nesmabadr, line 293: Similarly:
`When skip-reconciliation label is removed from Manifest`
`Then the module gets reconciled`

internal/declarative/v2/test/v1/declarative_test.go
```diff
@@ -225,6 +225,101 @@ var _ = Describe("Test Manifest Reconciliation for module deletion", Ordered, fu
...
+
+		EventuallyDeclarativeStatusShould(
+			ctx, key, testClient,
+			BeInState(shared.StateReady),
+			HaveConditionWithStatus(ConditionTypeResources, metav1.ConditionTrue),
+			HaveConditionWithStatus(ConditionTypeInstallation, metav1.ConditionTrue),
+		)
+
+		Expect(testClient.Get(ctx, key, obj)).To(Succeed())
+		Expect(obj.GetStatus()).To(HaveAllSyncedResourcesExistingInCluster(ctx, testClient))
```
nesmabadr, line 275: Please assert here that the resources got deployed

LeelaChacha: [The SyncedResourcesExistingMatcher matcher already does that](https://github.com/LeelaChacha/kyma-lifecycle-manager/blob/9724fb69d595a111d20c4757e111800e8e7c8bf4/internal/declarative/v2/test/v1/helper_test.go#L125). It looks at the status and gets all the resources via object key and returns an error if not found.

Issue #1020, opened 2023-11-09
Reorganise unit and integration tests


ameteiko:
### Description

For proper testing coverage assessment it is crucial to split the unit and integration test suites.

### Acceptance Criteria

- [x] Unit and Integration test suites are separated
- [x] GH Actions run test suites separately
- [ ] Coverage is generated only for the unit test suite
- [ ] Overall average coverage metric is calculated for the unit test suite
- [ ] Collect the code quality metrics

### Feature Testing

Irrelevant due to no changes in the codebase.

PR #1021, opened 2023-11-09
test(refactor): Split unit and integration test suites

ameteiko:
**Description**

- Split unit and integration test suites
- Introduce `unittest` Makefile target

**Related issue(s)**
#1020


tests/integration/controller/manifest/manifesttest/manifesttest.go
nesmabadr, line 1: I think we should rename this file to `manifest_test`. Also I believe there is no need to have it in `manifesttest` folder, it could be inside `manifest` folder directly, wdyt?

ameteiko: Renaming the file to `manifest_test.go` is impossible due to how Golang treats test files (the ones with the `_test.go` suffix). Referencing objects and types declared in the test files from other packages is impossible. Only the test files from the same package can share entities declared in them.

ameteiko: Moved the `manifesttest.go` to the root testing package as requested.

nesmabadr (approved):

PR #1024, opened 2023-11-10
feat(watcher): Enable Watcher by default on control-plane

jeremyharisch:
**Description**
Enables runtime-watcher by default in `control-plane` Kustomize setup.

- Since Watcher is now included in `control-plane` setup, every watcher related resource also has prefix `klm-` -> renamed default istio gateway to `klm-watcher-gateway`
- Introduce transformer to make sure `klm-watcher-serving-cert` is in `istio-system` namespace
- Comment Watcher related parts in `default` and `load_test` kustomize, since Watcher is not enabled there
- Remove outdated comments
- Enable `domain-name-pinning` by default
- Set Watcher image in kustomize to follow pattern of landscape promotion pattern
- Move file-patches in `watcher_local-test` to beginning of patch list, for easier readability
- Remove now unneeded patches in `watcher_local-test`
- Adapted test-suits

**Related issue(s)**


config/default/kustomization.yaml
```diff
@@ -20,16 +20,18 @@ commonLabels:
...
-- patch: |-
-    - op: replace
-      path: /metadata/name
-      value: lifecycle-manager-watcher-gateway
-  target:
-    group: networking.istio.io
-    version: v1beta1
-    kind: Gateway
-    name: watcher-gateway
+# [WATCHER]
```
ruanxin, line 25: for the default profile, we don't need watcher, right? It's better we remove it completely.

config/load_test/kustomization.yaml
```diff
@@ -145,18 +145,19 @@ patches:
...
-        value: default
-      - op: replace
-        path: /metadata/name
-        value: lifecycle-manager-watcher-gateway
-    target:
-      group: networking.istio.io
-      version: v1beta1
-      kind: Gateway
-      name: watcher-gateway
+# [WATCHER]
```
ruanxin, line 148: any reason not delete it completely?

cmd/flags.go
```diff
@@ -27,7 +27,7 @@ const (
 	defaultMaxConcurrentManifestReconciles = 1
 	defaultMaxConcurrentKymaReconciles     = 1
 	defaultMaxConcurrentWatcherReconciles  = 1
-	defaultIstioGatewayName                = "lifecycle-manager-watcher-gateway"
```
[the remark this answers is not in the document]
jeremyharisch: I clarified with. SRE. Normally, ARGO takes care of removed those, if not we we need to contact SREs and tell them what to remove. This would be a one time action

config/load_test/kustomization.yaml
jeremyharisch: Nope, will remove

config/default/kustomization.yaml
jeremyharisch: removed it

ruanxin (approved):

PR #1084, opened 2023-11-24
refactor: Move metrics to internal package

ruanxin:
refactoring metrics related code into internal package.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -51,8 +51,7 @@ import (
 	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
 	"github.com/kyma-project/lifecycle-manager/internal"
 	"github.com/kyma-project/lifecycle-manager/internal/controller"
-	kymametrics "github.com/kyma-project/lifecycle-manager/internal/controller/kyma/metrics"   //nolint:importas // ad-hoc import collision resolution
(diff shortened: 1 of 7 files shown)


cmd/main.go
```diff
@@ -51,8 +51,7 @@ import (
 	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
 	"github.com/kyma-project/lifecycle-manager/internal"
 	"github.com/kyma-project/lifecycle-manager/internal/controller"
-	kymametrics "github.com/kyma-project/lifecycle-manager/internal/controller/kyma/metrics"   //nolint:importas // ad-hoc import collision resolution
-	purgemetrics "github.com/kyma-project/lifecycle-manager/internal/controller/purge/metrics" //nolint:importas // ad-hoc import collision resolution
+	"github.com/kyma-project/lifecycle-manager/internal/pkg/metrics"
```
lindnerby, line 54: Please the remove the `pkg` parent,  it's either `pkg` or `internal`. `metrics` is a nice enough package name for this

lindnerby (changes requested):

cmd/main.go
ruanxin: what do you mean the `pkg` parent?

lindnerby: What's the intent of creating a `pkg` subpackage within `internal`?

lindnerby: As of clarification we can indicate shared internal package to put them in `internal/pkg` 👍

lindnerby (approved):

PR #1086, opened 2023-11-24
refactor: Change metrics to using struct instead of global value

ruanxin:
introduce metrics struct for Kyma and Purge metrics


pkg/module/sync/runner_impl.go
```diff
@@ -138,9 +138,11 @@ func (r *RunnerImpl) setupModule(module *common.Module, kyma *v1beta2.Kyma) erro
 	return nil
 }
 
-func (r *RunnerImpl) SyncModuleStatus(ctx context.Context, kyma *v1beta2.Kyma, modules common.Modules) {
+func (r *RunnerImpl) SyncModuleStatus(ctx context.Context, kyma *v1beta2.Kyma, modules common.Modules,
+	kymaMetrics *metrics.KymaMetrics,
```
lindnerby, line 142: Can we create and pass an interface here to replace concrete KymaMetrics? So unit testing of `RunnerImpl` becomes a bit easier and also `SyncModuleStatus` does not need to know implementation details of kyma metrics.

lindnerby (commented):
Nice changes for reducing global values!
Just have one suggestion:

pkg/module/sync/runner_impl.go
lindnerby: In unit tests we might want to assert that within this function and more likely within the `DeleteNoLongerExistingModuleStatus` function successively, a function of this metrics dependency  (in the current case `RemoveModuleStateMetrics`) is called or not called depending on the scenario. Testing this is easy when providing an interface, because it can be mocked and asserted/verified on, hence my recommendation.
Since it makes testing easier, decouples dependencies and it's not really a big change to swap this signature with an interface, I can see some need.

pkg/module/sync/runner_impl.go
```diff
@@ -138,9 +137,11 @@ func (r *RunnerImpl) setupModule(module *common.Module, kyma *v1beta2.Kyma) erro
 	return nil
 }
 
-func (r *RunnerImpl) SyncModuleStatus(ctx context.Context, kyma *v1beta2.Kyma, modules common.Modules) {
+func (r *RunnerImpl) SyncModuleStatus(ctx context.Context, kyma *v1beta2.Kyma, modules common.Modules,
+	removeModuleStateMetrics func(kymaName, moduleName string),
```
lindnerby, line 141: Thing is, my first idea was to pass a function here. But then, since we finally strife to build up our unit test coverage now, how do you assert that in case of `moduleStatus.Manifest == nil` in `DeleteNoLongerExistingModuleStatus` that this function has been called? Is it possible to verify that?

pkg/module/sync/runner_impl.go
```diff
@@ -224,17 +225,17 @@ func stateFromManifest(obj client.Object) shared.State {
...
+func DeleteNoLongerExistingModuleStatus(ctx context.Context, kyma *v1beta2.Kyma, moduleFunc GetModuleFunc,
+	removeModuleStateMetrics func(kymaName, moduleName string),
 ) {
 	moduleStatusMap := kyma.GetModuleStatusMap()
 	moduleStatus := kyma.GetNoLongerExistingModuleStatus()
 	for idx := range moduleStatus {
 		moduleStatus := moduleStatus[idx]
 		if moduleStatus.Manifest == nil {
-			metrics.RemoveModuleStateMetrics(kyma, moduleStatus.Name)
+			if removeModuleStateMetrics != nil {
```
lindnerby, line 236: Also, we could start using guard clauses to be honest. If `removeModuleStateMetrics` is nil, there is not repercussion, so it seems to be fine to use this without it. But actually I would say it's not really optional, we expect the function to be called in this scenario where the `moduleStatus.Manifest == nil`. And if `removeModuleStateMetrics` is nil, it's actually a developer error and we want to know about it and fail fast if that's the case.
The usage of `DeleteNoLongerExistingModuleStatus` in the test file also indicates that this is designed like this to make usage there more "comfortable", because we can just pass nil. But we can instead pass a mock implementation of the interface and can even verify if it's called or not.
WDYT?

pkg/module/sync/runner.go
```diff
@@ -2,8 +2,270 @@ package sync
...
+
 type GetModuleFunc func(ctx context.Context, module client.Object) error
+
+type Runner struct {
+	client.Client
+	versioner machineryruntime.GroupVersioner
+	converter machineryruntime.ObjectConvertor
+}
+
+// ReconcileManifests implements Runner.Sync.
```
lindnerby, line 46: I don't get this comment. Can you explain if you know? If not, let's remove it.

pkg/module/sync/runner.go
```diff
@@ -2,8 +2,270 @@ package sync
...
+func (r *Runner) deleteManifest(ctx context.Context, module *common.Module) error {
+	err := r.Delete(ctx, module.Manifest)
+	if util.IsNotFound(err) {
+		return nil
+	}
+	return fmt.Errorf("failed to delete manifest: %w", err)
+}
+
+func (r *Runner) setupModule(module *common.Module, kyma *v1beta2.Kyma) error {
+	// set labels
```
lindnerby, line 132: Unnecessary comment

pkg/module/sync/runner.go
```diff
@@ -2,8 +2,270 @@ package sync
...
+	module.ApplyLabelsAndAnnotations(kyma)
+	refs := module.GetOwnerReferences()
+	if len(refs) == 0 {
+		// set owner reference
```
lindnerby, line 136: Same here

pkg/module/sync/runner.go
```diff
@@ -2,8 +2,270 @@ package sync
...
+			return fmt.Errorf("error setting owner reference on component CR of type: %s for resource %s %w",
+				module.GetName(), kyma.Name, err)
+		}
+	}
+
+	return nil
+}
+
+func (r *Runner) SyncModuleStatus(ctx context.Context, kyma *v1beta2.Kyma, modules common.Modules,
+	metrics ModuleMetrics,
```
lindnerby, line 147: Now since you did this great refactoring we could add `ModuleMetrics` as a field to `Runner` and pass it via `func New(*)` as well. Since it's used by 2 public methods here.  WDYT?

pkg/module/sync/runner.go
ruanxin: it's all coming from old code, will remove it.

pkg/module/sync/runner.go
ruanxin: I think we should not introduce `ModuleMetrics` to `Runner`, as I mentioned in our discussion, the metrics to me are more low-level utility and usually address cross-cutting concerns, it should be flexible enough to pass into the actual place as composition, but should not belong to any specific business object. It's similar to using a struct to carry logger, to me, it's all anti-pattern. Why does this runner need a metric?

Although it's convent to carry those utility objects, it increases code rigidity.

pkg/module/sync/runner.go
ruanxin: removed

pkg/module/sync/runner.go
ruanxin: removed

pkg/module/sync/runner.go
lindnerby: Ok let's try that approach then, fine by me 👍

Issue #1061, opened 2023-11-16
Watcher: Rotate Certificates if CA Certificate has been rotated


jeremyharisch:
### Description

Our current [architecture](https://github.com/kyma-project/runtime-watcher/blob/main/docs/architecture.md) relies on the certificate rotation of cert-manager. Unfortunately, the cert-manager has a huge design flaw.

We are leveraging the suggested architecture from cert-manager to build an own PKI. This includes an Issuer which uses a certificate as CA. Here comes the catch: The certificates used as CA also gets rotated, but all the certificates which are issues using the CA certificate are not getting rotated, thus they will be refused on a Watcher request, since they are signed by an "unknown" authority (the old CA, which has now been rotated).

What does that mean? Cert-Manager should have a feature that if a CA Certificates gets rotated, then also all leaf-certificates should be rotated. This feature is fully missing but wanted by the community - Whole story can be read here: https://github.com/cert-manager/cert-manager/issues/2478

Our solution to this problem:
In a Kyma reconciliation loop, when the corresponding certificate gets created, check the creation time of the certificate secret and compare it with the creation time of the creation time of CA Certificate. To have a better performance, the CA Certificate should be cached. The cache should be updated if the current time passed the timestamp of the `status.notBefore` of the CA Certificate.

### Reasons

If leaf-certificates are not getting rotated then Watcher will not work anymore when CA cert gets rotated (every 60 days).

### Acceptance Criteria

- [x] On Certificate Creation compare kyma Certificate secret with cached CA Certificate, if CA Certificate has been rotated, delete Kyma Certificate Secret -> Cert-Manager will take care of creating new one
- [x] Cache CA Certificate -> update cached CA Certificate if `status.notBefore` timestamp has been passed
- [x] Implement e2e test
- [x] Create Follow-Up Issue to aim for no-downtime solution

### Feature Testing

End-to-End tests

PR #1087, opened 2023-11-24
fix: Remove Secret only when the secret creation time is before the CA Certificate validity

nesmabadr:
**Description**

- Remove Secret only when the secret creation time is before the CA Certificate validity

**Related issue(s)**
#1061

### DIFF
--- a/pkg/watcher/certificate.go
+++ b/pkg/watcher/certificate.go
@@ -98,11 +98,8 @@ func (c *CertificateManager) Remove(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	err = c.RemoveSecret(ctx)
-	if err != nil {
-		return err
-	}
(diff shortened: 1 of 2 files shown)


pkg/watcher/certificate.go
```diff
@@ -94,16 +94,34 @@ func (c *CertificateManager) Create(ctx context.Context) error {
...
 func (c *CertificateManager) Remove(ctx context.Context) error {
+	err := c.RemoveCertificate(ctx)
+	if err != nil {
+		return err
+	}
+	err = c.RemoveSecret(ctx)
+	if err != nil {
+		return err
+	}
+	return nil
```
jeremyharisch, line 105:
```suggestion
	return c.RemoveSecret(ctx)
```

Issue #1007, opened 2023-11-06
Alerting: Watcher Certificate rotation


jeremyharisch:
### Description

The certificates created by the KLM are certs from cert-manager and used for the runtime-watcher. Currently, we are relying on the default values from cert-manager regarding the certificate rotation: expires after 90 days, gets rotated after 2/3 of the rotation time (here 60 days). To check if this rotation is working, an alert should be triggered after a specified threshold has been passed - i.e. 24h after original planned rotation time. This check can be done either on the SKR cluster or the KCP, since each certificate gets blindly copied over to the SKR in each kyma reconciliation loop.

### Reasons

An alert should be triggered if a certain threshold, after certificate rotation time, has been passed to verify certificate rotation mechanism.

### Acceptance Criteria

- [x] Check last rotation time, if threshold is passed trigger alert (can be extracted from the Certificate CR (status includes notAfter, notBefore and the last renewal time)
- [x] Add a proper metric to base the alerting on

### Feature Testing

End-to-End tests

PR #1088, opened 2023-11-24
feat: Detect Watcher certificate not renew 

ruanxin:
related: https://github.com/kyma-project/lifecycle-manager/issues/1007

- add e2e test for self signed certificate renew `tests/e2e_test/self_signed_certificate_rotation_test.go`
- refactor kyma and purge metrics
- add watcher metrics for `certNotRenewGauge`
- distinguish the CA Certificate with self signed Certificate
- seperate `CertificateManager` param into `SkrWebhookManagerConfig`, `CertificateConfig` and `GatewayConfig`


pkg/watcher/skr_webhook_manifest_manager.go
```diff
@@ -151,20 +131,51 @@ func (m *SKRWebhookManifestManager) Install(ctx context.Context, kyma *v1beta2.K
...
+) {
+	if certificate.Status.RenewalTime != nil &&
+		time.Now().Add(-m.certificateConfig.RenewBuffer).After(certificate.Status.RenewalTime.Time) {
+		m.WatcherMetrics.SetCertNotRenew(kyma.Name)
+	} else {
+		m.WatcherMetrics.CleanupMetrics(kyma.Name)
+	}
+}
+
+func verifyCACertRotation(ctx context.Context,
```
lindnerby, line 145: This could be a method on the `CertificateManager` instead

tests/integration/controller/withwatcher/suite_with_watcher_test.go
```diff
@@ -178,21 +178,34 @@ var _ = BeforeSuite(func() {
...
 		RemoteSyncNamespace:    controller.DefaultRemoteSyncNamespace,
-		CACertificateName:      caCertificateName,
 	}
 
-	caCertCache := watcher.NewCertificateCache(5 * time.Minute)
+	certificateConfig := watcher.CertificateConfig{
+		IstioNamespace:      istioSystemNs,
+		RemoteSyncNamespace: controller.DefaultRemoteSyncNamespace,
+		CACertificateName:   caCertificateName,
+		AdditionalDNSNames:  nil,
```
lindnerby, line 192: `AdditionalDNSNames` is nil here and initialized with empty list here tests/integration/watcher/certificate_test.go, can you align this? Maybe just not set it.

pkg/watcher/certificate_manager.go
```diff
@@ -42,15 +42,26 @@ type SubjectAltName struct {
...
+	// IstioNamespace represents the cluster resource namespace of istio
+	IstioNamespace string
+	// RemoteSyncNamespace indicates the sync namespace for Kyma and module catalog
+	RemoteSyncNamespace string
+	// CACertificateName indicates the Name of the CA Root Certificate in the Istio Namespace
+	CACertificateName string
+	// AdditionalDNSNames indicates the DNS Names which should be added additional to the Subject
+	// Alternative Names of each Kyma Certificate
+	AdditionalDNSNames []string
+	Duration           apimetav1.Duration
```
lindnerby, line 62: Why we need to use the `apimetav1` and not `time`?

pkg/watcher/skr_webhook_manifest_manager.go
```diff
@@ -151,20 +131,51 @@ func (m *SKRWebhookManifestManager) Install(ctx context.Context, kyma *v1beta2.K
 	return nil
 }
 
+func (m *SKRWebhookManifestManager) verifyCertNotRenew(certificate *certmanagerv1.Certificate,
```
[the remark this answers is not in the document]
lindnerby: Just some weird name, it actually just updates watcher metrics and does not verify anthing since it also doesnt' return anything. Can you rename it to tell more what it does?

pkg/watcher/skr_webhook_manifest_manager.go
lindnerby: It also does not really verify, it actually removes the secret when rotated. Can you rename it?

lindnerby (changes requested):
Nice changes, also extracting some `Config` and extending e2e for this scenario 👍

cmd/flags.go
```diff
@@ -4,39 +4,42 @@ import (
...
+	defaultCaCertCacheTTL                   time.Duration = 1 * time.Hour
+	defaultSelfSignedCertDuration           time.Duration = 90 * 24 * time.Hour
+	defaultSelfSignedCertRenewBefore        time.Duration = 60 * 24 * time.Hour
+	defaultSelfSignedCertificateRenewBuffer               = 24 * time.Hour
+	DefaultRemoteSyncNamespace                            = "kyma-system"
 )
 
 //nolint:funlen
-func defineFlagVar() *FlagVar {
+func DefineFlagVar() *FlagVar {
```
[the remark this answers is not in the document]
ruanxin: This change is the reason for this [issue](https://github.com/kyma-project/lifecycle-manager/issues/1097), I think it's not worth to change it back to private, the idea to put this flag in main package is wrong to be me.

cmd/flags.go
```diff
@@ -135,13 +138,21 @@ func defineFlagVar() *FlagVar {
...
-	flag.StringVar(&flagVar.remoteSyncNamespace, "sync-namespace", controller.DefaultRemoteSyncNamespace,
+	flag.StringVar(&flagVar.remoteSyncNamespace, "sync-namespace", DefaultRemoteSyncNamespace,
 		"Name of the namespace for syncing remote Kyma and module catalog")
 	flag.StringVar(&flagVar.caCertName, "ca-cert-name", defaultCaCertName,
 		"Name of the CA Certificate in Istio Namespace which is used to sign SKR Certificates")
 	flag.DurationVar(&flagVar.caCertCacheTTL, "ca-cert-cache-ttl", defaultCaCertCacheTTL,
 		"The ttl for the CA Certificate Cache")
-	flag.BoolVar(&flagVar.isKymaManaged, "is-kyma-managed", false, "indicates whether Kyma is managed")
+	flag.DurationVar(&flagVar.SelfSignedCertDuration, "self-signed-cert-duration", defaultSelfSignedCertDuration,
+		"The lifetime duration of self-signed certificate")
```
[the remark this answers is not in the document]
ruanxin: there is no limit of this unit, any valid time unit is fine, like `120s`, I added the minimum accept value from cert manager. ` minimum accepted duration is 1 hour.`

cmd/flags.go
```diff
@@ -192,5 +203,8 @@ type FlagVar struct {
 	caCertName                             string
 	caCertCacheTTL                         time.Duration
 	enableVerification                     bool
-	isKymaManaged                          bool
+	IsKymaManaged                          bool
```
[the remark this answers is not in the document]
ruanxin: like what I wrote [here](https://github.com/kyma-project/lifecycle-manager/pull/1088/files#r1412137248), this flag file should be in another package so this default value can be reused by other code. I think after that refactoring, all of the flag values should be defined as public.

internal/pkg/metrics/kyma.go
```diff
@@ -11,8 +11,8 @@ import (
 )
 
 const (
-	metricKymaState   = "lifecycle_mgr_kyma_state"
-	metricModuleState = "lifecycle_mgr_module_state"
+	MetricKymaState   = "lifecycle_mgr_kyma_state"
```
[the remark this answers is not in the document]
ruanxin: Actually I come across this reference const to test code in several conversations.

To be honest, the [reply](https://stackoverflow.com/a/23374335) from this guy is more or less aligned with my idea.

If we want to verify the const value should not be changed, then we should implement a test to specifically test the naming consistency, in this case, we indicate this purpose more specifically. 

I don't think reference this const value in other test code for other purpose is bad practice, why do we restrict it? Let's say if we need to reuse this value in 10 other different test scenario? How do we deal with it? define 10 different const in each individual test?

And in some complicated test scenario, if we just define another variable without reference the value in production code, it will even bring much more confusion to us, because we lack of the context of this value.

So it really depends on the test, we can't treat it as a general rule for all cases.

pkg/watcher/certificate_manager.go
ruanxin: Because it's the contract definition of this field in `certmanagerv1.CertificateSpec`

pkg/watcher/skr_webhook_manifest_manager.go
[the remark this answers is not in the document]
ruanxin: changed to `updateCertNotRenewMetrics`

pkg/watcher/certificate_manager.go
ruanxin: I changed to `time.Duration`, it's more suitable for this config

internal/pkg/metrics/kyma.go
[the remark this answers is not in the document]
ruanxin: Another argument from the code iteration point of view, we should not expect those values to never change.

Take these metrics as an example, it may have the request coming to rename it. Then if we need to adapt this change, with the reference approach, just need to rename the constant, adjust the naming consistent unit test, done. But if follow the idea never reference production code, it may end up into full text search and reply, even more complicate case like some value in test code is concatenated dynamically.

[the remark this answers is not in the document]
ruanxin: let's put it in another issue, to make it more awareness https://github.com/kyma-project/lifecycle-manager/issues/1109, also this task is not related to this issue.

lindnerby (approved):

pkg/watcher/certificate_manager.go
```diff
@@ -274,6 +284,28 @@ func (c *CertificateManager) GetCACertificateStatus(ctx context.Context) (certma
 	return cachedCertStatus, nil
 }
 
+func (c *CertificateManager) RemoveSecretAfterCARotated(ctx context.Context, kymaObjKey client.ObjectKey) error {
```
ruanxin, line 287: for review

pkg/watcher/certificate_manager.go
```diff
@@ -42,15 +43,26 @@ type SubjectAltName struct {
...
-	caCertName          string
-	additionalDNSNames  []string
+	kcpClient       client.Client
+	caCertCache     *CACertificateCache
+	certificateName string
+	secretName      string
+	config          CertificateConfig
+}
+
+type CertificateConfig struct {
```
ruanxin, line 53: for review

pkg/watcher/skr_webhook_manifest_manager.go
```diff
@@ -36,39 +41,27 @@ type SkrWebhookManagerConfig struct {
...
-	CACertificateCache *CertificateCache
 }
 
 const rawManifestFilePathTpl = "%s/resources.yaml"
 
 func NewSKRWebhookManifestManager(kcpConfig *rest.Config,
 	schema *machineryruntime.Scheme,
-	caCertificateCache *CertificateCache,
-	managerConfig *SkrWebhookManagerConfig,
+	caCertificateCache *CACertificateCache,
```
ruanxin, line 52: for review

tests/e2e_test/self_signed_certificate_rotation_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/kyma-project/lifecycle-manager/pkg/watcher"
+)
+
+var _ = Describe("Self Signed Certificate Rotation", Ordered, func() {
+	kyma := NewKymaWithSyncLabel("kyma-sample", "kcp-system", v1beta2.DefaultChannel,
+		v1beta2.SyncStrategyLocalSecret)
+	InitEmptyKymaBeforeAll(kyma)
+	CleanupKymaAfterAll(kyma)
+
+	Context("Given Kyma deployed in KCP", func() {
```
ruanxin, line 23: for review

internal/pkg/metrics/watcher.go
```diff
@@ -0,0 +1,39 @@
...
+
+const (
+	SelfSignedCertNotRenewMetrics = "lifecycle_mgr_self_signed_cert_not_renew"
+)
+
+type WatcherMetrics struct {
+	certNotRenewGauge *prometheus.GaugeVec
+}
+
+func NewWatcherMetrics() *WatcherMetrics {
```
ruanxin, line 17: for review

PR #1101, opened 2023-12-01
Enforce explain

jeremyharisch:
**Description**

- ...
- ...
- ...

**Related issue(s)**


internal/controller/purge_controller.go
```diff
@@ -45,55 +46,51 @@ type RemoteClientResolver func(context.Context, client.ObjectKey) (client.Client
 type PurgeReconciler struct {
 	client.Client
 	record.EventRecorder
+	logger                logr.Logger
```
lindnerby, line 49: Why is this change needed? This field isn't even initialized upon construction, but assigned every reconciliation now. You can either pass the logger to functions as a parameter where you moved the log statements to, or `logf.FromContext(ctx)` within them, or inject it when creating PurgeReconciler in main setup.

pkg/istio/client.go
```diff
@@ -231,7 +233,8 @@ func PrepareIstioHTTPRouteForCR(obj *v1beta2.Watcher) *istioapiv1beta1.HTTPRoute
 		Match: []*istioapiv1beta1.HTTPMatchRequest{
 			{
 				Uri: &istioapiv1beta1.StringMatch{
-					MatchType: &istioapiv1beta1.StringMatch_Prefix{ //nolint:nosnakecase
+					MatchType: &istioapiv1beta1.StringMatch_Prefix{
+						//nolint:nosnakecase // cannot be avoided
```
lindnerby, line 237:
```suggestion
						//nolint:nosnakecase // external type
```

pkg/log/logger.go
```diff
@@ -13,7 +13,7 @@ func ConfigLogger(level int8, syncer zapcore.WriteSyncer) logr.Logger {
 		level = -level
 	}
 	// The following settings is based on kyma community Improvement of log messages usability
-	//nolint:lll
+	//nolint:lll // link to logging documentation agreement
```
lindnerby, line 16: Let's generally exclude urls from lll linter in the config with a regex like this:
```
    # Exclude lll issues for long lines starting with an url
    - linters:
        - lll
      source: "^// http "
```

pkg/testutils/utils.go
```diff
@@ -1,4 +1,4 @@
-//nolint:wrapcheck
+//nolint:wrapcheck // errors do not need to be wrapped in unit and integration test usage
```
lindnerby, line 1: Is it possible to move this suppression to the linter config and exclude this file `pkg/testutils/utils.go` there?

pkg/types/file_error.go
```diff
@@ -6,7 +6,7 @@ import (
 
 // ParsedFile contains the parsed content and any error encountered during processing of a file.
 //
-//nolint:errname
+//nolint:errname // Error as part of struct for simplification
```
lindnerby, line 9: But could be an explicit field right? Let's change that then.

tests/e2e_test/suite_test.go
```diff
@@ -1,4 +1,4 @@
-//nolint:gochecknoglobals
+//nolint:gochecknoglobals // does not apply to unit and integration tests
```
lindnerby, line 1: Add path or pattern `_test.go` to linter config

tests/integration/controller/kyma/manifest_test.go
```diff
@@ -1,3 +1,4 @@
+//nolint:goerr113 // dynamic error creation in unit and integration tests is sufficient
```
lindnerby, line 1: Agreed, add to config instead?

lindnerby (changes requested):
Left some comments, especially for team discussion after daily.
We should add the `gochecknoglobals` exception to the linter config for test files/folders.

internal/controller/purge_controller.go
```diff
@@ -163,6 +158,17 @@ func (r *PurgeReconciler) dropPurgeFinalizer(ctx context.Context, kyma *v1beta2.
 	return nil
 }
 
+func (r *PurgeReconciler) deadlineReached(kyma *v1beta2.Kyma) (bool, time.Duration) {
```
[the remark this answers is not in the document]
jeremyharisch: I moved it to a seperate function since we hit the funlen of the Reconcile method, plus to simplify complexity. I would leave it like this since this is an internal function of this pkg and not a function which should be reused outside of it.

pkg/types/file_error.go
jeremyharisch: It is an explicit filed already, but the linter does not like that we name it `err`. Do you have a better name suggestion

internal/manifest/parse.go
```diff
@@ -20,10 +20,11 @@ import (
 	"github.com/kyma-project/lifecycle-manager/pkg/ocmextensions"
 )
 
-//nolint:gochecknoglobals
+//nolint:gochecknoglobals // in-memory cache used mutex
```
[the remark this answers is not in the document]
jeremyharisch: Will be covered in https://github.com/kyma-project/lifecycle-manager/issues/1125

pkg/cache/crd_cache.go
```diff
@@ -6,7 +6,7 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 )
 
-//nolint:gochecknoglobals
+//nolint:gochecknoglobals // in-memory cache used for CRDs
```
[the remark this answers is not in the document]
jeremyharisch: Will be covered in https://github.com/kyma-project/lifecycle-manager/issues/1125

api/v1beta2/moduletemplate_types.go
```diff
@@ -154,7 +154,7 @@ func (m *ModuleTemplate) GetDescriptor() (*Descriptor, error) {
 	return mDesc, nil
 }
 
-//nolint:gochecknoglobals
+//nolint:gochecknoglobals // in-memory cache used for descriptors
```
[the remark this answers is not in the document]
jeremyharisch: Will be covered in https://github.com/kyma-project/lifecycle-manager/issues/1125

pkg/remote/kyma_synchronization_context.go
```diff
@@ -23,8 +23,8 @@ import (
 
 type ClientFunc func() *rest.Config
 
-//nolint:gochecknoglobals
 var (
+	//nolint:gochecknoglobals // used for testing
```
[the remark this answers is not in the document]
jeremyharisch: Will be covered with https://github.com/kyma-project/lifecycle-manager/issues/1126

internal/controller/purge_controller.go
jeremyharisch: Reverted

PR #1105, opened 2023-12-04
feat: Configurable namespace for the Watcher VirtualService

Tomasz-Smelcerz-SAP:
**Description**

- ...
- ...
- ...

**Related issue(s)**


pkg/istio/client.go
```diff
@@ -30,27 +30,29 @@ var (
 
 type Client struct {
 	istioclient.Interface
-	eventRecorder record.EventRecorder
-	logger        logr.Logger
+	eventRecorder      record.EventRecorder
+	logger             logr.Logger
+	watcherVSNamespace string
```
nesmabadr, line 35: Why not just pass the watcherVsNamespace to the functions instead of having it in the Client interface? It is similar to the vsName

Tomasz-Smelcerz-SAP: You are right, for the `Client` interface we should expose namespace directly.

PR #1160, opened 2023-12-12
fix: Cache all secret instead of depends on label

ruanxin:
The owner of kyma secret changed to infrastructure-manager, and it changes the `operator.kyma-project.io/managed-by` from `lifecycle-manager` to it's own name, which brings the problem that LM can't find the secret, becasue it configured secret cache with this managed-by label. It indicates a problem that relies on a label managed by external component is not a reliable solution, and cached by this label also not brings so much performance benefits.

This PR did following changes:
1. changed cache option to allow cache all secrets.
2. removed [this](https://github.com/kyma-project/lifecycle-manager/pull/1160/files#diff-1c593a77c4fff452c608cb88f26717515a825e9bc04a1186655fb9e09ad6bfe7L19) unnecessary cache of istioingressgateway service. 
related issue: https://github.tools.sap/kyma/backlog/issues/4826


.github/workflows/lint-golangci.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v4
         with:
-          go-version: '1.21'
+          go-version-file: 'go.mod'
```
ruanxin, line 15: To reviewer:
I unified the way to get go version from go.mod, check the detail here https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file

nesmabadr:
I suggest removing the [Managedby label](https://github.com/kyma-project/lifecycle-manager/blob/087266465b2f53ee6640c94e9018d0c11a0d99e5/tests/e2e_test/utils_test.go#L107) from the secret created for the E2E test:

nesmabadr (approved):

PR #1171, opened 2023-12-14
feat: Safe background Kyma deletion

### LINKED ISSUE #1066: Implement safety measures for the Kyma background delete
### Description

Lifecycle Manager should be proofed against K8S background delete mode - this refers mainly to Kyma CR but other resources may also be affected.

In particular - we need to make sure that Manifest CRs are removed before Kyma CR is gone, this would require setting up an additional finalizer on Kyma CR

### Reasons

We want to avoid potential issues related to not using foreground delete mode in the Modularization setup.

### Acceptance Criteria

- Describe the **KCP Kyma CR** deletion
  - Given an **SKR Cluster**
    - And a **KCP Kyma CR** is created on the **KCP Cluster**
    - And a related **Manifest CR** is created on the **KCP Cluster**
  - When the **Kyma CR** is deleted using background mode
    - Then the **Manifest CR** is removed before **Kyma CR** is gone

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

nesmabadr:
**Description**

- on KCP kyma deletion, if the manifest CRs are not yet deleted, then delete them first before removing Kyma finalizer
- Added an E2E test

**Related issue(s)**
Resolves #1066


tests/e2e/background_kyma_deletion_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	. "github.com/kyma-project/lifecycle-manager/pkg/testutils"
+)
+
+var _ = Describe("Background Kyma Deletion", Ordered, func() {
```
ruanxin, line 14: I didn't do code review, but for the refactoring of e2e tests, I think for the existing e2e tests which delete kyma, it should use default `Background` deletion, and only have one dedicated for `Foreground`.

internal/controller/kyma_controller.go
```diff
@@ -405,21 +406,60 @@ func (r *KymaReconciler) handleDeletingState(ctx context.Context, kyma *v1beta2.
...
+	}
 
 	controllerutil.RemoveFinalizer(kyma, shared.KymaFinalizer)
+
 	return ctrl.Result{Requeue: true}, r.updateKyma(ctx, kyma)
 }
 
+func (r *KymaReconciler) deleteManifests(ctx context.Context, manifests []v1beta2.Manifest) error {
+	for i := range manifests {
+		if err := r.Delete(ctx, &manifests[i]); err != nil {
```
jeremyharisch, line 441:
```suggestion
		if err := r.Delete(ctx, &manifests[i]); client.IgnoreNotFound(err) != nil {
```

internal/controller/kyma_controller.go
```diff
@@ -405,21 +406,60 @@ func (r *KymaReconciler) handleDeletingState(ctx context.Context, kyma *v1beta2.
...
+		return ctrl.Result{}, err
+	}
+
+	if !r.relatedManifestCRsAreDeleted(relatedManifests) {
+		if err = r.deleteManifests(ctx, relatedManifests); err != nil {
+			r.enqueueWarningEvent(kyma, deletionError, err)
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{Requeue: true}, nil
+	}
```
jeremyharisch, line 432:
```suggestion
	relatedManifests, err := r.getRelatedManifestCRs(ctx, kyma)
	if err != nil {
		err = fmt.Errorf("error while trying to get manifests: %w", err)
		r.enqueueWarningEvent(kyma, deletionError, err)
		return ctrl.Result{}, err
	}

	if !r.relatedManifestCRsAreDeleted(relatedManifests) {
		if err = r.deleteManifests(ctx, relatedManifests); err != nil {
			r.enqueueWarningEvent(kyma, deletionError, err)
			return ctrl.Result{}, err
		}
		return ctrl.Result{Requeue: true}, nil
	}
	r.Metrics.CleanupMetrics(kyma.Name)
```

I would prefer first cleaning up manifests, before cleaning up the metrics

pkg/remote/kyma_synchronization_context.go
```diff
@@ -87,7 +87,9 @@ func RemoveFinalizerFromRemoteKyma(
 		return err
 	}
 
-	controllerutil.RemoveFinalizer(remoteKyma, shared.KymaFinalizer)
+	for _, finalizer := range remoteKyma.Finalizers {
+		controllerutil.RemoveFinalizer(remoteKyma, finalizer)
+	}
```
jeremyharisch, line 92: Just for clarification: Is this change needed because of your other changes? Since you did not introduce any new finalizers. This is kinda like a purge mechanism, if I got it correct.

jeremyharisch (commented):
Smaller comments, rest looks fine

pkg/remote/kyma_synchronization_context.go
nesmabadr: This is not related to my change, but we don't sync the finalizers to the remote kyma. I added this just to remove any finalizers on the remote kyma if there was any added by the user

tests/e2e/kyma_deletion_test.go
```diff
@@ -3,6 +3,8 @@ package e2e_test
 import (
 	"os/exec"
 
+	apimetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
ruanxin, line 6: Please rename this test to indicate it's testing delete kyma by foreground.

ruanxin: no, this test is not designed to test foreground, it's testing kyma get deleted and module under stuck, then Kyma CR only get deleted when Kubeconfig Secret get deleted. This one should also test against background delete.

internal/controller/kyma_controller.go
```diff
@@ -405,21 +406,61 @@ func (r *KymaReconciler) handleDeletingState(ctx context.Context, kyma *v1beta2.
...
 			err = fmt.Errorf("error while trying to remove finalizer from remote: %w", err)
 			r.enqueueWarningEvent(kyma, deletionError, err)
 			return ctrl.Result{}, err
 		}
 
-		logger.Info("removed remote finalizer")
+		logger.Info("removed remote finalizers")
+	}
+
+	relatedManifests, err := r.getRelatedManifestCRs(ctx, kyma)
```
ruanxin, line 418: I think it makes sense you introduce a function named `cleanupManifestCRs` to encapsulate all this code, they are highly cohesion.

internal/controller/kyma_controller.go
```diff
@@ -405,21 +406,73 @@ func (r *KymaReconciler) handleDeletingState(ctx context.Context, kyma *v1beta2.
...
 
+func (r *KymaReconciler) cleanupManifestCRs(ctx context.Context, kyma *v1beta2.Kyma) (bool, error) {
+	relatedManifests, err := r.getRelatedManifestCRs(ctx, kyma)
+	if err != nil {
+		err = fmt.Errorf("error while trying to get manifests: %w", err)
+		r.enqueueWarningEvent(kyma, deletionError, err)
+		return false, err
+	}
+
+	if !r.relatedManifestCRsAreDeleted(relatedManifests) {
```
ruanxin, line 441: if you check positive `if r.relatedManifestCRsAreDeleted(relatedManifests)` will make this code much clean.

```
if r.relatedManifestCRsAreDeleted(relatedManifests) {
  return false, nil
}
if err = r.deleteManifests(ctx, relatedManifests); err != nil {
	r.enqueueWarningEvent(kyma, deletionError, err)
	return true, err
}
return true, nil

```

tests/e2e/warning_status_propagation_test.go
```diff
@@ -18,10 +18,10 @@ var _ = Describe("Warning Status Propagation", Ordered, func() {
 	module := NewTemplateOperator(v1beta2.DefaultChannel)
 	moduleCR := NewTestModuleCR(remoteNamespace)
 
-	InitEmptyKymaBeforeAll(kyma)
-	CleanupKymaAfterAll(kyma)
-
 	Context("Given SKR Cluster", func() {
+		InitEmptyKymaBeforeAll(kyma)
```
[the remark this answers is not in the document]
nesmabadr: I forgot to revert it (was trying to use DescribeTable)

pkg/testutils/manifest.go
```diff
@@ -245,3 +246,41 @@ func ManifestNoDeletionTimeStampSet(ctx context.Context,
 	}
 	return nil
 }
+
+func AddFinalizerToManifestCR(ctx context.Context, clnt client.Client, kymaName, kymaNamespace, moduleName,
```
ruanxin, line 250: I can't find usage for these two function

internal/controller/kyma_controller.go
```diff
@@ -405,21 +406,73 @@ func (r *KymaReconciler) handleDeletingState(ctx context.Context, kyma *v1beta2.
...
 	controllerutil.RemoveFinalizer(kyma, shared.KymaFinalizer)
+
 	return ctrl.Result{Requeue: true}, r.updateKyma(ctx, kyma)
 }
 
+func (r *KymaReconciler) cleanupManifestCRs(ctx context.Context, kyma *v1beta2.Kyma) (bool, error) {
+	relatedManifests, err := r.getRelatedManifestCRs(ctx, kyma)
+	if err != nil {
+		err = fmt.Errorf("error while trying to get manifests: %w", err)
+		r.enqueueWarningEvent(kyma, deletionError, err)
```
ruanxin, line 437: this `r.enqueueWarningEvent` can be extracted to upper layer

```
manifestsExist, err := r.cleanupManifestCRs(ctx, kyma)
if err != nil {
    r.enqueueWarningEvent
   return ctrl.Result{}, err
}
```

I'm fine with this return `(bool, error)`, however for this function, if you introduce here a customized error when manifest CR still exists, you just need to return error, and you don't need to deal with `err` and `manifestsExist` specifically, wdyt?
```
if r.relatedManifestCRsAreDeleted(relatedManifests) {
    return ErrManifestCRStillExists
}
```

nesmabadr: But this will not work in the case there is no error but we call the delete manifests. We want to requeue in this case so that we are sure that the manifests are deleted before we can remove the Kyma finalizer. If we just rely on the error, we can't do this

ruanxin: sorry, I'm wrong, this `ErrManifestCRStillExists` is not returned in that condition, but in the end.
```
if r.relatedManifestCRsAreDeleted(relatedManifests) {
    return nil
}

if err = r.deleteManifests(ctx, relatedManifests); err != nil {
    return err
}
return ErrManifestCRStillExists
```

internal/controller/kyma_controller.go
```diff
@@ -405,21 +408,73 @@ func (r *KymaReconciler) handleDeletingState(ctx context.Context, kyma *v1beta2.
...
 	controllerutil.RemoveFinalizer(kyma, shared.KymaFinalizer)
+
 	return ctrl.Result{Requeue: true}, r.updateKyma(ctx, kyma)
 }
 
+func (r *KymaReconciler) cleanupManifestCRs(ctx context.Context, kyma *v1beta2.Kyma) error {
+	relatedManifests, err := r.getRelatedManifestCRs(ctx, kyma)
+	if err != nil {
+		err = fmt.Errorf("error while trying to get manifests: %w", err)
+		return err
```
ruanxin, line 440: you can `return fmt.Errorf("error while trying to get manifests: %w", err)`

internal/controller/kyma_controller.go
```diff
@@ -405,21 +408,73 @@ func (r *KymaReconciler) handleDeletingState(ctx context.Context, kyma *v1beta2.
...
+	}
+
+	if r.relatedManifestCRsAreDeleted(relatedManifests) {
+		return nil
+	}
+
+	if err = r.deleteManifests(ctx, relatedManifests); err != nil {
+		err = fmt.Errorf("error while trying to delete manifests: %w", err)
+		return err
```
ruanxin, line 449: same here

ruanxin (approved):

PR #1172, opened 2023-12-14
fix: Delete SKR Certificate and Secret only if they exist

nesmabadr:
**Description**

- Delete SKR Certificate and Secret only if they exist. The current behavior executed the deletion of the certificate/secret even if they are not found which cause infinite reconciliations when the Kyma gets deleted. 

**Related issue(s)**

### DIFF
--- a/pkg/watcher/certificate_manager.go
+++ b/pkg/watcher/certificate_manager.go
@@ -124,7 +124,7 @@ func (c *CertificateManager) RemoveCertificate(ctx context.Context) error {
 		return fmt.Errorf("failed to get certificate: %w", err)
 	}
 
-	if !util.IsNotFound(err) {
+	if err == nil {
 		if err = c.kcpClient.Delete(ctx, certificate); err != nil {
 			return fmt.Errorf("failed to delete certificate: %w", err)
 		}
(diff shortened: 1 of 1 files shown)


pkg/watcher/certificate_manager.go
```diff
@@ -116,33 +116,39 @@ func (c *CertificateManager) Remove(ctx context.Context) error {
...
 		Namespace: c.config.IstioNamespace,
-	}, certificate); err != nil && !util.IsNotFound(err) {
+	}, certificate)
+	if err != nil && !util.IsNotFound(err) {
 		return fmt.Errorf("failed to get certificate: %w", err)
 	}
 
-	if err := c.kcpClient.Delete(ctx, certificate); err != nil {
-		return fmt.Errorf("failed to delete certificate: %w", err)
+	if !util.IsNotFound(err) {
```
ruanxin, line 127: you already return error above when `!util.IsNotFound(err) `, is it possible to enter here?

nesmabadr: The function continues execution if the error is not nil and the !util.IsNotFound(err), so we don't want to do the delete call if the error is not found. Or I can have the check as `if err == nil`

ruanxin: yes, then you should not use `!util.IsNotFound(err)`, that's misleading.

jeremyharisch (approved):

Issue #1052, opened 2023-11-16
[Mandatory Modules] - Reconciliation loop


lindnerby:
### Description

Implement New Reconciliation Loop for Handling Mandatory Modules.

### Acceptance Criteria

- [x] Create an additional control loop for Kyma objects that will generate mandatory module Manifest CRs, ensuring the owner reference for the Manifest CR is related to the Kyma CR.
- [x] No changes should be required for the Manifest Controller. This should be verified via end-to-end tests.
- [x] Normal and mandatory modules are accurately differentiated, with mandatory modules not appearing in the Kyma Status.
- [x] If the 'skip reconciliation' condition applies, it should also apply to the mandatory modules.
- [x] When a user marks a custom module template as mandatory in the SKR Kyma or the KCP Kyma and enables the mandatory module in the Kyma CR, the LM should refuse to accept it. Instead, it should place the module in the error state within the status.modules list.
- [x] Sync to SKR must be excluded for mandatory modules.
- [x] Create a follow-up issue to implement metrics that will include all the relevant details.
    - [ ] https://github.com/kyma-project/lifecycle-manager/issues/1184
- [x] Create a follow-up issue to implement e2e test cases.
    - [ ] https://github.com/kyma-project/lifecycle-manager/issues/1185
- [x] All functionalities are thoroughly tested (integration test).
- [x] Create a follow-up issue to document the feature.
    - [ ] https://github.com/kyma-project/lifecycle-manager/issues/1151

### Feature Testing

End-to-End tests, Integration tests, Unit tests

ruanxin:
also don't forget to create documentation regarding this new controller https://github.com/kyma-project/lifecycle-manager/blob/main/docs/technical-reference/controllers.md

PR #1173, opened 2023-12-14
feat: Mandatory modules

jeremyharisch:
**Description**

- Implements #1052 
- Renamed `channel` pkg to `lookup`
- New integration test suite for mandatory modules
- New test case for Kyma integration test
- Requeue Kyma in mandatory Reconciler by Watching the following:
    - Kyma CR labels or generation changed
    - Mandatory ModuleTemplate labels or generation changed

**Related issue(s)**

#1052


pkg/lookup/lookup.go
```diff
@@ -82,6 +83,29 @@ func GetTemplates(
 	return templates
 }
 
+// GetMandatoryTemplates returns all ModuleTemplates for modules which are marked as mandatory with the given channel.
+func GetMandatoryTemplates(ctx context.Context, kymaClient client.Reader) (ModuleTemplatesByModuleName,
```
ruanxin, line 87: I don't think it's a good idea put this function inside this package. The main purpose for the original file is designed to deal with regular template, and we have complicated lookup strategy, that's why `GetTemplates` have so many private methods depends on, but your `GetMandatoryTemplates` basically depends on nothing within this file, which decreased the code cohesion.

Try to create a new file to put this code, and you can also think of renaming `GetTemplates` to something more specific like `GetNormalTemplates` to distinguish between them.

pkg/lookup/lookup.go
```diff
@@ -82,6 +83,29 @@ func GetTemplates(
...
+	error,
+) {
+	moduleTemplateList := &v1beta2.ModuleTemplateList{}
+	if err := kymaClient.List(ctx, moduleTemplateList); err != nil {
```
ruanxin, line 91: I suggest we create an issue to introduce a label for this mandatory moduletemplate, We can implement some logic in mutationwebhook, when a mandatory module created in cluster, add a label for it, then we can change this list based on label, so that we can also remove that customized filtering in the following steps.

pkg/lookup/lookup.go
```diff
@@ -82,6 +83,29 @@ func GetTemplates(
...
+		return nil, fmt.Errorf("could not list ModuleTemplates: %w", err)
+	}
+
+	mandatoryModules := make(map[string]*ModuleTemplateTO)
+	for _, moduleTemplate := range moduleTemplateList.Items {
+		moduleTemplate := moduleTemplate
+		if moduleTemplate.Spec.Mandatory {
+			mandatoryModules[moduleTemplate.Name] = &ModuleTemplateTO{
+				ModuleTemplate: &moduleTemplate,
+				DesiredChannel: moduleTemplate.Spec.Channel,
```
ruanxin, line 101: This `DesiredChannel` should not coming from moduletemplate, and even not assigned to a MandatoryTemplate, right? There is no channel required by MandatoryTemplate, we will probably make channel as optional field for `ModuleTemplate` in next API upgrade. To assign a DesiredChannel for MandatoryTemplate is confusion here.

pkg/module/parse/template_to_module.go
```diff
@@ -106,7 +108,69 @@ func (p *Parser) GenerateModulesFromTemplates(ctx context.Context,
 	return modules
 }
 
-func overwriteNameAndNamespace(template *channel.ModuleTemplateTO, name, namespace string) {
+func (p *Parser) GenerateMandatoryModulesFromTemplates(ctx context.Context,
```
ruanxin, line 111: compare the logic of this method with `GenerateModulesFromTemplates`, lots of code is copy pasted, think about extract shared code into functions to increase the code reusability.

pkg/watch/mandatory_template_change.go
```diff
@@ -0,0 +1,68 @@
...
+		err := h.List(ctx, kymas, listOptions)
+		if err != nil {
+			return requests
+		}
+
+		templateName := types.NamespacedName{
+			Namespace: template.GetNamespace(),
+			Name:      template.GetName(),
+		}
+		for _, kyma := range kymas.Items {
```
ruanxin, line 53: same as this `Watch()` function, think about some code can be shared with `(h *TemplateChangeHandler) Watch()`.

for example, to append `requests`, basically you can define a function like `enqueueChangedKyma( kyma []v1beta2.kyma), then it can be shared.

And we already know basically log message not help us in most case, please don't introduce more log, instead, think about use events or put into resources status. For this case, I suggest remove this info log, together with the one from `(h *TemplateChangeHandler) Watch()`

tests/integration/controller/kyma/helper_test.go
```diff
@@ -24,12 +25,24 @@ const (
...
 
 func RegisterDefaultLifecycleForKyma(kyma *v1beta2.Kyma) {
+	mandatoryTemplate := builder.NewModuleTemplateBuilder().
+		WithModuleName("mandatory-template-operator").
+		WithChannel("mandatory").
+		WithMandatory(true).
+		WithOCM(compdescv2.SchemaVersion).Build()
+
 	BeforeAll(func() {
+		Eventually(CreateCR, Timeout, Interval).
```
ruanxin, line 35: It's better you define a function `DeployMandatoryModuleTemplates`

tests/integration/controller/kyma/helper_test.go
```diff
@@ -24,12 +25,24 @@ const (
 )
 
 func RegisterDefaultLifecycleForKyma(kyma *v1beta2.Kyma) {
+	mandatoryTemplate := builder.NewModuleTemplateBuilder().
+		WithModuleName("mandatory-template-operator").
+		WithChannel("mandatory").
```
ruanxin, line 30: same as what described [here](https://github.com/kyma-project/lifecycle-manager/pull/1173/files#r1431678322), for mandatory template, the channel is an irrelevant field.

If you define `mandatory` channel as a test case, it leaves the impression that this channel is required by mandatory template, what you can do is define some variable which indicate this field can be ignored, like `dummy-channel`.

tests/integration/controller/kyma/kyma_test.go
```diff
@@ -176,6 +177,33 @@ var _ = Describe("Kyma enable one Module", Ordered, func() {
...
+var _ = Describe("Kyma enable one Mandatory Module", Ordered, func() {
+	kyma := NewTestKyma("mandatory-module-kyma")
+
+	RegisterDefaultLifecycleForKyma(kyma)
+
+	It("should result Kyma in Error state", func() {
+		By("enabling one mandatory Module")
+		kyma.Spec.Modules = append(kyma.Spec.Modules, v1beta2.Module{
+			Name:    "mandatory-template-operator",
+			Channel: "mandatory",
```
ruanxin, line 189: although this channel should be irrelevant, we should design a test based on the real use case, we don't have a channel `mandatory`, if assign a module to `spec.modules`, either use a default regular channel or simply not define the channel but relies on kyma global channel.

tests/integration/controller/mandatorymodules/mandatory_modules_controller_test.go
```diff
@@ -0,0 +1,151 @@
...
+	ErrNoMandatoryManifest = errors.New("manifest for mandatory Module not found")
+)
+
+var _ = Describe("Mandatory Module Installation", Ordered, func() {
+	Context("Given Kyma with no Module and one mandatory ModuleTemplate on Control-Plane", func() {
+		kyma := NewTestKyma("no-module-kyma")
+		registerControlPlaneLifecycleForKyma(kyma)
+
+		It("Then Kyma CR should result in a ready state immediately as there are no modules", func() {
+			Eventually(KymaIsInState, Timeout, Interval).
```
ruanxin, line 37: Let's introduce global timeout and interval to suite_test.go, like what we did in [e2e test](https://github.com/kyma-project/lifecycle-manager/blob/8b74fb092d3013565b6c10a5c297092aeda7fa57/tests/e2e/suite_test.go#L91-L94), and don't continue doing this individual configure to new test.

tests/integration/controller/mandatorymodules/mandatory_modules_controller_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+		It("And Manifest CR for the Mandatory Module should be created with correct Owner Reference", func() {
+			Eventually(checkMandatoryManifestForKyma).
+				WithContext(ctx).
+				WithArguments(kyma, "kyma-project.io/template-operator").
+				Should(Succeed())
+		})
+	})
+
+	Context("Given Kyma with no Module and one mandatory ModuleTemplate on Control-Plane", func() {
```
ruanxin, line 65: This one can be another test, so that we can run them in parrallel

tests/integration/controller/mandatorymodules/mandatory_modules_controller_test.go
```diff
@@ -0,0 +1,151 @@
...
+		It("When Kyma has 'skip-reconciliation' label, then no Mandatory Module Manifest should be created", func() {
+			Eventually(checkMandatoryManifestForKyma).
+				WithContext(ctx).
+				WithArguments(kyma, "kyma-project.io/template-operator").
+				Should(Equal(ErrNoMandatoryManifest))
+		})
+	})
+})
+
+func registerControlPlaneLifecycleForKyma(kyma *v1beta2.Kyma) {
```
ruanxin, line 79: Please don't continue to introduce this function into different integration test packages, instead think about a solution to make those tests share this code.

Basically, we have `RegisterDefaultLifecycleForKyma` and `RegisterDefaultLifecycleForKymaWithoutTemplate` in `integration/controller/kyma`, those two functions can be extracted to `integration` package under a `helper.go` file, if you just refactor the parameters to accept `context.Context` and `client.Client`, then you can make those two function used by each test packages, of cause, change will affect many existing code if you apply those new function into existing code, but then you can create followup issue for that refactoring dedicated.

And for your function here, just need to reuse `RegisterDefaultLifecycleForKymaWithoutTemplate`, and additionally deploy your mandatory template.

A general guide here, if you can simply copy some existing code to achieve, always think upfront if you can refactor it to be reused.

pkg/lookup/lookup.go
jeremyharisch: https://github.com/kyma-project/lifecycle-manager/issues/1198

Can be covered with this issue

tests/integration/controller/mandatorymodules/mandatory_modules_controller_test.go
jeremyharisch: Implemented

pkg/watch/mandatory_template_change.go
jeremyharisch: I removed the logger. But atm there is not any relevant code that can be shared. For example the append you mentioned is a one liner; do you think for `requests = append(requests, reconcile.Request{NamespacedName: kymaName})` we should have a function? But this is just a normal list append.

We can have a talk what you would like to have simplified

tests/integration/controller/kyma/helper_test.go
jeremyharisch: I dont get what this brings us as value 🤔 
The CreateCR function looks like
```
func CreateCR(ctx context.Context, clnt client.Client, obj client.Object) error {
	err := clnt.Create(ctx, obj)
	if !apierrors.IsAlreadyExists(err) {
		return err
	}
	return nil
}
```

And what would the `DeployMandatoryModuleTemplate` do differently? It should also take a ModuleTemplateCR and create it. I do not want to have a function with hardcoded mandatory module templates to stay flexible and want to make use of the templateBuilder

jeremyharisch:
> Please check my comments, mainly code quality concern, also an **e2e test** is missing.

E2E will be covered with https://github.com/kyma-project/lifecycle-manager/issues/1185

pkg/templatelookup/regular.go
```diff
@@ -203,7 +205,8 @@ type Lookup interface {
 	WithContext(ctx context.Context) (*ModuleTemplateTO, error)
 }
 
-func NewTemplateLookup(client client.Reader,
+// New returns a new instance of TemplateLookup.
+func New(client client.Reader,
```
ruanxin, line 209: If you renamed to `New`, it's too generic, and this function only used for regular template, but not mandatory. Since you put both under this  `templatelookup` package, there is no idea for the user calling `templatelookup.New`, is it for regular template, or mandatory, or both?

tests/integration/controller/kyma/helper_test.go
ruanxin: what I mean is have a function to make this `BeforeAll` block looks clearly, same pattern as that `DeployModuleTemplates`, `DeleteModuleTemplates`

Then this block can be like this, it's not prevent you reuse `CreateCR`
```
BeforeAll(func() {
  DeployMandatoryModuleTemplates(ctx, controlPlaneClient)
  DeployModuleTemplates(ctx, controlPlaneClient, kyma)
})

func DeployMandatoryModuleTemplates(ctx context.Context, kcpClient client.Client) error {
	mandatoryTemplate := builder.NewModuleTemplateBuilder().
		WithModuleName("mandatory-template-operator").
		WithChannel(mandatoryChannel).
		WithMandatory(true).
		WithOCM(compdescv2.SchemaVersion).Build()
	return CreateCR(ctx, kcpClient, mandatoryTemplate)
}
```

Makefile
```diff
@@ -3,7 +3,7 @@
 APP_NAME = lifecycle-manager
 IMG_REPO := $(DOCKER_PUSH_REPOSITORY)$(DOCKER_PUSH_DIRECTORY)
 IMG_NAME := $(IMG_REPO)/$(APP_NAME)
-IMG := $(IMG_NAME):$(DOCKER_TAG)
+IMG := IMG := $(IMG_NAME):$(DOCKER_TAG)
```
ruanxin, line 6: is this committed by mistake?

tests/integration/controller/kyma/helper_test.go
jeremyharisch: ok, will do that

pkg/watch/template_change.go
```diff
@@ -3,71 +3,62 @@ package watch
...
+				templateUsed = true
+				break
 			}
-
-			requests = append(requests, reconcile.Request{NamespacedName: kymaName})
 		}
-
-		return requests
+		if !templateUsed {
+			return nil
```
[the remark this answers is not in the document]
jeremyharisch: Thanks for the hint, but it is in general false XD We do not want to return nothgin if one Kyma does not use the template, but continue with the rest of the kymas to check, will fix this

tests/integration/controller/kyma/helper_test.go
```diff
@@ -71,9 +63,14 @@ func RegisterDefaultLifecycleForKymaWithoutTemplate(kyma *v1beta2.Kyma) {
 func DeleteModuleTemplates(ctx context.Context, kcpClient client.Client, kyma *v1beta2.Kyma) {
 	for _, module := range kyma.Spec.Modules {
 		template := builder.NewModuleTemplateBuilder().
+			WithName(createModuleTemplateName(module)).
 			WithModuleName(module.Name).
 			WithChannel(module.Channel).
 			WithOCM(compdescv2.SchemaVersion).Build()
+		Eventually(RegularModuleTemplateExists, Timeout, Interval).
```
ruanxin, line 69: I think we don't need this check, it can happen some integration test testing moduletemplate deletion, this `AfterAll` step is just clean up resources, it's valid case that some resources already gone.

ruanxin (approved):

Issue #1028, opened 2023-11-10
resources deployed by lifecycle manager will constantly change


k15r:
**Description**

resources deployed by the lifecycle-manager get constantly updated to even without any need for updating.

- enable eg. `serverless` on any cluster controlled by the lifecycle-manager
- run `k get sa -n kyma-system serverless-operator -w -ojsonpath='{.metadata.name}{"\t"}{.metadata.resourceVersion}{"\n"}'`
even if serverless is not used (no function configured) the resourceVersion of the serverless-operator serviceAccount constantly (~30sec) changes

this scenario applies to all resources applied by the lifecycle-manager (excluding the default-cr)

this probably puts way more load on the api-server than necessary.

**Expected result**

resources that do not need to be changed, dont change

**Actual result**

resources that do not need to be changed change regularly

**Steps to reproduce**

**Troubleshooting**

ruanxin:
it's expected behaviour, not a bug, the LM have an interval doing server side apply to patch all the module resources, it's not designed to only update resources when change happens. However, to update the resourceversion is decided by api server, I believe for the service account, api-server decided to update the resource in each patch no matter if there is difference between old and new resources. But for other resources like deploy, apiserver will not update it if there is no change.

pbochynski:
@ruanxin Can you double check the code with server side apply. From my own experience I can say that SSA does not generate new resource versions if you apply the same content several times. It was a bug in previous kubernetes versions but it should be fixed now.
https://github.com/kubernetes/kubernetes/issues/119304

janmedrek:
We leave the matter of upgrading the `resourceVersion` to the K8S API Server, we cannot really influence the way it is handled by the K8S. We will keep on monitoring the issue though.

ruanxin:
followup: https://github.com/kyma-project/lifecycle-manager/issues/1167

PR #1174, opened 2023-12-14
fix: Make sure ssa not update resourceVersion

ruanxin:
controller runtime client can do patch directly with unstructured object, doesn't need to convert to typed object, and by this conversion, it brings side effects that make API server treat the object as a different one, which leads to the bug reported here. https://github.com/kyma-project/lifecycle-manager/issues/1028
```
resource.Object = c.convertUnstructuredToTyped(resource.Object, resource.Mapping)
```


tests/e2e/module_consistency_test.go
```diff
@@ -0,0 +1,95 @@
...
+				WithArguments(runtimeClient, "template-operator-controller-manager", TestModuleResourceNamespace).
+				Should(Succeed())
+
+			By("And KCP Kyma CR is in \"Ready\" State")
+			Eventually(KymaIsInState).
+				WithContext(ctx).
+				WithArguments(kyma.GetName(), kyma.GetNamespace(), controlPlaneClient, shared.StateReady).
+				Should(Succeed())
+		})
+		It("Then synced module resources keep consistent by verify resourceVersion not update", func() {
```
nesmabadr, line 51: I think this could be renamed to `And synced module resources remain consistent with the same resourceVersion`. And this should be a By in the previous `It` block

tests/e2e/module_consistency_test.go
```diff
@@ -0,0 +1,95 @@
...
+					gvk)
+				Expect(err).Should(Succeed())
+				Consistently(IsResourceVersionSame).
+					WithContext(ctx).
+					WithArguments(runtimeClient, objectKey, gvk,
+						resourceInCluster.GetResourceVersion()).Should(BeTrue())
+			}
+		})
+
+		It("Then synced module resources keep consistent by revert modification", func() {
```
nesmabadr, line 69: I think this statement is not clear enough, and also this is should start with an And and be moved to the previous It block

pkg/testutils/unstructured.go
```diff
@@ -0,0 +1,116 @@
...
+	if err := clnt.Get(ctx, objectKey, obj); err != nil {
+		if util.IsNotFound(err) {
+			return obj, nil
+		}
+		return nil, err
+	}
+	return obj, nil
+}
+
+func IsResourceVersionSame(ctx context.Context, clnt client.Client,
```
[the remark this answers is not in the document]
ruanxin: for a function with boolean as return value, use `is` or `if` as prefix is a common naming pattern, you can search more detail https://www.bing.com/search?q=+boolean+function+naming+pattern&qs=n&form=QBRE&sp=-1&lq=0&pq=+boolean+function+naming+pattern&sc=11-32&sk=&cvid=425A6121A15A43898F423A6AD303CD29&ghsh=0&ghacc=0&ghpl=

tests/e2e/module_consistency_test.go
ruanxin: I agree with the title, but put this in the previous `It` block, it affects the readability, this big truck of code is worth a dedicated block, and previous `It` block is checks in general this module deployed, this block check if they are not changed.

pkg/testutils/unstructured.go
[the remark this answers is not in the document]
nesmabadr: Yes but this one is used for the tests. so we have this naming convention here because it would be used with Eventually or Consistently then it would be read as a statement. It is similar to CRIsInState for example.

tests/e2e/module_consistency_test.go
nesmabadr: If it's not placed in the previous It block, the e2e report will have incorrect indentation before this statement as far as I know. For the sake of this, we could extract the logic to check the resources version in a function and call the function here and then it would be more readable WDYT?

tests/e2e/module_consistency_test.go
ruanxin: Do you have a better suggestion for this? 😄 

Regarding this BDD syntax, I think we need to clarify here, it's not strictly defined that one test case only allow one `Then` keyword, I agree with `And` it makes the statement more fluidly, but multiple `Then` in most case is more strateforward. Check the `And, But` section in Gherkin [document](https://cucumber.io/docs/gherkin/reference/)

And same as the reason for this block, it checks specifically resource get reset after user modified, it's not good mix with first `It` block, they address different concern.

pkg/testutils/unstructured.go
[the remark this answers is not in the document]
ruanxin: something defined as **naming conversion** should be consider as a widely accepted concept. I don't think we should treat the naming of the function in test as a special case, so I can't say this `CRIsInState` is so called naming conversion.

But of course, we can have a team agreement that we have some teamwised consensus for certain naming rules, but then lets bring this to the team before get decision.

tests/e2e/module_consistency_test.go
nesmabadr: It's not strict but IMO if you read the whole test as what's generated in the report then you'd expect all the Thens are concatenated with And, except if there's a new When (or an action that took place in order for this result to happen).

pkg/testutils/manifest.go
```diff
@@ -142,19 +143,20 @@ func GetManifestResource(ctx context.Context,
 	return moduleInCluster.Spec.Resource, nil
 }
 
-func AddSkipLabelToManifest(
+func ConfigSkipLabelToManifest(
```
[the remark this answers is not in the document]
ruanxin: renamed to `SetSkipLabelInManifest`, the concern here is reusable of the function. With `AddSkipLabelToManifest`, then it's only limited the functionality of it to add skip label, but if you need to remove the label, then need to have another one `DeleteSkipLabelToManifest`, and the code is similar, which not make so much sense to me.

We have this topic discussed before, I understand your concern here is by reading the name you can understand the functionality, but this is one aspect while design the function, if this is the only rule you follow with, then the outcome will be we have many functions which did similar thing which increase the code rigidity.

Also to understand the function, naming is only one part, we should combine with all parameters together, with this in concern, if we name it as `SetSkipLabelInManifest(... ifSkip bool)`, then it's also clean to me, the purpose of this function is change the skip label.

nesmabadr (approved):

mmitoraj (approved):

Issue #1112, opened 2023-12-05
Requeue reason metric


lindnerby:
### Description

We should add a metric to track the reason of requeue in Kyma control loop.
Name: `lifecycle_mgr_requeue_reason`
Add labels for reason accordingly

### Reasons

As of now we can observe an increase of immediate requeues in lifecycle-manager's control loop of Kymas, as a result of returning `ctrl.Result{Requeue: true}` in various places.
We are interested in the reason of these requeues to figure out which ones are the most frequent ones and how to adapt them, because they may be flooding the worker queues for no valid reason.

### Acceptance Criteria

- [x] Introduce a new metric for tracking the requeue reasons (counter)
  - consider introducing a label for the requeue reason
- [x] Panel added (to the overview dashboard)

ruanxin:
success requeue on STAGE between 11.17.2023 to 03.12.2023

![image](https://github.com/kyma-project/lifecycle-manager/assets/2189071/2a331714-5c22-41b0-8fd0-7128f9532bc5)

PR #1192, opened 2023-12-21
feat: Kyma requeue reason metrics

nesmabadr:
**Description**

- Add metrics with label for the requeue of Kyma reconciliation

**Related issue(s)**
#1112

### DIFF
--- a/internal/controller/kyma_controller.go
+++ b/internal/controller/kyma_controller.go
@@ -347,12 +347,11 @@ func (r *KymaReconciler) handleInitialState(ctx context.Context, kyma *v1beta2.K
 
 func (r *KymaReconciler) handleProcessingState(ctx context.Context, kyma *v1beta2.Kyma) (ctrl.Result, error) {
 	logger := logf.FromContext(ctx)
-	var requeueReason metrics.RequeueReason
 	var errGroup errgroup.Group
 	errGroup.Go(func() error {
 		err := r.reconcileManifests(ctx, kyma)
 		if err != nil {
-			requeueReason = metrics.ManifestReconciliationError
+			r.Metrics.RecordRequeueReason(metrics.ManifestReconciliationError)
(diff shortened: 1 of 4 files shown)


internal/controller/kyma_controller.go
```diff
@@ -131,6 +133,7 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 			r.deleteRemoteClientCache(ctx, kyma)
 			r.enqueueWarningEvent(kyma, syncContextError, err)
```
ruanxin, line 134: before this enqueueWarningEvent should also have a reason, right?

internal/controller/kyma_controller.go
```diff
@@ -371,6 +390,7 @@ func (r *KymaReconciler) handleProcessingState(ctx context.Context, kyma *v1beta
 	}
 
 	if err := errGroup.Wait(); err != nil {
+		r.Metrics.RecordRequeueReason(requeueReason)
```
ruanxin, line 393: this `Metrics.RecordRequeueReason` should not be put here, there has race condition in each errGroup.Go routine, you'll lost some error, for example, the `ManifestReconciliationError` and `ModuleCatalogSyncError` may happens during same time period, however, the err returned here is the first one get returned by those two threads (goroutine), and you will lost another error.

Instead, put RecordRequeueReason(requeueReason) in each errGroup goroutine.

ruanxin (changes requested):
lgtm, have some comments

internal/controller/kyma_controller.go
nesmabadr: This is the same error as the SyncContextRetrievalError, the requeue will happen in line 137

ruanxin (approved):

ruanxin (approved):

PR #1195, opened 2023-12-22
feat: Add watcher img registry flag

lindnerby:
**Description**

- for e2e tests in runtime-watcher component we need to be able to reference the dev registry where PR images are located
- introduce flag for it
- refactor naming of SKR Webhook to (runtime)watcher

**Related issue(s)**


cmd/main.go
```diff
@@ -142,14 +143,18 @@ func setupManager(flagVar *FlagVar, cacheOptions cache.Options, scheme *machiner
 	}
 
 	options := controllerOptionsFromFlagVar(flagVar)
-	if flagVar.enableKcpWatcher && flagVar.skrWatcherImage == "" {
+	if flagVar.enableWatcher && flagVar.watcherImage == "" {
 		setupLog.Error(errMissingWatcherImage, "unable to start manager")
 		os.Exit(1)
 	}
+	if flagVar.enableWatcher && flagVar.watcherRegistry == "" {
```
ruanxin, line 150: why have a separate check for watcherRegistry? it can be combined in line 146

pkg/watcher/skr_webhook_manifest_manager.go
```diff
@@ -23,40 +23,39 @@ import (
 	"github.com/kyma-project/lifecycle-manager/pkg/util"
 )
 
-// SKRWebhookManifestManager is a SKRWebhookManager implementation that applies
+// WatcherManifestManager is a SKRWebhookManager implementation that applies
```
ruanxin, line 26: Without change the skr-webhook service name, this renaming is a bit misleading tbh, but please then enhance the description comment of this WatcherManifestManager, it still contains SKRWebhookManager in the context.

cmd/main.go
```diff
@@ -142,14 +143,18 @@ func setupManager(flagVar *FlagVar, cacheOptions cache.Options, scheme *machiner
...
+		setupLog.Error(errMissingWatcherRegistry, "unable to start manager")
+		os.Exit(1)
+	}
 	remoteClientCache := remote.NewClientCache()
-	var skrWebhookManager *watcher.SKRWebhookManifestManager
-	if flagVar.enableKcpWatcher {
-		watcherChartDirInfo, err := os.Stat(flagVar.skrWatcherPath)
+	var skrWebhookManager *watcher.WatcherManifestManager
```
ruanxin, line 155:
```suggestion
	var watcherManifestManager *watcher.WatcherManifestManager
```

cmd/main.go
```diff
@@ -229,7 +234,7 @@ func controllerOptionsFromFlagVar(flagVar *FlagVar) ctrlruntime.Options {
 }
 
 func setupKymaReconciler(mgr ctrl.Manager, remoteClientCache *remote.ClientCache, flagVar *FlagVar,
-	options ctrlruntime.Options, skrWebhookManager *watcher.SKRWebhookManifestManager,
+	options ctrlruntime.Options, skrWebhookManager *watcher.WatcherManifestManager,
```
ruanxin, line 237:
```suggestion
	options ctrlruntime.Options, watcherManifestManager *watcher.WatcherManifestManager,
```

cmd/main.go
```diff
@@ -266,15 +271,15 @@ func setupKymaReconciler(mgr ctrl.Manager, remoteClientCache *remote.ClientCache
...
-		watcher.SkrWebhookManagerConfig{
-			SKRWatcherPath:         flagVar.skrWatcherPath,
-			SkrWatcherImage:        flagVar.skrWatcherImage,
-			SkrWebhookCPULimits:    flagVar.skrWebhookCPULimits,
-			SkrWebhookMemoryLimits: flagVar.skrWebhookMemoryLimits,
-			RemoteSyncNamespace:    flagVar.remoteSyncNamespace,
+	return watcher.NewWatcherManifestManager(mgr.GetConfig(), mgr.GetScheme(), caCertificateCache,
+		watcher.WatcherManagerConfig{
+			WatcherResourcesPath: flagVar.watcherResourcesPath,
+			WatcherImage:         fmt.Sprintf("%s/%s", flagVar.watcherRegistry, flagVar.watcherImage),
```
ruanxin, line 279: can you rename this `WatcherManagerConfig.WatcherImage` to something which can distinguish with `flagVar.watcherImage`

pkg/watcher/skr_webhook_manifest_manager.go
lindnerby: I revert the naming changes. Can be done in a follow-up and more thorougly 👍

cmd/flags.go
```diff
@@ -89,14 +96,8 @@ func DefineFlagVar() *FlagVar {
 		"Enabling verify modules against their signature")
 	flag.BoolVar(&flagVar.enableWebhooks, "enable-webhooks", false,
 		"Enabling Validation/Conversion Webhooks.")
-	flag.BoolVar(&flagVar.enableKcpWatcher, "enable-kcp-watcher", false,
+	flag.BoolVar(&flagVar.enableWatcher, "enable-kcp-watcher", false,
```
ruanxin, line 99: this flag controls if we enable watcher controller, this kcp indicate the controller is on the KCP, but not in SKR, I'm not against renaming it, but then the actual flag value is still `enable-kcp-watcher`, can you check if we can also rename it without bring additional changes in management config chat? Also please consider improving the description a bit.

cmd/main.go
```diff
@@ -268,25 +266,37 @@ func setupKymaReconciler(mgr ctrl.Manager, remoteClientCache *remote.ClientCache
...
-			CACertificateName:   flagVar.caCertName,
-			AdditionalDNSNames:  strings.Split(flagVar.additionalDNSNames, ","),
-			Duration:            flagVar.SelfSignedCertDuration,
-			RenewBefore:         flagVar.SelfSignedCertRenewBefore,
-		}, watcher.GatewayConfig{
-			IstioGatewayName:          flagVar.istioGatewayName,
-			IstioGatewayNamespace:     flagVar.istioGatewayNamespace,
-			LocalGatewayPortOverwrite: flagVar.listenerPortOverwrite,
-		})
+	watcherImg := fmt.Sprintf("%s:%s", watcherRegProd, flagVar.watcherImageTag)
```
ruanxin, line 269: can we extract line 269 - 272 as a function to return `watcherImg`?

cmd/flags.go
lindnerby: you are right the actual flag is not renamed and is in use, so i just reverted it.
since we anyways could think about making this option the default to `true` because it's the actual usecase and the special case is false, we can shift this so another PR since it will be more changes and needs an agreement.
also i don't get why it needs to be so specific in name, because there is no `enableSkrWatcher` usecase 😄

cmd/main.go
```diff
@@ -268,25 +261,46 @@ func setupKymaReconciler(mgr ctrl.Manager, remoteClientCache *remote.ClientCache
...
+		certConfig,
+		gatewayConfig)
+}
+
+const (
+	watcherRegProd = "europe-docker.pkg.dev/kyma-project/prod/runtime-watcher-skr"
+	watcherRegDev  = "europe-docker.pkg.dev/kyma-project/dev/runtime-watcher"
+)
+
+func getWatcherImg(flagVar *FlagVar) string {
```
ruanxin, line 298: this just need a boolean as input, not necessary for whole FlagVar

ruanxin (dismissed):

ruanxin (approved):

PR #1234, opened 2024-01-12
feat: Invalidate manifest client cache

Tomasz-Smelcerz-SAP:
**Description**

Observed on DEV: Lifecycle Manager slowly loses connectivity to the SKR's. The error reported in the Manifest CR: `Unauthorized`.
This correlates with short life of the SKR secrets on DEV (100 minutes).
Restarting the LM Pod helps for ~ 100 minutes.
It looks like the clients used in the manifest-controller loop are not refreshed when the secret is rotated/expires.

This PR is an attempt to get more insight into the problem.
On `Unauthorized` error it should invalidate the cached client.
Additional logging is provided for debugging purposes, to be removed once we find the 100% valid fix for the issue.

- Invalidate manifest controller client cache entry upon `Unauthorized` errors

### DIFF
--- a/internal/declarative/v2/reconciler.go
+++ b/internal/declarative/v2/reconciler.go
@@ -87,7 +87,7 @@ func newResourcesCondition(obj Object) apimetav1.Condition {
 	}
 }
 
-//nolint:funlen,cyclop // Declarative pkg will be removed soon
+//nolint:funlen,cyclop,gocognit // Declarative pkg will be removed soon
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	obj, ok := r.prototype.DeepCopyObject().(Object)
 	if !ok {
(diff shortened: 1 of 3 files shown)


internal/declarative/v2/reconciler.go
```diff
@@ -164,11 +164,22 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 		}
+		if errors.Is(err, ErrClientUnauthorized) {
+			logf.FromContext(ctx).Info("Detected \"unauthorized\" error for: " + req.NamespacedName.String())
+			if r.ClientCacheKeyFn != nil {
+				clientsCacheKey, ok := r.ClientCacheKeyFn(ctx, obj)
+				if ok {
+					logf.FromContext(ctx).Info("Invalidating manifest-controller client cache entry for key: " + fmt.Sprintf("%#v", clientsCacheKey))
+					r.ClientCache.Delete(clientsCacheKey)
+				}
+			}
```
nesmabadr, line 181: In this case, shouldn't you do a requeue?

Tomasz-Smelcerz-SAP: The next statement after this `if` block is: `return r.ssaStatus(ctx, obj, metrics.ManifestSyncResources, metrics.UnexpectedRequeue)`
This function always requeues the Manifest.

nesmabadr (approved):

PR #1257, opened 2024-01-23
fix: No such host in purge reconciler

Tomasz-Smelcerz-SAP:
**Description**

This PR is intended to get more insight into the "host not found" errors that appear on DEV.
The error itself looks like a `net.DNSError`. I was unable to reproduce it locally.
The errors are produced by the purge reconciler in the following conditions:
- The SKR cluster is removed
- The SKR secret is removed
- The purge reconciler finalizer is still there (and it's the only finalizer)

- an attempt to catch (and handle) "host not found" error.

### DIFF
--- a/internal/controller/purge_controller.go
+++ b/internal/controller/purge_controller.go
@@ -83,7 +83,7 @@ func (r *PurgeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 
 	start := time.Now()
 	remoteClient, err := r.ResolveRemoteClient(ctx, client.ObjectKeyFromObject(kyma))
-	if util.IsNotFound(err) {
+	if util.IsNotFound(err) || util.IsNoSuchHost(err) {
 		if err := r.dropPurgeFinalizer(ctx, kyma); err != nil {
 			r.Metrics.UpdatePurgeError(ctx, kyma, metrics.ErrPurgeFinalizerRemoval)
 			return ctrl.Result{}, err
--- a/internal/controller/purge_controller.go
+++ b/internal/controller/purge_controller.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/go-logr/logr"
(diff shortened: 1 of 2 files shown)


internal/controller/purge_controller.go
```diff
@@ -111,6 +112,13 @@ func (r *PurgeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 	return ctrl.Result{}, nil
 }
 
+func noSuchHostCheck(err error, logger logr.Logger) bool {
+	lgr := func(msg string) {
+		logger.V(log.InfoLevel).Info(msg)
```
nesmabadr, line 117: why not just do logf.FromContext(ctx).V(log.InfoLevel).Info(msg)?

nesmabadr: Also maybe just DebugLevel not InfoLevel in this case

Tomasz-Smelcerz-SAP: I don't know if Debug level logs are enabled.

Tomasz-Smelcerz-SAP: As for the first comment:
I don't have access to `ctx` object here. I could pass it as an argument, but then would that be simpler? My impl. has fewer characters to type than your proposal ;)

pkg/util/error.go
```diff
@@ -10,6 +12,25 @@ import (
 	"k8s.io/client-go/discovery"
 )
 
+const (
+	NoSuchHostMsg = "no such host"
+)
+
+func IsNoSuchHost(err error, lgr func(string)) bool {
```
jeremyharisch, line 19: I dont like how this has been implemented with the logger. In the team we agreed on to avoid passing logger from function to function. And here you even had the need to write a wrapper function.

Normally, we try to just pass `ctx context.Context` and then use `logf.FromContext`, in this way you could also avoid introducing `"github.com/go-logr/logr"` into the `purge_controller.go`.

PR #1262, opened 2024-01-24
fix: Logger config

Tomasz-Smelcerz-SAP:
**Description**

The existing log level setting logic was inverting any positive value to a negative one, thus effectively making any logging level a "Debug" level.
See: https://github.com/uber-go/zap/blob/master/zapcore/level.go, the "Debug" level has value of `-1` and any negative value looks like it's mapped to this one. The other values are non-negative ("Info" level is zero).

- Replace "inverting" logic with the logic that keeps the provided value within allowable limits.
- Adjust logging levels in local testing kustomize files as necessary

**Related issue(s)**

### DIFF
--- a/pkg/log/logger.go
+++ b/pkg/log/logger.go
@@ -9,9 +9,13 @@ import (
 )
 
 func ConfigLogger(level int8, syncer zapcore.WriteSyncer) logr.Logger {
-	if level > 0 {
-		level = -level
+	if zapcore.Level(level) < zapcore.DebugLevel {
+		level = int8(zapcore.DebugLevel)
 	}
+	if zapcore.Level(level) > zapcore.FatalLevel {
+		level = int8(zapcore.FatalLevel)
+	}
+
 	// The following settings is based on kyma community Improvement of log messages usability
 	// https://github.com/kyma-project/community/blob/main/concepts/observability-consistent-logging/improvement-of-log-messages-usability.md#log-structure
 	atomicLevel := zap.NewAtomicLevelAt(zapcore.Level(level))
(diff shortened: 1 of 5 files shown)


c-pius (approved):

Tomasz-Smelcerz-SAP:
closing this in favor of: https://github.com/kyma-project/lifecycle-manager/issues/1266

pkg/log/logger.go
```diff
@@ -9,9 +9,13 @@ import (
 )
 
 func ConfigLogger(level int8, syncer zapcore.WriteSyncer) logr.Logger {
-	if level > 0 {
-		level = -level
```
c-pius, line 13: @Tomasz-Smelcerz-SAP 

I still don't fully get it, but seems that this was right. According to docs, `zap` is using log levels like Debug -1, Info 0, Warning 2, ... where higher level means higher importance and less verbosity.

We are logging through the `logr` API where `zapr` is the implementation using `zap`. `logr` does not have a notion of log levels. Instead, it just distinguishes `info` and `error` messages and assigns a verbosity. The verbosity means "it is a more or less important info log" for instance. Verbosities are positive where higher number means less importance (opposite of `zap`).

Now, `zapr` describes the relation here https://github.com/go-logr/zapr?tab=readme-ov-file#increasing-verbosity where they initialize with the negative levels. Here I am now a bit confused and still need to think about it further. Just wanted to share my thoughts already.

c-pius: There we go: https://pkg.go.dev/github.com/go-logr/zapr#hdr-Implementation_Details

*Levels in logr correspond to custom debug levels in Zap. Any given level in logr is represents by its inverse in zap (`zapLevel = -1*logrLevel`). For example V(2) is equivalent to log level -2 in Zap, while V(1) is equivalent to Zap's DebugLevel.*

PR #1270, opened 2024-01-25
chore: Invalidate client cache when connection is refused or unauthorized

nesmabadr:
**Description**

- Invalidate client cache when connection is refused or unauthorized

**Related issue(s)**


pkg/util/error.go
```diff
@@ -63,3 +63,17 @@ func IsNotFound(err error) bool {
...
 }
+
+func IsConnectionRefusedOrUnauthorized(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if strings.Contains(strings.ToLower(err.Error()),
+		"connection refused") || strings.Contains(strings.ToLower(err.Error()),
+		"unauthorized") {
```
Tomasz-Smelcerz-SAP, line 74:
```suggestion
         msg := strings.ToLower(err.Error())
         if strings.Contains(msg, "connection refused") ||
                 strings.Contains(msg, "unauthorized") {
                 return true
```

Tomasz-Smelcerz-SAP: Actually, you can improve it even further: Just return the boolean directly, without the `if` statement:
```
         msg := strings.ToLower(err.Error())
         return strings.Contains(msg, "connection refused") ||
                 strings.Contains(msg, "unauthorized")

```

Tomasz-Smelcerz-SAP (approved):

PR #1272, opened 2024-01-29
fix: Add container name to metrics to exclude metrics which are not related

jeremyharisch:
**Description**

- Added container name to max parallel workers metrics dashboard to represent the actual wanted workers we want to have at the dashboard. Because before the metrics from `runtime-monitoring-operator` have been included as well
- 

**Before:**
Screenshot 2024-01-29 at 15 00 56

**After:**

Screenshot 2024-01-29 at 15 00 40

Actuel configured workers
Screenshot 2024-01-29 at 15 03 59

### DIFF
--- a/config/grafana/overview.json
+++ b/config/grafana/overview.json
@@ -821,7 +821,7 @@
             "uid": "P1809F7CD0C75ACF3"
           },
           "exemplar": true,
-          "expr": "sum by (controller) (controller_runtime_max_concurrent_reconciles{controller=\"kyma\"})",
+          "expr": "sum by (controller) (controller_runtime_max_concurrent_reconciles{container=\"manager\",controller=\"kyma\" })",
           "interval": "",
           "legendFormat": "{{controller}}",
           "refId": "A"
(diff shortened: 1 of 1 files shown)


jeremyharisch:
Screenshot 2024-01-29 at 15 04 57

If we remove the sum function then we can actually see why kyma was 85 before, but since we are just interested in the KLM metrics, we should add the container name

nesmabadr (approved):

config/grafana/overview.json
```diff
@@ -1994,5 +1995,5 @@
   "timezone": "",
   "title": "Lifecycle Manager Overview",
   "uid": "O3DH7uunk",
-  "version": 7
-}
+  "version": 1
```
nesmabadr, line 1998: is this version correct? It was 7

jeremyharisch: Was set by Plutono itself. I just changed the two lines with adding the container name. The rest was done by plutono

jeremyharisch: So I guess, it should be correct

nesmabadr: Yeah I was just surprised that it was downgraded so was just making sure you didn't edit it by mistake

PR #1273, opened 2024-01-29
fix: Introduce a Go routine to remove no longer existing Kymas from the metrics

nesmabadr:
**Description**

- Introduce a Go routine to remove no longer existing Kymas from the metrics
- It runs every 15 minutes (can be configured)
- Add unit tests

**Related issue(s)**
Resolves https://github.com/orgs/kyma-project/projects/30/views/15?pane=issue&itemId=50368250


internal/pkg/metrics/kyma.go
```diff
@@ -141,3 +146,68 @@ func calcStateValue(state, newState shared.State) float64 {
 func (k *KymaMetrics) RecordRequeueReason(kymaRequeueReason KymaRequeueReason, requeueType RequeueType) {
 	k.requeueReasonCounter.WithLabelValues(string(kymaRequeueReason), string(requeueType)).Inc()
 }
+
+func (k *KymaMetrics) CleanupNonExistingKymaCrsMetrics(ctx context.Context, kcpClient client.Client) error {
+	currentLifecycleManagerLogs, err := FetchLifecycleManagerMetrics()
```
c-pius, line 151: Intended to name `Logs` and not `Metrics`?

internal/pkg/metrics/kyma.go
```diff
@@ -141,3 +146,68 @@ func calcStateValue(state, newState shared.State) float64 {
...
+	}
+	kymaNames := getKymaNames(kymaCrsList)
+	for _, m := range currentLifecycleManagerLogs {
+		currentKymaName := getKymaNameFromLabels(m)
+		if !slices.Contains(kymaNames, currentKymaName) {
+			k.KymaStateGauge.DeletePartialMatch(prometheus.Labels{
+				KymaNameLabel: currentKymaName,
+			})
+		}
+	}
```
c-pius, line 173: I think this works, but maybe a set based comparison would read a bit easier and even be a bit more performant? Something like

```go
kymaNames := getKymaNames(kymaCrsList) // returning a "kymaName->true" mapping (map[string]bool)
currentKymaName := getKymaNameFromLabels(m)

if _, exists := kymaNames[currentKymaName]; !exists {
  k.KymaStateGauge.DeletePartialMatch(prometheus.Labels{
    KymaNameLabel: currentKymaName,
  })
}
```

internal/pkg/metrics/kyma_test.go
```diff
@@ -0,0 +1,89 @@
...
+		KymaStateGauge: sampleGauge,
+	}
+	if err := k.CleanupNonExistingKymaCrsMetrics(context.TODO(), fakeClientBuilder); err != nil {
+		t.Errorf("CleanupNonExistingKymaCrsMetrics() error = %v", err)
+	}
+
+	resultingMetrics, _ = ctrlmetrics.Registry.Gather()
+	for _, metric := range resultingMetrics {
+		if metric.GetName() == metrics.MetricKymaState {
+			if !reflect.DeepEqual(metric.GetMetric(), wantResultingMetrics) {
```
c-pius, line 83: Maybe too picky, but I think this would also go green if we accidentially deleted all `MetricKymaState` metrics, right? Should we rather explicitly expect that `len(resultingMetrics) == 1` and that `resultingMetrics[0].GetName() == metrics.MetricKymaState`?

cmd/main.go
```diff
@@ -156,19 +157,34 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
...
-
 	if flagVar.EnableWebhooks {
 		enableWebhooks(mgr)
 	}
 
+	addHealthChecks(mgr)
+	if flagVar.DropStoredVersion != "" {
+		go func(version string) {
+			dropStoredVersion(mgr, version)
+		}(flagVar.DropStoredVersion)
```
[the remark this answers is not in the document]
nesmabadr: We had it earlier without passing the version argument (so the same way you are mentioning), but then Checkmarx detected that this is could cause a race condition

internal/pkg/metrics/kyma.go
nesmabadr: No,  thanks for noticing that :D

cmd/main.go
```diff
@@ -178,15 +194,21 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
...
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
-		setupLog.Error(err, "problem running manager")
-		os.Exit(1)
+}
+
+func runKymaMetricsCleanup(kymaMetrics *metrics.KymaMetrics, kcpClient client.Client,
+	cleanupIntervalInMinutes time.Duration) {
+	scheduler := gocron.NewScheduler(time.UTC)
+	_, scheduleErr := scheduler.Every(cleanupIntervalInMinutes).Minutes().Do(func() {
+		if err := kymaMetrics.CleanupNonExistingKymaCrsMetrics(context.TODO(), kcpClient); err != nil {
```
c-pius, line 203: After reading a bit about context:
- https://medium.com/@jamal.kaksouri/the-complete-guide-to-context-in-golang-efficient-concurrency-management-43d722f6eaea
- https://stackoverflow.com/questions/74239074/context-todo-or-context-background-which-one-should-i-prefer

Should we rather use `context.Background()` here? My understanding is that TODO would indicate that we replace it with some parent context at some point, but I don't think this will be the case here?

nesmabadr: yes I actually had this just as a placeholder while writing the function and was just looking into it :D

cmd/main.go
```diff
@@ -179,15 +197,25 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
...
+	scheduler := gocron.NewScheduler(time.UTC)
+	_, scheduleErr := scheduler.Every(cleanupIntervalInMinutes).Minutes().Do(func() {
+		if err := kymaMetrics.CleanupNonExistingKymaCrsMetrics(ctx, kcpClient); err != nil {
+			setupLog.Info(fmt.Sprintf("failed to cleanup non existing kyma crs metrics, err: %s", err))
+		}
+	})
+	if scheduleErr != nil {
+		setupLog.Info(fmt.Sprintf("failed to setup cleanup routine for non existing kyma crs metrics, err: %s",
+			scheduleErr))
 	}
```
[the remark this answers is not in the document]
c-pius: EDIT: I think there is also a mistake in my suggestion. If we have the `defer cancel()`, the cancellation is done after running this function, so basically we still have a cancelled context on every invocation. We probably either need to omit the "global" context or provide it from the outside so that it is cancelled only once the KLM terminates

internal/pkg/metrics/kyma.go
```diff
@@ -141,3 +146,69 @@ func calcStateValue(state, newState shared.State) float64 {
...
+
+	return ""
+}
+
+func getKymaNames(kymaCrs *v1beta2.KymaList) map[string]bool {
+	if len(kymaCrs.Items) == 0 {
+		return nil
+	}
+
```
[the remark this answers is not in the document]
nesmabadr: I think it shouldn't matter

internal/pkg/metrics/kyma.go
```diff
@@ -141,3 +146,69 @@ func calcStateValue(state, newState shared.State) float64 {
...
+}
+
+func getKymaNameFromLabels(metric *prometheusclient.Metric) string {
+	for _, label := range metric.GetLabel() {
+		if label.GetName() == KymaNameLabel {
+			return label.GetValue()
+		}
+	}
+
+	return ""
```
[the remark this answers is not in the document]
nesmabadr: hmmm, there shouldn't be a metric with no kyma_name but I can add an error as a double check

cmd/main.go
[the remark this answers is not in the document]
nesmabadr: hmm your words actually make sense. Let me think how I can fix this. Thanks for noting that!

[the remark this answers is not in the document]
lindnerby: Maybe I am wrong, but don't you need to move the context creation in the function?
```
_, scheduleErr := scheduler.Every(cleanupIntervalInMinutes).Minutes().Do(func() {
    // Move the context creation here so it resets with each function run
    ctx, cancel := context.WithTimeout(context.Background(), metricCleanupTimeout)
    defer cancel()

    if err := kymaMetrics.CleanupNonExistingKymaCrsMetrics(ctx, kcpClient); err != nil {
        setupLog.Info(fmt.Sprintf("failed to cleanup non existing kyma crs metrics, err: %s", err))
    }
})
```

internal/pkg/metrics/kyma.go
```diff
@@ -142,3 +147,72 @@ func calcStateValue(state, newState shared.State) float64 {
...
+	for _, m := range currentLifecycleManagerMetrics {
+		currentKymaName := getKymaNameFromLabels(m)
+		if currentKymaName == "" {
+			continue
+		}
+		if _, exists := kymaNames[currentKymaName]; !exists {
+			logs.FromContext(ctx).Info(fmt.Sprintf("Deleting a metric for non-existing Kyma: %s", currentKymaName))
+			k.KymaStateGauge.DeletePartialMatch(prometheus.Labels{
+				KymaNameLabel: currentKymaName,
+			})
```
c-pius, line 176: I just noticed that there is a `CleanupMetrics` function which seems to do the same thing + also deleting `lifecycle_mgr_module_state` for respective kyma name. Can this be re-used or do we not want to delete the module-related ones?

https://github.com/kyma-project/lifecycle-manager/blob/main/internal/pkg/metrics/kyma.go#L91-L98

nesmabadr: I didn't use it because of the module-related metrics. Right now only the lifecycle manager are the ones that remain, that's why I didn't cleanup the modules metrics

cmd/main.go
```diff
@@ -179,15 +197,24 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
...
+	cleanupIntervalInMinutes int,
+) {
+	scheduler := gocron.NewScheduler(time.UTC)
+	_, scheduleErr := scheduler.Every(cleanupIntervalInMinutes).Minutes().Do(func() {
+		ctx, cancel := context.WithTimeout(context.Background(), metricCleanupTimeout)
+		defer cancel()
+		if err := kymaMetrics.CleanupNonExistingKymaCrsMetrics(ctx, kcpClient); err != nil {
+			setupLog.Info(fmt.Sprintf("failed to cleanup non existing kyma crs metrics, err: %s", err))
+		}
+	})
```
c-pius, line 212: Just an idea, and not sure if this is common in Go... When testing locally, I was wondering if there is a possibility to see that the job was run but failed to clean up the metric without raising an error. It may be worth adding an Info or Debug log at the end of this function indicating that a clean up run has completed.

c-pius (commented):
LGTM and was able to verify the metric deletion locally. Just one more minor thought, otherwise good to approve

cmd/main.go
c-pius: Basically just for being able to see that the job is alive, even though no metrics have been purged in a while

PR #1275, opened 2024-01-29
chore: replace Unauthorized against SKR events with metrics

### LINKED ISSUE #1236: Supress event creation during SKR secret rotation
### Description

After recent changes in infrastructure configuration, SKR secrets in KCP are rotated every hour or so.
When the secrets are rotated, the Lifecycle-Manager experiences a burst of `Unauthorized` errors.
Two reconciliation loops are mostly affected: Kyma reconciliation loop and the Manifest reconciliation loop.
The LM produces an event on every error, and eventually there's an error for every Kyma / Manifest object.
This means LM produces a burst of `Unauthorized` events with a rate reaching 90 events / minute.
We reduced the event size in https://github.com/kyma-project/lifecycle-manager/pull/1222, and ensured the clients are refreshed as soon as possible in https://github.com/kyma-project/lifecycle-manager/pull/1234, to reduce the time `Unauthorized` errors occur.
This helped already, as we don't observe LM restarts due to long ETCD time-outs (over 2 minutes).
Unfortunately we still observe ETCD outages occasionally.
The next step towards a "well behaving controller" is to reduce the number of events.

Note: It may be harder than it sounds. Normally we emit an event for some Kyma object. If we "compact" events for many Kymas into one , which Kyma object would be the target for the event? Maybe we should just suppress such errors? How about occasional `Unauthorized` errors, not related to secret rotation then? How to distinguish these two conditions?

### Reasons

- There's no point in creating hundreds of otherwise identical events.
- The event bursts have negative impact on ETCD performance.

### Acceptance Criteria

- reduce the number of events created during secrets rotation.

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

c-pius:
**Description**

- tries to cover all places with communication against the SKR in Kyma and Manifest reconcile loops and capture Unauthorized events to report those as metric instead of events

**Related issue(s)**
partially resolves #1236


internal/declarative/v2/reconciler.go
```diff
@@ -139,7 +139,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+		// Suppress the creation of events based on an "unauthorized against the SKR cluster" condition
+		// to not put burst loads onto ETCD when doing credential rotation
+		if util.IsUnauthorized(err) {
+			r.Metrics.RecordRequeueReason(metrics.ManifestSyncUnauthorized, metrics.UnexpectedRequeue)
+		} else {
+			r.Event(obj, "Warning", "ClientInitialization", err.Error())
+		}
+
 		obj.SetStatus(obj.GetStatus().WithState(shared.StateError).WithErr(err))
 		return r.ssaStatus(ctx, obj, metrics.ManifestClientInit, metrics.UnexpectedRequeue)
```
c-pius, line 151: I was considering here to directly returning with requeue reason after recording the unauthorized metric. However, I think we need to update the status (even though it is a transient condition that will resolve itself within a couple of mins). If we wouldn't do so and there is a permanent condition leading to unauthorized errors, we would keep the state healthy and wouldn't be easily able to detect that there is an ongoing issue.

internal/controller/kyma_controller.go
```diff
@@ -129,6 +129,14 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	// Suppress the creation of further events based on an "unauthorized against the SKR cluster" condition
+	// to not put burst loads onto ETCD when doing credential rotation.
+	// This is observed to be the first place where such condition appears in the Kyma reconcile loop.
+	if util.IsUnauthorized(err) {
+		r.Metrics.RecordRequeueReason(metrics.KymaUnauthorized, metrics.UnexpectedRequeue)
```
nesmabadr, line 136: if it is unauthorized you should invalidate the cache as well

internal/controller/kyma_controller.go
```diff
@@ -129,6 +129,15 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	// Suppress the creation of further events based on an "unauthorized against the SKR cluster" condition
+	// to not put burst loads onto ETCD when doing credential rotation.
+	// This is observed to be the first place where such condition appears in the Kyma reconcile loop.
+	if util.IsUnauthorized(err) {
+		r.deleteRemoteClientCache(ctx, kyma)
+		r.Metrics.RecordRequeueReason(metrics.KymaUnauthorized, metrics.UnexpectedRequeue)
+		return ctrl.Result{Requeue: true}, nil
```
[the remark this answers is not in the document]
c-pius: I didn't do it because this would mean we are recording an event in [`requeueWithError->updateStatusWithError->`](https://github.com/kyma-project/lifecycle-manager/blob/main/internal/controller/kyma_controller.go#L569) which we try to avoid. But right, I missed that the status needs to be updated. Will think about how to best achieve that without writing the event.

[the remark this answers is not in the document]
c-pius: Addressed this now by a minor refactoring. See commit [0fb198f](https://github.com/kyma-project/lifecycle-manager/pull/1275/commits/0fb198f8d3e3c9d7c1f4e76b187f8e8d10de7aeb) which moves the writing of the event from `updateStatusWithError` to `requeueWithError`. The behavior should be the same as before.

But now, I can call `updateStatusWithError` without writing the event, see [d15f716](https://github.com/kyma-project/lifecycle-manager/pull/1275/commits/d15f716192191061a0bdb9dc4c6f39200320b5fc)

The resulting status of the KymaCR is now:

```yaml
apiVersion: operator.kyma-project.io/v1beta2
kind: Kyma
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"operator.kyma-project.io/v1beta2","kind":"Kyma","metadata":{"annotations":{"skr-domain":"example.domain.com"},"name":"kyma-sample","namespace":"kcp-system"},"spec":{"channel":"regular"}}
    kyma-kcp-crd-generation: "2"
    kyma-skr-crd-generation: "1"
    moduletemplate-kcp-crd-generation: "2"
    moduletemplate-skr-crd-generation: "1"
    skr-domain: example.domain.com
  creationTimestamp: "2024-02-06T12:13:29Z"
  finalizers:
  - operator.kyma-project.io/Kyma
  generation: 1
  labels:
    operator.kyma-project.io/managed-by: lifecycle-manager
  name: kyma-sample
  namespace: kcp-system
  resourceVersion: "5377"
  uid: 574e9225-fd22-4eb6-882f-a96f2b8cf36e
spec:
  channel: regular
status:
  activeChannel: regular
  conditions:
  - lastTransitionTime: "2024-02-06T12:49:12Z"
    message: modules state is unknown
    observedGeneration: 1
    reason: Ready
    status: Unknown
    type: Modules
  - lastTransitionTime: "2024-02-06T12:49:12Z"
    message: module templates synchronization state is unknown
    observedGeneration: 1
    reason: Ready
    status: Unknown
    type: ModuleCatalog
  - lastTransitionTime: "2024-02-06T12:49:12Z"
    message: skrwebhook is out of sync and needs to be resynchronized
    observedGeneration: 1
    reason: Ready
    status: Unknown
    type: SKRWebhook
  lastOperation:
    lastUpdateTime: "2024-02-06T12:49:12Z"
    operation: 'initializing sync context failed: failed to ensure remote namespace
      exists: failed to get API group resources: unable to retrieve the complete list
      of server APIs: v1: Unauthorized'
  modules:
  - channel: regular
    fqdn: kyma-project.io/module/template-operator
    manifest:
      apiVersion: operator.kyma-project.io/v1beta2
      kind: Manifest
      metadata:
        generation: 1
        name: kyma-sample-template-operator-2235966007
        namespace: kcp-system
    name: template-operator
    resource:
      kind: Sample
      metadata:
        name: sample-yaml
        namespace: kyma-system
    state: Ready
    template:
      apiVersion: operator.kyma-project.io/v1beta2
      kind: ModuleTemplate
      metadata:
        generation: 1
        name: template-operator-regular
        namespace: kcp-system
    version: 0.1.0
  state: Error
```

Behavior of events, logs and metrics is still as described in PR description

nesmabadr (approved):

pkg/util/error.go
```diff
@@ -43,15 +43,23 @@ func IsNotFound(err error) bool {
...
 
-func IsConnectionRefusedOrUnauthorized(err error) bool {
+func IsConnectionRefusedOrUnauthorizedOrAskingForCredentials(err error) bool {
 	if err == nil {
 		return false
 	}
 
 	msg := strings.ToLower(err.Error())
-	if strings.Contains(msg, "connection refused") || strings.Contains(msg, "unauthorized") {
+	if strings.Contains(msg, "connection refused") || strings.Contains(msg, "unauthorized") || strings.Contains(msg, "the server has asked for the client to provide credentials") {
```
ruanxin, line 52: use human readable error messages to decide specific error cases in general is bad practice. Because the dependent upstream library may change the context of those errors, it's not reliable, instead we should always try to find the typed error in the depend library so that if the definition of the error changed, we are aware in the code base.

The reason why we accept `connection refused` is because this error is coming from go `net/http` package, if you check the [source code](https://cs.opensource.google/go/go/+/master:src/net/http/socks_bundle.go;l=209), it's very low level error code, in the socket layer, basically the author will not change it.

However, the `"unauthorized"` is already a code smell, unauthorized is a standard HTTP status code - 401, you could find in the `net/http/status.go` [here](https://cs.opensource.google/go/go/+/master:src/net/http/status.go;l=37) have all status code with const value defined, instead of using string to compare, we should use this const value. Because this error response are coming from API server, so the related library to define the response is apimachinery, so a much more reliable way is to case this error to [`StatusError`](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go#L35), then we can extract the status reason to tell if this error is unauthorized or not.

Then let's see this newly introduced message, actually, if we did the Unauthorized error correctly, we don't need to introduce this comparing, if you look at where is this message defined, it's from apimachinery [here](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go#L454), and the HTTP status of this message is exactly `StatusReasonUnauthorized`

So please refactor this code, use HTTP status to check Unauthorized error but not continuing to introduce this message comparing bad practices.

@c-pius @nesmabadr, if you need more clarity for this refactoring, feel free to reach me.

Issue #1119, opened 2023-12-05
Run Storage Version Migrator


ruanxin:
Use the Storage Version Migrator to upgrade existing objects to a new stored version

For each landscape (Dev->Stage->Prod) do the following steps in order:
- [x] 1. Run the [storage Version migrator](https://github.com/kubernetes-sigs/kube-storage-version-migrator)
- [x] 2. Remove the old version from the CustomResourceDefinition status.storedVersions field.
  - change the flag value to `v1beta1` for [dropVersionFromStoredVersions](https://github.com/kyma-project/lifecycle-manager/blob/23262218a84bc27515b427ccaec43da7f3c90e56/cmd/main.go#L186)
- [x] 3. Verify that `v1beta1` is no longer listed in the CustomResourceDefinition `status.storedVersions`.

Landscapes:

- [x] DEV
- [x] STAGE
- [x] PROD

Ref: [official migration guide](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version)

PR #1289, opened 2024-02-06
fix: Adjust script based on SREs comments

nesmabadr:
**Description**

- Adjust the context in the migrator script in order not to enforce it

**Related issue(s)**
#1119

### DIFF
--- a/scripts/migrator_cleanup.sh
+++ b/scripts/migrator_cleanup.sh
@@ -8,9 +8,7 @@ do
         esac
 done
 
-if [[ -n "$context" ]]; then
-  kubectl config use-context $context
-fi
+test -n ${context} && kubectl config use-context $context
 
 kubectl delete clusterrolebinding storage-version-migration-migrator
 kubectl delete clusterrolebinding storage-version-migration-trigger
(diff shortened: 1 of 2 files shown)


lindnerby (dismissed):

scripts/migrator_cleanup.sh
```diff
@@ -8,7 +8,9 @@ do
         esac
 done
 
-kubectl config use-context $context
+if [[ -n "$context" ]]; then
+  kubectl config use-context $context
+fi
```
gczuczy, line 13: pleaseswap that with:
```
test -n ${context} && kubectl config use-context $context
```
Also, on L1 you just swapped `bash` for a posix sh, and `[[/]]` is a bash-extension, not part of the POSIX-sh standard.

scripts/storage_version_migrator.sh
```diff
@@ -18,7 +17,11 @@ done
 
 git clone https://github.com/kubernetes-sigs/kube-storage-version-migrator.git
 cd kube-storage-version-migrator
-kubectl config use-context $context
+
+if [[ -n "$context" ]]; then
+  kubectl config use-context $context
+fi
```
gczuczy, line 23: Please see the comment in the other file.

gczuczy (changes requested):

lindnerby (approved):

Issue #1288, opened 2024-02-06
Fix E2E test flakiness


ameteiko:
The GitHub action for E2E tests is now [flaky](https://github.com/kyma-project/lifecycle-manager/commits/main/).

Allegedly, most of the failures started happening after [this PR](https://github.com/kyma-project/lifecycle-manager/pull/1220) merged into the main.

**Acceptance Criteria:**

- [ ] The tests stability is restored
- [ ] Present the findings to the team to brainstorm how to tighten the code review process to prevent it from happening again

PR #1292, opened 2024-02-07
chore: Bump e2e CM version to align with prod

lindnerby:
**Description**

- e2e test suite should run with the same version from landscapes

**Related issue(s)**


c-pius:
LGTM, should we re-run the failed tests? Both seem to be flaky and I assume not directly related to this change.

lindnerby:
> LGTM, should we re-run the failed tests? Both seem to be flaky and I assume not directly related to this change.

Yes, `Given SKR Cluster Then Module CR and Manifest CR are removed` is very flakey recently. Very annoying.
We have an [issue ready to be picked up](https://github.com/kyma-project/lifecycle-manager/issues/1288).

c-pius (approved):

PR #1294, opened 2024-02-07
chore: Add log message for the metrics cleanup job

nesmabadr:
**Description**

- Add a log message to ensure that the metrics cleanup job ran successfully

### DIFF
--- a/internal/pkg/metrics/kyma.go
+++ b/internal/pkg/metrics/kyma.go
@@ -177,6 +177,8 @@ func (k *KymaMetrics) CleanupNonExistingKymaCrsMetrics(ctx context.Context, kcpC
 		}
 	}
 
+	logs.FromContext(ctx).Info("Finished running the metrics cleanup job")
+
 	return nil
 }
 


internal/pkg/metrics/kyma.go
```diff
@@ -177,6 +177,7 @@ func (k *KymaMetrics) CleanupNonExistingKymaCrsMetrics(ctx context.Context, kcpC
 		}
 	}
 
+	logs.FromContext(ctx).Info("Finished running the metrics cleanup job")
```
c-pius, line 180: Would be okay for me, but wondering if it makes more sense to log it here: https://github.com/kyma-project/lifecycle-manager/blob/main/cmd/main.go#L206-L212

The context that it is run as part of a job resides there, and this is also where we log that the job failed to execute or setup. Also not sure if we at some point would also call this function additionally from outside a job.

But just a thought that came to my mind, I am also fine with this proposal.

nesmabadr: I was actually thinking to put it there, but I thought it would not matter. and I also put it here just to make sure that it actually entered the cleanup function

c-pius (approved):

Issue #1290, opened 2024-02-06
Excessive logging in case of misconfigured Kymas/SKR-secrets


Tomasz-Smelcerz-SAP:
### Description

Observed on DEV: In case of an "orphaned" Kyma, the Lifecycle-Manager floods the log stream with identical error messages.
Currently there are many orphaned Kymas on DEV, and so the log collection mechanism is overloaded because of that behavior.

Details:
The orphaned Kyma is a Kyma for which the SKR doesn't exist anymore. The secret with the kubeconfig still exists though, preventing LM from deleting the Kyma object.
The particular log entry I observed: `E0204 01:59:00.001032 1 memcache.go:265] couldn't get current server API group list: Get "https://api.ab2ec24.dev.kyma.ondemand.com/api": dial tcp: lookup api.ab2ec24.dev.kyma.ondemand.com on 10.4.0.10:53: no such host`.
I counted 706 such log entries for that single Kyma produced within one minute, in a randomly chosen period.

A/C:
[] - reduce the number of log entries produced for a single Kyma in case of "host not found" errors
[] -  (stretch) try to design the mechanism to be easily extensible, so that LM can handle other similar errors (like `connection reset`), which may also occur in described conditions, in the same way.

### Steps to reproduce

TBD

### Environment Type

Managed

### Environment Info

Kubernetes Version: x.y.z
Docker Version: x.y.z
Kyma CLI Version: x.y.z
If used - k3d/k3s/minikube versions

c-pius:
❗ Not sure if this is related, so take with care...

One thing I observed while testing another thing: if I shut down the local skr cluster while the kcp cluster is running, logs and metrics are also going bonkers reaching hundreds of entries within no time. Log messages are different though (saying "connection reset by peer").

The metric that is sky rocketing is `lifecycle_mgr_requeue_reason_total{requeue_reason="manifest_sync_resources",requeue_type="unexpected"}`. If we look at that metric in Plutono for the last 30 days, we can see some crazy numbers as well (one pod was writing 15 million entries in less than 24 hours)

jeremyharisch:
Blocked by https://github.com/kyma-project/lifecycle-manager/pull/1297

PR #1297, opened 2024-02-07
feat: Introduce different requeue after interval to control the frequency of unexpected requeue events

ruanxin:
At the moment, manifest controller only do requeue after for success reconcilation, however for certain cases, if the requeue is unexpected, there is no need to requeue inmediately, which brings unneccesary high amount of requeue event. For example, currently we havemany unexpected `manifest_sync_resources` requeue

This PR address it by introduce another time interval, which reuse `KymaRequeueBusyInterval`, trying to reduce the unexpected requeue numbers.

I also dropped the `CtrlOnSuccess` options, instead introduce the  `queue.RequeueIntervals` directly to manifest Reconciler struct to have fine grained requeue control.

### DIFF
--- a/internal/declarative/v2/reconciler.go
+++ b/internal/declarative/v2/reconciler.go
@@ -18,6 +18,7 @@ import (
 	"github.com/kyma-project/lifecycle-manager/api/shared"
 	"github.com/kyma-project/lifecycle-manager/internal/pkg/metrics"
 	"github.com/kyma-project/lifecycle-manager/pkg/common"
+	"github.com/kyma-project/lifecycle-manager/pkg/queue"
 	"github.com/kyma-project/lifecycle-manager/pkg/util"
 )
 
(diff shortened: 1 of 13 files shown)


internal/declarative/v2/reconciler.go
```diff
@@ -615,23 +614,22 @@ func (r *Reconciler) configClient(ctx context.Context, obj Object) (Client, erro
...
 	resetNonPatchableField(obj)
-	r.Metrics.RecordRequeueReason(requeueReason, requeueType)
+	r.Metrics.RecordRequeueReason(requeueReason, queue.UnexpectedRequeue)
 	if err := r.Status().Patch(ctx, obj, client.Apply, client.ForceOwnership, r.FieldOwner); err != nil {
 		r.Event(obj, "Warning", "PatchStatus", err.Error())
 		return ctrl.Result{}, fmt.Errorf("failed to patch status: %w", err)
 	}
-	return ctrl.Result{Requeue: true}, nil
+
+	return ctrl.Result{RequeueAfter: r.RequeueIntervals.Busy}, nil
```
ruanxin, line 625: To Reviewer:

basically if we need to update status, all requeueType here is unexpected, that's why I removed the requeueType parameter, and since all of them are unexpected requeue, there is no need to requeue immediately, in most case, this unexpected behavior can't be resolved in a short period, that's why we add a requeue interval to reduce the frequency.

c-pius:
I tested this branch for what I described [here](https://github.com/kyma-project/lifecycle-manager/issues/1290#issuecomment-1932532060). Result is the same actually, if SKR cluster gets shut down, logs and metrics sky rocket indicating to me that something is still wrong with the reconcile loop.

lindnerby (approved):

PR #1298, opened 2024-02-07
chore: Improve detection of connection errors

c-pius:
**Description**

- Improves the detection of connection errors as proposed in https://github.com/kyma-project/lifecycle-manager/pull/1275#discussion_r1481305436

**Related issue(s)**

- resolves https://github.com/kyma-project/lifecycle-manager/issues/1236

### DIFF
--- a/pkg/util/error.go
+++ b/pkg/util/error.go
@@ -3,6 +3,7 @@ package util
 import (
 	"errors"
 	"strings"
+	"syscall"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -43,17 +44,16 @@ func IsNotFound(err error) bool {
 	return false
 }
 
-func IsConnectionRefusedOrUnauthorizedOrAskingForCredentials(err error) bool {
+func IsConnectionRefusedOrUnauthorized(err error) bool {
 	if err == nil {
 		return false
 	}
 
-	msg := strings.ToLower(err.Error())
-	if strings.Contains(msg, "connection refused") || strings.Contains(msg, "unauthorized") || strings.Contains(msg, "the server has asked for the client to provide credentials") {
+	if errors.Is(err, syscall.ECONNREFUSED) {
 		return true
 	}
 
-	return false
+	return apierrors.IsUnauthorized(err)
 }
 
 func IsUnauthorized(err error) bool {
@@ -61,5 +61,5 @@ func IsUnauthorized(err error) bool {
 		return false
 	}
 
-	return strings.Contains(strings.ToLower(err.Error()), "unauthorized")
+	return apierrors.IsUnauthorized(err)
 }
--- a/pkg/util/error.go
+++ b/pkg/util/error.go
@@ -49,11 +49,7 @@ func IsConnectionRefusedOrUnauthorized(err error) bool {
 		return false
 	}
 
-	if errors.Is(err, syscall.ECONNREFUSED) {
-		return true
-	}
-
-	return apierrors.IsUnauthorized(err)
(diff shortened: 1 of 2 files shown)


c-pius:
@nesmabadr tests of this on my machine look good. Also I think the `connection refused` cases were properly handled on your machine yesterday. So for me it would be okay to get that in already and do the verification on stage to not keep this ticket open forever.

Once Xin is back, we can also verify directly again if this is okay for him. If not, we can again refactor. WDYT?

pkg/util/error.go
```diff
@@ -43,23 +44,22 @@ func IsNotFound(err error) bool {
...
 
-func IsConnectionRefusedOrUnauthorizedOrAskingForCredentials(err error) bool {
+func IsConnectionRefusedOrUnauthorized(err error) bool {
 	if err == nil {
 		return false
 	}
 
-	msg := strings.ToLower(err.Error())
-	if strings.Contains(msg, "connection refused") || strings.Contains(msg, "unauthorized") || strings.Contains(msg, "the server has asked for the client to provide credentials") {
+	if errors.Is(err, syscall.ECONNREFUSED) {
```
nesmabadr, line 52:
```suggestion
	return errors.Is(err, syscall.ECONNREFUSED) || apierrors.IsUnauthorized(err)
```

nesmabadr (approved):

pkg/util/error.go
```diff
@@ -43,23 +44,18 @@ func IsNotFound(err error) bool {
...
 		return false
 	}
 
-	msg := strings.ToLower(err.Error())
-	if strings.Contains(msg, "connection refused") || strings.Contains(msg, "unauthorized") || strings.Contains(msg, "the server has asked for the client to provide credentials") {
-		return true
-	}
-
-	return false
+	return errors.Is(err, syscall.ECONNREFUSED) || apierrors.IsUnauthorized(err)
```
ruanxin, line 52: @c-pius, it's nice that you find these helper functions to check the error type.

Since this check is quite important, we need to have e2e test cover the case, can you create a follow up issue to design an e2e test regarding this? If need any help, feel free reach me.

pkg/util/error.go
```diff
@@ -43,23 +44,18 @@ func IsNotFound(err error) bool {
...
 }
 
 func IsUnauthorized(err error) bool {
```
ruanxin, line 55: Do we need this wrapper function? is `apierrors.IsUnauthorized(nil)` not return `false`?

c-pius: > Do we need this wrapper function?

Not necessarily I think. It could be considered a left over from our previous check.

>  is apierrors.IsUnauthorized(nil) not return false?

I think it does return `false` here.

If you prever that, I can remove this wrapper function and directly call `apierrors.IsUnauthorized`. The only place we use it so far is [here](https://github.com/kyma-project/lifecycle-manager/blob/main/internal/controller/kyma_controller.go#L135)

pkg/util/error.go
c-pius: > can you create a follow up issue to design an e2e test regarding this

✅  https://github.com/kyma-project/lifecycle-manager/issues/1310

pkg/util/error.go
ruanxin: exactly, basically, you can write a small unit test to verify it, something like:
```
func Test_unauth(t *testing.T) {
	assert.Equal(t, false, apierrors.IsUnauthorized(nil))
}
```

so let's remove this function, and use `apierrors.IsUnauthorized` directly.

c-pius: https://github.com/kyma-project/lifecycle-manager/pull/1313

> exactly, basically, you can write a small unit test to verify it, something like

did that and works as expected. I also assume that you just meant for trying this out, not actually putting a unit test like this into the project.

PR #1301, opened 2024-02-08
feat: Set Kyma Module Status to `Warning` if ModuleTemplate is not found

jeremyharisch:
**Description**

- Sets Module and Kyma Status to `Warning` if no template was found for a Module
- Adapt Test case
- Fixed indentation for `By`-statements
    - The statements related to `By` should be inside the `By` call, otherwise if a test fails, the errr message is missleading

**Related issue(s)**

### DIFF
--- a/tests/integration/controller/kyma/kyma_test.go
+++ b/tests/integration/controller/kyma/kyma_test.go
@@ -31,57 +31,60 @@ var _ = Describe("Kyma with no Module", Ordered, func() {
 	RegisterDefaultLifecycleForKyma(kyma)
 
 	It("Should result in a ready state immediately", func() {
-		By("having transitioned the CR State to Ready as there are no modules")
-		Eventually(KymaIsInState, Timeout, Interval).
-			WithContext(ctx).
-			WithArguments(kyma.GetName(), kyma.GetNamespace(), controlPlaneClient, shared.StateReady).
-			Should(Succeed())
+		By("having transitioned the CR State to Ready as there are no modules", func() {
+			Eventually(KymaIsInState, Timeout, Interval).
+				WithContext(ctx).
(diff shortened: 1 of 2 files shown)


tests/integration/controller/kyma/kyma_test.go
```diff
@@ -182,41 +193,47 @@ var _ = Describe("Kyma enable one Mandatory Module", Ordered, func() {
 
 	RegisterDefaultLifecycleForKyma(kyma)
 
-	It("should result Kyma in Error state", func() {
-		By("enabling one mandatory Module")
-		kyma.Spec.Modules = append(kyma.Spec.Modules, v1beta2.Module{
-			Name: "mandatory-template-operator",
+	It("should result Kyma in Warning state", func() {
```
c-pius, line 196: I am a bit puzzled to be honest why we are updating the test suite for **Mandatory** Modules.

My understanding is that one of the faulty cases that we are addressing with this is for instance that the customer specifies a wrong module name in the KymaCR where the according ModuleTemplate is missing and KymaCR is going to Warning state to indicate that the user needs to update this (I think we have a bug ticket for this somewhere).

For Mandatory Modules my understanding is however that the source of truth (declared state) is the presence of the ModuleTemplate. So if the ModuleTemplate is missing, the manifest should just be de-installed? And since the Mandatory Modules should be transparent to the user, also I would expect that the state of the KymaCR is unaffected, even if there is an error in the Mandatory Module (whether this is good or not is another discussion :))

TL;DR: why do we update the Mandatory Module test suite and not the regular module ones?

jeremyharisch: Lets have a quick sync about this, I think you are missing some points here :D

c-pius (approved):

Issue #1080, opened 2023-11-23
Flapping state of module in kyma-CR


k15r:
### Description

If a module goes into `warning` state during deletion the kyma-CR will report it alternating `Is Deleting` and `Warning`

### Steps to reproduce

eg:
install nats module
create resources on nats:
- port forward nats and use the nats cli
```
nats stream create foo
```
delete the nats module

it will go into warning state as the manager will not remove the reconciler (resources exist on the nats cluster)
the result can be seen in the attachment

### Environment Type

Managed

### Environment Info

Kubernetes Version: x.y.z
Docker Version: x.y.z
Kyma CLI Version: x.y.z
If used - k3d/k3s/minikube versions

### Attachments

https://github.com/kyma-project/lifecycle-manager/assets/292906/169ce74e-3bc6-41f7-b065-27fd10fdd38a

lindnerby:
We agreed on investigating the root cause of this behaviour estimate it to 1. Follow-up issue for the fix will be linked here as well.

jeremyharisch:
After investigating into this Issue I have the following conclusion:
- The reason for this flickering is
    1.  In the declarative pkg the status of the Manifest CR will be set to `Deleting` if the DeletionTimestamp has been set and the state is unequal to `Deleting`, then it triggers a new reconiliation
    2. Then if the DeletionTimestamp is set and the state is equal to `Deleting` it continues with the normal workflow of synchronising the resources, thus if the Module CR has the state `Warning` (even with set DeletionTimestamp), the state will be propagated to the Manifest CR -> Manifest has now `Warning` state
    3. Then it starts again with point 1.
- The scenario was "covered" in the E2E test, but due to a wrong implementation of the code, the test was green -> Test needs to be adapted + We need a change in the Template-Operator to simulate this scenario

**Acceptance Criterias:** 
//TODO

jeremyharisch:
Template-Operator Changes will be done in https://github.com/kyma-project/template-operator/issues/147
Unitl then this issue is blcoked.

PR #1302, opened 2024-02-09
fix: Fix flapping module state between Warning and Deleting

nesmabadr:
**Description**

- Fix flapping module state
- Adjust the E2E test

**Related issue(s)**
#1080

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -246,7 +246,7 @@ jobs:
         if: ${{ matrix.e2e-test == 'non-blocking-deletion' }}
         run: |
           kubectl apply -f tests/moduletemplates/moduletemplate_template_operator_v2_fast.yaml
-      - name: Create Template Operator Module with final state `Warning` and apply
+      - name: Create Template Operator Module with final state and final deletion state as `Warning` and apply
         working-directory: template-operator
         if: ${{ matrix.e2e-test == 'module-status-propagation'}}
         run: |
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -251,9 +251,13 @@ jobs:
         if: ${{ matrix.e2e-test == 'module-status-propagation'}}
         run: |
           pushd config/default
-          echo "- op: replace
+          echo \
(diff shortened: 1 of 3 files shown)


tests/e2e/warning_status_propagation_test.go
```diff
@@ -68,60 +70,18 @@ var _ = Describe("Warning Status Propagation", Ordered, func() {
...
-				WithArguments(runtimeClient, defaultRemoteKymaName, remoteNamespace, module).
-				Should(Succeed())
-		})
-
-		It("Then Kyma Module is in KCP Kyma CR spec", func() {
-			Eventually(ContainsModuleInSpec, Timeout, Interval).
-				WithContext(ctx).
-				WithArguments(runtimeClient, defaultRemoteKymaName, remoteNamespace, module.Name).
-				Should(Succeed())
-
```
c-pius, line 96: We have to remove this because we needed to adapt the template operator to have warning state, correct?

.github/workflows/test-e2e.yml
```diff
@@ -251,9 +251,13 @@ jobs:
...
           pushd config/default
-          echo "- op: replace
+          echo \
+          "- op: replace
             path: /spec/template/spec/containers/0/args/1
-            value: --final-state=Warning" >> warning_patch.yaml
+            value: --final-state=Warning
+          - op: replace
+            path: /spec/template/spec/containers/0/args/2
+            value: --final-deletion-state=Warning" >> warning_patch.yaml
```
c-pius, line 260: Minor: not sure if obvious or the stage name needs to be adapted?

```diff
- Create Template Operator Module with final state `Warning` and apply
+ Create Template Operator Module with final state + final deletion state `Warning` and apply
```

tests/e2e/warning_status_propagation_test.go
nesmabadr: yes this is at the end of the test. so right now there is the finalizer so the module CR shouldn't be removed and thus kyma not ready. so what happens is that the finalizer is removed at the end so the module CR is removed and kyma is back to ready

c-pius (approved):

PR #1312, opened 2024-02-13
feat: Deprecation of RemoteTargetRef in v1beta2

jeremyharisch:
**Description**

- Deletes the functionality of Kyma.Spec.Modules[]. RemoteTargetRef

**Related issue(s)**


pkg/templatelookup/regular.go
```diff
@@ -32,7 +30,9 @@ type ModuleTemplateInfo struct {
 	DesiredChannel string
 }
 
-func NewTemplateLookup(reader client.Reader, descriptorProvider *provider.CachedDescriptorProvider, syncEnabled bool) *TemplateLookup {
+func NewTemplateLookup(reader client.Reader, descriptorProvider *provider.CachedDescriptorProvider,
+	syncEnabled bool,
```
lindnerby, line 34: `syncEnabled` can now also be removed from `TemplateLookup` 👍

tests/integration/controller/controlplane/kyma_remote_sync_test.go
```diff
@@ -370,16 +305,16 @@ var _ = Describe("Kyma sync default module list into Remote Cluster", Ordered, f
 
 var _ = Describe("CRDs sync to SKR and annotations updated in KCP kyma", Ordered, func() {
 	kyma := NewTestKyma("kyma-test-crd-update")
-	moduleInKcp := NewTestModule("in-kcp", v1beta2.DefaultChannel)
-	kyma.Spec.Modules = []v1beta2.Module{moduleInKcp}
-
-	remoteKyma := &v1beta2.Kyma{}
+	moduleInKCP := NewTestModule("in-kcp", v1beta2.DefaultChannel)
```
lindnerby, line 308: Can you reiterate over this test case please.
For example:
`moduleInKCP` gets created with name and channel, but is nowhere used except its name and channel in next line, where even the list is initialized with a new `v1beta2.Module`.

lindnerby (changes requested):

pkg/templatelookup/regular.go
jeremyharisch: Nice, catch! Thanks

tests/integration/controller/controlplane/kyma_remote_sync_test.go
jeremyharisch: It is used, as you already emtnioned in the next line. We use the `NewTestModule` function to generate a unique name. It then gets inserted into the Kyma Spec and a few lines after with `registerControlPlaneLifecycleForKyma(kyma)` a template gets created for it and deployed on the cluster. 

Indeed, we had a duplication of creating the template twice, but I now removed it

lindnerby (approved):

tests/integration/controller/controlplane/kyma_remote_sync_test.go
lindnerby: Yeah but we already pass name and channel and never use the Module itself. The unique name generation is not having any effect. But nevermind. :)

mmitoraj (approved):

PR #1321, opened 2024-02-17
feat: Add Manifest Reconcile Duration Metric

### LINKED ISSUE #1069: Introduce metrics to track each Module Manifest reconcile performance
**Description:**

At the moment, LM has the internal implemented performance metrics for manifest CR reconcile, however, it lacks of detailed view to tell how is each Module individually performed, which makes the troubleshooting difficult if some problematic modules are introduced.

**AC Updates**

- [x] Introduce Gauge metrics `reconcile_duration_seconds` with Manifest name as label to track each Manifest success reconciliation duration time
  -  configure a duration threshold value (1mins), only add new metric above the threshold, also take care to delete existing metrics if duration time is lower than the threshold.

- [x] Add a new metric diagram in the Plutono dashboard.

LeelaChacha:
**Description**

- added histogram metric `reconcile_duration_seconds` to KLM with module name as label
- ...
- ...

**Related issue(s)**

Closes #1069


LeelaChacha:
Note: Due to problems with local setup of plutono dashboard, I was unable to create a new visualization for this metric. It shall be done once this PR merged using DEV landscape's dashboard. The visualization will follow in another PR.

internal/pkg/metrics/manifest.go
```diff
@@ -1,10 +1,19 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	"github.com/kyma-project/lifecycle-manager/pkg/queue"
+)
 
 type ManifestRequeueReason string
 
 const (
+	metricReconcileDuration               string                = "reconcile_duration_seconds"
```
nesmabadr, line 15: Can you add a test case to the E2E kyma_metrics_test that this metric has a non-zero value?

internal/pkg/metrics/manifest.go
```diff
@@ -28,12 +37,31 @@ const (
...
 
 func NewManifestMetrics(sharedMetrics *SharedMetrics) *ManifestMetrics {
-	return &ManifestMetrics{SharedMetrics: sharedMetrics}
+	metrics := &ManifestMetrics{
+		SharedMetrics: sharedMetrics,
+		reconcileDurationHistogram: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Name:    metricReconcileDuration,
+				Help:    "Histogram of reconcile duration for manifest reconciliation in seconds",
+				Buckets: prometheus.DefBuckets,
```
nesmabadr, line 50: We might need to revisit the buckets here when the visualization is there

LeelaChacha: aren't the buckets irrelevant if we use count and sum of the histogram metric to derive the latency visualisation? This is also how the latency metrics in current Kyma Overview dashboard are shown

internal/pkg/metrics/manifest.go
LeelaChacha: oh. thanks for the info, didn"t know of this test. will implement

pkg/testutils/metrics.go
```diff
@@ -69,12 +69,25 @@ func IsManifestRequeueReasonCountIncreased(ctx context.Context, requeueReason, r
...
+) {
+	bodyString, err := getMetricsBody(ctx)
+	if err != nil {
+		return false, err
+	}
+	re := regexp.MustCompile(
+		metrics.MetricReconcileDuration + `_count{` + metrics.MetricLabelModule +
+			`="` + manifestName + reMatchCount)
+	count, err := parseCount(re, bodyString)
+	return count >= 0, err
```
nesmabadr, line 88: `count > 0`, and the count in this case should be non int because I assume it would be 0. something

internal/pkg/metrics/manifest.go
LeelaChacha: yes, there is histogram metric provided by the controller runtime. it is called `controller_runtime_reconcile_time_seconds`. This is what we are trying to replace with this metric.
We display it in Kyma Overview `Reconcile Duration` Visualization. Here is the PromQL query for it:
```
rate(controller_runtime_reconcile_time_seconds_sum{controller="manifest"}[$__rate_interval])
/
rate(controller_runtime_reconcile_time_seconds_count{controller="manifest"}[$__rate_interval])
```
Therefore, we do not need the buckets.

pkg/testutils/metrics.go
LeelaChacha: good catch. The count of a histogram metric is an integer

internal/pkg/metrics/manifest.go
LeelaChacha: but I can change the buckets to match the controller runtime metric

nesmabadr: okay we can have them like this for now till we have the visualization.
in the above question, I meant histograms defined in the lifecycle manager not those of the controller runtime

nesmabadr (approved):

Issue #1287, opened 2024-02-05
Setting up proper owner references for VirtualServices managed by the KLM


jeremyharisch:
### Description

For the runtimeWatcher the KLM sets up VirtualServices during runtime. To have a better overview of what the KLM manages, we should set proper OwnerReferences to the VirtualServices owned by the KLM.

The VirtualServices are created here: https://github.com/kyma-project/lifecycle-manager/blob/main/internal/controller/watcher_controller.go#L147

### Reasons

- Better overview of what KLM is managing
- Avoid having `orphaned` resources

### Acceptance Criteria

- [x] When creating VirtualServices, an OwnerReference should be placed leading to the ~~KLM~~ Watcher
- [x] Adapt integration tests, to check whether OwnerReference is set

PR #1322, opened 2024-02-19
wip: Set `OwnerReferences` for watcher `VirtualService`

c-pius:
**Description**

- refactors `IstioClient` by carving out the part of creating the `VirtualService` object to a dedicated `IstioService` (see commit https://github.com/kyma-project/lifecycle-manager/pull/1322/commits/14377c5258925d4fb4aa2eb7e80b3190476bda7c)
- implements setting of the `OwnerReferences` of the `VirtualService` (see commit https://github.com/kyma-project/lifecycle-manager/pull/1322/commits/ffcef258188c5fbd9ce8d432f1e4ee93f5ce456c)
  - improved this initial idea then with a cached implementation based on the implementation for `Descriptor`(see commit https://github.com/kyma-project/lifecycle-manager/pull/1322/commits/947816145bcc60c651c28b49c062014f0accfaa2)
- some chore improvements of this code above (here https://github.com/kyma-project/lifecycle-manager/pull/1322/commits/8ea65fe263802e33210ad5cd52e3928225cdaedb and here https://github.com/kyma-project/lifecycle-manager/pull/1322/commits/0b264b63d7cd46fb3e621dbf934ee6a316a6b4b0)
- TODO: add integration test
- TODO: add unit tests

**Related issue(s)**

- resolves https://github.com/kyma-project/lifecycle-manager/issues/1287


internal/pkg/flags/flags.go
```diff
@@ -52,6 +52,8 @@ const (
 	DefaultWatcherResourceLimitsMemory                                  = "200Mi"
 	DefaultDropStoredVersion                                            = "v1alpha1"
 	DefaultMetricsCleanupIntervalInMinutes                              = 15
+	DefaultKlmControllerManagerName                                     = "klm-controller-manager"
```
ruanxin, line 55: we don't need to define owner as flag, owner can be a static variable, not necessary to be configured, please check [here](https://github.com/kyma-project/lifecycle-manager/blob/9be2538f0201fe61eb7ac1f620242137f6178c74/internal/controller/watcher_controller.go#L191) how we setup owner for other resources, use `shared.OperatorName`

internal/pkg/flags/flags.go
```diff
@@ -52,6 +52,8 @@ const (
...
+	DefaultKlmControllerManagerNamespace                                = "kcp-system"
```
ruanxin, line 56: we already have `DefaultIstioGatewayNamespace` use `kcp-system`, basically istio gateway and klm are deployed under same namespace, think about reuse existing flag, also consider renaming `DefaultIstioGatewayNamespace` since this namespace is not only for istio gateway anymore.

internal/pkg/flags/flags.go
c-pius: to be discussed with the team tomorrow 👍🏻

c-pius: Agreed that we are going to use `WatcherCR` as the owner instead. This will make a big part of this PR obsolete. Will consider to close this one in favor of another more streamlined one.

c-pius:
Closing in favor of https://github.com/kyma-project/lifecycle-manager/pull/1342 and TBD

PR #1330, opened 2024-02-20
Introduce Mandatory Modules Metrics

jeremyharisch:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -164,9 +164,10 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
 	sharedMetrics := metrics.NewSharedMetrics()
 	descriptorProvider := provider.NewCachedDescriptorProvider(nil)
 	kymaMetrics := metrics.NewKymaMetrics(sharedMetrics)
+	mandatoryModulesMetrics := metrics.NewMandatoryModulesMetrics()
 	setupKymaReconciler(mgr, remoteClientCache, descriptorProvider, flagVar, options, skrWebhookManager, kymaMetrics)
 	setupManifestReconciler(mgr, flagVar, options, sharedMetrics)
-	setupMandatoryModuleReconciler(mgr, descriptorProvider, flagVar, options)
+	setupMandatoryModuleReconciler(mgr, descriptorProvider, flagVar, options, mandatoryModulesMetrics)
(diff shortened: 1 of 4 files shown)


internal/declarative/v2/reconciler.go
```diff
@@ -114,6 +117,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return r.ssaStatus(ctx, obj, metrics.ManifestInit)
 	}
 
+	logf.FromContext(ctx).Info("Before setting the Mandatory Module Metrics")
+	if obj.GetLabels()[shared.IsMandatoryModule] == strconv.FormatBool(true) {
```
nesmabadr, line 121: This shouldn't happen but just in case can you add `if object.GetLabels() != nil ` to the check as well?

tests/e2e/mandatory_modules_metrics_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+			By("And count of Mandatory ModuleTemplates Metric is 1", func() {
+				Eventually(GetMandatoryModuleTemplateCountMetric).
+					WithContext(ctx).
+					Should(Equal(1))
+			})
+		})
+
+		It("When the mandatory ModuleTemplate is removed", func() {
+			Eventually(DeleteCR).
```
nesmabadr, line 57: There is the [DeleteModuleTemplate](https://github.com/kyma-project/lifecycle-manager/blob/165c9f578d846cf1a1aa48a8e9245e384f1c948c/pkg/testutils/moduletemplate.go#L74) method that you could use here instead of the DeleteCR

jeremyharisch: This we actually cannot use, since this `DeleteModuleTemplate` func uses the `GetModuleTemplate` func given by our test utils. And this uses the `TemplateLookup` to fetch the ModuleTemplate - but only REGULAR templates and not mandatory ModuleTemplates. So, we could either introduce a `DeleteMandatoryModuleTemplate` function, or adapt the existing with a new bool input parameter, to switch between different LookUp mechanisms (in my opinion to overengineered) or we just delete the CR, since we know what the name is.

Issue #1281, opened 2024-02-01
Blocking deletion mode


janmedrek:
### Description

Lifecycle Manager should wait until all the managed resources are gone before proceeding with deleting the module operator.

Right now, Lifecycle Manager waits until the default CR is gone and follows up with the deletion of the module operator. This scenario allows the module operator to clean up the resources properly (or go into the `Warning` state).

This behaviour can also cause the modules to block the deletion forever if any CRs were not cleaned up but the default one and the operator are gone - they will stop the deletion of CRD infinitely and thus - the module will not be removed and stay in the `Deleting` state.

The solution for that situation would be to introduce a blocking deletion procedure:
- KLM should monitor whether there are any managed resources left.
  - If there are none - it should remove the operator and CRDs
  - If there are any - wait until the resources are gone

The list of managed resources can be taken directly from the manifest (all the CRDs in the module manifest are the managed resources).

### Reasons

We want to ensure safe deletion for the modules, it will also make it easier to introduce the community modules integration in the future. 

### Acceptance Criteria

- Given a **Kyma Module**
  - And a SKR **Kyma CR** is created on the **SKR Cluster**
  - And a **Manifest CR** is created on the **KCP Cluster**
- When the deletion is requested for the **Kyma Module**
  - Then the default **Module CR** is still marked for deletion
  - Then the **[Module Specific CRs]** are being monitored
    - And the **Module Operator** and **[Module Specific CRDs]** are not deleted if there are any **[Module Specific CRs]**
      - And the **State** is set to **Warning**
      - And the list of **[Module Specific CRDs]** is listed in the **Status** 
    - And the **Module Operator** and **[Module Specific CRDs]** are deleted if all **[Module Specific CRs]** are gone

### Feature Testing

- Integration tests
  - New test implemented if required
  - Existing tests adapted if required 
- E2E tests
  - New test implemented
  - Existing tests adapted if required

janmedrek:
Blocked by https://github.com/kyma-project/lifecycle-manager/issues/1080

ruanxin:
**Module Operator** here is not only `Deployment` and `Service`, but should also include all resources which ensure module operator functional correctly, which include RBAC related resources, e.g: `ServiceAccount`, `Role`, `ClusterRole`, `RoleBinding`, `ClusterRoleBinding`.

Issue #1006, opened 2023-11-06
Remove `declarative` library


jeremyharisch:
### Description

We have introduced the `declarative` pkg since it was planned to reuse this pkg in each Kyma Module operator. Plans now have changed and this pkg already has been moved to the internal section. Let's remove this pkg and refactor the manifest reconciler to the operator pattern instead declarative.

### Reasons

- Easier to read and debug manifest reconciler
- Easier to implement new features for manifest reconciler
- `declarative` pkg is not a pkg which should be reused outside of KLM

### Acceptance Criteria

- [ ] Remove declarative pkg
- [ ] Refactor manifest reconciler to have the same behaviour as before
- [ ] Adapt test suits incl. unit- and integration-tests

PR #1336, opened 2024-02-21
fix: Split resource deletion

ruanxin:
related issue: https://github.com/kyma-project/lifecycle-manager/issues/1281

This PR address above issue by split diff resource during deletion, it trying to delete resources managed by module operator first, after those managed resources are deleted, then LM start to delete operator-related resources, e.g: `ServiceAccount`, `Role`, `ClusterRole`, `RoleBinding`, `ClusterRoleBinding`. to ensure the module operator not being affected during deletion.


tests/moduletemplates/moduletemplate_template_operator_v1_regular.yaml
```diff
@@ -5,8 +5,12 @@ metadata:
   namespace: kcp-system
```
ruanxin, line 5: To reviewer:

All these moduletemplates are include the managed resource `Managed` from this [PR](https://github.com/kyma-project/template-operator/pull/162), you can of cause download the content from oci registry to verify.

tests/integration/declarative/declarative_test.go
```diff
@@ -1,384 +0,0 @@
-package declarative_test
```
ruanxin, line 1: To reviewer:

The reason to remove this declarative test:
1. I explicitly cast the generic object to manifest in [here](https://github.com/kyma-project/lifecycle-manager/pull/1336/files#diff-45632cfa5b5d235890e597e29448382f7c61fc85a55d1629de012328056e69a2R496) which made this declarative test not working anymore, since it use a customized TestAPI to test the reconciler function.
2. Since we decided to completely [remove this declarative library](https://github.com/kyma-project/lifecycle-manager/issues/1006), there is no need to support such tests anymore.
3. We have integration test for the manifest controller which tests against manifest CR specifically.

internal/pkg/resources/cleanup.go
```diff
@@ -0,0 +1,134 @@
...
+		kind == "Role" ||
+		kind == "ClusterRole" ||
+		kind == "RoleBinding" ||
+		kind == "ControllerManagerConfig" ||
+		kind == "ClusterRoleBinding" ||
+		kind == "Service" ||
+		kind == "Deployment"
+}
+
+func (c *ConcurrentCleanup) Run(ctx context.Context, infos []*resource.Info) error {
```
lindnerby, line 98:
```suggestion
func (c *ConcurrentCleanup) run(ctx context.Context, infos []*resource.Info) error {
```

internal/pkg/resources/cleanup.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	if err := c.CleanupResources(ctx, operatorRelatedResources, status); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *ConcurrentCleanup) CleanupResources(
```
lindnerby, line 52:
```suggestion
func (c *ConcurrentCleanup) cleanupResources(
```

pkg/testutils/manifest.go
```diff
@@ -51,6 +51,7 @@ func NewTestManifest(prefix string) *v1beta2.Manifest {
 	}
 }
 
+// GetManifest is only used when manifest still been tracked in kyma.status.
```
lindnerby, line 54: Sorry, I don't get this comment 😄 It's a test util function that should only be used when? Or what is meant by it?

lindnerby (changes requested):
This is a good opportunity to increase unit test coverage.
There is no [internal/pkg/resources/cleanup_test.go](url). Let's add these and also lock this pkg in `unit-test-coverage.yaml`!
May be a bit of work and refactoring needed, and also concurrency involved internally, so not super trivial, but we agreed to increase the coverage.

tests/integration/declarative/declarative_test.go
lindnerby: Agreed 👌

tests/e2e/kyma_metrics_test.go
```diff
@@ -1,14 +1,16 @@
 package e2e_test
 
 import (
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
```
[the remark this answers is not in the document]
ruanxin: actually after check in my goland import setting, there is no such way to control dot import position, it's by package name. As you can see in current e2e tests, the dot import is not always in the bottom. If we don't have an automatic import rule, I don't think it's good idea we define some rules and have to sort them manually.

pkg/testutils/manifest.go
ruanxin: If you check the internal logic of this function, it uses `kyma.status.Modules` to get Manifest CR name, as you know the Manifest CR name contains a hash value, which calculated by kyma and module fqdn which coming from module template, there is not easy way to get this name directly, otherwise we need to parsing moduletemplate, to get it from component descriptor.

However, if the module is disabled, it will not exists in the `kyma.status`, that's the reason why I add this commit, maybe we can enhance the commit a bit, any suggestion?

tests/e2e/kyma_metrics_test.go
```diff
@@ -72,17 +74,20 @@ var _ = Describe("Manage Module Metrics", Ordered, func() {
 		})
 
 		It("When Kyma Module is disabled", func() {
+			manifestInCluster, err := GetManifest(ctx, controlPlaneClient, kyma.GetName(), kyma.GetNamespace(),
+				module.Name)
+			Expect(err).Should(Succeed())
 			Eventually(DisableModule).
```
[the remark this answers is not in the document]
ruanxin: what you suggest is in `module_deletion_test.go`, this one is for testing metrics, and why I have to modify this test is because of the reason [here](https://github.com/kyma-project/lifecycle-manager/pull/1336#discussion_r1515882254)

tests/e2e/kyma_metrics_test.go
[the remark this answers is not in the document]
lindnerby: Strange, I though according to the linter the dot imports even need to be as a last section:
https://github.com/kyma-project/lifecycle-manager/blob/ef6228415a38e5a534122a1fbf0471193b181db5/.golangci.yaml#L55

internal/pkg/resources/cleanup.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+		operatorManagedResources = append(operatorManagedResources, resource)
+	}
+
+	return operatorRelatedResources, operatorManagedResources, nil
+}
+
+func isOperatorRelatedResources(kind string) bool {
+	return kind == "CustomResourceDefinition" ||
+		kind == "ServiceAccount" ||
```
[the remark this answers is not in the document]
ruanxin: do you really think the value in the list need to be const, what value it brings? Actually what I expected orignailly is that we can have some const variable can be directly reference from `k8s.io/api` package, to prevent this resource name get changed in future (although in rare cases it will happen), but unforturnately, can't find any. 

But it's good you point out this, I create a unit test to make sure the kind name not get changed, please check.

internal/pkg/resources/cleanup.go
ruanxin: it's one of the implementation of SSA interface
https://github.com/kyma-project/lifecycle-manager/blob/386e15e46f384b43f92c8349c4b66bb52b1ff864/internal/declarative/v2/ssa.go#L26

tests/e2e/kyma_metrics_test.go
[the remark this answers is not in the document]
lindnerby: Alright

PR #1342, opened 2024-02-23
chore: Refactor `istio` pkg

c-pius:
**Description**

- refactors the construction of the `VirtualService` object
  - moves the object construction out of the istio client
  - removes the dependency on the client for fetching the gateways and instead receives the gateways as an argument
  - => we have a pure function where output is fully dependent on input that is easily unit testable
  - added some guard clauses / sanity checks
- adds unit tests for the `VirtualService` construction
- refactors the construction of the `HttpRoute` object
  - added some guard clauses / sanity checks
- adds unit tests for the `HttpRoute` construction
- aligns methods in `istio#client.go` wrt. method name, argument name, return values
- moves the `istio` package from `pkg` to `internal`

**Related issue(s)**

- Preparation for https://github.com/kyma-project/lifecycle-manager/issues/1287

<details>

<summary>Examples of VirtualService objects</summary>

Before and After look equal ✅ 

### Before

```yaml
kind: VirtualService                                        
metadata:                                                   
  creationTimestamp: "2024-02-23T10:56:53Z"                 
  generation: 1                                             
  name: klm-kyma-watcher                                    
  namespace: kcp-system                                     
  resourceVersion: "86048"                                  
  uid: de1e8e9e-358e-46ad-a994-8795276ae55d                 
spec:                                                       
  gateways:                                                 
  - kcp-system/klm-watcher-gateway                          
  hosts:                                                    
  - host.k3d.internal                                       
  http:                                                     
  - match:                                                  
    - uri:                                                  
        prefix: /v1/lifecycle-manager/event                 
    name: kcp-system/klm-kyma-watcher                       
    route:                                                  
    - destination:                                          
        host: klm-event-service.kcp-system.svc.cluster.local
        port:                                               
          number: 8082                                      
                                             
```

### After

```yaml
kind: VirtualService                                        
metadata:                                                   
  creationTimestamp: "2024-02-23T10:54:20Z"                 
  generation: 1                                             
  name: klm-kyma-watcher                                    
  namespace: kcp-system                                     
  resourceVersion: "85740"                                  
  uid: 6abd22b4-bcce-4c99-848b-d022db6a2299                 
spec:                                                       
  gateways:                                                 
  - kcp-system/klm-watcher-gateway                          
  hosts:                                                    
  - host.k3d.internal                                       
  http:                                                     
  - match:                                                  
    - uri:                                                  
        prefix: /v1/lifecycle-manager/event                 
    name: kcp-system/klm-kyma-watcher                       
    route:                                                  
    - destination:                                          
        host: klm-event-service.kcp-system.svc.cluster.local
        port:                                               
          number: 8082       
```

</details>

### DIFF
--- a/internal/controller/setup.go
+++ b/internal/controller/setup.go
@@ -142,7 +142,7 @@ func (r *WatcherReconciler) SetupWithManager(mgr ctrl.Manager, options ctrlrunti
 		return errRestConfigIsNotSet
 	}
 	var err error
-	r.IstioClient, err = istio.NewVersionedIstioClient(r.RestConfig, r.EventRecorder,
+	r.IstioClient, err = istio.NewIstioClient(r.RestConfig, r.EventRecorder,
 		ctrl.Log.WithName("istioClient"))
 	if err != nil {
 		return fmt.Errorf("unable to set istio client for watcher controller: %w", err)
--- a/internal/controller/setup.go
+++ b/internal/controller/setup.go
@@ -22,7 +22,7 @@ import (
 
 	"github.com/kyma-project/lifecycle-manager/api/shared"
 	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
-	"github.com/kyma-project/lifecycle-manager/pkg/istio"
+	"github.com/kyma-project/lifecycle-manager/internal/istio"
 	"github.com/kyma-project/lifecycle-manager/pkg/security"
 	"github.com/kyma-project/lifecycle-manager/pkg/watch"
 )
--- a/internal/controller/watcher_controller.go
+++ b/internal/controller/watcher_controller.go
@@ -33,7 +33,6 @@ import (
 	"github.com/kyma-project/lifecycle-manager/api/shared"
 	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
 	"github.com/kyma-project/lifecycle-manager/internal/istio"
-	"github.com/kyma-project/lifecycle-manager/internal/istio/resource"
 	"github.com/kyma-project/lifecycle-manager/pkg/log"
 	"github.com/kyma-project/lifecycle-manager/pkg/queue"
 	"github.com/kyma-project/lifecycle-manager/pkg/status"
@@ -150,7 +149,7 @@ func (r *WatcherReconciler) handleProcessingState(ctx context.Context,
 		return r.updateWatcherState(ctx, watcherCR, shared.StateError, err)
 	}
 
-	virtualSvc, err := resource.NewVirtualService(r.WatcherVSNamespace, watcherCR, gateways)
+	virtualSvc, err := istio.NewVirtualService(r.WatcherVSNamespace, watcherCR, gateways)
 	if err != nil {
 		return r.updateWatcherState(ctx, watcherCR, shared.StateError, err)
 	}
--- a/internal/controller/watcher_controller.go
+++ b/internal/controller/watcher_controller.go
@@ -129,7 +129,7 @@ func (r *WatcherReconciler) stateHandling(ctx context.Context, watcherCR *v1beta
 }
 
 func (r *WatcherReconciler) handleDeletingState(ctx context.Context, watcherCR *v1beta2.Watcher) (ctrl.Result, error) {
-	err := r.IstioClient.RemoveVirtualServiceForCR(ctx, client.ObjectKeyFromObject(watcherCR), r.WatcherVSNamespace)
+	err := r.IstioClient.DeleteVirtualService(ctx, watcherCR.GetName(), r.WatcherVSNamespace)
 	if err != nil {
 		vsConfigDelErr := fmt.Errorf("failed to delete virtual service (config): %w", err)
 		return r.updateWatcherState(ctx, watcherCR, shared.StateError, vsConfigDelErr)
--- a/internal/controller/watcher_controller.go
+++ b/internal/controller/watcher_controller.go
@@ -144,8 +144,10 @@ func (r *WatcherReconciler) handleDeletingState(ctx context.Context, watcherCR *
 func (r *WatcherReconciler) handleProcessingState(ctx context.Context,
 	watcherCR *v1beta2.Watcher,
 ) (ctrl.Result, error) {
-	gateways, err := r.IstioClient.LookupGateways(ctx, watcherCR)
-	if err != nil {
+	gateways, err := r.IstioClient.ListGatewaysByLabelSelector(ctx, &watcherCR.Spec.Gateway.LabelSelector)
+	if err != nil || len(gateways.Items) == 0 {
+		r.EventRecorder.Event(watcherCR, "Warning", "WatcherGatewayNotFound",
+			"Watcher: Gateway for the VirtualService not found")
 		return r.updateWatcherState(ctx, watcherCR, shared.StateError, err)
 	}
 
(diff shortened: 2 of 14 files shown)


internal/controller/setup.go
```diff
@@ -22,7 +22,7 @@ import (
 
 	"github.com/kyma-project/lifecycle-manager/api/shared"
 	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
-	"github.com/kyma-project/lifecycle-manager/pkg/istio"
```
nesmabadr, line 25: What is the reason behind moving istio from `pkg` to `internal`?

internal/istio/resource/virtualservice.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+const (
+	contractVersion = "v1"
+	prefixFormat    = "/%s/%s/event"
+)
+
+func NewVirtualService(targetNamespace string, watcher *v1beta2.Watcher, gateways []*istioclientapiv1beta1.Gateway) (*istioclientapiv1beta1.VirtualService, error) {
+	if targetNamespace == "" {
+		return nil, fmt.Errorf("%w targetNamespace", ErrInvalidArgument)
```
nesmabadr, line 21: Also print the value of targetNamespace (which is empty in this case), or rewrite the error that it shows that the namespace is empty

nesmabadr: Same for the watcher and watcher.Name.

internal/istio/client.go
```diff
@@ -68,34 +63,6 @@ func (c *Client) ListVirtualServices(ctx context.Context, namespace string) (*is
 	return virtualServiceList, nil
 }
```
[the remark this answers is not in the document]
c-pius: My intention was to separate the code concerned with the object itself (i.e. the construction in this case) from the client that is using the objects. Personally, I am preferring such separation as it makes it easier to understand what code deals with what concern. If you see it differently I am happy to discuss.

Thinking about it, I also noticed that `Client` in `istio/client.go` is as of now also not really an "IstioClient". Maybe it is worth also renaming this to `VirtualServiceClient`. But this is a slightly different discussion.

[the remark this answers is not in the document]
c-pius: Yeah mean like in the following?

```
- istio
  - resource
    - ...
  - istioclient.go
    - # Client struct
    - # NewVersionedIstioClient
  - virtualserviceclient.go
    - # GetVirtualService
    - # ListVirtualService
    - ...
```

It sounds doable (because of the use of receiver functions), but I unfortunately lack the experience whether this is common practice or not in go. I will try to check :)

internal/controller/setup.go
```diff
@@ -142,7 +142,7 @@ func (r *WatcherReconciler) SetupWithManager(mgr ctrl.Manager, options ctrlrunti
 		return errRestConfigIsNotSet
 	}
 	var err error
-	r.IstioClient, err = istio.NewVersionedIstioClient(r.RestConfig, r.EventRecorder,
+	r.IstioClient, err = istio.NewIstioClient(r.RestConfig, r.EventRecorder,
```
c-pius, line 145: Removed `versioned` as the constructor doesn't include any special semantics with regards to version.

internal/controller/watcher_controller.go
```diff
@@ -144,7 +144,14 @@ func (r *WatcherReconciler) handleDeletingState(ctx context.Context, watcherCR *
 func (r *WatcherReconciler) handleProcessingState(ctx context.Context,
 	watcherCR *v1beta2.Watcher,
 ) (ctrl.Result, error) {
-	virtualSvc, err := r.IstioClient.NewVirtualService(ctx, watcherCR, r.WatcherVSNamespace)
+	gateways, err := r.IstioClient.ListGatewaysByLabelSelector(ctx, &watcherCR.Spec.Gateway.LabelSelector)
+	if err != nil || len(gateways.Items) == 0 {
+		r.EventRecorder.Event(watcherCR, "Warning", "WatcherGatewayNotFound",
+			"Watcher: Gateway for the VirtualService not found")
+		return r.updateWatcherState(ctx, watcherCR, shared.StateError, err)
+	}
```
c-pius, line 152: I have moved the length check and event recording outside of the client. If there is no error in the client, it shouldn't need to worry about if there are objects or not. The caller needs to determine if an empty list is an error case or not. In this case yes, since this is the behavior we had before.

internal/istio/client.go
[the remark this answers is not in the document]
c-pius: I thought about it some time and moved the virtual service part out of `resources` again and made it part of the `istio` package directly. The API surface of `istio` package is still fairly small. If in the future we see that it is growing too big, we can introduce sub-packages.

Wrt. to moving the CRUD functions to a virtual service client: I looked a bit closer at the client and was able to carve out further functions that did not really belong to client. Now we have a couple of VirtualService related methods and one method related to gateways. For simplicity, I opted now for keeping it a `istio` level client since it includes VS and GW related methods. We could of course split it into two clients but it may be a bit overkill to handle two client dependencies. Also the resulting client file is now fairly small and concise, so it would be okay for me to keep it as is. But if you prefer to move the receiver functions out to separate files I can also do so

tests/integration/controller/withwatcher/watcher_controller_helper_test.go
```diff
@@ -283,3 +284,32 @@ func listenerHTTPRouteExists(ctx context.Context, clt *istio.Client, namespace s
 
 	return errRouteNotFound
 }
+
+func isRouteConfigEqual(route1 *istioapiv1beta1.HTTPRoute, route2 *istioapiv1beta1.HTTPRoute) bool {
```
c-pius, line 288: Used to be part of client, but was only used in this test file. So I opted for moving it into the test.

internal/istio/client.go
```diff
@@ -0,0 +1,112 @@
...
+	err := c.NetworkingV1beta1().
+		VirtualServices(namespace).
+		Delete(ctx, name, apimetav1.DeleteOptions{})
+	if err != nil {
+		return errors.Join(ErrFailedToDeleteVirtualService, err)
+	}
+	return nil
+}
+
+func (c *Client) ListGatewaysByLabelSelector(ctx context.Context, labelSelector *apimetav1.LabelSelector) (*istioclientapiv1beta1.GatewayList,
```
c-pius, line 93: This was previously `GatewayLookup` which received the watcher and returned the list of gateway items. I opted for directly using the LabelSelector argument now since we were only getting this exact thing from the watcher CR, and return the `GatewayList` instead for consistency with the other list functions.

internal/istio/virtualservice.go
```diff
@@ -0,0 +1,93 @@
...
+
+	if watcher == nil {
+		return nil, fmt.Errorf("watcher must not be nil: %w", ErrInvalidArgument)
+	}
+
+	if watcher.GetName() == "" {
+		return nil, fmt.Errorf("watcher.Name must not be empty: %w", ErrInvalidArgument)
+	}
+
+	if gateways == nil {
```
nesmabadr, line 32: These 2 checks could me merged into one.

`if gateways == nil || len(gateways.Items)`

internal/istio/httproute.go
```diff
@@ -0,0 +1,65 @@
...
+
+	istioapiv1beta1 "istio.io/api/networking/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+)
+
+func NewHTTPRoute(watcher *v1beta2.Watcher) (*istioapiv1beta1.HTTPRoute, error) {
+	if watcher == nil {
+		return nil, fmt.Errorf("watcher must not be nil: %w", ErrInvalidArgument)
```
nesmabadr, line 14: I suggest moving these checks to another helper function that would only validate the watcher then call it here

nesmabadr (approved):

Issue #1125, opened 2023-12-06
Refactoring: Global In-Memory Caches


jeremyharisch:
### Description

Currently, three of our in-memory Cache implementation are solved by using a global variable in the corresponding package. A better design would to be get rid of global variables in packages and implement the in-memory cache in for example a constructor pattern.

Using global variables as in-memory cache should be avoided.

### Reasons

Avoid using global variables in packages.

### Acceptance Criteria

- [x] Refactor `descriptorCache` in [v1beta2 pkg](https://github.com/kyma-project/lifecycle-manager/blob/aef1529f43842c6aa2e14d5a52ca8b5049fad7f4/api/v1beta2/moduletemplate_types.go#L158) to not be implemented as global variabel
- [x] Refactor `fileMutexMap` in [manifest pkg ](https://github.com/kyma-project/lifecycle-manager/blob/aef1529f43842c6aa2e14d5a52ca8b5049fad7f4/internal/manifest/parse.go#L26) to not be implemented as global variabel
- [x] Refactor `crdCache` in [cache pkg ](https://github.com/kyma-project/lifecycle-manager/blob/aef1529f43842c6aa2e14d5a52ca8b5049fad7f4/pkg/cache/crd_cache.go#L10) to not be implemented as global variabel

### Feature Testing

Unit tests

PR #1344, opened 2024-02-26
chore: Extract filemutex cache

lindnerby:
**Description**

- Extract the locker caching when manifest parsing
- Add coverage and tests

**Related issue(s)**
* https://github.com/kyma-project/lifecycle-manager/issues/1125

### DIFF
--- a/internal/manifest/filemutex/cache.go
+++ b/internal/manifest/filemutex/cache.go
@@ -1,0 +1,33 @@
+package filemutex
+
+import (
+	"errors"
+	"sync"
+)
+
+var ErrMutexConversion = errors.New("failed to convert cached value to mutex")
+
+type MutexCache struct {
+	cache *sync.Map
+}
+
+func NewMutexCache(concurrMap *sync.Map) *MutexCache {
+	if concurrMap == nil {
+		return &MutexCache{cache: &sync.Map{}}
+	}
+	return &MutexCache{cache: concurrMap}
+}
+
+// GetLocker always returns the same sync.Locker instance for given key.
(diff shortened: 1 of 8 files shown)


internal/manifest/filemutex/cache_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/kyma-project/lifecycle-manager/internal/manifest/filemutex"
+)
+
+func TestNewMutexCache_WhenCalled_NotNil(t *testing.T) {
```
c-pius, line 13: The other PR has an explicit test for the constructor with a cache object being passed in. I consider it tested implicitly with all the tests below. But it probably wouldn't harm to have it explicitly included here as well.

internal/manifest/filemutex/cache.go
```diff
@@ -0,0 +1,33 @@
...
+type MutexCache struct {
+	cache *sync.Map
+}
+
+func NewMutexCache(concurrMap *sync.Map) *MutexCache {
+	if concurrMap == nil {
+		return &MutexCache{cache: &sync.Map{}}
+	}
+	return &MutexCache{cache: concurrMap}
+}
```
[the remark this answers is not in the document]
lindnerby: Yes, unfortunately limitation of the language itself. Would also prefer optional params. 🙏

internal/manifest/filemutex/cache_test.go
lindnerby: True, added.

PR #1345, opened 2024-02-26
chore: Extract and test CRD caching

lindnerby:
**Description**

- Extract global crd cache into testable struct
- Add tests and coverage

**Related issue(s)**
* https://github.com/kyma-project/lifecycle-manager/issues/1125

### DIFF
--- a/internal/crd/cache_test.go
+++ b/internal/crd/cache_test.go
@@ -55,7 +55,6 @@ func TestGet_WhenInCache_TypeIsRight(t *testing.T) {
 	cachedCrd, ok := cache.Get(key)
 
 	assert.True(t, ok)
-	assert.Equal(t, someCrd.Name, cachedCrd.Name)
 	assert.Equal(t, someCrd, cachedCrd)
 }
 
@@ -71,6 +70,5 @@ func TestAdd_WhenCalled(t *testing.T) {
 	assert.True(t, ok)
 	cachedCrd, ok := cachedValue.(apiextensionsv1.CustomResourceDefinition)
 	assert.True(t, ok)
-	assert.Equal(t, someCrd.Name, cachedCrd.Name)
 	assert.Equal(t, someCrd, cachedCrd)
 }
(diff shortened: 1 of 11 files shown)


internal/crd/cache_test.go
```diff
@@ -0,0 +1,76 @@
...
+	cache := crd.NewCache(internalCache)
+	someCrd := apiextensionsv1.CustomResourceDefinition{}
+	someCrd.Name = "some-crd"
+	internalCache.Store(key, someCrd)
+
+	cachedCrd, ok := cache.Get(key)
+
+	assert.True(t, ok)
+	assert.Equal(t, someCrd.Name, cachedCrd.Name)
+	assert.Equal(t, someCrd, cachedCrd)
```
c-pius, line 59: why comparing the names explicitly? Aren't structs "value types" where each value is compared individually, so it's covered by below check anyway?

internal/crd/cache_test.go
```diff
@@ -0,0 +1,76 @@
...
+	someCrd.Name = "some-crd"
+
+	cache.Add(key, someCrd)
+
+	cachedValue, ok := internalCache.Load(key)
+	assert.True(t, ok)
+	cachedCrd, ok := cachedValue.(apiextensionsv1.CustomResourceDefinition)
+	assert.True(t, ok)
+	assert.Equal(t, someCrd.Name, cachedCrd.Name)
+	assert.Equal(t, someCrd, cachedCrd)
```
c-pius, line 75: > `assert.Equal(t, someCrd.Name, cachedCrd.Name)`

same comment as above

> `cachedCrd, ok := cachedValue.(apiextensionsv1.CustomResourceDefinition)`

Here I am additionally wondering if this isn't covered implictly by the `assert.Equal(t, someCrd, cachedCrd)` check below. But doesn't harm to have it I guess.

internal/crd/cache_test.go
lindnerby: Ye the last one is enough! Forgot to delete the first assertion while trying things. Thanks, removed. 
Same for other comment.

c-pius (approved):

PR #1349, opened 2024-02-28
feat: Set `OwnerReference` for `VirtualService`

### LINKED ISSUE #1287: Setting up proper owner references for VirtualServices managed by the KLM

c-pius:
**Description**

- adds an `OwnerReference` pointing to `Watcher` for `VirtualServices`
  - reference is set without `ownerReferences.blockOwnerDeletion` (*field that takes a boolean value and controls whether specific dependents can block garbage collection from deleting their owner object*)

**Related issue(s)**

- resolves #1287 

**Resulting Example `VirtualService`**

```yaml
kind: VirtualService
metadata:
  creationTimestamp: "2024-02-28T12:45:20Z"
  generation: 1
  name: klm-kyma-watcher
  namespace: kcp-system
  ownerReferences:
    - apiVersion: operator.kyma-project.io/v1beta2
      kind: Watcher
      name: klm-kyma-watcher
      uid: dc06cfc4-95a8-4162-8fda-b0d721d1d2e1
  resourceVersion: "164953"
  uid: 69f2d259-3c63-4e33-9dee-7458d421f08a
spec:
  gateways:
    - kcp-system/klm-watcher-gateway
  hosts:
    - host.k3d.internal
  http:
    - match:
        - uri:
            prefix: /v1/lifecycle-manager/event
      name: kcp-system/klm-kyma-watcher
      route:
        - destination:
            host: klm-event-service.kcp-system.svc.cluster.local
            port:
              number: 8082
```


internal/istio/virtualservice_test.go
```diff
@@ -161,58 +193,109 @@ func Test_NewVirtualService_SetsCorrectHosts(t *testing.T) {
...
 
+	watcher.APIVersion = "operator.kyma-project.io/v1beta2"
+	watcher.Kind = "Watcher"
+	watcher.UID = uuid.NewUUID()
+
 	return watcher
 }
 
-func getSimpleNamespace() string {
+func setupNamespace() string {
```
lindnerby, line 289: Not sure about the renaming to prefix with "setup" here. Maybe it's just me, but setup indicates to me changes on a third party component, e.g. cluster setup before tests.
Maybe "get" or "create" are a bit better?

internal/istio/virtualservice_test.go
```diff
@@ -161,58 +193,109 @@ func Test_NewVirtualService_SetsCorrectHosts(t *testing.T) {
...
 	return "bar"
```
lindnerby, line 290: Also this could be a const? It's not creating a v1.Namespace, but using a function indicates it.

internal/istio/virtualservice_test.go
```diff
@@ -161,58 +193,109 @@ func Test_NewVirtualService_SetsCorrectHosts(t *testing.T) {
...
+	scheme := machineryruntime.NewScheme()
+	if err := api.AddToScheme(scheme); err != nil {
+		assert.Fail(t, "failed to setup scheme")
+	}
+
+	return scheme
+}
+
+func setupWatcher() *v1beta2.Watcher {
 	watcher := &v1beta2.Watcher{}
```
lindnerby, line 267: Can you think about moving some of this to `pkg/testutils/builder` and creating a testbuilder for Watcher objects, that always return you a usable watcher for test scenarios.

internal/controller/watcher_controller.go
```diff
@@ -151,7 +152,7 @@ func (r *WatcherReconciler) handleProcessingState(ctx context.Context,
 		return r.updateWatcherState(ctx, watcherCR, shared.StateError, err)
 	}
 
-	virtualSvc, err := istio.NewVirtualService(r.WatcherVSNamespace, watcherCR, gateways)
+	virtualSvc, err := r.VirtualServiceFactory.NewVirtualService(r.WatcherVSNamespace, watcherCR, gateways)
```
lindnerby, line 155: Watcher and VirtualService will not be applied to the same namespace or?

internal/istio/errors.go
```diff
@@ -13,6 +13,8 @@ var (
 	ErrFailedToConvertLabelSelector      = errors.New("failed to convert label selector to selector")
 	ErrFailedToGetGatewayByLabelSelector = errors.New("failed to get gateway by label selector")
 
+	ErrFailedToAddOwnerReference = errors.New("failed to add owner reference")
+
```
lindnerby, line 17: I think if we are doing grouping via empty lines, since empty lines are not self-explanatory there should be comments for each group, in general a least.
Since this is all about istio, I don't think we need that detailed description about errors so let's just remove the implicit grouping here. It's just noise and interrupting the natural flow of reading. WDYT?

internal/controller/watcher_controller.go
c-pius: There is a dedicated FlagVar [`flagVar.IstioGatewayNamespace`](https://github.com/kyma-project/lifecycle-manager/blob/main/cmd/main.go#L392) used. Right now it is configured to default to [`kcp-system`](https://github.com/kyma-project/lifecycle-manager/blob/main/internal/pkg/flags/flags.go#L37). So Watcher and VirtualService reside in the same namespace as of now, but not sure what the origin reason for the flagvar is and if this may need to change in the future.

c-pius: Just also remembered: https://github.com/kyma-project/lifecycle-manager/pull/1349#discussion_r1506042184

Actually, we have an implicit constraint now that `r.WatcherVSNamespace == watcher.Namespace`. So we may even get rid of this argument. Maybe also the flagvar needs to be renamed entirely

internal/istio/errors.go
c-pius: Done with https://github.com/kyma-project/lifecycle-manager/pull/1349/commits/52c2806a62491fa612a6d09e48c999f672b49f0e

internal/istio/virtualservice_test.go
c-pius: Done with https://github.com/kyma-project/lifecycle-manager/pull/1349/commits/016f9609061df109e3ad4c9a4fa60949ab981009

internal/istio/virtualservice_test.go
c-pius: Done with https://github.com/kyma-project/lifecycle-manager/pull/1349/commits/f4b560e3c89211b97965e4f693bb91bab45b4409

internal/istio/virtualservice_test.go
c-pius: Done as part of introducing the watcher builder https://github.com/kyma-project/lifecycle-manager/pull/1349/commits/016f9609061df109e3ad4c9a4fa60949ab981009

pkg/testutils/random/random.go
```diff
@@ -0,0 +1,25 @@
...
+		b[i] = randomNameCharSet[rand.Intn(len(randomNameCharSet))]
+	}
+	return string(b)
+}
+
+// Port creates a random int64 in range [1, 65535].
+func Port() int64 {
+	//nolint:gosec // random number generator sufficient for testing purposes
+	return int64(rand.Intn(randomPortUpperBoundary) + 1)
+}
```
c-pius, line 25: I have moved the existing `RandomName()` func to this helper for better re-use (commit https://github.com/kyma-project/lifecycle-manager/pull/1349/commits/afc3a0a46699e7f06ed62fd2b9b52aef17cd3da4).

It was part of `KymaBuilder` but already used elsewhere. Originally, I wanted to promote it to `testutils` package, but this wasn't simply possible because it would have introduced circular dependencies. Hence, I put it into a separate `testutils/random` package.

internal/controller/watcher_controller.go
lindnerby: That's why I am asking yes. 😄 

Seems the flag is then obsolet even and can be removed. And Watcher namespace always dictates the VS namespace, as you said.

c-pius: Removed explicit namespace arg with https://github.com/kyma-project/lifecycle-manager/pull/1349/commits/443c8b87fcf54c486833401ae256729153110e38

Overall `flagVar.IstioGatewayNamespace` to be discussed in the daily.

lindnerby (approved):

PR #1353, opened 2024-03-01
feat: Add cache metrics

lindnerby:
**Description**

- Adds metrics to for how many items are in caches for descriptors, CRDs, mutex locks and remote clients:
    - `lifecycle_mgr_cache_size_descriptor`
    - `lifecycle_mgr_cache_size_crd`
    - `lifecycle_mgr_cache_size_filemutex`
    - `lifecycle_mgr_cache_size_client`
- Add dashboard

**Related issue(s)**


internal/crd/cache.go
```diff
@@ -33,4 +44,13 @@ func (c *Cache) Get(key string) (apiextensionsv1.CustomResourceDefinition, bool)
...
+	if c.metrics == nil {
+		return
+	}
+	length := 0
+	c.cache.Range(func(key, value interface{}) bool {
+		length++
+		return true
+	})
+	c.metrics.UpdateCrdTotal(length)
 }
```
c-pius, line 56: Was just curious and took a quick glimpse already. It may be possible to replace `cache.Store` with [`cache.Swap`](https://pkg.go.dev/sync#Map.Swap). Then the `length` may be a property of the `Cache` struct that is increased whenever the `loaded` return value returns `false` => we get rid of iterating over the whole cache each time we add a value.

`length` would need to be a thread-safe counter though

lindnerby: That is an excellent idea. Thanks!

internal/crd/cache.go
```diff
@@ -6,16 +6,28 @@ import (
...
-	cache *sync.Map
+	cache   *sync.Map
+	metrics Metrics
+	size    int
 }
 
-func NewCache(internalCache *sync.Map) *Cache {
-	if internalCache == nil {
-		return &Cache{cache: &sync.Map{}}
+func NewCache(cache *sync.Map, metrics Metrics) *Cache {
```
ruanxin, line 19: I don't get the point of why we need to pass sync.Map as a param, the only usage in unit test is also meaningless to me.
And since we touch the refactoring of cache, I suggest we design it as a singleton.

internal/pkg/metrics/cache.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (m *CacheSizeMetrics) UpdateDescriptorTotal(size int) {
+	m.descriptorTotalGauge.Set(float64(size))
+}
+
+func (m *CacheSizeMetrics) UpdateCrdTotal(size int) {
+	m.crdTotalGauge.Set(float64(size))
+}
+
+func (m *CacheSizeMetrics) UpdateFilemutexTotal(size int) {
```
ruanxin, line 58: There is no usage found `UpdateFilemutexTotal`

ruanxin (changes requested):
please see my comments

internal/crd/cache.go
lindnerby: Yes, it's for unit testing, and therefore not meaningless. But I guess we can reach an acceptable coverage by using methods to verify. Will have a look.

Do you have an idea on how to design a singleton without using globals?

lindnerby:
Closing this PR for now, due to missing use case.

PR #1357, opened 2024-03-04
feat: Add Manifest Reconcile Duration Metric To Plutono

### LINKED ISSUE #1069: Introduce metrics to track each Module Manifest reconcile performance

LeelaChacha:
**Description**

- Add Visualisation for `manifest_reconcile_duration_seconds` metric

**Related issue(s)**

Closes #1069

### DIFF
--- a/config/grafana/overview.json
+++ b/config/grafana/overview.json
@@ -21,7 +21,7 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
-  "id": 21,
+  "id": 15,
   "links": [],
   "panels": [
     {
@@ -78,7 +78,7 @@
         "alertThreshold": true
       },
       "percentage": false,
-      "pluginVersion": "7.5.28",
+      "pluginVersion": "7.5.29",
       "pointradius": 5,
       "points": false,
       "renderer": "flot",
@@ -160,13 +160,11 @@
       },
       "yaxes": [
         {
-          "$$hashKey": "object:133",
           "format": "ops",
           "logBase": 1,
           "show": true
         },
         {
-          "$$hashKey": "object:134",
           "format": "ops",
           "logBase": 1,
           "show": true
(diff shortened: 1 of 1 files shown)


config/grafana/overview.json
```diff
@@ -715,9 +709,9 @@
             "uid": "P1809F7CD0C75ACF3"
           },
           "exemplar": true,
-          "expr": "rate(controller_runtime_reconcile_time_seconds_sum{controller=\"manifest\"}[$__rate_interval])\n/\nrate(controller_runtime_reconcile_time_seconds_count{controller=\"manifest\"}[$__rate_interval])",
+          "expr": "rate(reconcile_duration_seconds_sum[$__rate_interval])\n/\nrate(reconcile_duration_seconds_count[$__rate_interval])",
```
nesmabadr, line 712: Shouldn't we keep both `reconcile_duration_seconds_sum` and `controller_runtime_reconcile_time_seconds_sum`?

LeelaChacha: hmmm,  I would say they both convey the same information so it is unnecessary. Plus since we have more control over our custom metrics, it is open for extension in the future as opposed to the built-in metric

nesmabadr: hmm makes sense. But please change the value of the module name label to actually reflect the module name not the manifest name.

LeelaChacha:
After deploying the histogram metric, we realized that the it is not a feasible solution in production as it generates huge amount of data. Therefore, this metric will be not be released and a gauge metric will be used instead. Look at update issue #1069

Issue #1329, opened 2024-02-20
Investigation: GH-Action Pipeline to detect diff in Kustomize Setup


jeremyharisch:
### Description

Since we switched to Helm Charts on our landscape deployments, but we still stick to Kustomize in our OpenSource project, we need to create a Pipeline which "reminds" us to propagate Kustomize changes to our internal Helm charts.

### Reasons

- Reminder to propagate Setup changes to Helm charts
- Pipeline should only be green, if one Codeowner confirmed that Kustomize changes have been propagated to internal Helm Chart setup
- Hinders us to merge unwanted setup changes

### Acceptance Criteria

- [x] Investigate:
    - Is it possible to retrigger a PR-based GH-Action-Pipeline with an additional input-parameter
    - Should be possible the same way as it works with input-parameters for pipelines with manual trigger
    - If possible, continue with next ACs
    - If it is not working with additional input parameter then proper label should be set on the PR
- [x] Implement new GH-Action Pipeline
    - Pipeline should get triggered on PRs opened to `main` branch
    - Should render kustomize for main branch and for PR-branch; compare both; if no diff has been found; pipeline is green
    - If diff was found pipeline should be red
    - If pipeline was red, and then retriggered with input parameter set to `Changes Propagated = True`, then Pipeline is green, eventho diff have been detected

c-pius:
> Is it possible to retrigger a PR-based GH-Action-Pipeline with an additional input-parameter

I am afraid this is not possible. At least from the answers [here](https://stackoverflow.com/questions/63288356/how-to-link-a-github-actions-manual-run-to-a-pr) (question is different but answers seem to apply to our case as well) it doesn't seem possible. Also, we have a similar setup with our TestSuite E2E. TestSuite E2E has PR and Workflow triggers. Looking at the PR checks, we don't find the option to re-trigger the check based on the manual workflow dispatch. We can only re-trigger via standard "re-run" jobs. When looking at the Actions tab, we find the manual trigger for TestSuite E2E, but the question is how this trigger would link the run to the PR (as also mentioned on the stack question above and [here](https://stackoverflow.com/questions/76464269/how-to-rerun-a-github-action-workflow-from-the-command-line-and-have-the-status)). It may be possible to add the latest commit hash of the PR as argument, or to build some scripting that looks this up from a passed PR number. But both seem inpractical.

Therefore looking into a label-based alternative now.

c-pius:
[...]

PR #1358, opened 2024-03-04
chore: Verify diff in manifests

c-pius:
**Description**

- creates a new GH pipeline that
  - creates the manifests based on the current PR
  - creates the manifests based on main
  - calculates the diff between those manifests
- if there is a diff, the pipeline goes red and the author must confirm that related Helm charts updates have been taken care of
  - confirmation is done by adding the `confirm/helm-update` label which will skip the pipeline

**Related issue(s)**

- resolves https://github.com/kyma-project/lifecycle-manager/issues/1329

**Automatically removing the `confirm/helm-update` label on pushes**

This is unfortunately not easily doable, and I would propose that we skip this part. If it turns out to lead to issues, we can re-open the task and find a proper solution.

Background:

In order to unlabel a PR, we need to have write permissions to the PR. For this, we would need to change the event to `pull_request_target`. [`pull_request_target`](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) has some security implications and it is firmly recommended to NOT checkout the head of the branch and operate on it. This in turn prevents us from generating the manifests according to the state of the PR head.

To navigate around this, I have tried to separate the unlabeling from the manifest creation and diffing. The idea was to have a dedicated workflow on `pull_request_target` that removes the `confirm/helm-update` label. The action of removing the label should have triggered the manifest creation and diffing workflow in turn. However, this does not work as ["An action in a workflow run can’t trigger a new workflow run"](https://github.com/orgs/community/discussions/25565).

Another option to synchronize both workflows would be [`workflow_run`](https://docs.github.com/en/webhooks/webhook-events-and-payloads#workflow_run). However, with this we are missing the proper PR context to checkout and create the manifests for. Therefore this is also a non viable solution.

### DIFF
--- a/.github/workflows/check-manifest-diffs.yaml
+++ b/.github/workflows/check-manifest-diffs.yaml
@@ -1,5 +1,9 @@
 name: "Check for diff in manifests"
 
+env:
+  PR_CACHE_KEY: pr-manifests-${{ github.run_id }}-${{ github.run_attempt }}
+  MAIN_CACHE_KEY: main-manifests-${{ github.run_id }}-${{ github.run_attempt }}
+
 on:
   pull_request:
     branches: [ "main" ]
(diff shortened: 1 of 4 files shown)


lindnerby (approved):
LGTM!

Issue #1310, opened 2024-02-12
Cover "Unauthorized" detection in e2e tests


c-pius:
### Description

As suggested by Xin [here](https://github.com/kyma-project/lifecycle-manager/pull/1298/files#r1486122982), detection of "Unauthorized" errors is quite important and should be covered somewhere in e2e tests. The aim is to find a proper place in e2e tests where we can cover this situation and design the appropriate test.

### Reasons

We need to make sure that those are covered correctly in the future, even though code may change.

### Acceptance Criteria

- [ ] When KymaCR is ready
  - [ ] And template-operator is ready
  - [ ] And connection to SKR cluster returns "unauthorized" / "connection refused"
  - [ ] Then:
    - [ ] `lifecycle_mgr_requeue_reason_total{requeue_reason="kyma_unauthorized",requeue_type="unexpected"}` and `lifecycle_mgr_requeue_reason_total{requeue_reason="manifest_unauthorized",requeue_type="unexpected"}` shall be written
    - [ ] NO events for KymaCR and ManifestCR shall be written

PR #1363, opened 2024-03-06
fix: Cache Invalidation for DNS Lookup Errors

### LINKED ISSUE #1290: Excessive logging in case of misconfigured Kymas/SKR-secrets

LeelaChacha:
**Description**

Changes proposed to the MANIFEST Reconciler this pull request:

- Invalidate remote client cache when dns lookup fails
- test client when new one is created to prevent excessive logging downstream

**Related issue(s)**

Closes #1290


pkg/util/error.go
```diff
@@ -44,10 +45,28 @@ func IsNotFound(err error) bool {
 	return false
 }
 
-func IsConnectionRefusedOrUnauthorized(err error) bool {
+func IsConnectionOrHostError(err error) bool {
```
ruanxin, line 48: please rename this function to express it's for determining connection errors, something like `IsConnectionRelatedError`

internal/declarative/v2/reconciler.go
```diff
@@ -609,9 +609,24 @@ func (r *Reconciler) configClient(ctx context.Context, obj Object) (Client, erro
 	if err != nil {
 		return nil, err
 	}
+
+	err = testClient(ctx, clnt)
```
ruanxin, line 613: it's not a good approach test client connection inside the configuration, the purpose of this `configClient` is just for config client, but not take care if the connection is fine or not. To handle the connection error is the response in the follow-up process, but not here.

pkg/util/error.go
ruanxin: Please also think how to test this scenario, you can think about have it as e2e test, to simulate the situation have unauthorized error, no such host error. check the existing e2e how it simulate connection refused error by delete skr cluster [here](https://github.com/kyma-project/lifecycle-manager/blob/b8e3fec0e7da1d1f493b99203a87379994a89f8c/tests/e2e/deletion_test.go#L79)

ruanxin (changes requested):
please check my comments

pkg/util/error.go
ruanxin: unit test only this function is pointless, basically what you do here is mock a fake connection error, then test if it return true or false. But first we don't know if this is the same as the actual connection error, more critically, what the consequence this error brings is not tested.

So what's expected to be tested here: when this connection error happens, we need to verify this [cache evict](https://github.com/kyma-project/lifecycle-manager/blob/b2fe4ec711869034bba7c77ffdf7f3f2791d2ec1/internal/declarative/v2/reconciler.go#L152-L155) and reset new valid remote client cache works as expected. I‘m fine if you can design an integration test for this scenario, but an e2e test is much more straightforward in this case.

please also remove this [block](https://github.com/kyma-project/lifecycle-manager/blob/b2fe4ec711869034bba7c77ffdf7f3f2791d2ec1/internal/declarative/v2/reconciler.go#L457-L461), this event does not bring too much value for us.

LeelaChacha: Hmm, maybe this could be a separate issue because this also handles the testing of connection refused errors which is an existing functionality before this CR. plus we can get the feature merged faster that way. either way, it's your call

internal/declarative/v2/reconciler.go
LeelaChacha: ok the function call has been moved. I decided to keep it in the getTargetClient function as the error value from that funciton creates a ClientIntialization error event. which is fitting because if testing the client fails, it should count as initialization failure

pkg/util/error.go
ruanxin: Let's follow the engineering standard on how to properly do bug fixing/new features. Ideally, all features should be tested before being accepted, no exceptions. If you saw some function without a test covered, that's not something you should follow with, instead, as you suggest, create a followup issue or tested immediately, but the followup issue should only accepted when this issue is urgent with time pressure and approved by the manual fix as a compromise but ideally should avoid it to postpone the test, for example, we have this [issue](https://github.com/kyma-project/lifecycle-manager/issues/1310) to address that unauthorized connection error.

As I said, the connection refused errors have been already covered in the e2e test [here](https://github.com/kyma-project/lifecycle-manager/blob/b8e3fec0e7da1d1f493b99203a87379994a89f8c/tests/e2e/deletion_test.go#L79).

For enterprise software, seeking speed is not our goal, especially when the component has been running in the production environment. Instead, the code quality and making sure bug free is the first priority.

internal/declarative/v2/reconciler.go
```diff
@@ -592,6 +600,10 @@ func (r *Reconciler) getTargetClient(ctx context.Context, obj Object) (Client, e
 			return nil, fmt.Errorf("failed to patch namespace: %w", err)
 		}
 	}
+	err = testClient(ctx, clnt)
```
ruanxin, line 603: If you testClient here, it means for all client connection, you add additionally a request, which is not necessary, because anyway, if the connection have problems, the process afterwards which doing real meaningful requests, like deploy resources, get module data will be failed and return back error, you don't have to test this connection upfront, besides, even if it pass the connection test in this step, can you make sure the followup process not failed with connect problem? With this additional tests, however it brings some side effects to affect the performance, e.g: slow down the reconciliation time, add networking costs.

pkg/util/error.go
c-pius: Wrt. testing: maybe it is possible to re-use what is done [here](https://github.com/kyma-project/lifecycle-manager/blob/b8e3fec0e7da1d1f493b99203a87379994a89f8c/tests/e2e/utils_test.go#L100-L113) (called in `InitEmptyKymaBeforeAll`), but instead of replacing localhost with the k3d internal host, we replace it with some not reachable dummy value.

internal/declarative/v2/reconciler.go
LeelaChacha: That is on purpose to test the connection upfront. because in case of orphaned kyma, the downstream process generates a huge number og logs before the reconciliation is finished. that is the problem that this PR is trying to solve

ruanxin (approved):

ruanxin (approved):

PR #1380, opened 2024-03-08
feat: Add manifest reconciliation metric to Plutono Dashboard

nesmabadr:
**Description**

- Add manifest reconciliation metric to Plutono Dashboard

**Related issue(s)**


c-pius:
I have tried to visualize this in Plutono by copying the JSON but don't get any data. I assume this is related to the 1m threshold that was set for writing the metric? Does this make sense and if so, do we have any other way of checking if it works?

c-pius (approved):

PR #1392, opened 2024-03-12
chore: Refactor and increase purge controller monitoring

c-pius:
**Description**

- aligns log/error messages
- refactors the `Reconcile` function
  - it is the "brains" for orchestrating the reconcile and delegates handling of certain situations to "handlers" now
- monitoring tasks (logs/metrics/events) have been taken out of the "helper" functions and promoted to the handlers
  - it will be easier to see what is written for a given situation (e.g., sometimes we write logs in the helper functions but metrics/events in the outer function)
- adds new info level logs for when finalizers have been removed (either from Kyma or other resources)

**Related issue(s)**

### DIFF
--- a/internal/controller/purge_controller.go
+++ b/internal/controller/purge_controller.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
@@ -85,9 +86,13 @@ func (r *PurgeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 	remoteClient, err := r.ResolveRemoteClient(ctx, client.ObjectKeyFromObject(kyma))
 	if err != nil {
 		if util.IsNotFound(err) {
-			if err := r.dropPurgeFinalizer(ctx, kyma); err != nil {
+			dropped, err := r.dropPurgeFinalizer(ctx, kyma)
+			if err != nil {
 				r.raiseRemovingPurgeFinalizerFailedError(ctx, kyma, err)
 				return ctrl.Result{}, err
+			}
+			if dropped {
+				logger.Info(fmt.Sprintf("Removed purge finalizer for Kyma %s", req.NamespacedName))
 			}
 			return ctrl.Result{Requeue: true}, nil
 		}
@@ -96,19 +101,28 @@ func (r *PurgeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 	}
 
 	r.Metrics.UpdatePurgeCount()
-	if err := r.performCleanup(ctx, remoteClient); err != nil {
+
+	handledResources, err := r.performCleanup(ctx, remoteClient)
+	if err != nil {
 		logger.Error(err, fmt.Sprintf("failed purge cleanup for Kyma %s", req.NamespacedName))
 		r.Metrics.UpdatePurgeError(ctx, kyma, metrics.ErrCleanup)
 		return ctrl.Result{}, err
 	}
 
-	if err := r.dropPurgeFinalizer(ctx, kyma); err != nil {
+	if len(handledResources) > 0 {
+		logger.Info("Removed purge finalizer for resources " + strings.Join(handledResources, ", "))
+	}
+
+	dropped, err := r.dropPurgeFinalizer(ctx, kyma)
+	if err != nil {
 		r.raiseRemovingPurgeFinalizerFailedError(ctx, kyma, err)
 		return ctrl.Result{}, err
 	}
+	if dropped {
+		logger.Info(fmt.Sprintf("Removed purge finalizer for Kyma %s", req.NamespacedName))
+	}
 
 	r.Metrics.UpdatePurgeTime(time.Since(start))
-
 	return ctrl.Result{}, nil
 }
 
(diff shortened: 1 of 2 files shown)


internal/controller/purge_controller.go
```diff
@@ -61,60 +61,61 @@ func (r *PurgeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
 			return ctrl.Result{}, err
 		}
-		return ctrl.Result{Requeue: false}, nil
+		return ctrl.Result{}, nil
 	}
 
-	requeueAfter := r.calculateRequeueAfterTime(ctx, kyma)
-	if requeueAfter != 0 {
+	if requeueAfter := r.calculateRequeueAfterTime(kyma); requeueAfter != 0 {
+		logger.Info(fmt.Sprintf("Purge reconciliation for Kyma %s will be requeued after %s", req.NamespacedName, requeueAfter))
```
c-pius, line 80: Reviewer note: This message is promoted from Debug to Info. I am not entirely sure if this is necessary, but this should not re-occur frequently anyway and therefore not spam the infrastructure. And having it we have a hint in the logs that purging was started on info level.

internal/controller/purge_controller.go
```diff
@@ -231,12 +220,22 @@ func dropFinalizers(ctx context.Context, remoteClient client.Client,
...
 
-func (r *PurgeReconciler) setFinalizerWarningEvent(kyma *v1beta2.Kyma, err error) {
-	r.Event(kyma, "Warning", "SettingPurgeFinalizerError", err.Error())
+func (r *PurgeReconciler) raiseSettingPurgeFinalizerFailedEvent(kyma *v1beta2.Kyma, err error) {
+	r.Event(kyma, "Warning", "SettingPurgeFinalizerFailed", err.Error())
+}
+
+func (r *PurgeReconciler) raiseRemovingPurgeFinalizerFailedEvent(kyma *v1beta2.Kyma, err error) {
+	r.Event(kyma, "Warning", "RemovingPurgeFinalizerFailed", err.Error())
+}
```
c-pius, line 235: Reviewer Note: I separated this into two events as before we didn't distinguish whether the error occurred when setting or removing the finalizer. According to API conventions events should be specific: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#events

internal/controller/purge_controller.go
ruanxin: I think this log not helpful too much, basically it means when the kyma is under delete, but the time is still not reached `PurgeFinalizerTimeout`, it gets re-queued. With this log message, what we can get is only this kyma CR is still pending to do purge, however we still don't know if this kyma has been deleted during this timeout or it's indeed entered into purge process.

What I expected to see is some logs from the actual purge process, for example, if the log recorded the time that it's purge controller removed the purge finalizer before kyma controller did all finalizers cleanup, then we know this purge process works.

internal/controller/purge_controller.go
```diff
@@ -61,60 +61,61 @@ func (r *PurgeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
 
 	if err := r.dropPurgeFinalizer(ctx, kyma); err != nil {
-		r.Metrics.UpdatePurgeError(ctx, kyma, metrics.ErrPurgeFinalizerRemoval)
+		r.raiseRemovingPurgeFinalizerFailedError(ctx, kyma, err)
 		return ctrl.Result{}, err
 	}
-	duration := time.Since(start)
-	r.Metrics.UpdatePurgeTime(duration)
+
+	r.Metrics.UpdatePurgeTime(time.Since(start))
```
ruanxin, line 110: This metrics actually only record happy path, but for those kyma end up into error cleanup, we don't have any metrics reflect it, if we can improve this metrics to update purge time also include error case, it also help us identify those kyma not perform cleanup within expected time.

ruanxin (changes requested):
please see my comments

internal/controller/purge_controller.go
c-pius: Apologies, not sure if I can fully follow you. For the unhappy path, we do have:

- log + metric + event, if dropping the purge finalizer from Kyma fails:
  - [here](https://github.com/kyma-project/lifecycle-manager/blob/371c5e32cd551a507d41da74d9262a3b850081a1/internal/controller/purge_controller.go#L89), and
  - [here](https://github.com/kyma-project/lifecycle-manager/blob/371c5e32cd551a507d41da74d9262a3b850081a1/internal/controller/purge_controller.go#L106)
- metric, if the cleanup failed
  - [here](https://github.com/kyma-project/lifecycle-manager/blob/371c5e32cd551a507d41da74d9262a3b850081a1/internal/controller/purge_controller.go#L101)

The metric is Kyma specific, so we can derive the time of trying the purge from the time the counter was increased, e.g., here:
image

internal/controller/purge_controller.go
c-pius: added with https://github.com/kyma-project/lifecycle-manager/pull/1392/commits/8e705b62df18921f0646675bd6b3ad12b57b6166

internal/controller/purge_controller.go
c-pius: Added the reset of the value here: https://github.com/kyma-project/lifecycle-manager/pull/1392/commits/d668ab42f9cd87d79808e9af1ea649532acf99c8

internal/pkg/metrics/purge.go
```diff
@@ -57,7 +57,15 @@ func (p *PurgeMetrics) UpdatePurgeTime(duration time.Duration) {
 	p.purgeTimeGauge.Set(duration.Seconds())
 }
 
-func (p *PurgeMetrics) UpdatePurgeError(ctx context.Context, kyma *v1beta2.Kyma, purgeError PurgeError) {
+func (p *PurgeMetrics) SetPurgeError(ctx context.Context, kyma *v1beta2.Kyma, purgeError PurgeError) {
+	p.updatePurgeError(ctx, kyma, purgeError, 1)
+}
+
+func (p *PurgeMetrics) UnsetPurgeError(ctx context.Context, kyma *v1beta2.Kyma, purgeError PurgeError) {
```
ruanxin, line 64: We should use another approach to unset the metrics, we can remove this metrics entry completely, a metric with 0 doesn't provide too much value, but it still increases the size of the metric, remember we have metrics size limit for Prometheus server to fetch with, so we need to prevent introduce useless metrics.

you can reference [here](https://github.com/kyma-project/lifecycle-manager/blob/f6a6d354f7993da4b730ffac0d14ef6f1f2c6a6e/internal/pkg/metrics/kyma.go#L95) how we clean up other metrics.

c-pius: Done with https://github.com/kyma-project/lifecycle-manager/pull/1392/commits/cf58689cafa1242f0643748ea67fc5caeeff4b02

PR #1403, opened 2024-03-18
fix: Add wait for cache sync for the storage version dropper

### LINKED ISSUE #1383: DropStoredVersion should support storage versions based on CRD kind
### Description

As a preparation for the Sync Resource feature, the drop stored version should handle different versions based on the CRD kind.

### Reasons

After the introduction of the Sync Resource, there will be different CRD versions (v1alpha1 for the SyncResource and v1beta2 for the Manifest, Kyma, Watcher and ModuleTemplate), so we should make the version to be dropped configurable per CRD kind

### Acceptance Criteria

- [ ] Support different storage versions to be dropped for different CRD kinds.

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

nesmabadr:
**Description**

- the client.list in the storage version dropper sometimes throws this error "the cache is not started, can not read objects", that's why I added the wait for sync cache so that we wait for the cache to be initialized before we run the dropper.

**Related issue(s)**
Resolves #1383

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -178,8 +178,12 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
 	addHealthChecks(mgr)
 	if flagVar.DropCrdStoredVersionMap != "" {
 		go func(crdVersions string) {
-			clnt := mgr.GetClient()
-			crd.DropStoredVersion(clnt, crdVersions)
+			if mgr.GetCache().WaitForCacheSync(context.Background()) {
+				clnt := mgr.GetClient()
+				crd.DropStoredVersion(clnt, crdVersions)
+			} else {
+				setupLog.V(log.InfoLevel).Error(err, "unable to run storage version dropper")
+			}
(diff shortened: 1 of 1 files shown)


jeremyharisch (approved):

c-pius:
@nesmabadr Checkmarx reports potential race conditions on `mgr` and `err` objects. Looking at the code, I think the usage of the `err` object when logging the error seems like a mistake. It has been assigned earlier unrelated to dropping the storage versions.

Regarding `mgr`, would it be possible to extract the `mgr.GetCache().WaitForCacheSync(context.Background())` to before invoking to goroutine and then pass `mgr.GetClient()` as argument? This may resolve the vulnerability as it also doesn't flag the same usage below on `go runKymaMetricsCleanup`. Btw., not sure why it hasn't occurred yet but the `runKymaMetricsCleanup` goroutine may be subject to the "the cache is not started, can not read objects" error as well. So maybe a good idea to anway wait for the cache to sync before invoking any of these.

PR #1408, opened 2024-03-18
fix: Avoid unnecessary ssa for the manifest

### LINKED ISSUE #1384: Avoid update unnecessary last Operation `lastUpdateTime`
### Description

At the moment, the amount of SSA of Manifest CR has big impact to KCP API Server. However, most of the SSA diffs are just `lastUpdateTime` field, if we can have a way to detect those unnecessary update before sending the SSA, it can reduce the request dramatically. The solution here is implement a client side diff detection based on:
 1. `status.lastOperation.operation` message field,
 2. `status.state`
 If above fields to be updated is not changed, we can assume there is no other changes have to be updated to Manifest CR.

### Reasons

Reduce the SSA requests to API Server.

### Acceptance Criteria

- [ ] Stop sending Patch request if mentioned fields keep identical in the end of manifest reconciliation
- [ ] Cover functionality in tests as described below

### Feature Testing

- [ ] Cover functionality in unit and integration test
- [ ] Enhance existing E2E test to include this scenario

### Testing approach

_No response_

### Attachments

_No response_

nesmabadr:
**Description**

- Only do ssa if the status.state or status.lastOperation.operation have changed

**Related issue(s)**
Resolves #1384


internal/declarative/v2/reconciler.go
```diff
@@ -622,19 +624,26 @@ func (r *Reconciler) configClient(ctx context.Context, obj Object) (Client, erro
 	return clnt, nil
 }
 
-func (r *Reconciler) ssaStatus(ctx context.Context, obj client.Object,
-	requeueReason metrics.ManifestRequeueReason,
+func (r *Reconciler) ssaStatusIfDiffExist(ctx context.Context, obj Object,
+	requeueReason metrics.ManifestRequeueReason, currentObjStatus shared.Status,
```
Tomasz-Smelcerz-SAP, line 628: Naming: I'd say that the "current" status is defined in the `obj.GetStatus()` and the `currentObjStatus` represents "previous", already existing status.
Thus I'd vote for renaming the parameter to `previousObjStatus`

internal/declarative/v2/reconciler_test.go
```diff
@@ -156,3 +157,42 @@ func Test_hasDiff(t *testing.T) {
 		})
 	}
 }
+
+func TestReconciler_hasStatusDiff_WhenSameStatus(t *testing.T) {
```
Tomasz-Smelcerz-SAP, line 161: I would add a test case for an empty (uninitialized) LastOperation. And maybe consider then joining the three test cases into a single table-driven test.

tests/e2e/manifest_reconciliation_test.go
```diff
@@ -106,5 +107,41 @@ var _ = Describe("Manifest Skip Reconciliation Label", Ordered, func() {
 				WithArguments(defaultRemoteKymaName, RemoteNamespace, runtimeClient, shared.StateReady).
 				Should(Succeed())
 		})
+
+		It("When a blocking finalizer is added to the Manifest CR", func() {
+			Eventually(AddFinalizerToManifest).
```
Tomasz-Smelcerz-SAP, line 112: Is there a reason to use `Eventually` here, instead of just failing on error? Are errors expected when adding Finalizers?

tests/e2e/manifest_reconciliation_test.go
```diff
@@ -106,5 +107,41 @@ var _ = Describe("Manifest Skip Reconciliation Label", Ordered, func() {
...
+				WithContext(ctx).
+				WithArguments(controlPlaneClient, kyma.GetName(), kyma.GetNamespace(), module.Name,
+					"blocking-finalizer").
+				Should(Succeed())
+
+			By("And Manifest CR has deletion timestamp set")
+			Eventually(DeleteManifest).
```
Tomasz-Smelcerz-SAP, line 119: Is there any reason to use `Eventually` here? Are errors expected during the delete?

tests/e2e/manifest_reconciliation_test.go
nesmabadr: I used Eventually here and in the below statement because sometimes updating the objects can cause a "conflict" if it was being reconciled at the same time or any other update is happening at the same time

internal/declarative/v2/reconciler.go
nesmabadr: yeah makes sense I think it would make it less confusing

PR #1413, opened 2024-03-20
feat: Add SyncResource CRD

ruanxin:


api/v1alpha1/syncresource_types.go
```diff
@@ -0,0 +1,114 @@
...
+package v1alpha1
+
+import (
+	apimetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/kyma-project/lifecycle-manager/api/shared"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
nesmabadr, line 26: Please remove these lines

config/rbac/syncresource_editor_role.yaml
nesmabadr, line 1: Should these rbac files be referenced anywhere?

ruanxin: They are automatically generated by kubebuilder, basically kubebuilder offers two ways to config `ClusterRole`, reference resource role (by `xxx_editor_role`, `xxx_viewer_role`) specifically or use `role.yaml` https://github.com/kyma-project/lifecycle-manager/blob/e47431db6cedd2b74d4fd66cff40f6468e2766b9/config/rbac/kustomization.yaml#L11 which include all resources, let's have another issue align about which way prefer, then do the cleanup.

ruanxin: I changed my mind, we need to only keep `role.yaml`, there are problems in those editor and viewer role, check the followup [issue](https://github.com/kyma-project/lifecycle-manager/issues/1424) for detail.

PR #1423, opened 2024-03-21
fix: Remove some privileges from KLM 

### LINKED ISSUE #1415: Revisit ServiceAccount privileges for the KLM
### Description

Right now KLM has an elevated set of privileges to run in the K8S cluster. It should be reduced to the minimum necessary for it to operate correctly.

### Reasons

We want to keep the KCP cluster secure in case the KLM pod is compromised.

### Acceptance Criteria

- [x] KLM is running with a minimal set of privileges
  - [x] remove unnecessary verbs/resources for the KLM ClusterRole
- [x] Align with the SRE team on the requirements and expectations
- [ ] Introduce a e2e verification for plain checking minimal required RBAC privileges

nesmabadr:
**Description**

- Use namespaced role bindings for the control plane profile
- Use cluster role bindings for the default and other profiles

**Related issue(s)**
Resolves #1415


config/control-plane/kustomization.yaml
```diff
@@ -68,6 +68,9 @@ patches:
       - op: add
         path: /spec/template/spec/containers/0/args/-
         value: --is-kyma-managed
+      - op: add
```
nesmabadr, line 71: Note to reviewer: This will not be introduced here. This will be added to the landscapes in the management charts repo.

So when the review for this PR is finalized, the steps to merge it will be as follows:
- introduce a PR to lifecycle manager that only adds the flag without references or any other code
- add the flag value in the management charts repo
- Add the rest of the code here in another PR

internal/util.go
```diff
@@ -71,12 +71,17 @@ func GetResourceLabel(resource client.Object, labelName string) (string, error)
...
+	for _, v := range namespaces {
+		defaultNamespaces[v] = cache.Config{}
+	}
 	return cache.Options{
 		ByObject: map[client.Object]cache.ByObject{
 			&apicorev1.Secret{}: {
 				Label: k8slabels.Everything(),
 			},
 		},
+		DefaultNamespaces: defaultNamespaces,
```
ruanxin, line 85: Do we really need to define the namespace? What happens if we keep like before? If it's just for performance gain, I  don't think it is worth the effort to introduce such options to offer the namespace to be cached. It increase the complex to config LM.

config/control-plane/kustomization.yaml
```diff
@@ -110,6 +113,83 @@ transformers:
     kind: ConfigMap
     name: klm-dashboard-status
     version: v1
+- |-
```
ruanxin, line 116: If we consider `control-plane` profile purely for KCP landscape, can we get rid of this `PatchTransformer`?

internal/util.go
nesmabadr: No it has to be there because otherwise it requires the role bindings to be cluster scope not namespace

config/manager/kustomization.yaml
```diff
@@ -15,5 +15,5 @@ configMapGenerator:
 
 images:
 - name: controller
-  newName: europe-docker.pkg.dev/kyma-project/prod/lifecycle-manager
-  newTag: latest
+  newName: europe-docker.pkg.dev/kyma-project/dev/lifecycle-manager
```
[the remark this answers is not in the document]
nesmabadr: pushed by mistake

config/control-plane/kustomization.yaml
nesmabadr: which PatchTransformer are you referring to here?

nesmabadr: But the thing is we have the namespacetransformer, that's why we still need to patch the namespaces for these resources

internal/util.go
ruanxin, line 85: What happens if the `namespaces` param is empty? Do we need this cache option then?

ruanxin (dismissed):

internal/util.go
ruanxin: ok, should be fine if passing an empty map to `DefaultNamespaces`

config/rbac/common/gateway_clusterrole.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
ruanxin, line 2: do we need `gateways` to be cluster wised? is it only for `istio-system` namespace?

nesmabadr: We list the gateways in all namespaces https://github.com/kyma-project/lifecycle-manager/blob/b2e06b62b67b29841bbbc8b11ec617125e4b7e12/internal/istio/client.go#L103

ruanxin: OK, this we need to improve, not make sense to list all gateways, but we can do it in a followup issue.

tests/integration/controller/withwatcher/watcher_controller_test.go
```diff
@@ -41,7 +41,8 @@ func gatewayUpdated(customIstioClient *istio.Client) error {
 	if err != nil {
 		return err
 	}
-	gateways, err := customIstioClient.ListGatewaysByLabelSelector(suiteCtx, &watcher.Spec.Gateway.LabelSelector)
+	gateways, err := customIstioClient.ListGatewaysByLabelSelector(suiteCtx, &watcher.Spec.Gateway.LabelSelector,
+		kcpSystemNs)
```
ruanxin, line 45: why use `kcpSystemNs`, but not istio-system?

ruanxin: the test better reflect to actual use case.

ruanxin (approved):

PR #1426, opened 2024-03-26
refactor: Remote pkg & sync strategy label

lindnerby:
**Description**

- Remove client caching from context and replace skr context retrieval with explicit interface
- Remove need for sync strategy annotation
- Remove global var localClientFunc  and function implementations to be replaced with context related interface implementations
- Move remote pkg to internal

**Related issue(s)**
* https://github.com/kyma-project/lifecycle-manager/pull/1426


internal/controller/kyma/controller.go
```diff
@@ -115,62 +114,45 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
-
-	if !kyma.DeletionTimestamp.IsZero() && errors.Is(err, remote.ErrAccessSecretNotFound) {
-		logger.Info("access secret not found for kyma, assuming already deleted cluster")
-		r.Metrics.CleanupMetrics(kyma.Name)
-		r.removeAllFinalizers(kyma)
-
-		if err := r.updateKyma(ctx, kyma); err != nil {
-			r.Metrics.RecordRequeueReason(metrics.KymaUnderDeletionAndAccessSecretNotFound, queue.UnexpectedRequeue)
-			return ctrl.Result{}, err
+	if r.SyncKymaEnabled(kyma) {
```
ruanxin, line 117: can we create an issue to introduce another controller for the separation of single/dual kyma reconciliation?

lindnerby: Sure, created it [here](https://github.com/kyma-project/lifecycle-manager/issues/1525). Still needs to be fully refined.

internal/controller/kyma/controller.go
```diff
@@ -115,62 +114,45 @@ func (r *KymaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
 	}
 
 	return r.reconcile(ctx, kyma)
 }
 
-func (r *KymaReconciler) deleteRemoteClientCache(ctx context.Context, kyma *v1beta2.Kyma) {
+func (r *Reconciler) deleteRemoteClientCache(ctx context.Context, kyma *v1beta2.Kyma) {
 	logger := logf.FromContext(ctx)
 	logger.Info("connection refused, assuming connection is invalid and resetting cache-entry for kyma")
 	r.RemoteClientCache.Delete(client.ObjectKeyFromObject(kyma))
```
ruanxin, line 152: you can also use `kyma.GetNamespacedName()`, right?

internal/remote/client_factory.go
```diff
@@ -0,0 +1,107 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/clientcmd"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kyma-project/lifecycle-manager/api/shared"
+)
+
+type SkrContextFactory interface {
+	Get(kyma types.NamespacedName) (*SkrContext, error)
+	Init(ctx context.Context, kyma types.NamespacedName) error
```
ruanxin, line 22: I think we can also introduce `DeleteCache(kyma types.NamespacedName)` to this interface, and then we can get rid of that `RemoteClientCache` in `KymaReconciler struct`

internal/remote/client_factory.go
```diff
@@ -0,0 +1,107 @@
...
+	restConfig.Burst = k.kcpClient.Config().Burst
+
+	clnt, err := client.New(restConfig, client.Options{Scheme: k.kcpClient.Scheme()})
+	if err != nil {
+		return fmt.Errorf("failed to create lookup client: %w", err)
+	}
+
+	skrClient := NewClientWithConfig(clnt, restConfig)
+
+	namespace := &apicorev1.Namespace{
```
ruanxin, line 75: this whole "ensures remote namespace exists" logic can be extracted to the main reconciler logic, then you don't need to implement such logic in `DualClusterFactory`, `SingleClusterFactory` for the integration test. It's not a logic of this init function should care about.

internal/controller/kyma/controller.go
ruanxin: check this comment for further improvement https://github.com/kyma-project/lifecycle-manager/pull/1426/files#r1594129327

internal/remote/client_factory.go
ruanxin: something like this in kyma reconciler
```
if r.SyncKymaEnabled(kyma) {
    err := r.SkrContextFactory.Init(ctx, kyma.GetNamespacedName())
    skrContext, err := r.SkrContextFactory.Get(kyma.GetNamespacedName())
    if err != nil {
	    return ctrl.Result{}, err
    }
    if err := r.ensureRemoteNamespaceExits(ctx, skrContext); err != nil {
	    return ctrl.Result{}, err
    }
...
```

internal/controller/kyma/controller.go
```diff
@@ -115,41 +114,71 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-			r.removeAllFinalizers(kyma)
-
-			if err := r.updateKyma(ctx, kyma); err != nil {
-				r.Metrics.RecordRequeueReason(metrics.KymaUnderDeletionAndAccessSecretNotFound, queue.UnexpectedRequeue)
-				return ctrl.Result{}, err
-			}
-			r.Metrics.RecordRequeueReason(metrics.KymaUnderDeletionAndAccessSecretNotFound, queue.IntendedRequeue)
-			return ctrl.Result{Requeue: true}, nil
+		result, err := r.handleRemoteInitialization(ctx, kyma)
+		if err != nil {
```
ruanxin, line 118: you didn't catch this case `return ctrl.Result{Requeue: true}, nil`

ruanxin: I mean [here](https://github.com/kyma-project/lifecycle-manager/pull/1426/files/834af8e799d6a3cf317c9898980788141aa0a8a1..b1dad117f6ef192357ecc852d7b3a42026af0671#diff-4d414901850a616d4e06a120615896904639aef683f0a4d8860abf65c76179d4R139)

ruanxin (approved):

mmitoraj (approved):

PR #1435, opened 2024-03-27
chore: Sum count to avoid having multiple gauges

jeremyharisch:
**Description**

When KLM restarts, then we have multiple Gauges which is not needed.

Before:
Screenshot 2024-03-27 at 13 23 50

After: 
Screenshot 2024-03-27 at 13 23 28

**Related issue(s)**


Tomasz-Smelcerz-SAP (approved):

Tomasz-Smelcerz-SAP:
Move to blocked until it's clear how we proceed with overall changes in the monitoring dashboards (Helm vs kustomize)

PR #1436, opened 2024-03-27
chore: Namespace transformer

nesmabadr:
**Description**

- Adjust namespace transformer to have kcp-system only for those resources without a namespace set

**Related issue(s)**


config/default/namespace_delete.yaml
```diff
@@ -0,0 +1,5 @@
+$patch: delete
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: system
```
ruanxin, line 5: can we delete all the `system` namespace so that there is no needs for this delete patch?

config/control-plane/kustomization.yaml
```diff
@@ -90,6 +87,26 @@ patches:
 # Note1: Now as the 'patchesJson6902' is deprecated, the direct use of the 'PatchTransformer' is the only way to change a resource namespace to something different from the value configured by the global namespace transformer.
 # Note2: the 'grafana_dashboard' label could be applied using a standard JSON patch, but a namespace must be patched here and it's better to have all related customizations in a single place. 
 transformers:
```
c-pius, line 89: I think the comments above need to be updated. We don't use `the global namespace transformer` anymore.

We also had a comment from SRE wrt. usage of the `kyma-system` namespace [here](https://github.tools.sap/kyma/management-plane-charts/pull/858) (see heading `Dashboard CMs`). Do we need to alter that namespace anyway (maybe also in a follow-up)?

config/control-plane/kustomization.yaml
```diff
@@ -90,6 +87,26 @@ patches:
...
+- |-
+  apiVersion: builtin
+  kind: NamespaceTransformer
+  metadata:
+    name: add-resources-to-kcp-system
+    namespace: kcp-system
+  unsetOnly: true
```
c-pius, line 96: Also related to above comment wrt. `kyma-system` namespace. If we need to keep it, could we still:

Since we are only setting the unset-ones, could the `kyma-system` namespace `PatchTransformers` be replaced by directly setting the namespace and making the rest a regular patch?

config/default/kustomization.yaml
```diff
@@ -33,10 +30,12 @@ patches:
 # 'CERTMANAGER' needs to be enabled to use ca injection
 # the following config is for teaching kustomize how to do var substitution
 - path: webhookcainjection_patch.yaml
+- path: namespace_delete.yaml
 
 resources:
 - ../manager
 - ../certmanager
+- namespace_create.yaml
```
c-pius, line 38: My understanding here is that we first delete the `system` namespace, and then create the `kcp-system` namespace. For my understanding, why are we doing this now and have not done this before?

config/default/kustomization.yaml
```diff
@@ -51,82 +50,50 @@ components:
...
+  kind: NamespaceTransformer
+  metadata:
+    name: adds-resources-to-kcp-system
+    namespace: kcp-system
+  unsetOnly: true
+  setRoleBindingSubjects: allServiceAccounts
+  fieldSpecs:
+  - kind: Namespace
+    path: metadata/name
+    create: true
```
c-pius, line 65: NOt sure if I can follow what this does. According to docs:

> will update the metadata/name field of all Namespace objects

But what does this mean in our case here?

config/control-plane/kustomization.yaml
nesmabadr: But was this agreed upon? But let's do this in a followup not here because I would prefer not to change any namespaces in this PR to make sure everything is working fine as is first

config/control-plane/kustomization.yaml
nesmabadr: you mean setting the Kyma-system namespace directly in the ConfigMap? If this is the case, I was trying to do so but couldn't find a way to set a namespace for `ConfigMapGenerator`

config/default/kustomization.yaml
nesmabadr: because previously we had the system namespace, and in the kustomization we had the global namespace transformer so it renames the system to scp-system. With the new namespace transformer, it only sets the resources namespaces not the namespace name itself.

config/default/kustomization.yaml
nesmabadr: This one is no longer needed here. I was trying to use this to rename the system namespace but didn't work. will remove it

config/control-plane/kustomization.yaml
c-pius: Yeah good point, I also don't know for sure. There are some examples where they set the namespace in a config map generator. But there also seems to be some caution necessary. In doubt, fine to leave it as is.

https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/configmapgenerator/#overriding-base-configmap-values

config/manager/kustomization.yaml
```diff
@@ -15,5 +15,4 @@ configMapGenerator:
 
 images:
 - name: controller
-  newName: europe-docker.pkg.dev/kyma-project/prod/lifecycle-manager
-  newTag: latest
+  newName: /lifecycle-manager
```
c-pius, line 18: Intentional?

config/webhook/manifests.yaml
```diff
@@ -9,7 +9,6 @@ webhooks:
   clientConfig:
     service:
       name: webhook-service
-      namespace: system
```
c-pius, line 12: For me this one gets re-added when I run `make dry-run-control-plane`

c-pius (commented):
Just two minor things. I also diffed the generated manifests locally which looks good (no changes). Just wondering why the diff job in the PR doesn't start 🤔

config/manager/kustomization.yaml
nesmabadr: no pushed by mistake. Thanks!

config/control-plane/kustomization.yaml
nesmabadr: I tried this earlier but I think this only works for the Kustomization not the Component probably because it produced error when I tried to have it

c-pius (approved):

PR #1440, opened 2024-04-02
feat: Add makefile target for building kustomize manifests

gczuczy:
**Description**

This PR adds a makefile target to build the kustomize files and save the output - which should be part of the commits. This is needed for internal control-plane deployment integration.

**Related issue(s)**


manifests/all.yaml
```diff
@@ -0,0 +1,6484 @@
+apiVersion: apiextensions.k8s.io/v1
```
ruanxin, line 1: why we need this?

Makefile
```diff
@@ -44,10 +44,23 @@ help: ## Display this help.
...
+KUSTOMIZE_SRC?=config/control-plane/
+MANIFESTS_ALL=manifests/all.yaml
+MANIFESTS_TARGETS=
+
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
+$(MANIFESTS_ALL): manifests kustomize
+	$(KUSTOMIZE) build $(KUSTOMIZE_SRC) > $@
```
ruanxin, line 57: What will be the downstream process in internal control-plane chart repo? You use this `all.yaml` as a base to extract CRD and RBAC related resources?

gczuczy: Exactly. It solves the problem that you cannot do a `kustomize build https://blah` remotely, but you can simply fetch a yaml file and use parts selectively - this is quite easy.

ruanxin: but then it requires each PR to be merged in the LM repo to update this `all.yaml`, right? Otherwise, how can you keep them up-to-date? And I think it's not good practice that you build yaml based on latest commit, you must have some version control for the config file.

Why not like currently in management-plane-charts, introduce another kustomize yaml, and remotely reference to a fixed commit? We can improve further here to reference with release tag.

Example
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - github.com/kyma-project/lifecycle-manager//config/control-plane?ref=0.1.1
```

gczuczy: That's rather simple, because kustomize doesn't take argument it simply doesn't have the ability to. Tracking the version in multiple files in the same repository is very bad practice. The version is tracked in the chart's `Chart.yaml`, and kustomize is uanble to take it from there. Also, the ref to fetch this from has to be provided through the makefile framework, having some kind of a `REF` variable, defaulting to to the `Chart.yaml`'s `.appVersion`, and if you are working on upstream or something, you can override it when you are calling the target.

This would require each and every call to re-assemble a kustomization.yaml which is absolutely error-prone. Alternative is, you have to manually maintain this file in the chart, which is also very error-prone. Also, kustomize doesn't support supplying the `kustomization.yaml` in the command line, and that has to be worked around.

Many other projects i've encountered are doing this, simply hooked up to the topevel codegen target.

An alterntive would be to require git in our internal toolchain, check out the lifecycle-manager repo with the proper ref (git does accept this from the CLI), and build the kustomize there. Despite possible, this is horribly overbloats and overcompliates a simple thing, I would like to avoid this.

ruanxin: after align with @gczuczy , we agreed to use kustomize remote reference to generate control-plane manifests on management plane charts repo based on `appVersion` on the fly.

PR #1443, opened 2024-04-03
chore: Configure leader election timeout

### LINKED ISSUE #1351: chore: Increase leader election timeout
**Observation**: We are facing process crashes due to API server outages that last over 2 minutes. If the Lifecycle-Manager Pod cannot renew the "leader election" lease in that time, it crashes.

the LM Pod crash increases the total load in the control-plane, because all the caches must be re-loaded etc.
We should try to reduce the number of such events.

**Investigation**:
I can't directly explain the timings observed in the KCP. The flags controlling the leader election time-outs are not configured at this moment, defaulting to something around 10 seconds. The details below:

The leader election mechanism is configured in [cmd/main.go](https://github.com/kyma-project/lifecycle-manager/blob/9c8f61c002791f1b06b12a53523ac1b689f68f93/cmd/main.go#L142)
You can add there two options:
- `LeaseDuration *time.Duration` 
- `RenewDeadline *time.Duration`

The `RenewDeadline` configures the time after which the controller "gives up" and returns with an error. It defaults to 10 seconds.
The `LeaseDuration` configures the duration of the lease. It [MUST](https://github.com/kubernetes-sigs/controller-runtime/pull/1761#pullrequestreview-857173088) be greater than configured  `RenewDeadline`.

Example scenario I performed on a local k3d:
- given: the `LeaseDuration` is set to 120 seconds and the `RenewDeadline` is set to 90 seconds
- when:  The API server is killed
- then: The LM process exits after ~90 seconds

The error I got is: 
`
{"level":"ERROR","date":"2024-03-28T16:54:08.2667308+01:00","logger":"setup","caller":"cmd/main.go:160","msg":"problem running manager","context":{"error":"leader election lost"},"stacktrace":"..."}
`

AC:
[ ] Introduce two new flags to configure leader election timeouts
[ ] Return an error if the `RenewDeadline` is configured as greater than `LeaseDuration`
[ ] Set the defaults so that the timeout for the `RenewDeadline` is about 90 seconds
[ ] Enhance the message produced on error - it could be: `problem running manager: <internal error message>`
[ ] [Return different exit code](https://github.com/kyma-project/lifecycle-manager/blob/9c8f61c002791f1b06b12a53523ac1b689f68f93/cmd/main.go#L148), e.g: `2` when the controller exits due to internal problems.

Tomasz-Smelcerz-SAP:
**Description**

Fixes #1351

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -66,7 +66,11 @@ import (
 	// +kubebuilder:scaffold:imports
 )
 
-const metricCleanupTimeout = 5 * time.Minute
+const (
+	metricCleanupTimeout    = 5 * time.Minute
+	bootstrapFailedExitCode = 1
+	runtimeProblemExitCode  = 2
+)
 
 var (
 	scheme       = machineryruntime.NewScheme() //nolint:gochecknoglobals // scheme used to add CRDs
@@ -100,7 +104,7 @@ func main() {
 	setupLog.Info("starting Lifecycle-Manager version: " + buildVersion)
 	if err := flagVar.Validate(); err != nil {
 		setupLog.Error(err, "unable to start manager")
-		os.Exit(1)
+		os.Exit(bootstrapFailedExitCode)
 	}
 	if flagVar.Pprof {
 		go pprofStartServer(flagVar.PprofAddr, flagVar.PprofServerTimeout)
@@ -144,12 +148,14 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
 			HealthProbeBindAddress: flagVar.ProbeAddr,
 			LeaderElection:         flagVar.EnableLeaderElection,
 			LeaderElectionID:       "893110f7.kyma-project.io",
+			LeaseDuration:          &flagVar.LeaderElectionLeaseDuration,
+			RenewDeadline:          &flagVar.LeaderElectionRenewDeadline,
 			Cache:                  cacheOptions,
 		},
 	)
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
-		os.Exit(1)
+		os.Exit(bootstrapFailedExitCode)
 	}
 
 	var skrWebhookManager *watcher.SKRWebhookManifestManager
@@ -157,7 +163,7 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
 	if flagVar.EnableKcpWatcher {
 		if skrWebhookManager, err = createSkrWebhookManager(mgr, flagVar); err != nil {
 			setupLog.Error(err, "failed to create skr webhook manager")
-			os.Exit(1)
+			os.Exit(bootstrapFailedExitCode)
 		}
 		setupKcpWatcherReconciler(mgr, options, flagVar)
 	}
@@ -185,7 +191,7 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
 
 	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
 		setupLog.Error(err, "problem running manager")
-		os.Exit(1)
+		os.Exit(runtimeProblemExitCode)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


Tomasz-Smelcerz-SAP:
How to test:

1. Start the local `k3d` cluster
2. Run the LM locally: `go run cmd/main.go --leader-elect=true -leader-election-renew-deadline 20s`
3. Kill the local `k3d` cluster
4. The LM process should terminate with exit code `2` after 20 seconds.

Example run:
```
E0403 13:30:12.213201 1006013 leaderelection.go:332] error retrieving resource lock default/893110f7.kyma-project.io: Get "https://0.0.0.0:43177/apis/coordination.k8s.io/v1/namespaces/default/leases/893110f7.kyma-project.io": dial tcp 0.0.0.0:43177: connect: connection refused
E0403 13:30:14.212719 1006013 leaderelection.go:332] error retrieving resource lock default/893110f7.kyma-project.io: Get "https://0.0.0.0:43177/apis/coordination.k8s.io/v1/namespaces/default/leases/893110f7.kyma-project.io": dial tcp 0.0.0.0:43177: connect: connection refused
E0403 13:30:16.212918 1006013 leaderelection.go:332] error retrieving resource lock default/893110f7.kyma-project.io: Get "https://0.0.0.0:43177/apis/coordination.k8s.io/v1/namespaces/default/leases/893110f7.kyma-project.io": dial tcp 0.0.0.0:43177: connect: connection refused
E0403 13:30:18.213662 1006013 leaderelection.go:332] error retrieving resource lock default/893110f7.kyma-project.io: Get "https://0.0.0.0:43177/apis/coordination.k8s.io/v1/namespaces/default/leases/893110f7.kyma-project.io": dial tcp 0.0.0.0:43177: connect: connection refused
E0403 13:30:20.212556 1006013 leaderelection.go:332] error retrieving resource lock default/893110f7.kyma-project.io: client rate limiter Wait returned an error: context deadline exceeded
I0403 13:30:20.212574 1006013 leaderelection.go:285] failed to renew lease default/893110f7.kyma-project.io: timed out waiting for the condition
{"level":"ERROR","date":"2024-04-03T13:30:20.212595966+02:00","logger":"setup","caller":"cmd/main.go:194","msg":"problem running manager","context":{"error":"leader election lost"},"stacktrace":"main.setupManager\n\t/[...]/src/github.com/kyma-project/lifecycle-manager/cmd/main.go:194\nmain.main\n\t/[...]/src/github.com/kyma-project/lifecycle-manager/cmd/main.go:105\nruntime.main\n\t/usr/lib/go/src/runtime/proc.go:271"}
exit status 2
```
_Note that the time difference between the first log and last log entry in the example above is not exactly 20 seconds but more like 18 seconds. This is because it takes about 2 seconds for the controller to "realize" the API Server connectivity is gone_

Tomasz-Smelcerz-SAP:
> Thanks for the good description, I can replicate the behavior as you described.
> 
> Two general things I am wondering about:
> 
> a) when does the leader election actually get used? I looked at Argo for stage and prod and can see that we only have one pod there. Are we only using it when a new version of KLM is deployed?
A quote from some k8s book: `Some master components, such as the scheduler and the controller manager, can't have multiple instances active at the same time. This will be chaos, as multiple schedulers try to schedule the same pod into multiple nodes or multiple times into the same node. The correct way to have a highly-scalable Kubernetes cluster is to have these components run in leader election mode. This means that multiple instances are running, but only one is active at a time, and if it fails, another one is elected as leader and takes its place.`
In our case this mechanism is mainly used to prevent two instances of KLM running in parallel due to misconfiguration.

> b) in the ticket you described that API server outages are about 2 mins. Why are we setting the renew deadline to 90 seconds then and not something slightly above 2 mins?
Short answer: I think shorter is better, but we decided to "test" how it behaves for longer time-outs...
Long anser: We raise this mainly to "see what happens" during ETCD outages, hoping the KLM will survive the short ones. Other than that we don't have any strict recommendation. I think the 90 seconds is OK - it is almost one order of magnitude bigger than the default 10 seconds. Raising it requires also raising the lease time and this has consequences - two KLM instances running in parallel will "detect" that situation with time resolution determined by the lease time. So the longer the lease time, the greater risk of having two KLMs reconciling the same cluster.

c-pius (commented):
Thanks for the explanations! Good for me, just one thing I just remembered. We once agreed to add all flag defaults to the tests file which has not been done yet: https://github.com/kyma-project/lifecycle-manager/blob/main/internal/pkg/flags/flags_test.go

c-pius (approved):

Issue #1239, opened 2024-01-16
Ensure TTL for in-memory client cache(s)


Tomasz-Smelcerz-SAP:
### Description

Ensure every SKR client cache used in the system has a properly configured time-to-live.
We have two cache implementations, both custom code based on `sync.Map`:
- `pkg/remote/client_cache.go`
- `internal/declarative/v2/client_cache.go`

Somehow related issue: https://github.com/kyma-project/lifecycle-manager/issues/684

### Reasons

We cache SKR clients because these are costly to create. We have two caches: one for the Kyma reconciliation loop, one for the Manifest reconciliation loop. The clients are configured a little bit differently, that's why we use two separate ones.
For now, we are keeping the clients forever, which is not good, because:
- There's no guarantee that a client is removed from the cache after the SKR cluster is deleted.
- A small memory leak in the client, multiplied by a number of cached clients (a few hundreds) results in a big memory leak overall.

### Acceptance Criteria

- [x] Introduce TTL cache per client with a jitter

Issue #1126, opened 2023-12-06
Refactoring: Remove `localClient` global Variable in `remote` pkg


jeremyharisch:
### Description

To have some of our code testable in EnvTest, we introduced a global variable called `localClient` in the [`remote` package.](https://github.com/kyma-project/lifecycle-manager/blob/aef1529f43842c6aa2e14d5a52ca8b5049fad7f4/pkg/remote/kyma_synchronization_context.go#L28)

But, we should not have production ready code including code which is just used for testing, instead the code should be written in a way that it is testable without such modifications.

The global var `localClient` is set during EnvTests when creating a new SKRCluster, and thus when the LifecycleManager tries to resolve the SKRClient, it instead used the set `localClient`. With the current implementation each test scenario using the [`NewSKRCluster` func](https://github.com/kyma-project/lifecycle-manager/blob/aef1529f43842c6aa2e14d5a52ca8b5049fad7f4/pkg/testutils/utils.go#L124) runs in its own virtual SKR cluster, it would be great to keep such behaviour to have separated test cases.

### Reasons

- Not have test code in production.
- Reduce tech dept.
- Avoided unwanted behaviour by overwriting the global variable.

### Acceptance Criteria

- [ ] Refactor the implementation of `restConfigFromStrategy` in such a way that it does not use a global variable to determine if a localClient should be used
- [ ] Adapt test suits
- [ ] Test cases should still run in own virtual SKR cluster

### Feature Testing

Integration tests

PR #1445, opened 2024-04-04
feat: TTL client caches

lindnerby:
**Description**

- Replace internal caches of `pkg/remote/client_cache` and `internal/declarative/v2/client_cache` with TTL caches
- TTL for clients is between 23h-25h

**Related issue(s)**
* https://github.com/kyma-project/lifecycle-manager/issues/1239


pkg/remote/client_cache_test.go
```diff
@@ -0,0 +1,75 @@
+package remote_test
```
nesmabadr, line 1: Could you please maybe add this package to the unit-test-coverage file?

lindnerby: Sure, the coverage is very low, because the remote package contains a lot. On another branch concerning [this](https://github.com/kyma-project/lifecycle-manager/issues/1126) refactoring, the package it is moved and nicely structured and also separated in subpackages so the coverage can be adapted then.

nesmabadr (approved):

nesmabadr (approved):

PR #1451, opened 2024-04-05
chore: Introduce shared GitHub Actions

### LINKED ISSUE #1224: E2E config improvement
### Description
Both E2E workflows of lifecycle-manager and runtime-watcher have overlapping config and setup requirements. Let's extract and share the setup requirements steps. Have a look at [re-using workflows](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-an-action-inside-a-different-private-repository-than-the-workflow).
Possible further investigation: 
    * https://docs.github.com/en/actions/creating-actions/sharing-actions-and-workflows-from-your-private-repository
    * https://docs.github.com/en/actions/creating-actions/sharing-actions-and-workflows-with-your-organization

### Reasons

We want to update versions and make changes to prerequisites of e2e tests only one time and not per repository.

### Acceptance Criteria

- [ ] Setup steps in e2e tests are shared between lifecycle-manager and runtime-watcher: install kubectl, checkout repository, setup go, install istio, install kyma cli, install certmanager cli, install k3d, provision clusters
- [ ] Split the setup into three separate workflows: "install prerequisites" and "create clusters" (which includes kubeconfig handling and etc/hosts patching), "template-operator setup" (only required for certain matrix entries, but includes, repository checkout, module creation, apply module). Feel free to come up with a better split of workflows if applicable.
- [ ] Align steps from runtime-watcher towards -> lifecycle-manager because in runtime-watcher they have been reworked already quite a bit!

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

c-pius:
**Description**

- splits the E2E Test workflow into single actions that can be composed and re-used between our repositories
- adds some improvements, see the inline comments

**Related issue(s)**

- partially closes #1224

**Comments**

~~To ease the review allowing to compare the new workflow runs against the old ones, the old `test-e2e.yaml` workflow is still there unchanged and the new workflow is in `test-e2e-junk.yaml`. Once the changes are good to go, the old file will be updated with the new workflow and the separate junk file will be deleted.~~

~~When reviewing, it is probably best to start from `test-e2e-junk.yaml` and from there look into the used helper actions.~~

Once this PR is merged, we can re-use the actions in watcher. Right now it is not possible because policies restrict using actions from forked repositories.


.github/actions/deploy-lifecycle-manager/action.yaml
```diff
@@ -0,0 +1,31 @@
...
+  steps:
+    - name: Deploy LM local testing kustomize
+      working-directory: lifecycle-manager
+      shell: bash
+      run: |
+        maxRetry=5
+        for retry in $(seq 1 $maxRetry)
+        do
+          if make local-deploy-with-watcher IMG=europe-docker.pkg.dev/kyma-project/${{ inputs.klm_image_repo }}/lifecycle-manager:${{ inputs.klm_version_tag }}; then
+            kubectl wait pods -n kcp-system -l app.kubernetes.io/name=lifecycle-manager --for condition=Ready --timeout=90s
```
c-pius, line 21: comment: this wait is not there in watcher. Not sure if really needed or if we could omit it.

.github/actions/get-configuration/action.yaml
```diff
@@ -0,0 +1,38 @@
...
+        echo "k8s_version=${{ github.event.inputs.k8s_version || '1.28.7' }}" >> $GITHUB_OUTPUT
+        echo "istio_version=1.20.3" >> $GITHUB_OUTPUT
+        echo "k3d_version=5.6.0" >> $GITHUB_OUTPUT
+        echo "cert_manager_version=1.13.3" >> $GITHUB_OUTPUT
+        echo "klm_version_tag=PR-${{ github.event.pull_request.number }}" >> $GITHUB_OUTPUT
+        echo "klm_image_repo=dev" >> $GITHUB_OUTPUT
+    - name: Expose environment variables
+      shell: bash
+      run: |
+        echo "GOSUMDB=off" >> $GITHUB_ENV
```
c-pius, line 38: comment: I opted to have everything just needed in the test setup as "regular" vars, and to only export those things as env vars which are needed downstream by the tests.

.github/actions/get-configuration/action.yaml
```diff
@@ -0,0 +1,38 @@
...
+    - name: Define variables
+      id: define-variables
+      shell: bash
+      run: |
+        echo "k8s_version=${{ github.event.inputs.k8s_version || '1.28.7' }}" >> $GITHUB_OUTPUT
+        echo "istio_version=1.20.3" >> $GITHUB_OUTPUT
+        echo "k3d_version=5.6.0" >> $GITHUB_OUTPUT
+        echo "cert_manager_version=1.13.3" >> $GITHUB_OUTPUT
+        echo "klm_version_tag=PR-${{ github.event.pull_request.number }}" >> $GITHUB_OUTPUT
+        echo "klm_image_repo=dev" >> $GITHUB_OUTPUT
```
c-pius, line 34: comment: these here used to be a conditional based on whether the job is running for a `main` build or a `PR` build. Since we removed the post `main` build, I removed this conditional now.

.github/workflows/test-e2e-junk.yaml
```diff
@@ -0,0 +1,92 @@
...
+jobs:
+  wait-for-image-build:
+    name: "Wait for image build"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/wait-for-image-build
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          statusName: pull-lifecycle-mgr-build
```
c-pius, line 21: comment: these here used to be a conditional based on whether the job is running for a `main` build or a `PR` build. Since we removed the post `main` build, I removed this conditional now.

.github/actions/deploy-lifecycle-manager/action.yaml
jeremyharisch: As discussed in person, this is needed and we should definitely keep it to ensure the pods is running in the first retry to have working webhooks

jeremyharisch (approved):

jeremyharisch (approved):

PR #1458, opened 2024-04-10
fix: Update Manifest lastOperation when an error occurs

### LINKED ISSUE #1453: Manifest CR skip updating last operation when error occurs
### Description

At the moment, Manifest controller only update last operation field with some defined operations. However, for unexpected errors, KLM not record it, which makes troubleshooting sometimes difficult.

One example, the ssaStatusIfDiffExist not record the error while doing ssa to status subresources. 
https://github.com/kyma-project/lifecycle-manager/blob/36caa68020f6dd5c60913bb5fe0e803ffb244fb9/internal/declarative/v2/reconciler.go#L141-L148

### Steps to reproduce

we have one incident: #5371, the manifest CR requeued with error, but there is no lastOperation message indicating the error message.

### Environment Type

Managed

### Environment Info

Kubernetes Version: x.y.z
Docker Version: x.y.z
Kyma CLI Version: x.y.z

### Actual Result
Last Operation in Manifest CR is not updated to reflect the error that occurred.

### Expected Result
Last Operation in Manifest CR should contain info on what was the last error.

nesmabadr:
**Description**

- Update Manifest LastOperation when an error occurs

**Related issue(s)**
Resolves #1453

### DIFF
--- a/internal/declarative/v2/reconciler.go
+++ b/internal/declarative/v2/reconciler.go
@@ -309,6 +309,7 @@ func (r *Reconciler) renderResources(
 	converter := NewResourceToInfoConverter(ResourceInfoConverter(clnt), r.Namespace)
 
 	if target, err = r.renderTargetResources(ctx, clnt, converter, obj, spec); err != nil {
+		obj.SetStatus(status.WithState(shared.StateError).WithErr(err))
 		return nil, nil, err
 	}
 
@@ -348,7 +349,7 @@ func (r *Reconciler) syncResources(ctx context.Context, clnt Client, obj Object,
 		if obj.GetDeletionTimestamp().IsZero() {
 			obj.SetStatus(status.WithState(shared.StateProcessing).WithOperation(ErrWarningResourceSyncStateDiff.Error()))
 		} else if status.State != shared.StateWarning {
-			obj.SetStatus(status.WithState(shared.StateDeleting).WithOperation("manifest should be deleted"))
+			obj.SetStatus(status.WithState(shared.StateDeleting).WithOperation(ErrWarningResourceSyncStateDiff.Error()))
 		}
 		return ErrWarningResourceSyncStateDiff
 	}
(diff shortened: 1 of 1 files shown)


internal/declarative/v2/reconciler.go
```diff
@@ -432,6 +433,7 @@ func (r *Reconciler) removeModuleCR(ctx context.Context, clnt Client, obj Object
 			if err := preDelete(ctx, clnt, r.Client, obj); err != nil {
 				r.Event(obj, "Warning", "PreDelete", err.Error())
 				// we do not set a status here since it will be deleting if timestamp is set.
```
Tomasz-Smelcerz-SAP, line 435: Is this comment still valid?

internal/declarative/v2/reconciler.go
```diff
@@ -514,9 +517,16 @@ func (r *Reconciler) pruneDiff(
...
 	manifest, ok := obj.(*v1beta2.Manifest)
 	if !ok {
+		obj.SetStatus(obj.GetStatus().WithErr(v1beta2.ErrTypeAssertManifest))
 		return v1beta2.ErrTypeAssertManifest
 	}
-	return resources.NewConcurrentCleanup(clnt, manifest).DeleteDiffResources(ctx, diff)
+	err = resources.NewConcurrentCleanup(clnt, manifest).DeleteDiffResources(ctx, diff)
+	if err != nil {
+		obj.SetStatus(obj.GetStatus().WithErr(err))
+		return err
```
Tomasz-Smelcerz-SAP, line 526: You can save one line of code and move this return to the line 529 :)

internal/declarative/v2/reconciler.go
nesmabadr: yes because I don't set a new status, I only update the error with the current status

Tomasz-Smelcerz-SAP (approved):

PR #1461, opened 2024-04-10
chore: Remove manifest rolebinding in kyma-system and istio-system namespaces 

nesmabadr:
**Description**

- Remove manifest rolebinding in kyma-system and istio-system namespaces 
- Use ByObject instead of DefaultNamespaces for controller runtime cache options

**Related issue(s)**

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -15,5 +15,5 @@ configMapGenerator:
 
 images:
 - name: controller
-  newName: europe-docker.pkg.dev/kyma-project/prod/lifecycle-manager
-  newTag: latest
+  newName: europe-docker.pkg.dev/kyma-project/dev/lifecycle-manager
+  newTag: PR-1461
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -15,5 +15,5 @@ configMapGenerator:
 
 images:
 - name: controller
-  newName: europe-docker.pkg.dev/kyma-project/dev/lifecycle-manager
-  newTag: PR-1461
+  newName: europe-docker.pkg.dev/kyma-project/prod/lifecycle-manager
+  newTag: latest
--- a/internal/cache_options.go
+++ b/internal/cache_options.go
@@ -19,7 +19,10 @@ type DefaultCacheOptions struct {
 }
 
 type KcpCacheOptions struct {
-	CacheOptions cache.Options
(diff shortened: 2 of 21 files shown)


config/manager/kustomization.yaml
```diff
@@ -15,5 +15,5 @@ configMapGenerator:
 
 images:
 - name: controller
-  newName: europe-docker.pkg.dev/kyma-project/prod/lifecycle-manager
-  newTag: latest
+  newName: europe-docker.pkg.dev/kyma-project/dev/lifecycle-manager
+  newTag: PR-1461
```
ruanxin, line 19: this should not be committed

internal/cache_options.go
```diff
@@ -0,0 +1,92 @@
...
+	apicorev1 "k8s.io/api/core/v1"
+	k8slabels "k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+)
+
+type ConfigCacheOptions interface {
+	GetConfigCacheOptions() cache.Options
```
ruanxin, line 14: I would rename it to `GetCacheOptions`, it's aligned with what gets returned.

config/manager/kustomization.yaml
nesmabadr: yes pushed by mistake

internal/cache_options.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+}
+
+func (c *KcpCacheOptions) GetCacheOptions() cache.Options {
+	return cache.Options{
+		ByObject: map[client.Object]cache.ByObject{
+			&apicorev1.Secret{}: {
+				Label: k8slabels.Everything(),
+				Namespaces: map[string]cache.Config{
+					"kcp-system":   {},
```
ruanxin, line 41: These namespace should coming from flag value instead of hard-coded, we have `flagVar.IstioGatewayNamespace`, `flagVar.IstioNamespace`, `flagVar.RemoteSyncNamespace` for those namespace, so that they are configurable.

ruanxin (changes requested):
the namespace of cache options should be configurable based on actual given flag.

ruanxin (approved):

mmitoraj (approved):

PR #1465, opened 2024-04-11
test: Add check for module version in Kyma.Modules.Status in E2E test

### LINKED ISSUE #1456: Avoid ssa for Manifest CR when no update needed
### Description

At the moment, kyma controller will do ssa for all enabled modules, no matter if there is change or not, and in most cases, the related manifest CRs only need to be updated if there is update in the moduletemplate.

Solution:

Kyma `status.modules[].version` tracking the current enabled manifest CR version, before doing ssa of manifest, we should compare it with the version of the current moduletemplate, only update manifest when the version get updated.

Include the check of comparing the `status.modules[].channel` as well.

https://github.com/kyma-project/lifecycle-manager/blob/bb2bf933d260730000b6e7f56b50fa0e5dfd0809/pkg/module/sync/runner.go#L121-L132

### Reasons

reduce requests to API server.

### Acceptance Criteria
- [x] Only do manifest CR update, by extending this [function](https://github.com/kyma-project/lifecycle-manager/blob/bb2bf933d260730000b6e7f56b50fa0e5dfd0809/pkg/module/sync/runner.go#L162), when: 
    - [x] There is new version for the moduletemplate 
    - [x] Or the channel has been switched 
    - [x] Or Manifest CR status.state has been changed compared to KymaCr.Status.Modules[x].state 
- [x] Cover scenario in tests.
- [x] Extend [this test](https://github.com/kyma-project/lifecycle-manager/blob/a0c49436f3d11d03c9a7556ec11c7c9f69d621d9/tests/e2e/module_upgrade_new_version_test.go#L12) to check Module Version  in `kyma.status` and ManifestCR gets updated in E2E

### Feature Testing

Unit Tests, check if possible for integration or E2E test

### Testing approach

_No response_

### Attachments

_No response_

nesmabadr:
**Description**

- Update E2E test for module version check on update
- Avoid ssa for Manifest CR when no update needed

**Related issue(s)**
Resolves #1456


pkg/module/sync/runner.go
```diff
@@ -110,23 +110,37 @@ func (r *Runner) updateManifests(ctx context.Context, kyma *v1beta2.Kyma,
...
 	return nil
 }
 
-func (r *Runner) doUpdateWithStrategy(ctx context.Context, owner string, isEnabledModule bool,
+func (r *Runner) DoUpdateWithStrategy(ctx context.Context, owner string, isEnabledModule bool,
 	manifestObj *v1beta2.Manifest,
 ) error {
+	manifestInCluster := &v1beta2.Manifest{}
+	var err error
+	if err = r.Get(ctx, client.ObjectKey{Namespace: manifestObj.GetNamespace(), Name: manifestObj.GetName()},
```
ruanxin, line 125: we should not do `Get` here, which means that we introduce an additional request API server for each `manifestObj`, which means around 10k requests more for the current PROD landscape, please think twice when introducing new requests to the API server, it will affect performance hugely.

pkg/module/sync/runner.go
```diff
@@ -110,23 +110,37 @@ func (r *Runner) updateManifests(ctx context.Context, kyma *v1beta2.Kyma,
...
+		manifestInCluster); err != nil && !util.IsNotFound(err) {
+		return fmt.Errorf("error get manifest %s: %w", client.ObjectKeyFromObject(manifestObj), err)
+	}
+
+	if !NeedToUpdate(manifestInCluster, manifestObj) {
```
ruanxin, line 130: as mentioned in the original [solution](https://github.com/kyma-project/lifecycle-manager/issues/1456), the version and module channel should coming from `kyma.status`, but not from actual `manifestInCluster`

ruanxin (changes requested):
please check my comment

pkg/module/sync/runner.go
```diff
@@ -142,25 +156,30 @@ func (r *Runner) patchManifest(ctx context.Context, owner string, manifestObj *v
...
-		return fmt.Errorf("error get manifest %s: %w", client.ObjectKeyFromObject(manifestObj), err)
-	}
-	if !needToUpdate(manifestInCluster, manifestObj) {
-		return nil
-	}
+func (r *Runner) updateAvailableManifestSpec(ctx context.Context, manifestInCluster, manifestObj *v1beta2.Manifest,
+) error {
 	manifestInCluster.Spec = manifestObj.Spec
 	if err := r.Update(ctx, manifestInCluster); err != nil {
 		return fmt.Errorf("error update manifest %s: %w", client.ObjectKeyFromObject(manifestObj), err)
```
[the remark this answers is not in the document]
nesmabadr: No longer passed to the function

pkg/module/sync/runner.go
```diff
@@ -142,25 +146,46 @@ func (r *Runner) patchManifest(ctx context.Context, owner string, manifestObj *v
...
-	return manifestInCluster.Spec.Version != manifestObj.Spec.Version
+func NeedToUpdate(kymaStatus v1beta2.KymaStatus, manifestObj *v1beta2.Manifest) bool {
+	moduleStatus := getModuleStatus(kymaStatus, manifestObj.GetName())
+
+	if moduleStatus == nil {
+		return true
+	}
+
+	return moduleStatus.Version != manifestObj.Spec.Version ||
+		moduleStatus.Channel != getChannelLabel(manifestObj)
```
ruanxin, line 172: Can you create an issue to introduce channel as an optional field in manifest CR v1beta2, we need to get rid of this get channel from label logic asap.

pkg/module/sync/runner.go
```diff
@@ -142,25 +146,46 @@ func (r *Runner) patchManifest(ctx context.Context, owner string, manifestObj *v
...
 	manifestInCluster.Spec = manifestObj.Spec
 	if err := r.Update(ctx, manifestInCluster); err != nil {
 		return fmt.Errorf("error update manifest %s: %w", client.ObjectKeyFromObject(manifestObj), err)
 	}
 	return nil
 }
 
-func needToUpdate(manifestInCluster, manifestObj *v1beta2.Manifest) bool {
-	return manifestInCluster.Spec.Version != manifestObj.Spec.Version
+func NeedToUpdate(kymaStatus v1beta2.KymaStatus, manifestObj *v1beta2.Manifest) bool {
```
ruanxin, line 164: think about rename this `manifestObj` to something indicate it's lastest manifest cr to be updated.

pkg/module/sync/runner.go
```diff
@@ -110,24 +110,28 @@ func (r *Runner) updateManifests(ctx context.Context, kyma *v1beta2.Kyma,
...
+func (r *Runner) DoUpdateWithStrategy(ctx context.Context, owner string, isEnabledModule bool,
 	manifestObj *v1beta2.Manifest,
 ) error {
 	if isEnabledModule {
 		return r.patchManifest(ctx, owner, manifestObj)
 	}
+
 	// For disabled module, the manifest CR is under deleting, in this case, we only update the spec when it's still not deleted.
-	if err := r.updateAvailableManifestSpec(ctx, manifestObj); err != nil && !util.IsNotFound(err) {
+	if err := r.updateAvailableManifestSpec(ctx, manifestObj); client.IgnoreNotFound(err) != nil {
```
ruanxin, line 131: do you still need this logic? the `manifestObj` should contains disabled module, right?

pkg/module/sync/runner.go
nesmabadr: https://github.com/orgs/kyma-project/projects/30/views/13?pane=issue&itemId=60056896

pkg/module/sync/runner.go
```diff
@@ -142,25 +131,31 @@ func (r *Runner) patchManifest(ctx context.Context, owner string, manifestObj *v
...
+}
+
+func getModuleStatus(status v1beta2.KymaStatus, moduleName string) *v1beta2.ModuleStatus {
+	for idx := range status.Modules {
+		if status.Modules[idx].Name == moduleName {
+			return &status.Modules[idx]
+		}
 	}
 	return nil
 }
```
c-pius, line 152: should this rather be a method on the `KymaStatus` type? Something like: `func(status KymaStatus) GetModuleStatus(name string) (ModuleStatus, err)`?

pkg/testutils/manifest.go
```diff
@@ -588,3 +589,17 @@ func CredSecretLabelSelector(labelValue string) *apimetav1.LabelSelector {
 		MatchLabels: map[string]string{OCIRegistryCredLabelKeyForTest: labelValue},
 	}
 }
+
+func ManifestVersionIsCorrect(ctx context.Context, clnt client.Client, kymaName,
+	kymaNamespace,
+	moduleName, version string,
+) error {
```
c-pius, line 596: is this formatting intended? Looks a bit odd

pkg/module/sync/runner.go
nesmabadr: you mean the whole updateAvailableManifestSpec? yes we still need this if the module is being disabled

pkg/testutils/manifest.go
nesmabadr: That's the result of gofumpt

pkg/module/sync/runner.go
nesmabadr: Hmmm I don't think it makes a difference here TBH. Why would you prefer it like this?

pkg/module/sync/runner.go
ruanxin: why? if `manifestObj` contains all latest content, you can do the `patchManifest` to disabled module also.

ruanxin: the point is we need to ignore the error returned by `patchManifest` if manifest CR is not exists any more.

pkg/module/sync/runner.go
nesmabadr: hmm makes sense. But my point was that it is not used anywhere else so I just had it as a helper method here. But it makes sense to have it on the KymaStatus yes

c-pius (commented):
would be good from my side, but would leave the approve to Xin

ruanxin (dismissed):

pkg/testutils/modulecr.go
```diff
@@ -23,6 +23,7 @@ const (
 	ModuleManagedCRName                = "template-operator-managed-resource"
 	ModuleDeploymentNameInNewerVersion = "template-operator-v2-controller-manager"
 	ModuleDeploymentNameInOlderVersion = "template-operator-v1-controller-manager"
+	NewModuleVersion                   = "2.4.2-e2e-test"
```
ruanxin, line 26: this version better read from moduletemplate, it's more flexible in case we update the test moduletemplate later.

pkg/testutils/moduletemplate.go
```diff
@@ -85,3 +88,24 @@ func DeleteModuleTemplate(ctx context.Context,
 	}
 	return nil
 }
+
+func ReadModuleVersionFromModuleTemplate(moduleTemplateFilePath string) (string, error) {
```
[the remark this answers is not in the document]
nesmabadr: no strong reason here, but just to make sure that it is the actual new moduleTemplate that we intend to apply

ruanxin (approved):

mmitoraj (approved):

PR #1473, opened 2024-04-16
remove global value in main.go

ruanxin:


cmd/main.go
```diff
@@ -70,18 +71,15 @@ const (
 	metricCleanupTimeout    = 5 * time.Minute
 	bootstrapFailedExitCode = 1
 	runtimeProblemExitCode  = 2
+	buildVersion            = "not_provided"
```
Tomasz-Smelcerz-SAP: This will break the contract for setting the build version in the Dockerfile (the target must be a `var`, not a `const`)
I noticed though that it doesn't work anymore for other reasons anyway - no build version is present in latest image builds.
It looks like something has changed in the image-builder. I will create a follow-up issue to clean this up.

Tomasz-Smelcerz-SAP (approved):

PR #1495, opened 2024-04-23
fix: Return error when doing ssa for diff in declarative library

### LINKED ISSUE #1399: "Operator Controller Runtime Reconcile Error" metric is broken
### Description

There are two problems with the metric:

1. The errors for the `manifest` controller is reported with a value of zero. Checked on DEV and STAGE. It's hard to believe there are no errors at all. The `kyma` controller reports some errors. See the attached screenshot.
    - Currently we are returning nil in `declarative` pkg if error happens
    - Should return error in error case
    - Adapt existing test-cases
2. The panel plots graphs for all controllers in the given environment, there's no proper filtering that narrows the graph only to the the metrics produced by Lifecycle Manager.    - Probably already fixed, but should be double checked

### Steps to reproduce

See the plutono dashboard on DEV, STAGE

### Environment Type

Managed

### Environment Info

n/a

### Attachments

image

nesmabadr:
**Description**

- Return error when doing ssa for diff in declarative library

**Related issue(s)**
Resolves #1399


internal/declarative/v2/reconciler.go
```diff
@@ -144,7 +144,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			r.cleanUpMandatoryModuleMetrics(obj)
 			return r.removeFinalizers(ctx, obj, []string{r.Finalizer}, metrics.ManifestRemoveFinalizerWhenParseSpec)
```
Tomasz-Smelcerz-SAP, line 145: I think here we're also ignoring the original error. It is clearly visible in the `removeFinalizers` function, in the line 256.

internal/declarative/v2/reconciler.go
```diff
@@ -640,7 +640,7 @@ func (r *Reconciler) configClient(ctx context.Context, obj Object) (Client, erro
 }
 
 func (r *Reconciler) ssaStatusIfDiffExist(ctx context.Context, obj Object,
-	requeueReason metrics.ManifestRequeueReason, previousStatus shared.Status,
+	requeueReason metrics.ManifestRequeueReason, previousStatus shared.Status, err error,
```
Tomasz-Smelcerz-SAP, line 643: Small suggestion: This parameter name: `err` is later shadowed by the variable defined in the line 647.
Shadowing may be confusing: one has to read the code very carefully to understand the scopes of the "original" vs "shadowed" variable.
I'd vote for renaming this parameter to "originalErr" or "externalErr" to better describe the purpose of this parameter and avoid shadowing.

Tomasz-Smelcerz-SAP (changes requested):
Please see my comments

internal/declarative/v2/reconciler.go
```diff
@@ -253,7 +253,7 @@ func (r *Reconciler) removeFinalizers(ctx context.Context, obj Object, finalizer
 		obj.SetStatus(obj.GetStatus().WithState(shared.StateDeleting).WithOperation(msg))
 	}
 
-	return r.ssaStatusIfDiffExist(ctx, obj, requeueReason, currentObjStatus)
+	return r.ssaStatusIfDiffExist(ctx, obj, requeueReason, currentObjStatus, nil)
```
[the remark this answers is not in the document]
nesmabadr: ah good point. I was confused by the deletion condition that's why I missed that error

Tomasz-Smelcerz-SAP (approved):

Tomasz-Smelcerz-SAP (approved):

Tomasz-Smelcerz-SAP (approved):

Issue #1352, opened 2024-02-29
chore: Get event creation statistics for environments


Tomasz-Smelcerz-SAP:
### Description

We create many events in the control-plane. These events are meant to deliver additional information about "history" of errors on a given Kyma object etc.
We got the feedback that the event creation rate is too high and that the LM is overloading the API Server.
To reduce the number of events we have to find out how many events of each type are created in given environment (DEV, STAGE, PROD).
Then we can decide if want to get rid of some of the events, or maybe find a way to suppress the creation ratio of them. 

### Reasons

Reduce the load on the API Server
Increase stability of the control-plane
Improve LM observability

### Acceptance Criteria

- [ ] Find out how many events the LM is creating in a reasonable time frame (few hours/day)
- [ ] Find out how many event "types" are created - events created with similar message or because of similar condition.
- [ ] Produce a statistics that allow to see which events are created most often. A Histogram?
- [ ] Provide the above for every environment: DEV, STAGE, PROD

ebensom:
Statistics prod PROD environment
----

[Number of events per kind during the last 7 days](https://console.cloud.google.com/monitoring/metrics-explorer;startTime=2024-04-02T11:00:00.000Z;endTime=2024-04-09T11:00:00.000Z?pageState=%7B%22xyChart%22:%7B%22constantLines%22:%5B%5D,%22dataSets%22:%5B%7B%22plotType%22:%22LINE%22,%22prometheusQuery%22:%22sum%20by%20(kind)(increase(logging_googleapis_com:user_klm_events%7Bmonitored_resource%3D%5C%22k8s_cluster%5C%22%7D%5B7d%5D))%22,%22targetAxis%22:%22Y1%22,%22unitOverride%22:%22%22%7D%5D,%22options%22:%7B%22mode%22:%22COLOR%22%7D,%22y1Axis%22:%7B%22label%22:%22%22,%22scale%22:%22LINEAR%22%7D%7D%7D&project=sap-ti-dx-kyma-mps-prod):

| Kind | No. of events | 
| -- | -- |
| Manifest | 141241 |
| Kyma | 11982 |

[...]

PR #1508, opened 2024-04-29
chore: Remove useless event

ruanxin:
This PR is a follow up to https://github.com/kyma-project/lifecycle-manager/issues/1352, remove some useless event which brings not so much value and reduce traffic cost to API server.


internal/controller/kyma_controller.go
```diff
@@ -426,35 +411,28 @@ func (r *KymaReconciler) handleDeletingState(ctx context.Context, kyma *v1beta2.
 
 	if r.WatcherEnabled(kyma) {
 		if err := r.SKRWebhookManager.Remove(ctx, kyma); err != nil {
-			// error is expected, try again
-			r.enqueueNormalEvent(kyma, webhookChartRemoval, err.Error())
-			return ctrl.Result{RequeueAfter: r.RequeueIntervals.Busy}, nil
```
lindnerby, line 431: Here the busy interval requeue gets replaced with immediate requeue. What's the reasoning behind?

internal/controller/kyma_controller.go
```diff
@@ -426,35 +411,28 @@ func (r *KymaReconciler) handleDeletingState(ctx context.Context, kyma *v1beta2.
...
+			return r.requeueWithError(ctx, kyma, err)
 		}
 
 		r.RemoteClientCache.Delete(client.ObjectKeyFromObject(kyma))
 		if err := remote.RemoveFinalizersFromRemoteKyma(ctx, r.RemoteSyncNamespace); client.IgnoreNotFound(err) != nil {
-			err = fmt.Errorf("error while trying to remove finalizer from remote: %w", err)
-			r.enqueueWarningEvent(kyma, deletionError, err)
 			r.Metrics.RecordRequeueReason(metrics.FinalizersRemovalFromRemoteKyma, queue.UnexpectedRequeue)
-			return ctrl.Result{}, err
+			return r.requeueWithError(ctx, kyma, err)
```
lindnerby, line 429: Here and in the line 423 you are introducing updateStatus calls now, which weren't required before. Is this expected?

internal/declarative/v2/reconciler.go
```diff
@@ -469,26 +458,19 @@ func (r *Reconciler) renderTargetResources(
...
-			r.Event(obj, "Warning", "PostRenderTransform", err.Error())
 			obj.SetStatus(status.WithState(shared.StateError).WithErr(err))
 			return nil, err
 		}
 	}
 
 	target, err := converter.UnstructuredToInfos(targetResources.Items)
 	if err != nil {
-		// Prevent ETCD load bursts during secret rotation
-		if !util.IsConnectionRelatedError(err) {
```
lindnerby, line 488: Is this util function still used somewhere else now?

ruanxin: yes, it's a critical function actually.

internal/controller/kyma_controller.go
ruanxin: yes, when this situation happens, it helps us know the reason why this kyma stuck under deletion by check lastoperation field.

internal/controller/kyma_controller.go
ruanxin: to compliance with requeue standard, when err occurs, requeue with error, there is no needs to control requeue interval for error, it's controlled by rate limit.

lindnerby (approved):

PR #1576, opened 2024-05-23
chore: RequeueAfter should not mixed with requeue with error

ruanxin:
This PR fixed the requeue behaviour, if requeue with error, the requeueAfter will not working, we should not mixed those two senario.

Also renamed some functions to clarified the responsibility.


lindnerby (approved):

internal/declarative/v2/reconciler.go
```diff
@@ -198,23 +185,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 	if requireUpdateSyncedOCIRefAnnotation(obj, spec.OCIRef) {
 		updateSyncedOCIRefAnnotation(obj, spec.OCIRef)
-		return r.updateObject(ctx, obj, metrics.ManifestUpdateSyncedOCIRef, queue.IntendedRequeue)
+		return r.updateObject(ctx, obj, metrics.ManifestUpdateSyncedOCIRef)
 	}
 
 	if !obj.GetDeletionTimestamp().IsZero() {
 		r.cleanUpMandatoryModuleMetrics(obj)
 		r.ManifestMetrics.RemoveManifestDuration(req.Name)
-		return r.removeFinalizers(ctx, obj, []string{r.Finalizer}, metrics.ManifestRemoveFinalizerInDeleting, nil)
```
ruanxin, line 214: To reviewer:
previous e2e failing is due to this line, the error pass to `removeFinalizers` is `nil, which leads to requeue without error.
I refactored the `removeFinalizers` to not take care requeue logic.

Issue #1247, opened 2024-01-19
Research on tests for API Changes options


nesmabadr:
### Description

Since we can have multiple API versions supported (now we have v1beta1 and v1beta2), we need to investigate a way to add tests to validate that any changes/fields added to an API version get added to the other version as well.

### Reasons

Since we can have multiple API versions supported, it can likely happen that we only add a field to the new API version forgetting about the old ones, and this can introduce issues in the Lifecycle Manager

### Acceptance Criteria

- [x] Think of a way to test that API changes are consistent among the different versions.
- [x] Investigate whether such tests could be implemented/how to do that
- [x] Create a follow-up issue for the implementation

c-pius:
## Idea 1

Use the generated OpenAPISpecification (OAS) v3 schema and compare old vs new version for differences in a GH Action.

TL;DR: the only viable option seems to be option **3**

> For testing, the schema of our KymaCRD is taken and the following additional changes are done to `v1beta2`:
> 
> - [ ] various descriptions are changed (this should be ignored)
> - [ ] some property is removed (needs to be detected)
> - [ ] some property is added (needs to be detected)
> - [ ] some enum entry is added (needs to be detected)
> - [ ] some type is changed (needs to be detected)

### 1. [`dyff`](https://github.com/homeport/dyff)

Plain YAML diff tool.

```diff
+ can work directly on the same file with two different root paths
+ large numbers of stars (1.2k)
+ allows to filter out description changes ('--exclude-regexp ".*description"')
+ nice output describing differences (value change, key addition, key removal, ...)
- filtering/exclusion options do not allow filtering on certain keys added/removed from a map (e.g., to filter out the removed entry 'properties.spec.properties.sync', we would need to filter out the whole 'properties.spec.properties' path)
- large number of open and unaddressed issues
- moderate maintenance (one commit every month or so)
```

<details>

<summary>Example</summary>

```sh
dyff between ./config/crd/bases/operator.kyma-project.io_kymas.yaml ./config/crd/bases/operator.kyma-project.io_kymas.yaml --chroot-of-from "spec.versions.0.schema.openAPIV3Schema" --chroot-of-to "spec.versions.1.schema.openAPIV3Schema" --exclude-regexp ".*description"
     _        __  __
   _| |_   _ / _|/ _|  between ./config/crd/bases/operator.kyma-project.io_kymas.yaml, YAML root was changed to spec.versions.0.schema.openAPIV3Schema
[...]

PR #1602, opened 2024-06-03
feat: Add Workflow to Check API Changes in PR

### LINKED ISSUE #1492: Create GH Action that detects API changes
### Description

Follow-up from https://github.com/kyma-project/lifecycle-manager/issues/1247

We want to create a GH Action that detects changes between API versions. We chose to go with [Option 3 of Idea 1](https://github.com/kyma-project/lifecycle-manager/issues/1247#issuecomment-2066782823). 

### Reasons

Since we can have multiple API versions supported, it can likely happen that we only add a field to the new API version forgetting about the old ones, and this can introduce issues in the Lifecycle Manager

### Acceptance Criteria

- [x] GH Action added that fails when differences are reported
- [x] When the description of fields changes
  - [x] NO difference is reported
- [x] When a field is removed
  - [x] a difference is reported
- [x]  When a field is added
  - [x] a difference is reported
- [x]  When a field type is changed
  - [x] a difference is reported
- [x] When a field's enum values are changed
  - [x] a difference is reported

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

LeelaChacha:
Closes #1492


LeelaChacha:
TODO: make new workflow required

.github/workflows/check-api-changes.yaml
```diff
@@ -0,0 +1,101 @@
...
+            main_file="./main-branch-crds/$file"
+            pr_file="./pr-crds/$file"
+            versions=$(yq e '.spec.versions[].name' "$pr_file")
+
+            for version in $versions; do
+              compare_schemas "$main_file" "$pr_file" "$version"
+            done
+          done
+
+          echo "API schemas match in all CRD files"
```
c-pius, line 101: There are a couple of things that need to be changed here:

- a) we need to compare different versions of the APIs, e.g. `v1beta1` against `v1beta2` (see also the example here https://github.com/kyma-project/lifecycle-manager/issues/1247#issuecomment-2066782823). The reason is that we need to ensure that when K8s stores one version of the CRD in etcd, it must be able to translate this version to other served versions. Hence we want to be alerted when the API changes between versions so that we can check if we need to backport something to the older API. And for the known changes and okay ones, we need to do the pre-processing (e.g., as in the example remove the `sync` field from `v1beta1` as this is a known and okay change).
Since we may have more than two API versions in the future, I would propose that we take the latest version and compare it against the predecessor, then take the predecessor and compare it against its predecessor and so on. With this change, we then should also be able get rid of the `Compare names of API versions in CRD files` step (as we may introduce new versions, we just need to take care about the compatiblity), and we also may get rid of the label to accept a change (accepting a change is done by pre-processing a known difference so that it doesn't yield a difference output)
- b) use `dyff` instead of `diff`. Apologies, this was discussed in the team when looking at https://github.com/kyma-project/lifecycle-manager/issues/1247#issuecomment-2066782823, but not stated clearly as part of this ticket. Main reasons are the license, better formatting of the output, and the option to ignore descriptions.
- c) I would propose to move the preprocessing to another step if possible. This step should be named something like "preprocessing known differences" so it is obvious that we have a set of differences that we need to pre-process and where to add future pre-processing

c-pius (changes requested):
I really like the well written and structured scripts! There is just one misconception that we need to change. See the detailed comments. If you have questions, just ping me anytime to discuss this in detail.

.github/workflows/check-api-changes.yaml
```diff
@@ -0,0 +1,101 @@
...
+      - name: Copy PR CRDs to temp dir
+        run: mkdir -p /tmp/pr-crds && cp ./config/crd/bases/* /tmp/pr-crds/
+
+      - name: Checkout main of lifecycle-manager
+        uses: actions/checkout@v4
+        with:
+          ref: main
+
+      - name: Copy main CRDs to temp dir
+        run: mkdir -p /tmp/main-branch-crds && cp ./config/crd/bases/* /tmp/main-branch-crds/
```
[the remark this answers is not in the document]
LeelaChacha: that was my first instict but the "main" in main-crds felt like a modifier of the crds instead of referring to the branch. But I guess, that was just overthinking

.github/workflows/check-api-changes.yaml
```diff
@@ -0,0 +1,101 @@
...
+      - name: Compare names and number of CRD files
+        working-directory: /tmp
+        run: |
+          main_files=$(ls ./main-branch-crds)
+          pr_files=$(ls ./pr-crds)
+          
+          if [[ "$pr_files" != "$main_files" ]]; then
+            echo "Number of CRD files do not match"
+            echo "Differences between main-branch-crds and pr-crds:"
+            diff -r main-branch-crds pr-crds || true
```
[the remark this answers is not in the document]
LeelaChacha: it shows the diff of the two directories in case the strings don't match (API(s) was added or removed)

.github/workflows/check-api-changes.yaml
LeelaChacha: oh I guess I misunderstood the issue a bit. I wrote the workflow to compare crd changes between main and pr. I will modify it for intra-crd version comparison.

LeelaChacha:
@c-pius: I would rather not split the job into preprocessing and diff. It would increase the code size and complexity. For eg. generating the files in a different step and the dyff'ing the correct version pairs for all the crds. 
I tried to keep the script as simple as possible, avoiding overengineered sed commands,etc. This allows better readability and maintainability.
WDYT?

api-version-compatibility-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+    v1beta2:
+      - .spec.properties.modules.x-kubernetes-list-map-keys
+      - .spec.properties.modules.x-kubernetes-list-type
+operator.kyma-project.io_moduletemplates.yaml:
+  exclusions:
+    v1beta1:
+      - .spec.properties.target
+      - .spec.required
+    v1beta2:
+      - .spec.required
```
c-pius, line 24:
```suggestion
# The configuration file is used by the GitHub Actions workflow defined in .github/workflows/check-api-compatibility.yaml.
# It is used to define exclusions when comparing different versions of CRD schemas and
# specifies the fields that should be excluded from the comparison for each version of a CRD.
# The structure of the file is as follows:
#
# <crd-filename>:
#   exclusions:
#     <version>:
#       - <exclusion-path> (may be a yq operator without spaces)

operator.kyma-project.io_kymas.yaml:
  exclusions:
    v1beta1:
      - .spec.properties.sync
    v1beta2:
      - .spec.properties.modules.x-kubernetes-list-map-keys
      - .spec.properties.modules.x-kubernetes-list-type
operator.kyma-project.io_moduletemplates.yaml:
  exclusions:
    v1beta1:
      - .spec.properties.target
      - .spec.required[]|select(.=="target")

```

We can actually also supply a yq operator to the `del` operator and explicitly remove the "target" entry from the required array. We only can't use spaces as the parsing of the exlucsions does not properly detect this.

c-pius (changes requested):
Just some minor suggestions left 👍🏻

api-version-compatibility-config.yaml
LeelaChacha: I tried do that but I didn't work, but your the no space trick of yours should work

c-pius (approved):
Nice job!

PR #1620, opened 2024-06-10
feat: Avoid Redundant SSA for Manifest Patching

### LINKED ISSUE #1456: Avoid ssa for Manifest CR when no update needed

LeelaChacha:
Closes #1456


LeelaChacha:
NOTE: One integration test is failing as it is invalidated now. It is supposed to check that a manifest is reset if manually overwritten. However, now manifest patching only works when new template version, channel, or kyma module status is present.
How to proceed with the test? Should we reverse the check to show that manifest ssa is not?

c-pius (commented):
Thanks for the re-work! Can confirm that the flickering is now gone 👍🏻 

> NOTE: One integration test is failing as it is invalidated now. It is supposed to check that a manifest is reset if manually overwritten. However, now manifest patching only works when new template version, channel, or kyma module status is present.
How to proceed with the test? Should we reverse the check to show that manifest ssa is not?

This is an important finding. We actually changed parts of our handling, hence the test fails. I also tested this manually. Before, when e.g. changing `Spec.Install.Source.Repo` of a ManifestCR, we reconciled it back to the original state. Now this is not happening anymore except we also change the version, or channel. I guess this would be acceptable as ManifestCRs reside in KCP and can only be changed by us or SRE. I still want to double check... @ruanxin is this something we explicitly accept?

ruanxin:
> Thanks for the re-work! Can confirm that the flickering is now gone 👍🏻
> 
> > NOTE: One integration test is failing as it is invalidated now. It is supposed to check that a manifest is reset if manually overwritten. However, now manifest patching only works when new template version, channel, or kyma module status is present.
> > How to proceed with the test? Should we reverse the check to show that manifest ssa is not?
> 
> This is an important finding. We actually changed parts of our handling, hence the test fails. I also tested this manually. Before, when e.g. changing `Spec.Install.Source.Repo` of a ManifestCR, we reconciled it back to the original state. Now this is not happening anymore except we also change the version, or channel. I guess this would be acceptable as ManifestCRs reside in KCP and can only be changed by us or SRE. I still want to double check... @ruanxin is this something we explicitly accept?

Yes, normally modifying manifest CR spec is controlled by us, it's fine to skip this overwritten, however, it still has benefits that allow this overwritten.
one use case for troubleshooting:
1. add skip reconciliation label for one Kyma CR
2. modify related manifest CR spec for testing or troubleshooting
3. afterwards, remove skip reconciliation label
4. manifest CR get reset.

And it's possible to support this feature, just need to do a client diff check between the spec of manifest in cluster (`manifestInCluster`) and converted from moduletemplate (`manifestObj`), but we should monitor the kyma reconciliation time after release, to see how expensive this diff check.

c-pius:
Thanks for chiming in! Just not sure if I understood you correctly since you explained both... Are you suggesting to accept it as proposed in this PR (don't overwrite manual changes), or are you suggesting to add the diff check as an additional evaluation in `NeedToUpdate` (and overwrite manual changes)?

ruanxin:
> Thanks for chiming in! Just not sure if I understood you correctly since you explained both... Are you suggesting to accept it as proposed in this PR (don't overwrite manual changes), or are you suggesting to add the diff check as an additional evaluation in `NeedToUpdate` (and overwrite manual changes)?

I suggest adding this diff check in this PR, otherwise, we lose the support for that use case.

c-pius:
> > Thanks for chiming in! Just not sure if I understood you correctly since you explained both... Are you suggesting to accept it as proposed in this PR (don't overwrite manual changes), or are you suggesting to add the diff check as an additional evaluation in `NeedToUpdate` (and overwrite manual changes)?
> 
> I suggest adding this diff check in this PR, otherwise, we lose the support for that use case.

Thanks for clarifying!

@LeelaChacha would you mind checking if this can be added to the `NeedToUpdate`?

LeelaChacha:
@ruanxin @c-pius does a diff of the spec suffice?

c-pius:
> @ruanxin @c-pius does a diff of the spec suffice?
> 
> Test is now green, btw

I guess checking the spec part should suffice. However, I am not sure if `reflect.DeepEqual` is the way to go. As mentioned [here](https://github.com/kubernetes-sigs/kubebuilder/issues/592), K8s may provide default values for unset fields. When doing so, it will be unequal all the time for certain manifests. Maybe this `equality.Semantic.DeepDerivative` is a good alternative. Would you mind checking with some debug comments whether it works?

And one other thing for @ruanxin. I vaguely remember that somebody mentioned that we transitioned to using Server Side Apply to transfer the compute load of comparing the resources from our controller to the API server (not sure if I remember this correctly). We are now re-adding this load on the controller. Are we sure we want to do that?

c-pius (approved):

internal/descriptor/provider/provider.go
```diff
@@ -70,6 +71,20 @@ func (c *CachedDescriptorProvider) GetDescriptor(template *v1beta2.ModuleTemplat
...
+		return ErrTemplateNil
+	}
+	template.Spec.Descriptor.Object = descriptor
+	encoded, err := compdesc.Encode(descriptor.ComponentDescriptor, compdesc.DefaultJSONCodec)
+	template.Spec.Descriptor.Raw = encoded
+	if err != nil {
+		return errors.Join(ErrEncode, err)
+	}
+	return nil
+}
```
c-pius, line 86: This is production code that is only used for testing. Such things should be avoided in general (sometimes not possible though). Here is a proposal for how we can fix the integration test without adding new production code:

https://github.com/LeelaChacha/kyma-lifecycle-manager/pull/1

c-pius (changes requested):

internal/descriptor/provider/provider.go
LeelaChacha: that is a really clean solution! The only downside I see is the extra file, but objectively better than extra functions only for test.

c-pius (approved):

docs/technical-reference/api/manifest-cr.md
```diff
@@ -2,6 +2,16 @@
...
 
+## Patching
+
+The [Runner](../../../pkg/module/sync/runner.go) is responsible for creating and updating manifests via Server Side Apply (SSA). An update is only performed when one of the following conditions is met:
+
+1. The version of the manifest object is different from the version in the Kyma module status.
+2. The channel of the manifest object is different from the channel in the Kyma module status.
+3. The state of the manifest object is different from the state in the Kyma module status.
+
+*NOTE:* The module status is not present in Kyma for mandatory modules, hence their manifest is updated via SSA in every reconcile loop.
```
mmitoraj, line 13:
```suggestion
>[!NOTE] 
>The module status is not present in Kyma for mandatory modules, hence their manifest is updated using SSA in every reconcile loop.
```
in Kyma? meaning Kyma CR?

docs/technical-reference/api/manifest-cr.md
```diff
@@ -2,6 +2,16 @@
 
 The [Manifest custom resource (CR)](../../../api/v1beta2/manifest_types.go) is our internal representation of what results from the resolution of a ModuleTemplate CR in the context of a single cluster represented by a Kyma CR. Thus, a lot of configuration elements are similar or entirely equivalent to the data layer found in a ModuleTemplate CR.
 
+## Patching
+
+The [Runner](../../../pkg/module/sync/runner.go) is responsible for creating and updating manifests via Server Side Apply (SSA). An update is only performed when one of the following conditions is met:
+
+1. The version of the manifest object is different from the version in the Kyma module status.
```
mmitoraj, line 9:
```suggestion
1. The Manifest CR version differs from the Kyma CR's module status version.
```

docs/technical-reference/api/manifest-cr.md
```diff
@@ -2,6 +2,16 @@
...
+2. The channel of the manifest object is different from the channel in the Kyma module status.
```
mmitoraj, line 10:
```suggestion
2. The Manifest CR channel differs from the Kyma CR's module status channel.
```

docs/technical-reference/api/manifest-cr.md
```diff
@@ -2,6 +2,16 @@
...
+3. The state of the manifest object is different from the state in the Kyma module status.
```
mmitoraj, line 11:
```suggestion
3. The Manifest CR state differs from the Kyma CR's module status state.
```

c-pius (approved):

mmitoraj (approved):

Issue #1554, opened 2024-05-16
Stop aggregate module cr state into manifest cr


ruanxin:
### Description

Related epic: https://github.com/kyma-project/lifecycle-manager/issues/1471

At the moment, KLM aggregates module cr state from skr cluster into manifest cr, which makes manifest cr state mixed with module cr, in order to achieve the goal of the epic, we should stop this aggregate process.

### Reasons

not mix the module status with manifest status

### Acceptance Criteria

- [x] Remove the aggregation logic, code reference: https://github.com/kyma-project/lifecycle-manager/blob/407b873a1d7f3de86446b193e1f772dec3f45f0c/internal/declarative/v2/reconciler.go#L389
- [x] Adjust existing tests aligned with this new requirement
- [x] synchronised with Busola team so that all the states are displayed correctly

### Feature Testing

Unit tests, Integration tests, End-to-End tests

#### required e2e tests:
1. test module CR state not affect manifest CR
**Given** **Template Operator ModuleTemplate** on the **KCP Cluster**
**And** the **Sample CR** final state has been configured as `Warning`
**When** the **Template Operator Module** is enabled
**Then** the **Manifest CR** should be in `Ready` state
**Then** the **Sample CR** should be in `Warning` state

2. test the module Deployment failing state propagated into manifest CR
**Given** **Template Operator ModuleTemplate** on the **KCP Cluster**
**When** the **Template Operator Module** is enabled
**And** the **Template Operator Deployment** configured in error state
**Then** the **Manifest CR** should be in `Error` state
**Then** the **Sample CR** should be in `Ready` state

c-pius:
Re-opened this as we have noticed undesired behavior during verification on STAGE. We can observe that the state in `ManifestCR` is separated from `DefaultCR` ✅ 

However, we also noticed that on initial installation, `ManifestCR` and therefore also `KymaCR` go first to `Error` state before becoming `Ready`. Previously, this has been `Processing` instead of `Error` for the transient initializing state.

c-pius:
### Some observations:

- ManifestCR state goes Processing->Error->Processing. If second processing is not reached within 5s, it stays in error for 5 mins.
- ManifestCR sometimes transitions Processing before the Deployment is ready

### Reproduce on local test cluster:

[...]

PR #1628, opened 2024-06-12
feat: Separate business status from module status

jeremyharisch:
**Description**

Implements https://github.com/kyma-project/lifecycle-manager/issues/1554

- New Status behaviour for Manifest CRs:
    - If the Module defined in the Manifest CR is applied successfully and is running, the status is Ready.
    - If the Module is still being applied or starting, the status is Processing.
    - If the Module fails to start or the Manifest application fails, the status is Error.
    - If the Manifest CR is marked for deletion, the status is Deleted.
- Covered new behaviour in E2E Test

**Related issue(s)**

Resolves https://github.com/kyma-project/lifecycle-manager/issues/1554


docs/technical-reference/api/manifest-cr.md
```diff
@@ -87,7 +87,16 @@ The resource is the default data that should be initialized for the module and i
 
 ### **.status**
 
-The Manifest CR status is an unmodified version of the [declarative status](../../../internal/declarative/README.md#resource-tracking), so the tracking process of the library applies. There is no custom API for this.
+The status of the Manifest Custom Resource (CR) is a direct, unaltered reflection of the [declarative status](../../../internal/declarative/README.md#resource-tracking), which means it follows the tracking process implemented by the internal library. There isn't a custom API specifically designed for this status tracking. Instead, the existing mechanisms provided by the library are utilized to monitor and update the status of the Manifest CR.
```
ruanxin, line 90: thb, the first part is hard to understand for me (generated by GPT?), and we will soon remove the declarative pattern, there is no need to align this declarative status. I suggest removing the first paragraph.

docs/technical-reference/api/manifest-cr.md
```diff
@@ -87,7 +87,16 @@ The resource is the default data that should be initialized for the module and i
...
+The status of the Manifest Custom Resource (CR) is a direct, unaltered reflection of the [declarative status](../../../internal/declarative/README.md#resource-tracking), which means it follows the tracking process implemented by the internal library. There isn't a custom API specifically designed for this status tracking. Instead, the existing mechanisms provided by the library are utilized to monitor and update the status of the Manifest CR.  
+
+The Manifest CR status is set based on the following logic, managed by the manifest reconciler: 
+
+- If the Module defined in the Manifest CR is successfully applied and the deployed Module is up and running, the status of the Manifest CR is set to `Ready`.
+- While the Manifest is being applied and the Deployment is still starting, the status of the Manifest CR is set to `Processing`.
+- If the Deployment is unable to start (for example, due to an `ImagePullBackOff` error) or if the application of the Manifest fails, the status of the Manifest CR is set to `Error`.
+- If the Manifest CR is marked for deletion, the status of the Manifest CR is set to `Deleted`.
```
ruanxin, line 97: We don't have `Deleted` state.

internal/declarative/v2/reconciler.go
```diff
@@ -394,7 +394,7 @@ func (r *Reconciler) checkTargetReadiness(
 
 	resourceReadyCheck := r.CustomReadyCheck
 
-	crStateInfo, err := resourceReadyCheck.Run(ctx, clnt, manifest, target)
+	crStateInfo, err := resourceReadyCheck.Run(ctx, clnt, target)
```
ruanxin, line 397: `crStateInfo` previously means module CR state, with your change, it's the deployment state. please consider changing to a proper name.

internal/declarative/v2/reconciler.go
```diff
@@ -406,6 +406,10 @@ func (r *Reconciler) checkTargetReadiness(
 		return ErrInstallationConditionRequiresUpdate
 	}
 
+	if !manifest.GetDeletionTimestamp().IsZero() {
```
ruanxin, line 409: this condition is related to manifest CR, it's not related to `checkTargetReadiness`, should not inside this function, and also consider renaming `checkTargetReadiness` to something more intuitive, like `checkDeploymentState`

internal/manifest/ready_check.go
```diff
@@ -2,227 +2,67 @@ package manifest
...
 
-const (
-	ToWarningDuration                = 5 * time.Minute
-	customResourceStatePath          = "status.state"
-	ModuleCRWithCustomCheckWarning   = "module CR state not found or given customStateCheck.jsonPath is not exists"
-	ModuleCRWithNoCustomCheckWarning = "module CR state not found"
-)
-
-// NewCustomResourceReadyCheck creates a readiness check that verifies that the Resource in the Manifest
+// NewDeploymentReadyCheck creates a readiness check that verifies that the Resource in the Manifest
```
ruanxin, line 18: the definition of this function should be updated

internal/manifest/ready_check.go
```diff
@@ -2,227 +2,67 @@ package manifest
...
-	err := clnt.Get(ctx, client.ObjectKeyFromObject(moduleCR), moduleCR)
-	if err != nil {
-		if util.IsNotFound(err) && !manifest.DeletionTimestamp.IsZero() {
-			return declarativev2.StateInfo{State: shared.StateDeleting}, nil
-		}
-		return declarativev2.StateInfo{State: shared.StateError}, fmt.Errorf("failed to fetch resource: %w", err)
-	}
-	return HandleState(manifest, moduleCR)
+	deploymentState := getDeploymentState(clnt, resources)
+	return declarativev2.StateInfo{State: deploymentState}, nil
```
ruanxin, line 31: why always return a `nil` error?

internal/manifest/ready_check.go
```diff
@@ -2,227 +2,67 @@ package manifest
...
-		foundStateInCR = true
-		_, found := stateResult[stateCheck.MappedState]
-		if found {
-			stateResult[stateCheck.MappedState] = stateResult[stateCheck.MappedState] && stateFromCR == stateCheck.Value
-		} else {
-			stateResult[stateCheck.MappedState] = stateFromCR == stateCheck.Value
-		}
+func getDeploymentState(clt declarativev2.Client, resources []*resource.Info) shared.State {
+	deploy, found := findDeployment(clt, resources)
+	// Not every module operator use Deployment by default, e.g: StatefulSet also a valid approach
```
ruanxin, line 36: Let's have a followup issue to support `StatefulSet` state check.

internal/manifest/ready_check.go
```diff
@@ -232,3 +72,31 @@ func isDeploymentReady(clt declarativev2.Client, resources []*resource.Info) boo
...
 	return false
 }
+
+func getPodsForDeployment(clt declarativev2.Client, deploy *apiappsv1.Deployment) (*apicorev1.PodList, error) {
+	podList := &apicorev1.PodList{}
+	listOptions := &client.ListOptions{
+		Namespace:     deploy.Namespace,
+		LabelSelector: k8slabels.SelectorFromSet(deploy.Spec.Selector.MatchLabels),
+	}
+	if err := clt.List(context.TODO(), podList, listOptions); err != nil {
```
ruanxin, line 82: you need to use context from `Run(ctx context.Context, ...`

internal/manifest/ready_check.go
```diff
@@ -232,3 +72,31 @@ func isDeploymentReady(clt declarativev2.Client, resources []*resource.Info) boo
...
+		return nil, fmt.Errorf("failed to list pods: %w", err)
+	}
+	return podList, nil
+}
+
+func getPodsState(podList *apicorev1.PodList) shared.State {
+	for _, pod := range podList.Items {
+		for _, condition := range pod.Status.ContainerStatuses {
+			switch {
+			case *condition.Started && condition.Ready:
```
ruanxin, line 92: why not use containerStates? https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-states

`ContainerStatus.State` contains three state, which can be mapped to our shared state.

docs/technical-reference/api/manifest-cr.md
jeremyharisch: Typo, `Deleting` it is

internal/declarative/v2/reconciler.go
jeremyharisch: I will refactor the function to split the deployment check and the manifest status setting into two different functions

internal/manifest/ready_check.go
jeremyharisch: Because in this implementation for the interface there is no return Error needed, if an error appears then the State will be set to `Error`.
There is a second implementation for the `ReadyCheck` which has an error case and is needed for testing.
Check here: https://github.com/kyma-project/lifecycle-manager/blob/8766d1503157a6a05027266eb02c77bd388936a7/internal/declarative/v2/ready_check.go#L31

internal/manifest/ready_check.go
jeremyharisch: Good catch!

internal/manifest/ready_check.go
jeremyharisch: Because I had the Problem when using ContainerStatus, then it was flickering between `Waiting` and `Terminated`. When a container gets terminated, the container will be restarted and thus have again the status `Waiting` since it tries to fetch the image. If for example the image does not exist (ImagePullBackOff) or the image crashes it terminates again and restart. This is reflected in the restart count.

Using the conditions is in my opinion the better way to map the current behaviour of the deployment to our states without any flickering.

internal/declarative/v2/reconciler.go
```diff
@@ -387,31 +392,47 @@ func hasDiff(oldResources []shared.Resource, newResources []shared.Resource) boo
...
+	ctx context.Context, clnt Client, obj Object, target []*resource.Info,
+) (shared.State, error) {
 	resourceReadyCheck := r.CustomReadyCheck
 
-	crStateInfo, err := resourceReadyCheck.Run(ctx, clnt, manifest, target)
+	deploymentStateInfo, err := resourceReadyCheck.Run(ctx, clnt, target)
 	if err != nil {
-		manifest.SetStatus(status.WithState(shared.StateError).WithErr(err))
-		return err
+		return "", err
```
ruanxin, line 402: there is no `""` shared.State, should be `StateError` although it's not in use.

internal/manifest/ready_check.go
ruanxin: ok, then I'm fine with this `condition.Started`, however, it's an optional field, how about those `nil` case, I mean if the containerstatues doesn't have Started condition, your current implementation will have NPE.

You use two boolean variables (`*condition.Started` and `condition.Ready`) in the switch case, it should have 4 cases, but only have three in your implementation. The fourth case may never occured, but for consistent purpose, we should not ignore it. I suggest the rest two cases by default return `Error` state

```
case *condition.Started && condition.Ready:
  return shared.StateReady
case *condition.Started && !condition.Ready:
  return shared.StateProcessing
default:
  return shared.StateError
```

and please write a unit test for this `getPodsState` function.

internal/manifest/ready_check.go
ruanxin: then please just return `getDeploymentState(clnt, resources), nil`

internal/manifest/ready_check.go
jeremyharisch: https://github.com/kyma-project/lifecycle-manager/issues/1649

internal/declarative/v2/reconciler.go
```diff
@@ -387,31 +392,47 @@ func hasDiff(oldResources []shared.Resource, newResources []shared.Resource) boo
...
+		return shared.StateError, err
+	}
+
+	if deploymentStateInfo.State == shared.StateProcessing {
+		return shared.StateProcessing, nil
 	}
 
-	if crStateInfo.State == shared.StateProcessing {
-		waitingMsg := "waiting for resources to become ready: " + crStateInfo.Info
+	if !obj.GetDeletionTimestamp().IsZero() {
```
ruanxin, line 409: this `obj` is manifest, and you did this check in `setManifestState`

internal/declarative/v2/reconciler.go
ruanxin: check the comment here https://github.com/kyma-project/lifecycle-manager/pull/1628/files#r1658759549

docs/technical-reference/api/manifest-cr.md
```diff
@@ -87,7 +87,14 @@ The resource is the default data that should be initialized for the module and i
 
 ### **.status**
 
-The Manifest CR status is an unmodified version of the [declarative status](../../../internal/declarative/README.md#resource-tracking), so the tracking process of the library applies. There is no custom API for this.
+The Manifest CR status is set based on the following logic, managed by the manifest reconciler: 
+
+- If the Module defined in the Manifest CR is successfully applied and the deployed Module is up and running, the status of the Manifest CR is set to `Ready`.
```
grego952, line 92:
```suggestion
- If the module defined in the Manifest CR is successfully applied and the deployed module is up and running, the status of the Manifest CR is set to `Ready`.
```

docs/technical-reference/api/manifest-cr.md
```diff
@@ -87,7 +87,14 @@ The resource is the default data that should be initialized for the module and i
...
+- While the Manifest is being applied and the Deployment is still starting, the status of the Manifest CR is set to `Processing`.
```
grego952, line 93:
```suggestion
- While the manifest is being applied and the Deployment is still starting, the status of the Manifest CR is set to `Processing`.
```

docs/technical-reference/api/manifest-cr.md
```diff
@@ -87,7 +87,14 @@ The resource is the default data that should be initialized for the module and i
...
+- If the Deployment is unable to start (for example, due to an `ImagePullBackOff` error) or if the application of the Manifest fails, the status of the Manifest CR is set to `Error`.
```
grego952, line 94:
```suggestion
- If the Deployment is unable to start (for example, due to an `ImagePullBackOff` error) or if the application of the manifest fails, the status of the Manifest CR is set to `Error`.
```

docs/technical-reference/api/manifest-cr.md
```diff
@@ -87,7 +87,14 @@ The resource is the default data that should be initialized for the module and i
...
+- If the Manifest CR is marked for deletion, the status of the Manifest CR is set to `Deleting`.
+
+This status provides a reliable way to track the state of the Manifest CR and the associated Module, offering insights into the deployment process and any potential issues, while being decoupled of the Module business logic.
```
grego952, line 97:
```suggestion
This status provides a reliable way to track the state of the Manifest CR and the associated module, offering insights into the deployment process and any potential issues, while being decoupled from the module business logic.
```

docs/technical-reference/api/manifest-cr.md
```diff
@@ -87,7 +87,14 @@ The resource is the default data that should be initialized for the module and i
 
 ### **.status**
 
-The Manifest CR status is an unmodified version of the [declarative status](../../../internal/declarative/README.md#resource-tracking), so the tracking process of the library applies. There is no custom API for this.
+The Manifest CR status is set based on the following logic, managed by the manifest reconciler: 
+
+- If the module defined in the Manifest CR is successfully applied and the deployed module is up and running, the status of the Manifest CR is set to `Ready`.
+- While the manifest is being applied and the Deployment is still starting, the status of the Manifest CR is set to `Processing`.
+- If the Deployment is unable to start (for example, due to an `ImagePullBackOff` error) or if the application of the manifest fails, the status of the Manifest CR is set to `Error`.
```
IwonaLanger, line 94:
```suggestion
- If the Deployment cannot start (for example, due to an `ImagePullBackOff` error) or if the application of the manifest fails, the status of the Manifest CR is set to `Error`.
```

docs/technical-reference/api/manifest-cr.md
```diff
@@ -87,7 +87,14 @@ The resource is the default data that should be initialized for the module and i
...
+- If the Manifest CR is marked for deletion, the status of the Manifest CR is set to `Deleting`.
+
+This status provides a reliable way to track the state of the Manifest CR and the associated module, offering insights into the deployment process and any potential issues, while being decoupled from the module business logic.
```
IwonaLanger, line 97:
```suggestion
This status provides a reliable way to track the state of the Manifest CR and the associated module. It offers insights into the deployment process and any potential issues while being decoupled from the module's business logic.
```

IwonaLanger (approved):

ruanxin (approved):

PR #1635, opened 2024-06-14
feat: Plutono configuration for displaying Enabled Modules in KCP

### LINKED ISSUE #1528: Monitoring - dashboard for Module Manifests
### Description

There should be an option to look into our KCP situation - see how many instances of each Kyma Module are enabled. Right now it is only possible by executing kubectl commands on the KCP cluster. Dashboard would provide a more transparent and convenient way of accessing that data, both for KLM and Module teams.

### Reasons

We want to have a more detailed overview of how many modules are enabled.

### Acceptance Criteria

- [x] There should be a dashboard that would include:
  - Information on how many instances of each module are enabled (manifests)
  - Include it in here: https://plutono.cp.stage.kyma.cloud.sap/d/eniERxEVz/lifecycle-manager-kyma-status?orgId=1
  - Do not hardcode module names, should group them automatically by module name 

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

amritanshusikdar:
**Description**
Added the configuration for the Plutono dashboard under the `Kyma State Overview` to include the new view, which contains all the enabled modules and their respective active instances.

**Related issue(s)**
Resolves #1528

### DIFF
--- a/config/grafana/status.json
+++ b/config/grafana/status.json
@@ -34,7 +34,7 @@
     },
     {
       "datasource": null,
-      "description": "This panel provides information on Which modules are enabled on the KCP and the number of their respective instances.",
+      "description": "This panel provides information on which modules are enabled on the KCP and their respective instance counts.",
       "fieldConfig": {
         "defaults": {
           "color": {
@@ -47,10 +47,6 @@
               {
                 "color": "green",
                 "value": null
-              },
-              {
-                "color": "red",
-                "value": 250
               }
             ]
           }
--- a/config/grafana/status.json
+++ b/config/grafana/status.json
@@ -33,6 +33,66 @@
       "type": "row"
     },
     {
+      "datasource": null,
+      "description": "This panel provides information on Which modules are enabled on the KCP and the number of their respective instances.",
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "thresholds"
+          },
+          "mappings": [],
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green",
+                "value": null
+              },
+              {
+                "color": "red",
+                "value": 250
+              }
+            ]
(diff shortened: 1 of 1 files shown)


config/grafana/status.json
```diff
@@ -32,6 +32,66 @@
...
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green",
+                "value": null
+              },
+              {
+                "color": "red",
+                "value": 250
+              }
```
jeremyharisch, line 54: Any specific Reason why the graph should switch to the color red if there are more than 250 modules enabled?

Lets all have them at the same color. Red normally shows that something goes into a direction which we do not want to have.

config/grafana/status.json
```diff
@@ -32,6 +32,66 @@
       "title": "Kyma State Overview",
       "type": "row"
     },
+    {
+      "datasource": null,
+      "description": "This panel provides information on Which modules are enabled on the KCP and the number of their respective instances.",
```
jeremyharisch, line 37:
```suggestion
      "description": "This panel provides information on which modules are enabled on the KCP and their respective instance counts.",
```

jeremyharisch (commented):
Left two small comments, other than that looks good to me

config/grafana/status.json
amritanshusikdar: The initial idea was to have the whole count category as red, if even a single instance of that module is in error state or not in the desired state. But I left it, since it was out of the scope of this issue. So let's just keep everything as green. :)

config/grafana/status.json
amritanshusikdar: Done :)

amritanshusikdar:
Thanks, all comments addressed :D

LeelaChacha (approved):
LGTM. just label was added.

PR #1665, opened 2024-07-05
chore: Bump k8s version for e2e to 1.29.6

lindnerby:
**Description**

- Bump k8s version to 1.29.6 for smoke and e2e test suites
- Introduce wait to img build on main builds for dispatch workflows

**Related issue(s)**

### DIFF
--- a/.github/actions/get-configuration/action.yaml
+++ b/.github/actions/get-configuration/action.yaml
@@ -26,7 +26,7 @@ runs:
       id: define-variables
       shell: bash
       run: |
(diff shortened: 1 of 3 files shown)


c-pius:
LGTM. Am just wondering if we should bump to 1.29 already considering that we will be on 1.28 in KCPs til 10th of August. On the other hand, this will give us quite some time to monitor the behavior. How did we do it in the past?

ruanxin (approved):

PR #1675, opened 2024-07-10
feat(Deletion Modes): Populate associated resources field in Manifest

### LINKED ISSUE #1651: [Deletion Modes] Implement Logic to Extract and Insert Associated Resources into Manifest CR `associatedResources`
### Description

With the issue described in [lifecycle-manager #1582](https://github.com/kyma-project/lifecycle-manager/issues/1582), a new field called associatedResources is added to the `Manifest` Custom Resource. This change requires implementing logic to extract the list of managed resources from the ModuleTemplate and insert it into the `manifest.spec. associatedResources `.

The list of managed resources is an additional layer in the OCM specification. An example of this can be found in the implementation described in [template-operator #200](https://github.com/kyma-project/template-operator/issues/200).

After implementing this logic, the Manifest CR should look like this:

```yaml
[...]
spec:
  install:
    name: raw-manifest
    source:
      [...]
  remote: true
  resource:
    kind: Sample
    metadata:
      name: sample-yaml
      namespace: kyma-system
    spec:
      resourceFilePath: ./module-data/yaml
  associatedResources: 
     - serverless.kyma-project.io/v1alpha2/functions
     - operator.kyma-project.io/v1alpha1/serverlesses
  version: 1.1.1
```

### Reasons

- Provide logic to extract managed resources from `ModuleTemplate` OCM Layer
- Fill in `manifest.spec. associatedResources ` list

### Acceptance Criteria

- [x] Kyma Reconciliation Loop:
    - [x] Extract managed resources from OCM layer (`ModuleTemplate`)
    - [x] While creating Manifest CRs insert extracted managed resources into the `manifest.spec. associatedResources `
- [x] Cover newly introduced functions in Unit-Tests
- [x] Cover new functionality of Kyma reconciliation loop in integration test

nesmabadr:
**Description**

- Populate AssociatedResources field in the Manifest CR
- Add unit and integration tests

**Related issue(s)**
Resolves #1651


pkg/module/parse/template_to_module.go
```diff
@@ -81,7 +92,7 @@ func (p *Parser) GenerateMandatoryModulesFromTemplates(ctx context.Context,
 	return modules
 }
 
-func (p *Parser) appendModuleWithInformation(module v1beta2.AvailableModule, kyma *v1beta2.Kyma,
+func (p *Parser) appendModuleWithInformation(ctx context.Context, module v1beta2.AvailableModule, kyma *v1beta2.Kyma,
```
lindnerby, line 95: Let's refactor this part a bit!
The reason to pass modules list as arg here is to append to it. Appending can be done by consumer, let's change the result of this function to return only the module. and split responsibilities. Then this func also needs a new name.

Also we may try to move all the logic from `p.newManifestFromTemplate()` into a separate dependency, let's say a `ManifestCreator` that may be tested separately.

pkg/module/parse/template_to_module.go
```diff
@@ -241,3 +263,54 @@ func insertLayerIntoManifest(
 
 	return nil
 }
+
+func (p *Parser) fetchAssociatedResourcesField(ctx context.Context, layer *img.OCI, version string) ([]string,
+	error,
+) {
+	associatedResourcesLayer := v1beta2.ImageSpec{
```
lindnerby, line 270: Creating this seems a bit redundant since the only value used for a dependency is `layer.CredSecretSelector` for the `LookupKeyChain()`

lindnerby (changes requested):

Issue #1568, opened 2024-05-21
[Deletion Modes] Implement module resources deletion logic


nesmabadr:
### Description

In order to support https://github.com/kyma-project/lifecycle-manager/issues/1211, we need to adapt our current module deletion logic in the declarative library to wait until all managed resources get deleted before the deletion of the other module resources as well as the Manifest. 

We should rely on the newly introduced associatedResources field in the Manifest spec, if the manifest is marked for deletion, we list any resources defined in the associatedResources in SKR cluster and wait for their deletion first before proceeding.

### Reasons

To support the new blocking deletion mode and the unmanaged modules

### Acceptance Criteria

- [ ] Implement changes required in the declarative reconciler
     - [ ] If the module is marked for deletion, then wait for the associatedResource deletion before performing the deletion of resources and Manifest. I.e., for all CRDs listed in `associatedResources `, we check that no CRs are existing for those CRDs. The CRDs themselves may still exist.
     - [ ] As long as CRs are still existing, then Manifest CR should be in `Deleting` State
 - [ ] Implement E2E tests

### Feature Testing

End-to-End tests, Unit tests, Integration tests

### Testing approach

E2E Test:

Given SKR Cluster
When Kyma Module is enabled on SKR Kyma CR with Managed set to true
Then SKR Kyma CR is in Ready State 
And Manifest CR is in Ready State
When Kyma Module is disabled from Kyma CR 
Then all modules resources from the synced field in Manifest are in SKR Cluster
And Manifest CR is in a "Deleting" State
When associatedResources are manually deleted from SKR Cluster
Then all other module resources are removed from SKR Cluster
And Manifest is removed from KCP Cluster
And Module is no longer in Kyma CR Status

c-pius:
> We should rely on the newly introduced managedResources field in the Manifest spec, if the manifest is marked for deletion, we list any resources defined in the managedResources in SKR cluster and wait for their deletion first before proceeding.

@kyma-project/jellyfish Sorry to open this up again... but do we have any written statement that the managedResources over all modules are mutually exclusive? Going through https://github.com/kyma-project/community/issues/870 and https://raw.githubusercontent.com/kyma-project/community-modules/main/model.json didn't make this obvious to me. Just wanting to be sure since if they are not mutually exclusive, this would not be considered in the blocking deletion implementation.

PR #1676, opened 2024-07-10
refactor: Simplify declarative reconciler

lindnerby:
**Description**

Cleanup before adding to deletion behavior.

- Delete obsolete fields from Options
- Use Manifest type directly instead of custom Object interface, remove type casts
- Simplify SpecResolver

**Related issue(s)**
* https://github.com/kyma-project/lifecycle-manager/issues/1568
* https://github.com/kyma-project/lifecycle-manager/issues/1006


internal/declarative/v2/inmemory_rendered.go
```diff
@@ -62,6 +62,5 @@ func (c *InMemoryManifestCache) Parse(spec *Spec,
 }
 
 func generateCacheKey(spec *Spec) string {
-	file := filepath.Join(ManifestFilePrefix, spec.Path, spec.ManifestName)
-	return fmt.Sprintf("%s-%s", file, spec.Mode)
+	return filepath.Join(ManifestFilePrefix, spec.Path, spec.ManifestName) + "-raw"
```
ruanxin, line 65: let's get rid of this magic `-raw` suffix, if there is no other `spec.Mode`, then it does not bring any value to contribute cache key.

lindnerby: Sure, in the sense of true refactoring I didn't want to change the behavior at all. If you think it's fine to remove, will do.

ruanxin: leave it like this will make it more difficult to understand in the future, no one knows what's the meaning of this `-raw`. If you want to keep the existing implementation, then it's better we also refactor it to following best practice, e.g: define a const value with a meaningful name to indicate what it means.

However, it's just a cache key, a fixed suffix not brings any difference, I mean you can't messed up while generate new cache. And all existing cache value will be anyway evicted after new release. So you don't need to care about compatible problem here.

ruanxin (approved):

PR #1690, opened 2024-07-17
chore: Configure different requeue intervals for Manifest reconciliation

### LINKED ISSUE #1684: Config different requeue interval based on Manifest CR state
### Description

atm, in Manifest CR controller, in the happy path scenario,  the reconcile will requeue with success interval (5m): 
https://github.com/kyma-project/lifecycle-manager/blob/46d1e4b685fc1bfccf0240ceb19e38ba858c64e7/internal/declarative/v2/reconciler.go#L655

This frequency is acceptbale by user when Manifest CR is in ready state, however, when Manifest CR is in different state, say `processing`, `error`, as a user, often expect quick state change, that's the reason why we design different requeue interval in Kyma CR controller.

This Issue is aiming to apply same pattern to Manifest CR controller to improve user expierence.

### Reasons

Improve user experience.

### Acceptance Criteria

- [ ] Define additional requeue interval for Manifest controller: `Busy`, `Warning`, `Error`
- [ ] Reuse `DetermineRequeueInterval`: 
https://github.com/kyma-project/lifecycle-manager/blob/4f97326af4a7d3f42341fab3569628e3a0517b8c/pkg/queue/requeue_intervals.go#L16-L31

### Feature Testing

Unit tests

### Testing approach

_No response_

### Attachments

_No response_

nesmabadr:
**Description**

- Configure different requeue intervals for Manifest reconciliation
- Add unit test for the newly added flags

**Related issue(s)**
Resolves #1684


internal/declarative/v2/reconciler.go
```diff
@@ -219,7 +220,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	return r.finishReconcile(ctx, manifest, metrics.ManifestReconcileFinished, manifestStatus, nil)
 }
 
-func (r *Reconciler) cleanupManifest(ctx context.Context, req ctrl.Request, manifest *v1beta2.Manifest, manifestStatus shared.Status,
+func (r *Reconciler) cleanupManifest(ctx context.Context, req ctrl.Request, manifest *v1beta2.Manifest,
+	manifestStatus shared.Status,
```
lindnerby, line 224: All through this file there are random newline changes introduced. Can you align them to either all params being in one line or each param in a separate line?

pkg/queue/requeue_intervals.go
```diff
@@ -21,10 +21,10 @@ func DetermineRequeueInterval(state shared.State, intervals RequeueIntervals) ti
 		fallthrough
 	case shared.StateProcessing:
 		return intervals.Busy
-	case shared.StateReady:
-		fallthrough
 	case shared.StateWarning:
 		return intervals.Warning
+	case shared.StateReady:
+		fallthrough
```
lindnerby, line 27: Can you instead of using `fallthrough` here, either return the interval value that is assigned to the state directly or remove the case of that specific state where is anyways handled by the default case then?
Usually `fallthrough` is used to add a piece of logic to the specific case before the fallthrough case, other than that  it's not really needed imo.

lindnerby (changes requested):
Just two cosmetic remarks! PTAL

lindnerby (approved):

PR #1695, opened 2024-07-17
chore: Refactor NewCachedDescriptorProvider

ruanxin:
remove DescriptorCache as input paremeter

### DIFF
--- a/internal/descriptor/provider/provider_test.go
+++ b/internal/descriptor/provider/provider_test.go
@@ -14,7 +14,7 @@ import (
 )
 
 func TestGetDescriptor_OnEmptySpec_ReturnsErrDecode(t *testing.T) {
-	descriptorProvider := provider.NewCachedDescriptorProvider() // assuming it handles nil cache internally
+	descriptorProvider := provider.NewCachedDescriptorProvider()
 	template := &v1beta2.ModuleTemplate{}
 
 	_, err := descriptorProvider.GetDescriptor(template)
--- a/internal/descriptor/provider/provider_test.go
+++ b/internal/descriptor/provider/provider_test.go
@@ -14,7 +14,7 @@ import (
 )
 
 func TestGetDescriptor_OnEmptySpec_ReturnsErrDecode(t *testing.T) {
-	descriptorProvider := provider.NewCachedDescriptorProvider(nil) // assuming it handles nil cache internally
+	descriptorProvider := provider.NewCachedDescriptorProvider() // assuming it handles nil cache internally
 	template := &v1beta2.ModuleTemplate{}
 
 	_, err := descriptorProvider.GetDescriptor(template)
@@ -24,7 +24,7 @@ func TestGetDescriptor_OnEmptySpec_ReturnsErrDecode(t *testing.T) {
 }
 
 func TestAdd_OnNilTemplate_ReturnsErrTemplateNil(t *testing.T) {
-	descriptorProvider := provider.NewCachedDescriptorProvider(nil)
+	descriptorProvider := provider.NewCachedDescriptorProvider()
 
 	err := descriptorProvider.Add(nil)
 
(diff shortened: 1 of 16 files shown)


internal/descriptor/provider/provider_test.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 func TestGetDescriptor_OnEmptySpec_ReturnsErrDecode(t *testing.T) {
-	descriptorProvider := provider.NewCachedDescriptorProvider(nil) // assuming it handles nil cache internally
+	descriptorProvider := provider.NewCachedDescriptorProvider() // assuming it handles nil cache internally
```
nesmabadr, line 17: Please remove this comment, it is not longer applicable

mmitoraj (approved):

nesmabadr (approved):

mmitoraj (approved):

lindnerby (changes requested):
Can you explain what is the motivation  to reduce the coverage, remove encapsulation and promote modification of the descriptor cache directly? 
I don't see a reason for it. You could introduce a specific constructor function instead, so we have one with auto generated cache and one for cache injection for the tests.

ruanxin:
> Can you explain what is the motivation to reduce the coverage, remove encapsulation and promote modification of the descriptor cache directly? I don't see a reason for it. You could introduce a specific constructor function instead, so we have one with auto generated cache and one for cache injection for the tests.

the reason is by previous coverage, it fails the test coverage check (from 68 to 66.7) of `provider_test.go`, https://github.com/kyma-project/lifecycle-manager/actions/runs/9980148681/job/27580851622#step:6:278, and by this refactoring, it's not changing(adding or remove) any existing behaviour of this `CachedDescriptorProvider`, also spend time to add more test for this to get a higher coverage also not in this scope of this refactoring task. 

With the existing test cases and anonymous function constructor, IMO, it should be configured with current percentage: `66`.

I don't get the reason why needs to inject a cache to this `CachedDescriptorProvider`, you already tested the `GetDescriptor` as well as `Add`, what's the purpose of injecting a cache? If you need to verify the content inside the cache, we can use `provider.DescriptorCache` directly.

ruanxin:
I think we are mixing several concept here. The initial reason for this refactoring is just trying to get rid of initializing this provider with nil parameter `NewCachedDescriptorProvider(nil)`, currently spreading in the production code, which is meaningless.

Regarding the answer you provided that you want to test Provider with different cache, I don't see this in your original code tbh (that's why I don't get the reason in the beginning). If you want to achieve this goal, why give `*cache.DescriptorCache` as a member, it's a `struct`, the only thing you can provide is a single implementation which is [here](https://github.com/kyma-project/lifecycle-manager/blob/d6c42e009612bd729131af18f9ac1dfab57cb34b/internal/descriptor/cache/cache.go#L9), it's not an `interface`.
```
type CachedDescriptorProvider struct {
	DescriptorCache *cache.DescriptorCache
}

``` 
Then I would design like this below:
```
type DescriptorCache interface {
	Get(key DescriptorKey) *v1beta2.Descriptor
	Set(key DescriptorKey, value *v1beta2.Descriptor)
}

type CachedDescriptorProvider struct {
	descriptorCache DescriptorCache
}
```
I would also design a dedicated constructor for inject different `DescriptorCache` implementation, since currently on the production code, there is no other implementation, if really in the production code, we relies on different cache implementation, then it make sense get rid of the one without parameter.
```
// this is default which used in production
func NewCachedDescriptorProvider() *CachedDescriptorProvider
// this is for providing different implemention, probably useful for testing purpose
func NewCachedDescriptorProviderForTest(descriptorCache DescriptorCache) *CachedDescriptorProvider
```

Secondly, regarding address encapsulate of the `descriptorCache` and consider this as an internal member of this Provider, which I agree with also, then I would refactor this Provider to offer an additional function `GetDescriptorByCacheKey()` to make this function work  https://github.com/kyma-project/lifecycle-manager/pull/1695/files#diff-36cc16a1916ddd61fd387154219e8a413c14e5cff7cb6c0839016cff7140ef62R227, but the reason way I think it's ok to public this member is I don't think we as developer is stupid enough to directly assign a different `DescriptorCache` to this provider after initlized. And regarding chain of get calls, it's also no harm here, the chain call for this provider is basically `provider.DescriptorCache.Get(key)`, `provider.DescriptorCache.Set()`, I see certain valid usage of them, you can of cause argue that we design different method to encapsulate them.
 
last but not least, I think we should apply the design principle which aligned with current state of the code, also consider the trade off, take your Law of Demeter as an example, it also brings [disadvantage](https://en.wikipedia.org/wiki/Law_of_Demeter#Disadvantages) if we take it strictly. So I think we should not use some extreme examples to influence the current state.

PR #1698, opened 2024-07-18
feat: Support "application/x-tar" media type in the module template

### LINKED ISSUE #1650: [Deletion Modes] Make KLM support `application/x-tar` as access mediaType for `raw-manifest` and `default-cr` resource
In order to make [Kyma Module as OCM component](https://github.com/kyma-project/lifecycle-manager/issues/1587) compatible with [ocm-controller](https://github.com/open-component-model/ocm-controller), the `default-cr` and `raw-manifest` resources in Kyma Module should be defined as `application/x-tar` mediaType (as Directory). 

Currently KLM is possible to parsing `raw-manifest` as `application/octet-stream` mediaType (as file), however it is not possible to deploy by ocm-controller.
https://github.com/kyma-project/lifecycle-manager/blob/a0c49436f3d11d03c9a7556ec11c7c9f69d621d9/tests/moduletemplates/moduletemplate_template_operator_v2_fast.yaml#L34-L48

new kyma module ocm component looks like below:
```
resources:
  - access:
      imageReference: europe-docker.pkg.dev/kyma-project/prod/template-operator:1.0.0
      type: ociArtifact
    digest:
      hashAlgorithm: SHA-256
      normalisationAlgorithm: ociArtifactDigest/v1
      value: 03a194e1dca2421755cec5ec1e946de744407e6e1ca3b671f715fee939e8d1fb
    name: module-image
    relation: external
    type: ociArtifact
    version: 1.0.0
  - access:
      localReference: sha256:ad76888e8aee1c337eed330443cae54b4f912824e9c17c0f876e48d12f8ad65d
      mediaType: application/x-tar
      type: localBlob
    digest:
      hashAlgorithm: SHA-256
      normalisationAlgorithm: genericBlobDigest/v1
      value: ad76888e8aee1c337eed330443cae54b4f912824e9c17c0f876e48d12f8ad65d
    name: raw-manifest
    relation: local
    type: directory
    version: 1.0.0
  - access:
      localReference: sha256:7b747b3bbc6504c3198c52c764508f0217a39f44cb4ee52ed5f67b3b0fb6df9e
      mediaType: application/x-tar
      type: localBlob
    digest:
      hashAlgorithm: SHA-256
      normalisationAlgorithm: genericBlobDigest/v1
      value: 7b747b3bbc6504c3198c52c764508f0217a39f44cb4ee52ed5f67b3b0fb6df9e
    name: default-cr
    relation: local
    type: directory
    version: 1.0.0
  sources: []
  version: 1.0.0
```

## AC
- [x] **Manifest Controller:** KLM is able to deploy `raw-manifest` resource from access mediaType `application/x-tar` and `application/octet-stream` (for backward compatibility)
- [x] (❌ ) **Kyma Controller:** Add that KLM is able to deploy `default-cr` resource from access mediaType `application/x-tar` and `moduletemplate.spec.data` (for backward compatibility) during Manifest CR creation
- [x] Cover with integration test
    - [x] Hint: https://github.com/kyma-project/lifecycle-manager/blob/93a3385641053bfbe04ca6261ba4d14a2af25f2c/tests/integration/controller/manifest/manifest_controller_test.go#L17
- [x] Adapt existing E2E test if necessary

## AC updated:
- [ ] Default CR keep in ModuleTemplate `moduletemplate.spec.data`, so that KLM is able to deploy it directly, but not read from `default-cr` layer.

LeelaChacha:
closes #1650


ruanxin (changes requested):
We need to have an e2e test which can deploy module based on this new ModuleTemplate format `https://github.com/kyma-project/lifecycle-manager/blob/feat/deletion-mode/tests/moduletemplates/moduletemplate_template_operator_regular_new_ocm.yaml`

api/v1beta2/moduletemplate_types.go
```diff
@@ -69,6 +69,13 @@ func (d *Descriptor) DeepCopyObject() machineryruntime.Object {
 	return &Descriptor{ComponentDescriptor: d.Copy()}
 }
 
+type MediaTypeMetadata string
+
+const (
+	MediaTypeFile MediaTypeMetadata = "application/octet-stream"
```
ruanxin, line 75: These two mediatype you'd better import directly from ocm dependencies, to keep consistency.
it's here https://github.com/open-component-model/ocm/blob/releases/v0.12.0/pkg/mime/types.go#L11-L24

config/samples/component-integration-installed/operator_v1beta2_moduletemplate_kcp-module-cred-label.yaml
```diff
@@ -46,7 +46,7 @@ spec:
             - name: oci-registry-cred
               value:
                 operator.kyma-project.io/oci-registry-cred: operator-regcred
-          name: template-operator
+          name: raw-manifest
```
ruanxin, line 49: you can't directly change the content here, this component descriptor is just a local copy, KLM download the actually content from remote oci registry.

````
ocm get resource europe-west3-docker.pkg.dev/sap-kyma-jellyfish-dev/template-operator//kyma-project.io/template-operator -c 1.6.0 -owide
NAME              VERSION IDENTITY TYPE       RELATION ACCESSTYPE   ACCESSSPEC
config            v1.6.0           yaml       local    localOciBlob {"digest":"sha256:db86408caca4c94250d8291aa79655b84146f9cc45e0da49f05a52b3722d74a0"}
template-operator v1.6.0           helm-chart local    localOciBlob {"digest":"sha256:8f01a545b7f53f91dd9f0bb7b74b30dfa2e571dae8d38fd0f56af66f809378fd"}
````
 
This file should only used for the integration test [here](https://github.com/kyma-project/lifecycle-manager/blob/d51230b9959a0b8a9ae592684e3406f5244909ff/tests/integration/controller/kyma/moduletemplate_test.go#L128).

I assume this outdated file should not affect the change you did, please create a followup issue to update this outdated moduletemplate.

File: internal/manifest/parse.go [hunk 3]
```diff
@@ -33,14 +40,38 @@ func NewPathExtractor(cache *filemutex.MutexCache) *PathExtractor {
...
 
-func (p PathExtractor) GetPathFromRawManifest(ctx context.Context,
+func (p PathExtractor) FetchLayerToFile(ctx context.Context,
+	imageSpec v1beta2.ImageSpec,
+	keyChain authn.Keychain,
+	layerName string,
+) (string, error) {
+	switch imageSpec.Type {
+	case v1beta2.OciRefType:
+		return p.getPathForFetchedLayer(ctx, imageSpec, keyChain, v1beta2.RawManifestLayerName+".yaml")
```
ruanxin, line 50: why you not use `layerName` here?

pkg/img/layer.go
```diff
@@ -10,8 +10,10 @@ import (
 type LayerName string
 
 const (
-	ConfigLayer LayerName = "config"
-	CRDsLayer   LayerName = "crds"
+	ConfigLayer      LayerName = "config"
+	CRDsLayer        LayerName = "crds"
+	DefaultCRLayer   LayerName = "default-cr"
+	RawManifestLayer LayerName = "raw-manifest"
```
ruanxin, line 16: this one should relocated to `api` module, and this rawmanifestlayer is already defined as `v1beta2.RawManifestLayerName`

internal/manifest/parse.go
```diff
@@ -33,14 +40,38 @@ func NewPathExtractor(cache *filemutex.MutexCache) *PathExtractor {
...
+	case v1beta2.OciDirType:
+		tarFile, err := p.getPathForFetchedLayer(ctx, imageSpec, keyChain, layerName+".tar")
```
ruanxin, line 52: This logic here should be improved, you should not always decompress the content, instead:
1. download the layer as `raw-manifest.tar`, but only once, this is already done in `getPathForFetchedLayer`
2. check if the decompressed file `raw-manifest.yaml` exits, if exists, return
3. if not exists, decompress the file to a folder

internal/manifest/parse.go
```diff
@@ -115,3 +146,60 @@ func pullLayer(ctx context.Context, imageRef string, keyChain authn.Keychain) (c
 func getFsChartPath(imageSpec v1beta2.ImageSpec) string {
 	return filepath.Join(os.TempDir(), fmt.Sprintf("%s-%s", imageSpec.Name, imageSpec.Ref))
 }
+
+func untarLayer(tarPath string) (string, error) {
```
ruanxin, line 150: This logic should be improved, you need to consider concurrency here, it will be multiple threads trying to read and uncompress the file, we configured multiple Manifest workers.

Please reference how the `getPathForFetchedLayer` is implemented, use `filemutex.MutexCache` could help in your case as well.

internal/manifest/parse.go
```diff
@@ -115,3 +146,60 @@ func pullLayer(ctx context.Context, imageRef string, keyChain authn.Keychain) (c
...
+	for {
+		header, err := tarReader.Next()
+		if errors.Is(err, io.EOF) {
+			break
+		}
+		if err != nil {
+			return "", fmt.Errorf("failed to read tar: %w", err)
+		}
+
+		if strings.HasPrefix(header.Name, "._") {
```
ruanxin, line 168: what's the purpose of this check?

internal/manifest/parse.go
```diff
@@ -115,3 +146,60 @@ func pullLayer(ctx context.Context, imageRef string, keyChain authn.Keychain) (c
...
+
+		if _, err := os.Stat(extractedFilePath); err == nil {
+			return extractedFilePath, nil
+		}
+
+		outFile, err := os.Create(extractedFilePath)
+		if err != nil {
+			return "", fmt.Errorf("failed to create extracted file: %w", err)
+		}
+		defer outFile.Close()
```
ruanxin, line 185: use defer inside for loop is not the best practice. `Possible resource leak, 'defer' is called in the 'for' loop `, instead you can simply put this file close in the end before return extractedFilePath

tests/integration/controller/manifest/manifest_controller_test.go
```diff
@@ -155,6 +163,43 @@ var _ = Describe(
...
+						serverAddress, false, false), standardTimeout, standardInterval).
+						WithArguments(manifest).
+						Should(Succeed())
+					By("Then Manifest CR is in Ready State", func() {
+						Eventually(testutils.ExpectManifestStateIn(ctx, kcpClient, shared.StateReady),
+							standardTimeout, standardInterval).
+							WithArguments(manifest.GetName()).
+							Should(Succeed())
+					})
+					By("And OCI-Sync-Ref Annotation exists", func() {
```
ruanxin, line 188: Please don't introduce this `"sync-oci-ref"` check, it's annotation will deprecated soon.

internal/manifest/parse.go
```diff
@@ -20,7 +22,12 @@ import (
 	"github.com/kyma-project/lifecycle-manager/pkg/ocmextensions"
 )
 
-var ErrImageLayerPull = errors.New("failed to pull layer")
+var (
+	ErrImageLayerPull       = errors.New("failed to pull layer")
+	ErrInvalidImageSpecType = errors.New("invalid image spec type provided, only 'oci-ref' 'oci-dir' are allowed")
```
ruanxin, line 27: use you defined RefTypeMetadata for `oci-dir`, `oci-ref` .

internal/manifest/parse.go
```diff
@@ -115,3 +146,60 @@ func pullLayer(ctx context.Context, imageRef string, keyChain authn.Keychain) (c
...
+		extractedFilePath, err := sanitizeArchivePath(filepath.Dir(tarPath), header.Name)
+		if err != nil {
+			return "", fmt.Errorf("failed to sanitize archive path: %w", err)
+		}
+
+		if _, err := os.Stat(extractedFilePath); err == nil {
+			return extractedFilePath, nil
+		}
+
+		outFile, err := os.Create(extractedFilePath)
```
ruanxin, line 181: I think here you'd better check the file type, if it's a folder, should return with error. check the switch case here: https://medium.com/@skdomino/taring-untaring-files-in-go-6b07cf56bc07

```
// check the file type
		switch header.Typeflag {
```

internal/manifest/parse.go
```diff
@@ -115,3 +146,60 @@ func pullLayer(ctx context.Context, imageRef string, keyChain authn.Keychain) (c
...
+		if err != nil {
+			return "", fmt.Errorf("failed to create extracted file: %w", err)
+		}
+		defer outFile.Close()
+
+		var maxBytes int64 = 1024 * 1024
+		if _, err := io.CopyN(outFile, tarReader, maxBytes); err != nil && !errors.Is(err, io.EOF) {
```
ruanxin, line 188: why not use `io.Copy`?

internal/manifest/parse.go
ruanxin: Also, please write a test for this function, what's expected for this test:
1. use your `rendered.tar` as input,
2. initailze several goruntine, execute this untarLayer function.
3. verify the file:
   1.  only created once
   2. the completeness of the content in the file
   3. no modification after created 
 

I think you can use `FileInfo.ModTime` to verify 3.iii, it's returned by `io.OpenFile`
```
type FileInfo interface {
	Name() string       // base name of the file
	Size() int64        // length in bytes for regular files; system-dependent for others
	Mode() FileMode     // file mode bits
	ModTime() time.Time // modification time
	IsDir() bool        // abbreviation for Mode().IsDir()
	Sys() any           // underlying data source (can return nil)
}
```

internal/manifest/parse.go
LeelaChacha: That is what I did first but a linter had a problem with this - since zip bombs exist, a file might have a small compressed size but huge size when uncompressed so the bytes must be limited

config/samples/component-integration-installed/operator_v1beta2_moduletemplate_kcp-module-cred-label.yaml
LeelaChacha: this does affect the new changes since we switch on the layer name, and default case is not do anything. And because "template-operator" layer is not called "raw-manifest", it is not rendered by the new changes

LeelaChacha: the local copy is indeed used here: https://github.com/kyma-project/lifecycle-manager/blob/2dcdb77176b7192743822204f50e6cb77b3df96e/tests/integration/controller/kyma/moduletemplate_test.go#L128

internal/manifest/parse.go
ruanxin: But then it means the content in this tar file already invalid, we don't need to tolerate this misconfigured tar file, please don't introduce such hacking, instead, we should prevent it while generating this layer.

config/samples/component-integration-installed/operator_v1beta2_moduletemplate_kcp-module-cred-label.yaml
ruanxin: ok, but then we still need to update remote content.

internal/manifest/parse_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/stretchr/testify/assert"
+	"os"
+	"path/filepath"
+	"sync"
+	"testing"
+	"time"
+)
+
+func TestPathExtractor_untarLayer(t *testing.T) {
+	tempDir, err := os.MkdirTemp("", "untar-test")
```
ruanxin, line 15: Please think about extracting the following code into function, like `createTar` or something similar.

internal/manifest/parse_test.go
```diff
@@ -0,0 +1,80 @@
...
+		}(i)
+	}
+	wg.Wait()
+
+	assert.NoError(t, resultErr)
+	assert.NotEmpty(t, results)
+	assert.NotEmpty(t, modTimes)
+
+	for i := 1; i < numGoroutines; i++ {
+		assert.Equal(t, results[0], results[i])
```
ruanxin, line 73: this verification can be potentially false positive, it could possible that your go runtine not assign any value to `results[0]` or `modTimes[0]`, which means all of them are kept in the initial state.

Please use the fileInfo returned by `os.Stat` to verify all results and modTimes

internal/manifest/parse.go
ruanxin: but CopyN only copy the maxBytes, you can't make sure the file get fully copied.

internal/manifest/parse.go
ruanxin: I left a comment regarding this, it should only affect in Mac OS

lindnerby (commented):
All good. Just some small suggestions.

internal/manifest/parser/template_to_module.go
```diff
@@ -210,25 +206,16 @@ func translateLayersAndMergeIntoManifest(
...
-		if !ok {
-			return fmt.Errorf("%w: not an OCIImage", ErrDefaultConfigParsing)
-		}
-		manifest.Spec.Config = &v1beta2.ImageSpec{
-			Repo:               ociImage.Repo,
-			Name:               ociImage.Name,
-			Ref:                ociImage.Ref,
-			Type:               v1beta2.OciRefType,
-			CredSecretSelector: ociImage.CredSecretSelector,
+	case v1beta2.DefaultCRLayer:
```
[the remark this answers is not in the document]
ruanxin: all cases should be presented, otherwise, lint will not happy.

internal/manifest/img/parse.go
```diff
@@ -111,19 +112,26 @@ func parseLayersByName(repo *genericocireg.RepositorySpec, descriptor *compdesc.
...
 	labels ocmmetav1.Labels,
 ) (*OCI, error) {
-	layerRef := OCI{
-		Type: string(v1beta2.OciRefType),
+	layerRef := OCI{}
+
+	switch accessSpec.MediaType {
+	case mime.MIME_TAR:
+		layerRef.Type = string(v1beta2.OciDirType)
+	case mime.MIME_OCTET:
```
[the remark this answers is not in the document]
ruanxin: changed to if condition

internal/manifest/img/layer.go
```diff
@@ -41,9 +39,23 @@ func (o *OCI) String() string {
...
 	Layer     struct {
-		LayerName
+		v1beta2.LayerName
 		LayerRepresentation
 	}
 )
 
 type Layers []Layer
+
+func (l Layer) ConvertToImageSpec() (*v1beta2.ImageSpec, error) {
```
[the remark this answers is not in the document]
ruanxin: this function here is actually for compatible purposes, the logic is basically just casting from `OCI` to `ImageSpec`, actually they are same, the thing we should think about is probabaly drop this `Layer` struct, I will think about follow up issue, but not covered here. Besides this function is used in `pathextractor_test.go`, it's partially tested.

tests/integration/controller/manifest/manifest_controller_test.go
```diff
@@ -14,9 +16,17 @@ import (
...
 
-func setupTestEnvironment(ociTempDir, installName string) {
+func setupTestEnvironment(ociTempDir, installName, mediaType string) {
 	It("setup OCI", func() {
-		err := testutils.PushToRemoteOCIRegistry(server, manifestFilePath, installName)
+		var err error
+		switch mediaType {
+		case mime.MIME_TAR:
+			err = testutils.PushToRemoteOCIRegistry(server, manifestTarPath, installName)
+		case mime.MIME_OCTET:
```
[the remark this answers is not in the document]
ruanxin: changed to if condition

internal/manifest/spec_resolver.go
```diff
@@ -5,22 +5,21 @@ import (
...
 	"sigs.k8s.io/yaml"
 
 	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
 	declarativev2 "github.com/kyma-project/lifecycle-manager/internal/declarative/v2"
-	"github.com/kyma-project/lifecycle-manager/pkg/ocmextensions"
+	"github.com/kyma-project/lifecycle-manager/internal/manifest/auth"
+	"github.com/kyma-project/lifecycle-manager/internal/manifest/img"
 )
 
 type SpecResolver struct {
```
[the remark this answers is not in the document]
ruanxin: Honestly, test this function is not so useful, basically, the critical logic is `manifestPathExtractor.FetchLayerToFile`, which have test covered, the rest is just copy some fields to `&declarativev2.Spec`

But for the purpose to make code testable, I created a followup issue to address it: https://github.com/kyma-project/lifecycle-manager/issues/1781

lindnerby (approved):

ruanxin (approved):

Issue #1590, opened 2024-05-31
[Module catalogue improvements] Update ModuleTemplate API version to support fixed module versions


Tomasz-Smelcerz-SAP:
### Description

**Note: This issue is just about API change, the change in the controller logic is expected to be implemented [here](https://github.com/kyma-project/lifecycle-manager/issues/1591)**

As described [here](https://github.com/kyma-project/lifecycle-manager/issues/1472), ModuleTemplate CR must be extended.

What we essentially want follows the ideas from the [community-modules](https://raw.githubusercontent.com/kyma-project/community-modules/4067085235bdb3519a9f593b9c473106471e200e/model.json) repository.
In general, we want that:
- Every ModuleTemplate describes a single module version
- Every such version may be selected by user(s) for installation
- Every such version may be assigned to a set of channels - possibly empty. Module Team decides about versions-channels mapping.

In addition, to support [Deletion modes](https://github.com/kyma-project/lifecycle-manager/issues/1211) we need a `managedResources` list describing the list of Group/Version/Kinds that are defined by the Module's Team as "managed" by the module.
There is a [decision record for that](https://github.com/kyma-project/lifecycle-manager/issues/1587). Assuming that list is embedded, no further API changes are required on the ModuleTemplate CR. 

The set of ModuleTemplate CRs in the control-plane defines the Module Catalog for the Lifecycle-Manager.

The above requirements can be implemented by releasing a new ModuleTemplate API version with the following changes:
- Deprecate the `channel` attribute
- Add an optional `version` attribute. At the moment it is an annotation: `"operator.kyma-project.io/module-version": "0.1.2"`. This attribute will be required in the future
- Add an optional `name` attribute. At the moment it is a label: `"operator.kyma-project.io/module-name": "template-operator"`. It is not strictly required, it's just to make the API more robust. It will be required in the future.

### Reasons

We need that to support features described in #1472 

### The Data Model

Example: Consider a module: `foo`.
It will be released in two versions available for user to install: `1.0.0`, `2.0.0`.
Additionally there is a `2.2.0` version that is only available in the `experimental` channel.
The `1.0.0` is released in the `regular` channel, the `2.0.0` version is released in the `fast` channel

We need five ModuleTemplate CRs to model that:

1. ModuleTemplate for the version 1.0.0
```
name: "foo"
version: "1.0.0"
descriptor: [ OCM descriptor for version 1.0.0 ]
```

2. ModuleTemplate for the version 2.0.0
```
[...]

PR #1699, opened 2024-07-18
feat: Explicit version for module template

Tomasz-Smelcerz-SAP:
**Description**

Add explicit version and moduleName to the ModuleTemplateCR
- new attributes: version and moduleName
- CRD-level validations for new attributes
- integration tests for validations

**Related issue(s)**
#1590


api/v1beta2/moduletemplate_types.go
```diff
@@ -78,6 +78,18 @@ type ModuleTemplateSpec struct {
...
 
+	// Version identifies the version of the Module. It can be empty, or a semantic version.
+	// +optional
+	// +kubebuilder:validation:Pattern:=^((0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)?)?$
+	// +kubebuilder:validation:MaxLength:=32
+	Version string `json:"version"`
+
+	// ModuleName is the name of the Module. Can be empty.
+	// +optional
+	// +kubebuilder:validation:Pattern:=^([a-z][a-z][a-z]+(-[a-z][a-z][a-z]+)*)?$
```
c-pius, line 89: Maybe this alternative would be a bit simpler to read `^([a-z]{3,}(-[a-z]{3,})*)?$`

api/v1beta2/moduletemplate_types.go
```diff
@@ -78,6 +78,18 @@ type ModuleTemplateSpec struct {
 	// +kubebuilder:validation:MinLength:=3
 	Channel string `json:"channel"`
 
+	// Version identifies the version of the Module. It can be empty, or a semantic version.
```
c-pius, line 81:
```suggestion
	// Version identifies the version of the Module. Can be empty, or a semantic version.
```

To be aligned with the description of ModuleName?

c-pius (commented):
LGTM. One minor comment wrt. to the description. Smoke tests failed due to api quota for docker, should be green when re-running.

The API diff check noticed a difference. Do we need to add the field to the v1beta1 API for conversion between both or should we add the change to the exclusions list?

c-pius (approved):

Issue #1427, opened 2024-03-26
[Deletion Modes] Implement logic for "unmanaging" a Module


janmedrek:
### Description

There should be an option to unmanage a module. When unmanaged, the KLM deletes the module's ManifestCR and stops reconciling the module. However, the workload (module resources deployed to the SKR) is not affected.

> 

### Reasons

In some cases, users would like to have full control over their Modules' lifecycle. 

### Acceptance Criteria

- [x] When the user sets `.spec.modules[].managed: false`
  - [x] KLM deletes module's ManifestCR without affecting deployed resources
  - [x] module's workload (resources deployed to the SKR) remain to exist
  - [x] KLM does NOT reconcile the module's workload any longer
  - [x] KLM sets KymaCR `.status.state` and `.status.modules[]` to
    ```yaml
    status:
      state: Ready
      modules:
      - name: some-module
        state: Unmanaged
    ```
- [x] When the KymaCR is:
    ```yaml
    spec:
      modules:
      - name: some-module
        managed: false
    status:
      state: Ready
      modules:
      - name: some-module
        state: Unmanaged
    ```
    And the user removes the module completely from `spec.modules[]`
  - [x] KLM deletes module from KymaCR `.status.modules[]`
  - [x] module's workload (resources deployed to the SKR) remain to exist
  - [x] KLM does NOT reconcile the module's workload
  - [x] KLM sets KymaCR `.status.state` to `Ready`
- [x] Documentation clearly states that
  - [x] Before removing an entry from `.spec.modules[]` that has `managed: false` set, the related `.status.modules[].state` MUST be `Unmanaged`
  - [x] If the entry is removed before `.status.modules[].state` transitioned to `Unmanaged`, the behavior of the deletion is unpredictable (i.e., either behaves like deletion of a managed or unmanaged module) and may lead to erroneous states

### Feature Testing

End-to-End tests

### Testing approach

E2E Test:

  - Given a **Kyma Module** with managed set to true
    - And a **SKR Kyma CR** is created on the **SKR Cluster**
    - And a **Manifest CR** is created on the **KCP Cluster**
    - And - **Module Operator**, **Module CR**, and **[Module Managed Resources]** are created on the **SKR Cluster**
    - And **Kyma Status** is in a `Ready` State
  - When the **Managed** field is set to false for the **Kyma Module**
    - Then the **Kyma Module** is no longer reconciled
      - **Module Status** for that **Kyma Module** becomes **Unmanaged**
[...]

PR #1717, opened 2024-07-25
feat: Unmanage module

lindnerby:
**Description**

- If remote modules spec gets set managed to `false` remove manifest in control-plane, but keep resources
- Kyma module status set to new state "Unmanaged" and remove resource, template and manifest

**Related issue(s)**
* https://github.com/kyma-project/lifecycle-manager/issues/1427


internal/controller/watcher/controller.go
```diff
@@ -122,6 +122,8 @@ func (r *Reconciler) stateHandling(ctx context.Context, watcher *v1beta2.Watcher
 		return r.handleProcessingState(ctx, watcher)
 	case shared.StateReady, shared.StateWarning:
 		return r.handleProcessingState(ctx, watcher)
+	case shared.StateUnmanaged:
+		panic("unmanaged not a valid state for watcher")
```
ruanxin, line 126: same reason as [here](https://github.com/kyma-project/lifecycle-manager/pull/1717/files#r1735776926), btw, Kyma and Watcher CR are all controlled by us, no one will change the state to unmanaged manually, fallthrough this condition and stop requeue is enough already.

tests/e2e/module_unmanage_test.go
```diff
@@ -0,0 +1,94 @@
...
+				Should(Succeed())
+
+			By("And Module Operator Deployment is not removed on SKR cluster")
+			Consistently(CheckIfExists).
+				WithContext(ctx).
+				WithArguments(ModuleResourceName, TestModuleResourceNamespace,
+					"apps", "v1", "Deployment", skrClient).
+				Should(Succeed())
+
+			By("And Manifest CR is removed")
```
ruanxin, line 63: I would put this check after `SetModuleUnmanaged`, then check module CR is not removed and so on.

tests/e2e/module_unmanage_test.go
```diff
@@ -0,0 +1,94 @@
...
+		})
+
+		It("When Kyma Module is set to managed again", func() {
+			Eventually(SetModuleManaged).
+				WithContext(ctx).
+				WithArguments(skrClient, defaultRemoteKymaName, RemoteNamespace, module.Name).
+				Should(Succeed())
+		})
+
+		It("Then Manifest is re-created", func() {
```
ruanxin, line 83: I would at least add another check to make sure "And KCP Kyma CR is in \"Ready\" State" again

pkg/module/sync/runner.go
```diff
@@ -180,17 +188,21 @@ func (r *Runner) updateAvailableManifestSpec(ctx context.Context, manifestObj *v
...
 	}
-	if moduleStatus.Template != nil && moduleStatus.Template.GetGeneration() != moduleTemplateGeneration {
+	if moduleStatus.Template != nil && moduleStatus.Template.GetGeneration() != module.Template.GetGeneration() {
 		return true
 	}
 	return manifestObj.Spec.Version != moduleStatus.Version ||
 		manifestObj.Labels[shared.ChannelLabel] != moduleStatus.Channel ||
-		moduleStatus.State != manifestInCluster.Status.State
+		moduleStatus.State != manifestInCluster.Status.State ||
+		module.IsUnmanaged != manifestInCluster.IsUnmanaged()
```
ruanxin, line 205: I think unmanaged condition need to be treated separately, if `module.IsUnmanaged`, we only update manifest cr if it's not contains `UnmanagedAnnotation`, but doesn't update if other conditions are different.

internal/controller/watcher/controller.go
lindnerby: Yes, I wanted to indicate that this state should never happen and if it happens it's a developer error and not an error we should handle. Rather bubble it up as quickly as possible.

internal/controller/kyma/controller.go
```diff
@@ -292,6 +293,8 @@ func (r *Reconciler) processKymaState(ctx context.Context, kyma *v1beta2.Kyma) (
 		return r.handleProcessingState(ctx, kyma)
 	case shared.StateReady, shared.StateWarning:
 		return r.handleProcessingState(ctx, kyma)
+	case shared.StateUnmanaged:
+		panic("unmanaged not a valid state for kyma")
```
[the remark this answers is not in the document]
lindnerby: See other comment

tests/e2e/module_unmanage_test.go
lindnerby: Sorry I forgot to push the extension of the e2e tests. I added a lot more to this suite. PTAL

pkg/module/sync/runner_test.go
```diff
@@ -266,7 +314,43 @@ func TestNeedToUpdate(t *testing.T) {
...
+					},
+				}, &common.Module{
+					Template: &templatelookup.ModuleTemplateInfo{
+						ModuleTemplate: &v1beta2.ModuleTemplate{
+							ObjectMeta: apimetav1.ObjectMeta{
+								Generation: trackedModuleTemplateGeneration,
+							},
+						},
+					},
+					IsUnmanaged: true,
```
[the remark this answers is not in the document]
lindnerby: It should be the case inbetween, when the module is set to unmanaged but there still is a cluster manifest, which may not have the updated annotation yet, right? 🤔

pkg/testutils/kyma.go
```diff
@@ -173,6 +173,46 @@ func DisableModule(ctx context.Context, clnt client.Client,
...
+		}
+	}
+	err = clnt.Update(ctx, kyma)
+	if err != nil {
+		return fmt.Errorf("update kyma failed: %w", err)
+	}
+	return nil
+}
+
+func SetModuleManaged(ctx context.Context, clnt client.Client,
```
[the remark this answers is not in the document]
lindnerby: Changed the signature and remove the other.

internal/controller/watcher/controller.go
ruanxin: But one of the key aspect of error handling is reporting, this also aplies to fail fast principle, it expects that failures are detected early and reported immediately.  The current use of panic only stops the process from continuing but does not properly handle the reporting.

If the Kyma CR is in this state, using panic will not cause the KLM to crash (which is also a poor design choice in my opinion), but will only stop the current worker. This means that without checking the logs, it is difficult to identify that such an error exists.

As I mentioned [here](https://github.com/kyma-project/lifecycle-manager/pull/1717#discussion_r1735776926), a proper handling approach would be to write an error message in the Kyma CR and **prevent it from requeuing**. This would avoid interfering with other Kyma CRs and make the KLM more robust.

lindnerby: Alright! Will change to returning an error instead 👍

pkg/module/sync/runner.go
lindnerby: Can you check again? Current logic should be sufficient in my opinion?

ruanxin (approved):

PR #1722, opened 2024-07-26
chore: Limit access for remote and istio namespaces

### LINKED ISSUE #1613: Investigate need for `kyma-system` and `istio-system` RoleBindings to `klm-manager-role` ClusterRole
### Description

As of now we have three bindings to `klm-manager-role` ClusterRole for `kcp-system`, `kyma-system` and `istio-system` namespaces: https://github.com/kyma-project/lifecycle-manager/blob/main/config/rbac/namespace_bindings/role_binding.yaml

While testing the helm setup, we tried to remove the ones for `kyma-system` and `istio-system`, but this leads to errors like:
```
klm-controller-manager-7d846d6545-5hqq9 manager W0605 12:53:04.754801       1 reflector.go:539] pkg/mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: failed to list *v1beta2.Watcher: watchers.operator.kyma-project.io is forbidden: User "system:serviceaccount:kcp-system:klm-controller-manager" cannot list resource "watchers" in API group "operator.kyma-project.io" at the cluster scope
klm-controller-manager-7d846d6545-5hqq9 manager E0605 12:53:04.754828       1 reflector.go:147] pkg/mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: Failed to watch *v1beta2.Watcher: failed to list *v1beta2.Watcher: watchers.operator.kyma-project.io is forbidden: User "system:serviceaccount:kcp-system:klm-controller-manager" cannot list resource "watchers" in API group "operator.kyma-project.io" at the cluster scope
```

Those errors are surprising because KLM should be working in `kcp-system` namespace on KCP, and `kyma-system` namespace only on SKR which is accessed through the related kubeconfig file.

We need to find out why those bindings are needed and if it can be reduced to not need those anymore.

### Reasons

Keeping RBAC as restricted and clean as possible

### Acceptance Criteria

- [x] **Time-boxed** Research One day:
    - [x] Check what the current behaviour of KLM is if we remove corresponding RoleBindings
    - [x] Find the reasoning if KLM has errors
    - [x] Propose possible solutions to the team
- [ ] Create Follow-Up issue

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

nesmabadr:
**Description**

- Limit access for remote and istio namespaces

**Related issue(s)**
Resolves #1613


config/rbac/namespace_bindings/istio_namespace_role.yaml
```diff
@@ -0,0 +1,33 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
ruanxin, line 3: Why not define this as a `Role` only, all the resources is in `istio-namespace`, right?

config/rbac/namespace_bindings/remote_namespace_role.yaml
```diff
@@ -0,0 +1,48 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
ruanxin, line 3: Same here, if the granted resources is namespaced, we better define a `Role` instead of `ClusterRole`, any specific reason choose `ClusterRole` with `RoleBinding`?

ruanxin (changes requested):
Notes: we also need to update management plane charts repo accordingly.

config/rbac/namespace_bindings/remote_namespace_role.yaml
```diff
@@ -0,0 +1,48 @@
...
+metadata:
+  name: manager-role-remote-namespace
+rules:
+- apiGroups:
+    - ""
+  resources:
+    - secrets
```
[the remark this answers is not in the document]
nesmabadr: It was there in the cache_options, didn't notice that it is not required. Thanks!

config/rbac/namespace_bindings/istio_namespace_role.yaml
nesmabadr: yeah good point since it is already configured manually not using the kubebuilder markers

ruanxin (approved):

Issue #1591, opened 2024-05-31
[Module catalogue improvements] Implement new Module Catalog support


Tomasz-Smelcerz-SAP:
### Description

Implement new logic to support use cases described [here](https://github.com/kyma-project/lifecycle-manager/issues/1472)
The API changes are introduced by these two issues: https://github.com/kyma-project/lifecycle-manager/issues/1590, https://github.com/kyma-project/lifecycle-manager/issues/1589.

**Note: At this moment it is not 100% clear which scenarios related to module management must be supported. Precisely, what happens if:**
- A Module `foo` is installed from a channel `fast`. After a long time, the Module Catalog changes and there is no more ModuleTemplate for `foo` in the `fast` channel What happens then?

Such scenarios must be either excluded (it will not happen because...) or implemented as part of this issue.

### Reasons

We need it to support use cases described [here](https://github.com/kyma-project/lifecycle-manager/issues/1472)

### Acceptance Criteria

- [x] Resolve doubts related to Module Catalog changes (as described) - this requires a separate issue related to module sunset procedure. We do NOT implement it  here.
- [ ] The Lifecycle-Manager must support both the old ModuleTemplates and a new ones (with extended attributes) - we must provide a smooth migration path
- [ ] Installing an explicit module version is supported (this is new)
- [ ] Installing a module using a channel name is supported (this logic already exists)
- [ ] Automatic upgrade of a module installed using a channel name is supported (this logic already exists)
- [ ] No module downgrade is supported (based on semantic versioning)

### Feature Testing
unit test, integration test, e2e test

### Testing approach

Ensure in the e2e test that the old ModuleTemplate attributes (as of now) and the new ones (as implemented by this issue) are supported.

### Related issue(s)
#1472, https://github.com/kyma-project/lifecycle-manager/issues/1590

Issue #1589, opened 2024-05-31
[Module catalogue improvements ] Extend Kyma CR module list with an explicit version


Tomasz-Smelcerz-SAP:
### Description

**Note: This issue is just about API change, the change in the controller logic is expected to be implemented [here](https://github.com/kyma-project/lifecycle-manager/issues/1591)**

In order to support [explicit module version](https://github.com/kyma-project/lifecycle-manager/issues/1472), we need to add an optional `version` attribute to the  Kyma CR module list.

##### Use case 1 / existing: User wants to enable a module in their cluster with the default channel.
```
[...]
channel: regular
modules:
  - name: my-module
[...]
```

##### Use case 2 / existing: User wants to enable a specific channel of a module in their cluster.
```
[...]
modules:
  - name: my-module
    channel: fast
[...]
```

##### Use case 3 / new: User wants to enable a specific version of a module in their cluster.
```
[...]
modules:
  - name: my-module
    version: 1.0.0
[...]
```

### Reasons

We need this because users must be given an option to install a specific version of the module.
If the user wants to enable a specific module version, a ModuleTemplate CR corresponding to that version must exist in the control-plane.

### Acceptance Criteria

- [x] Add new optional `version` attribute is to the module list in Kyma CR, which must be a semantic version.
- [x] Add a validation: either `channel` or `version` is present - it is an error to provide both 
  - should be validated in the reconciliation loop and present an error in the state for the user
  - can open schema validation support such case? **No**
- [x] Add a validation that `channel` must not be set to `none`

#### Examples

```yaml
# KymaCR
spec:
  channel: regular
  modules:
    - name: template-operator
      version: 2.3.1
# => resolves to template-operator-2.3.1
---
# KymaCR
spec:
  channel: regular
  modules:
    - name: template-operator
# => resolves to template-operator-regular
---
# KymaCR
spec:
  channel: regular
  modules:
    - name: template-operator
      channel: fast
# => resolves to template-operator-fast
---
# KymaCR
spec:
  channel: regular
  modules:
    - name: template-operator
      channel: fast
      version: 2.3.1
# => ERROR: must not have both channel and version set
---
# KymaCR
spec:
  channel: regular
  modules:
    - name: template-operator
      channel: none
# => ERROR: must not use none as a value
```

### Feature Testing
e2e test for the validation. Can be done with local EnvTest.

### Related issue(s)
https://github.com/kyma-project/lifecycle-manager/issues/1590

[...]

PR #1748, opened 2024-08-07
feat: Module catalog improvements implementation

Tomasz-Smelcerz-SAP:
**Description**

Fixes https://github.com/kyma-project/lifecycle-manager/issues/1591

- add support for explicit module version handling
- ensure compatibility with existing logic
- extend tests

**Related issue(s)**
#1589 , #1590 

**Compatibility**
The following scenarios has been tested ([pkg/templatelookup/regular_test.go](https://github.com/kyma-project/lifecycle-manager/pull/1748/files#diff-e02f3b4956fd4ff60c0808f889a64ded170e5324ebf45fa04f6b7da0c221b940)):
- installation of a module by version, then upgrade
- installation of a module by version, then downgrade: Not allowed
- change from channel to version (no upgrade)
- change from channel to version (upgrade)
- change from channel to version (downgrade): Not allowed
- change from version to channel (no upgrade)
- change from version to channel (upgrade)
- change from version to channel (downgrade): Not allowed


mmitoraj (approved):

api/v1beta2/kyma_types.go
```diff
@@ -387,24 +387,42 @@ func (kyma *Kyma) IsBeta() bool {
 
 type AvailableModule struct {
 	Module
-	Enabled bool
-	Valid   bool
+	Enabled         bool
+	ValidationError string
```
ruanxin, line 391: why not declare it as `error`?

api/v1beta2/kyma_types.go
```diff
@@ -387,24 +387,42 @@ func (kyma *Kyma) IsBeta() bool {
...
+}
+
+func (a AvailableModule) IsValid() bool {
```
ruanxin, line 394: I don't recommend this way to indicate `IsValid`,  if you declare `ValidationError` as `Error`, then you can return `a.ValidationError == nil`

api/v1beta2/kyma_types.go
```diff
@@ -419,18 +437,18 @@ func (kyma *Kyma) GetAvailableModules() []AvailableModule {
...
 		})
 	}
 	return modules
 }
 
-func determineModuleValidity(moduleStatus ModuleStatus) bool {
+func determineModuleValidity(moduleStatus ModuleStatus) string {
 	if moduleStatus.Template == nil {
-		return false
+		return "no template defined"
```
ruanxin, line 449: you can define it as a specific error

api/v1beta2/moduletemplate_types.go
```diff
@@ -189,18 +189,38 @@ func (m *ModuleTemplate) IsInternal() bool {
 
 var ErrInvalidVersion = errors.New("can't find valid semantic version")
 
-func (m *ModuleTemplate) GetVersion() (*semver.Version, error) {
+// GetVersionLegacy() returns the version of the ModuleTemplate from the annotation on the object.
+// Remove once shared.ModuleVersionAnnotation is removed
+func (m *ModuleTemplate) GetVersionLegacy() (string, error) {
```
ruanxin, line 194: not need to make it public, it's a private method now.

internal/controller/kyma/controller.go
```diff
@@ -264,6 +298,14 @@ func (r *Reconciler) syncStatusToRemote(ctx context.Context, kcpKyma *v1beta2.Ky
 	return nil
 }
 
+// ValidateKymaSpec validates the Kyma spec.
+func (r *Reconciler) ValidateKymaSpec(kyma *v1beta2.Kyma) error {
```
ruanxin, line 302: please consider rename it to indicate this function only validate global channel, but `ValidateKymaSpec` is too generic.

internal/controller/kyma/controller.go
```diff
@@ -131,6 +134,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		}
 	}
 
+	err := r.ValidateKymaSpec(kyma)
```
ruanxin, line 137: I don't think we need to validate kyma here, this kyma is kcp kyma, the channel is generated and managed by us, the upstream process (keb) should make sure the correctness of channel.

internal/controller/kyma/controller.go
```diff
@@ -209,6 +207,42 @@ func (r *Reconciler) reconcile(ctx context.Context, kyma *v1beta2.Kyma) (ctrl.Re
 	return res, err
 }
 
+func (r *Reconciler) replaceKymaFromRemote(ctx context.Context, kyma *v1beta2.Kyma) ReconciliationLoopResult {
```
ruanxin, line 210: to be honest, I don't think encapsulating the following logic into `replaceKymaFromRemote` is a good idea, first, it violates single responsibility principle. `SyncRemoteCrd` is nothing to do with `replaceKymaFromRemote` secondly, you complicated the result to return. 

I think it's perfectly fine keep  `SyncRemoteCrds` and `replaceSpecFromRemote` separated as before, and you don't have to introduce such `ReconciliationLoopResult` struct only for data transfer purpose, besides, the usage is only scoped in this function, I don't see the value.

In general, I don't like the idea we have multiple layer to transfer the `ctrl.Result, error`, this better always keep in the reconcile logic so that clearly indicate us in which step we stop reconciliation with what behavior (requeue need? contains error? ...),

internal/controller/kyma/controller.go
```diff
@@ -209,6 +207,42 @@ func (r *Reconciler) reconcile(ctx context.Context, kyma *v1beta2.Kyma) (ctrl.Re
...
+			// if the spec is invalid, we should not continue with processing, but make the user aware of the issue
+			if sErr := r.updateStatusWithError(ctx, kyma, err); sErr != nil {
+				r.Metrics.RecordRequeueReason(metrics.StatusSyncToRemote, queue.UnexpectedRequeue)
+				return newReconcileResult(r.requeueWithError(ctx, kyma, fmt.Errorf("could not synchronize remote kyma status: %w", sErr)))
+			}
+			if sErr := r.syncStatusToRemote(ctx, kyma); sErr != nil {
+				r.Metrics.RecordRequeueReason(metrics.StatusSyncToRemote, queue.UnexpectedRequeue)
+				return newReconcileResult(r.requeueWithError(ctx, kyma, fmt.Errorf("could not synchronize remote kyma status: %w", sErr)))
+			}
+			return newReconcileResult(ctrl.Result{}, nil)
```
ruanxin, line 236: if you return `ctrl.Result{}, nil`, then this kyma will not reconcile anymore

internal/controller/kyma/controller.go
```diff
@@ -209,6 +207,42 @@ func (r *Reconciler) reconcile(ctx context.Context, kyma *v1beta2.Kyma) (ctrl.Re
...
+			return newReconcileResult(r.requeueWithError(ctx, kyma, fmt.Errorf("could not update kyma annotations: %w", err)))
+		}
+		r.Metrics.RecordRequeueReason(metrics.CrdAnnotationsUpdate, queue.IntendedRequeue)
+		return newReconcileResult(ctrl.Result{Requeue: true}, nil)
+	}
+	// update the control-plane kyma with the changes to the spec of the remote Kyma
+	if err = r.replaceSpecFromRemote(ctx, kyma); err != nil {
+		if errors.Is(err, ErrInvalidKymaSpec) {
+			// if the spec is invalid, we should not continue with processing, but make the user aware of the issue
+			if sErr := r.updateStatusWithError(ctx, kyma, err); sErr != nil {
```
ruanxin, line 228: please don't use `updateStatusWithError` here, we have the pattern always update status when requeue happens, that's why before your implementation, the only usage of `updateStatusWithError` is in `requeueWithError`.

internal/controller/kyma/controller.go
```diff
@@ -584,3 +631,30 @@ func (r *Reconciler) SyncKymaEnabled(kyma *v1beta2.Kyma) bool {
 func (r *Reconciler) IsKymaManaged() bool {
 	return r.IsManagedKyma
 }
+
+// ReconciliationLoopResult is a struct that allows helper functions used in reconciliation process to indicate that the main reconciliation loop should be finished now using certain values.
+type ReconciliationLoopResult struct {
```
ruanxin, line 636: As I explained, this data transfer struct is not useful, but makes the return logic hard to understand imo.

internal/controller/kyma/controller.go
Tomasz-Smelcerz-SAP: In my opinion, we should always protect ourselves from user input, regardless of how much we trust the "source".
A controller shouldn't expect the data is always correct - invalid data may be produced because of an error in other systems. We shouldn't just blindly "trust" these systems.

internal/controller/kyma/controller.go
Tomasz-Smelcerz-SAP: Yes, I introduced this function only to get rid of linter errors that complexity of other functions increased too much.
I cannot move this code back - because of it.
I would have to introduce a bigger refactoring which I wanted to avoid in THIS PR.
I have created [an issue to improve this](https://github.com/kyma-project/lifecycle-manager/issues/1750):

internal/controller/kyma/controller.go
```diff
@@ -274,7 +316,12 @@ func (r *Reconciler) replaceSpecFromRemote(ctx context.Context, controlPlaneKyma
 		}
 		return err
 	}
-	remote.ReplaceModules(controlPlaneKyma, remoteKyma)
+
+	if err := r.ValidateKymaSpec(remoteKyma); err != nil {
```
ruanxin, line 320: If we do not validate kyma spec channel, but just let the reconciliation happens, anyway, if user given an invalid global channel, it will still have the error messages in each module status, that this module not exists in that invalid channel.

But have a validating logic here, it complicates the reconciliation, which is this huge update status logic you introduced [here](https://github.com/kyma-project/lifecycle-manager/pull/1748/files#diff-4d414901850a616d4e06a120615896904639aef683f0a4d8860abf65c76179d4R226-R237)

pkg/templatelookup/regular.go
```diff
@@ -52,12 +53,23 @@ func (t *TemplateLookup) GetRegularTemplates(ctx context.Context, kyma *v1beta2.
 		if found {
 			continue
 		}
-		if !module.Valid {
-			templates[module.Name] = &ModuleTemplateInfo{Err: fmt.Errorf("%w: invalid module", ErrTemplateNotValid)}
+		if !module.IsValid() {
+			if module.Enabled {
+				templates[module.Name] = &ModuleTemplateInfo{Err: fmt.Errorf("%w: %s", ErrModuleInSpecNotValid, module.ValidationError)}
```
ruanxin, line 58: what's the purpose to distinguish those error in spec and in status?

api/v1beta2/kyma_types.go
Tomasz-Smelcerz-SAP: Unfortunately this makes controller-tools generator complain: `src/github.com/kyma-project/lifecycle-manager/api/v1beta2/kyma_types.go:392:2: invalid field type: interface{Error() string}`
I haven't found any easy solution yet.

pkg/templatelookup/regular_test.go
```diff
@@ -212,25 +213,281 @@ func TestTemplateLookup_GetRegularTemplates_WhenSwitchModuleChannel(t *testing.T
...
+		wantErrContains          string
+	}{
+		{
+			name: "When staying with the same version, then result contains no error",
+			kyma: builder.NewKymaBuilder().
+				WithEnabledModule(moduleToInstall).
+				WithModuleStatus(v1beta2.ModuleStatus{
+					Name:    moduleToInstall.Name,
+					Channel: "regular",
+					Version: "1.1.0",
```
ruanxin, line 311: can you define those versions for testing `1.1.0`, `1.0.0` as a const value so that better for us to understand which field to test here?

api/v1beta2/kyma_types.go
Tomasz-Smelcerz-SAP: I think that external API is more important than internal implementation. Because we have this type in the `api/v1beta2` package, controller-gen is trying to process it and fails. I don't know to disable this other than moving this type to some other package.

tests/e2e/module_install_by_version_test.go
```diff
@@ -0,0 +1,50 @@
...
+				Should(Succeed())
+
+			By("And Module Operator Deployment exists")
+			Eventually(DeploymentIsReady).
+				WithContext(ctx).
+				WithArguments(runtimeClient, ModuleDeploymentNameInNewerVersion, TestModuleResourceNamespace).
+				Should(Succeed())
+
+			By("And KCP Kyma CR is in \"Ready\" State")
+			Eventually(KymaIsInState).
```
ruanxin, line 44: it's better also verify module status is expected

ruanxin (changes requested):
besides those comment, please also cover this missing AC in this one or followup issue.
> The Lifecycle-Manager must support both the old ModuleTemplates and a new ones (with extended attributes) - we must provide a smooth migration path

internal/controller/kyma/controller.go
ruanxin: I'd rather suppress linter if currently can't manage to make the `reconcile` logic shorter, but don't want to introduce such complicated encapuslation only for linter. but as I described here, I don't think you need to change this existing logic https://github.com/kyma-project/lifecycle-manager/pull/1748#discussion_r1711371963

internal/controller/kyma/controller.go
ruanxin: as I said,
1. it's not user input, we can control it, if upstream process messed up, for example, keb generate kyma with an invalid global channel, then the situation you described still exists, all kyma will in error state, that can be prevented.

2. ideally, as I described [here](https://github.com/kyma-project/lifecycle-manager/pull/1748#discussion_r1711371963), if you skip this process, in the end, the existing logic should also report in the status that each enabled module configured an invalid global channel.

What you do here makes the reconciliation logic more complicated.

api/v1beta2/kyma_types.go
ruanxin: did you also fixed related code regarding `ValidationError`? you have many place assign it with string.

internal/controller/kyma/controller.go
ruanxin: update: I don't think you need to change this existing logic https://github.com/kyma-project/lifecycle-manager/pull/1748#discussion_r1711371963

internal/controller/kyma/controller.go
ruanxin: update: I don't think you need to change this existing logic https://github.com/kyma-project/lifecycle-manager/pull/1748#discussion_r1711371963

internal/controller/kyma/controller.go
ruanxin: update: I don't think you need to change this existing logic https://github.com/kyma-project/lifecycle-manager/pull/1748#discussion_r1711371963

pkg/templatelookup/regular.go
```diff
@@ -248,17 +273,63 @@ func (t *TemplateLookup) getTemplate(ctx context.Context, name, desiredChannel s
...
+	}
+	return filteredTemplates[0], nil
+}
+
+func TemplateNameMatch(template *v1beta2.ModuleTemplate, name string) bool {
+	if len(template.Spec.ModuleName) > 0 {
+		return template.Spec.ModuleName == name
+	}
+
+	// Drop the legacyCondition once the label 'shared.ModuleName' is removed
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: But it triggers the linter :)

pkg/templatelookup/regular.go
Tomasz-Smelcerz-SAP: Well, only to provide more accurate error message so that we know what has caused the problem. I have found the existing "given module template is not valid": invalid module" error totally useless considering it was used to describe a problem with missing manifest reference in the Kyma.Status.

pkg/templatelookup/regular_test.go
Tomasz-Smelcerz-SAP: I thought that just version numbers serve as a better "name" than the name itself - they're usually shorter and visually distinct. But If you think conts like "oldVersion" "newVersion" are better, I'll do that.

Tomasz-Smelcerz-SAP:
> besides those comment, please also cover this missing AC in this one or followup issue.
> 
> > The Lifecycle-Manager must support both the old ModuleTemplates and a new ones (with extended attributes) - we must provide a smooth migration path

There's a unit test with over 10 scenarios in `pkg/templatelookup/regular_test.go`.
It kind-of tests this AC: It tests if certain configuration of ModuleTemplates are allowed in the Spec. It doesn't test the installation, but the Module Catalog Improvement doesn't change much related to the installation process (and there is an e2e test for that).
Do you think we need to add another e2e test just to cover the upgrade scenarios?

api/v1beta2/kyma_types.go
Tomasz-Smelcerz-SAP: I'll try to move the function and the type out of the api package.

internal/controller/kyma/controller.go
Tomasz-Smelcerz-SAP: I agree it's too early. I'll remove it.

internal/controller/kyma/controller.go
Tomasz-Smelcerz-SAP: I'll try to not Validate Kyma Spec as you described in your other comment. If not possible, I'll move that logic back into main reconciliation loop and supress the linter.

internal/controller/kyma/controller.go
Tomasz-Smelcerz-SAP: I'll try to to that.

pkg/templatelookup/regular.go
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: I'll create an issue for that deprection and provide it's name here. Or use an existing one.

tests/e2e/module_install_by_version_test.go
Tomasz-Smelcerz-SAP: I'll add this Expectation

tests/moduletemplates/moduletemplate_template_operator_v2_direct_version.yaml
```diff
@@ -0,0 +1,78 @@
...
+  moduleName: template-operator
+  version: 2.4.2-e2e-test
+  data:
+    apiVersion: operator.kyma-project.io/v1alpha1
+    kind: Sample
+    metadata:
+      name: sample-2.4.2-e2e-test
+    spec:
+      resourceFilePath: "./module-data/yaml"
+  descriptor:
```
Tomasz-Smelcerz-SAP, line 21: The descriptor here must be copied to be in a new format

pkg/templatelookup/regular_test.go
```diff
@@ -212,25 +213,281 @@ func TestTemplateLookup_GetRegularTemplates_WhenSwitchModuleChannel(t *testing.T
...
+					Name:    moduleToInstall.Name,
+					Channel: "fast",
+					Version: "2.1.0",
+					Template: &v1beta2.TrackingObject{
+						PartialMeta: v1beta2.PartialMeta{
+							Generation: 1,
+						},
+					},
+				}).Build(),
+			availableModuleTemplates: (&ModuleTemplateListBuilder{}).
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: I'll try to do that.

api/v1beta2/kyma_types.go
Tomasz-Smelcerz-SAP: moved

api/v1beta2/kyma_types.go
Tomasz-Smelcerz-SAP: `IsValid()` is removed.

api/v1beta2/kyma_types.go
Tomasz-Smelcerz-SAP: done.

api/v1beta2/moduletemplate_types.go
Tomasz-Smelcerz-SAP: done.

internal/controller/kyma/controller.go
Tomasz-Smelcerz-SAP: renamed to `ValidateDefaultChannel`

internal/controller/kyma/controller.go
Tomasz-Smelcerz-SAP: removed.

internal/controller/kyma/controller.go
Tomasz-Smelcerz-SAP: done.

internal/controller/kyma/controller.go
Tomasz-Smelcerz-SAP: reconcile logic reverted.

internal/controller/kyma/controller.go
Tomasz-Smelcerz-SAP: removed, reconcile logic reverted.

internal/controller/kyma/controller.go
Tomasz-Smelcerz-SAP: I reverted the reconcile logic, but I am leaving this validation here. This has an advantage of providing more useful error message and it doesn't introduce any additional complication to the reconciliation logic.

pkg/templatelookup/regular.go
Tomasz-Smelcerz-SAP: Done: I am now returning an error directly, so this conditional could be removed.

pkg/templatelookup/regular.go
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: https://github.com/kyma-project/lifecycle-manager/issues/1796

pkg/templatelookup/regular_test.go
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: done.

tests/e2e/module_install_by_version_test.go
Tomasz-Smelcerz-SAP: Added

pkg/templatelookup/availableModules.go
```diff
@@ -0,0 +1,76 @@
...
+func (a AvailableModule) configuredWithVersionInSpec() bool {
+	return a.Enabled && a.Version != "" && a.Channel == ""
+}
+
+// installedwithVersionInStatus returns true if the Module installed using a specific version (instead of a channel) is reported in Status.
+func (a AvailableModule) installedwithVersionInStatus() bool {
+	return !a.Enabled && shared.NoneChannel.Equals(a.Channel) && a.Version != ""
+}
+
+func GetAvailableModules(kyma *v1beta2.Kyma) []AvailableModule {
```
ruanxin, line 36: If you refactor this `GetAvailableModules` not belong to kyma struct, I think it's better to rename it, not use `GetXXX`, it's not aligned with the naming conversion.

tests/moduletemplates/moduletemplate_template_operator_v2_direct_version.yaml
ruanxin: it's not necessary that fixed version must use new OCM format, both new and old format should compatible with fixed version moduletemplate.

pkg/templatelookup/regular.go
```diff
@@ -239,26 +258,72 @@ func (t *TemplateLookup) getTemplate(ctx context.Context, name, desiredChannel s
...
+
+	if filteredTemplates[0].Spec.Mandatory {
+		return nil, fmt.Errorf("%w: for module %s in channel %s",
+			ErrTemplateMarkedAsMandatory, name, desiredChannel)
+	}
+
+	return filteredTemplates[0], nil
+}
+
+func (t *TemplateLookup) getTemplateByVersion(ctx context.Context, name, version string) (
```
ruanxin, line 284: The current implementation is fine, but we should plan a follow-up refactor to optimize the `get template` function. Instead of using `List`, we should retrieve the template directly by its `moduleTemplate` name. Since the `moduleTemplate` names follow a strict naming convention of `[module name] + [channel or version]`, it will be straightforward to implement a `Get` request for direct fetching. This change will improve performance, especially when dealing with a large number of resources, as `List` is not efficient in such scenarios.

Can you create a followup issue?

Tomasz-Smelcerz-SAP: https://github.com/kyma-project/lifecycle-manager/issues/1805

pkg/templatelookup/availableModules.go
Tomasz-Smelcerz-SAP: done.

ruanxin (approved):

Issue #1762, opened 2024-08-09
KymaCR.status.modules not updated on module disabling


c-pius:
### Description

When disabling a module, the module's manifest gets deleted quite swiftly. Watcher logs also indicate that the change is sent from SKR to KCP immediately. However, even after the manifest got deleted, the KymaCR still lists the module under `.status.modules[]` in Ready state, seemingly until the next scheduled requeue. When this requeue happens, the entry is fully removed from `.status.modules[]`.
When adding a module, the change is immediately reflected in `.status.modules[]`.

### Steps to reproduce

> reproducable on DEV, STAGE and PROD

- create a cluster on stage
- watch the KymaCR on KCP
- watch the manifests for that kyma on KCP
- disable a module
- observe that the manifest for that module is deleted almost immediately after disabling the module
- observe that the KymaCR continues to list the module under `.spec.modules[]` in Ready state for a couple of mins (even if the manifest is already completely gone)

### Environment Type

Managed

### Environment Info

Kubernetes Version: x.y.z
Docker Version: x.y.z
Kyma CLI Version: x.y.z
If used - k3d/k3s/minikube versions

### Acceptance Criterias
- [ ] Reproduce this locally and investigate if next AC can be implemented as stated
- [ ] Implement [this comment](https://github.com/kyma-project/lifecycle-manager/issues/1762#issuecomment-2283796055) for the following scenario
    - Disabling module, thus corresponding Manifest CR gets marked for deletion
    - While Manifest is not deleted, Kyma CR should be in a `Processing` state
    - This should already be covered by this test: https://github.com/kyma-project/lifecycle-manager/blob/d51230b9959a0b8a9ae592684e3406f5244909ff/tests/e2e/module_deletion_upgrade_test.go#L45
    - Revisit https://github.com/kyma-project/lifecycle-manager/blob/e1ded83f86ae4023cd100da1d921ae60b246acf0/api/v1beta2/kyma_types.go#L328 to actively set the `Processing` when Module gets deleted, so `Processing` update gets not skipped if deletion is "too" fast

jeremyharisch:
[...]

PR #1792, opened 2024-08-23
fix: Replace RestrictedEnqueueRequestForOwner with EnqueueRequestForOwner

ruanxin:
**Description**

We have a bug in the current `RestrictedEnqueueRequestForOwner` implementation, check detail [here](https://github.com/kyma-project/lifecycle-manager/pull/1792#discussion_r1728922633).
The reason why we previously introduced this customized  `RestrictedEnqueueRequestForOwner` is because we want to limit kyma CR enqueue only triggered by update event which makes manifest CR state changes (we use SSA, which update manifest very frequently before). However, since we limited the SSA in this https://github.com/kyma-project/lifecycle-manager/issues/1456, there is not necessary to have this customized version, we can directly use the native `EnqueueRequestForOwner` to achieve the same goal.

This PR also replaced `GenerationChangedPredicate`, `LabelChangedPredicate` with `ResourceVersionChangedPredicate` so that all watched resources can trigger enqueue based on subresource changes. For example, Manifest CR status change can trigger owner kyma CR enqueue. 

**Related issue(s)**
[#1762](https://github.com/kyma-project/lifecycle-manager/issues/1762)


pkg/watch/restricted_enqueue_request_for_owner.go
```diff
@@ -1,227 +0,0 @@
...
-	oldIfAny, object client.Object,
-	result map[reconcile.Request]any,
-	ref apimetav1.OwnerReference,
-	refGV schema.GroupVersion,
-) {
-	// Compare the OwnerReference Group and Kind against the OwnerType Group and Kind specified by the user.
-	// If the two match, create a Request for the objected referred to by
-	// the OwnerReference.  Use the Name from the OwnerReference and the Namespace from the
-	// object in the event.
-	if ref.Kind != e.groupKind.Kind || refGV.Group != e.groupKind.Group {
```
ruanxin, line 160: To reviewer:

this condition prevent following logic to enqueue related kyma CR, due to `e.groupKind` is empty

here is some log based on this [commit](https://github.com/kyma-project/lifecycle-manager/pull/1792/commits/2609f056d261942ed675e2f7b6800b143b02a610#diff-d43b6cad816120c3cd2afecb269bcc63171dfa42da998f1e196c6cc344181f56R162)
```
{"level":"INFO","date":"2024-08-23T11:17:04.906351175Z","caller":"watch/restricted_enqueue_request_for_owner.go:162","msg":"RestrictedEnqueueRequestForOwner","context":{"groupKind.Kind":"","groupKind.Group":"","ref.Kind":"Kyma","refGV.Group":"operator.kyma-project.io"}}
```

pkg/watch/restricted_enqueue_request_for_owner.go
```diff
@@ -1,227 +0,0 @@
...
-		componentOld, _ := machineryruntime.DefaultUnstructuredConverter.ToUnstructured(oldIfAny)
-		componentNew, _ := machineryruntime.DefaultUnstructuredConverter.ToUnstructured(object)
-		oldState, okOld, _ := unstructured.NestedString(componentOld, "status", "state")
-		newState, okNew, _ := unstructured.NestedString(componentNew, "status", "state")
-
-		if err != nil || !okNew || !okOld {
-			e.Log.Error(err, "error getting owner")
-		}
-
-		if oldState != newState {
```
lindnerby, line 189: Do you think it's fine to use the default implementation that does not check state changes and simply enqueues a reconciliation request based on the ownership relationships. I mean to declare the state change detection as not necessary?
I don't have enough insight to judge if this will result in more enqueues or not.

pkg/watch/restricted_enqueue_request_for_owner.go
```diff
@@ -1,227 +0,0 @@
...
-	}
-}
-
-// parseOwnerTypeGroupKind parses the OwnerType into a Group and Kind and caches the result.  Returns false
-// if the OwnerType could not be parsed using the scheme.
-func (e *RestrictedEnqueueRequestForOwner) parseOwnerTypeGroupKind(scheme *machineryruntime.Scheme) error {
-	// Get the kinds of the type
-	kinds, _, err := scheme.ObjectKinds(e.OwnerType)
-	if err != nil {
-		e.Log.Error(err, "Could not get ObjectKinds for OwnerType", "owner type", fmt.Sprintf("%T", e.OwnerType))
```
lindnerby, line 114: We need to watch out here since the default implementation will [panic if the `OwnerType` parsing is erroring out](https://github.com/kubernetes-sigs/controller-runtime/blob/b901db121e1f53c47ec9f9683fad90a546688c3e/pkg/handler/enqueue_owner.go#L70). Let's keep this in mind when this change goes live.

pkg/watch/restricted_enqueue_request_for_owner.go
ruanxin: yes, I explained here:
> The reason why we previously introduced this customized RestrictedEnqueueRequestForOwner is because we want to limit kyma CR enqueue only triggered by update event which makes manifest CR state changes (we use SSA, which update manifest very frequently before). However, since we limited the SSA in this https://github.com/kyma-project/lifecycle-manager/issues/1456, there is not necessary to have this customized version, we can directly use the native EnqueueRequestForOwner to achieve the same goal.

pkg/watch/restricted_enqueue_request_for_owner.go
ruanxin: This panic makes sense, if the scheme doesn't contain Kyma GKV, it doesn't make sense to initiate KLM.

pkg/watch/restricted_enqueue_request_for_owner.go
lindnerby: Ok, missed that the first time. Thanks!

lindnerby (approved):
lgtm

Issue #1771, opened 2024-08-13
Increase generated cert key size to at least 3072 bits


janmedrek:
### Description

All certificate key lengths should be at least 3072 bits if RSA is used. This applies to all the certificates generated/used by both Lifecycle Manager and Runtime Watcher.

### Reasons

We want to fulfil security requirements and recommendations.

### Acceptance Criteria

- [x] Increase the key size for the generated certificates to 4096
- [x] Keep using the RSA algorithm
- [ ] Check if there will be some downtime, and if it resolves by itself
    - [ ] Communicate this information to the SREs
- [x] Create a follow-up ticket to investigate the possibility of using the ECC algorithm
  - Does it change anything on the client side?

### Attachments

We can ignore the Runtime Watcher downtime as long as it does not affect the KLM behaviour - we have 5 minutes interval so the impact would be minimal for affected clusters.

API ref: https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.PrivateKeyAlgorithm

Code changes:
- https://github.com/kyma-project/lifecycle-manager/blob/a180040ef2e6b8742a61ccc44eaaabae1dded8f5/config/watcher/certificate_setup.yaml#L27
- https://github.com/kyma-project/lifecycle-manager/blob/49a4feeb102523184b1945dc3147c008a9299e3a/pkg/watcher/certificate_manager.go#L208

jeremyharisch:
Places which need to be changed:
- https://github.com/kyma-project/lifecycle-manager/blob/a180040ef2e6b8742a61ccc44eaaabae1dded8f5/config/watcher/certificate_setup.yaml#L27
- https://github.com/kyma-project/lifecycle-manager/blob/49a4feeb102523184b1945dc3147c008a9299e3a/pkg/watcher/certificate_manager.go#L208
- Maybe the Issuers need to be updated

Please keep in mind, that we might have a downtime of the watcher

jeremyharisch:
Questions need to clarified:
- Is Watcher communication relevant for module requirements?
- If yes:
    - Which algorithm to use? Symentric/Asymetric?
    - What is min length? Since 3072 is invalid for RSA for private key length

Internal issues: 
- kyma/backlog/issues/5825
- kyma/backlog/issues/6006#issuecomment-7333169

amritanshusikdar:
[...]

PR #1794, opened 2024-08-23
chore: Increase cert key size to '4096'

c-pius:
**Description**

- increases root cert key size to 4096
- increases client cert key size to 4096
  - also makes this configurable by flag and prevent invalid / (currently) not supported values
- sets rotation policy to Always

**Related issue(s)**

- closes https://github.com/kyma-project/lifecycle-manager/issues/1771


lindnerby (approved):
lgtm

PR #1798, opened 2024-08-26
feat: Add Jitter to Manifest Reconciliation

### LINKED ISSUE #1688: Introduce a random jitter in the Manifest CR loop requeue interval
### Description

Currently we re-queue items in the Manifest Reconciliation loop using a fixed interval. This has a side-effect: Most of the objects are processed in fixed intervals, resulting in a fixed "processing frequency": we observe regular peaks in CPU utilization, queue depth, active workers count, etc.
It is inefficient: we require over 3 CPU cores at peaks, and less than 1 core between the peeks. On average, we consume up to 2 cores. If we "smooth out" our queuing schedule, we'll be able to reduce resource consumption (CPU count) by 1/3.

**Related issue:**
https://github.com/kyma-project/lifecycle-manager/issues/1684

### Reasons

- Improve system behaviour

### Acceptance Criteria

- [x] The scheduling algorithm is configurable via runtime flags
- [x] The requeue of items is more regular (no distinct peaks of processing). See the picture in the comment.

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

This is how it currently looks - inefficient CPU utilization:
image

LeelaChacha:
Closes #1688


LeelaChacha:
**NOTE** The "G404: Use of weak random number generator (math/rand instead of crypto/rand)" linter error was excluded because the crypto/rand library does not have a method to generate floats and the workarounds would add extra cpu cycles to the jitter calculations, twice. So, the math/rand lib is used to generate the float

LeelaChacha:
The Jitter probability and factor have been set to 10%. This is n intuition-based value, and should be tested in landscapes.

cmd/main.go
```diff
@@ -404,6 +404,13 @@ func setupManifestReconciler(mgr ctrl.Manager, flagVar *flags.FlagVar, options c
 			Busy:    flagVar.ManifestRequeueBusyInterval,
 			Error:   flagVar.ManifestRequeueErrInterval,
 			Warning: flagVar.ManifestRequeueWarningInterval,
+			Jitter: &queue.RequeueJitter{
+				Enabled:           true,
```
Tomasz-Smelcerz-SAP, line 408: I don't think we need an "Enabled" attribute. I don't even think we need such property (method). If the jitter value is zero we can just "apply it" - zero jitter means no jitter. So the value of zero for JitterPercentage naturally disables it, and anything greater than zero enables it.

internal/pkg/flags/flags.go
```diff
@@ -133,6 +136,16 @@ func DefineFlagVar() *FlagVar {
 	flag.DurationVar(&flagVar.ManifestRequeueBusyInterval, "manifest-requeue-busy-interval",
 		DefaultManifestRequeueBusyInterval,
 		"determines the duration a Manifest in Processing state is enqueued for reconciliation.")
+	flag.Float64Var(&flagVar.ManifestRequeueJitterProbability, "manifest-requeue-jitter-probability",
+		DefaultManifestRequeueJitterProbability,
+		"determines the probability that jitter is applied to the requeue interval.")
+	flag.Float64Var(&flagVar.ManifestRequeueJitterPercentage, "manifest-requeue-jitter-percentage",
```
Tomasz-Smelcerz-SAP, line 142: This must be limited. I'd argue that the jitter should be between [0...5%] (inclusive). I don't have any formula for the 5% value, it's just my gut feeling - the jitter should be almost "invisible" for the end user. If we allow something like 10% it is almost immediately visible for anyone who takes a look at the "reconciliation started" log entries.

internal/pkg/flags/flags.go
```diff
@@ -19,6 +19,9 @@ const (
 	DefaultManifestRequeueErrInterval                                   = 2 * time.Second
 	DefaultManifestRequeueWarningInterval                               = 30 * time.Second
 	DefaultManifestRequeueBusyInterval                                  = 5 * time.Second
+	DefaultManifestRequeueJitterProbability                             = 0.10
```
Tomasz-Smelcerz-SAP, line 22: 10% is too much. I think we agreed during the refinement of this issue, that the following params could be the default:
JitterProbablility: 2%,  JitterPercentage: 2%, always enabled.

internal/pkg/flags/flags.go
```diff
@@ -133,6 +136,16 @@ func DefineFlagVar() *FlagVar {
...
 		DefaultManifestRequeueBusyInterval,
 		"determines the duration a Manifest in Processing state is enqueued for reconciliation.")
+	flag.Float64Var(&flagVar.ManifestRequeueJitterProbability, "manifest-requeue-jitter-probability",
+		DefaultManifestRequeueJitterProbability,
+		"determines the probability that jitter is applied to the requeue interval.")
+	flag.Float64Var(&flagVar.ManifestRequeueJitterPercentage, "manifest-requeue-jitter-percentage",
+		DefaultManifestRequeueJitterPercentage,
+		"determines the percentage range for the requeue jitter applied to the requeue interval. "+
+			"E.g., 0.1 means +/- 10% of the interval.")
+	flag.DurationVar(&flagVar.ManifestRequeueJitterDisableAfter, "manifest-requeue-jitter-disable-after-seconds",
```
Tomasz-Smelcerz-SAP, line 146: This flag is not strictly necessary. Either remove it or configure it's behavior so that the value of zero disables it i.e. the Jitter is always enabled.

internal/pkg/flags/flags.go
```diff
@@ -133,6 +136,16 @@ func DefineFlagVar() *FlagVar {
 	flag.DurationVar(&flagVar.ManifestRequeueBusyInterval, "manifest-requeue-busy-interval",
 		DefaultManifestRequeueBusyInterval,
 		"determines the duration a Manifest in Processing state is enqueued for reconciliation.")
+	flag.Float64Var(&flagVar.ManifestRequeueJitterProbability, "manifest-requeue-jitter-probability",
```
Tomasz-Smelcerz-SAP, line 139: This must be limited between [0%...100%]

pkg/queue/requeue_intervals.go
```diff
@@ -11,23 +12,58 @@ type RequeueIntervals struct {
...
+		return interval
+	}
+
+	if time.Since(j.StartedAt) > j.DisableAfter {
+		j.Enabled = false
+		return interval
+	}
+
+	//nolint:gosec // No simple way to generate a float in crypto/rand lib and this is not a security-sensitive context
+	if rand.Float64() < j.JitterProbability {
```
Tomasz-Smelcerz-SAP, line 60: That looks OK when I look at it, please provide an unit test for it though.
In order to test it you externalize `rand.Float64` into a new field in `RequeueJitter` struct, say, `randFunc func() float64`
Then in your test you can configure the `RequeueJitter` struct with a function that always returns `0.02` for example.
In the "real" code you wire-up math.Rand() to this field (main.go: 407)
We don't have to "test" math.Rand() directly, it's assumed to be working fine.

Tomasz-Smelcerz-SAP (changes requested):
Overall looks OK, but there are some minor (but important) quality issues. Please take a look at my comments.

internal/pkg/flags/flags.go
LeelaChacha: makes sense

internal/pkg/flags/flags.go
```diff
@@ -306,7 +319,14 @@ func (f FlagVar) Validate() error {
...
 		8192: false, // see https://github.com/kyma-project/lifecycle-manager/issues/1793
 	}[f.SelfSignedCertKeySize] {
-		return errInvalidSelfSignedCertKeyLenght
+		return errInvalidSelfSignedCertKeyLength
+	}
+
+	if f.ManifestRequeueJitterProbability <= 0 || f.ManifestRequeueJitterProbability >= 0.05 {
+		return errInvalidManifestRequeueJitterPercentage
+	}
+	if f.ManifestRequeueJitterProbability <= 0 || f.ManifestRequeueJitterProbability >= 1 {
```
Tomasz-Smelcerz-SAP, line 328: Actually I think both zero and one are totally good values here.

LeelaChacha: Good catch. I somehow inverted the logic for the check in my head and was trying to include 0 and 1. I agree. They are valid opitons

Tomasz-Smelcerz-SAP (approved):

Issue #1719, opened 2024-07-25
Fix faulty ManifestCR state determination


c-pius:
### Description

With the [ADR Mapping module deployment state to ManifestCR state](https://github.com/kyma-project/lifecycle-manager/issues/1706) we defined how we want to map the ManifestCR state from the deployment of the module. This ADR now needs to be implemented in code.

This should be a good candidate for good code design. We already have the `DeploymentReadyCheck` struct. Consider to improve to revisit it's interface and dependencies so that we get a nicely testable "state machine service". If re-designing this, also consider that later on we want to add state checks for StatefulSet as well.

### Reasons

With https://github.com/kyma-project/lifecycle-manager/issues/1554 we introduced new ManifestCR state determination that proofed to be faulty.

### Acceptance Criteria

- [ ] ManifestCR state determination follows the Decision section of the ADR
- [ ] Existing E2E tests extended to verify the expected state transitions from the ADR
  - [ ] consider if we can include in the tests that the transitions happen quicker than the requeue success interval
- [ ] ManifestCR state determination unit tested with coverage > 80%

janmedrek:
HINT: KLM Code - https://github.com/kyma-project/lifecycle-manager/blob/294103d856cb03a2e148c960b56ce071b02d4372/internal/manifest/deployment_ready_check.go#L44

c-pius:
Re-estimated to 0 as we included it in the sprint closing on 2024-09-11.

PR #1807, opened 2024-08-28
fix: Fix faulty manifest state determination

nesmabadr:
**Description**

- Fix faulty manfiest state determination

**Related issue(s)**
#1719


internal/manifest/readycheck/deployment_ready_check.go
```diff
@@ -0,0 +1,59 @@
...
+type DeploymentReadyCheck struct{}
+
+func (c *DeploymentReadyCheck) Run(
+	deploy *apiappsv1.Deployment,
+) (shared.State, error) {
+	deploymentState := GetDeploymentState(deploy)
+	return deploymentState, nil
+}
+
+func GetDeploymentState(deploy *apiappsv1.Deployment) shared.State {
```
c-pius, line 32: I think this should not be an exported func (right now it is used in the test and Run only). In my opinion we can move this directly up into the Run (may be renamed) and test against the Run function.

internal/manifest/readycheck/deployment_ready_check.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// NewDeploymentReadyCheck creates a readiness check that verifies if a Deployment is ready.
+func NewDeploymentReadyCheck() *DeploymentReadyCheck {
+	return &DeploymentReadyCheck{}
+}
+
+type DeploymentReadyCheck struct{}
+
+func (c *DeploymentReadyCheck) Run(
```
c-pius, line 25: I think `c` could be `_` as we are not using it?

internal/manifest/readycheck/statefulset_ready_check.go
```diff
@@ -1,10 +1,13 @@
-package manifest
+package readycheck
```
c-pius, line 1: Just to confirm my understanding. We are going to tackle the stateful set part iwth the other ticket you created. Here the changes are necessary because we removed the helper funcs from the `ready_check.go`?

nesmabadr: Yes I moved the GetPodsState here because right now it is only used for the StatefulSet and so that it would be easier to remove this when we enhance the state determination for StatefulSets.

internal/manifest/readycheck/deployment_ready_check.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/kyma-project/lifecycle-manager/api/shared"
+)
+
+const (
+	NewRSAvailableReason    = "NewReplicaSetAvailable"
+	FoundNewRSReason        = "FoundNewReplicaSet"
+	NewReplicaSetReason     = "NewReplicaSetCreated"
+	ReplicaSetUpdatedReason = "ReplicaSetUpdated"
+)
```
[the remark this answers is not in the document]
nesmabadr: I already tried to use these but I was never able to add this package, had several compatibility issues with the packages we use. Even tried to bump some indirect packages but didn't work either

internal/manifest/README.md
```diff
@@ -8,4 +8,4 @@ It contains various aspects to:
 2. Look up the correct Client for Single or Dual Cluster Mode based on the [SKR Client Lookup](skr_client_lookup.go) and a [ClusterClient](client.go) to read Secret Data as if they were a kubeconfig
 3. Look up the correct Keychain to access image layers contained in private registries through a [Keychain](../../pkg/ocmextensions/cred.go)
 4. Interact with Default Data supplied to the Manifest CR for initializing a module with [Pre/Post-Hooks](custom_resource.go)
-5. Check for Module Readiness not only by looking at the resources supplied in the chart, but also to introspect the already mentioned CustomResource using a [custom ReadyCheck](ready_check.go)
+5. Check for Module Readiness not only by looking at the resources supplied in the chart, but also to introspect the already mentioned CustomResource using a [custom ReadyCheck](readycheck/ready_check.go)
```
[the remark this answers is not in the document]
nesmabadr: I think this was outdated, will change it

internal/manifest/readycheck/deployment_ready_check.go
```diff
@@ -0,0 +1,59 @@
...
+
+// NewDeploymentReadyCheck creates a readiness check that verifies if a Deployment is ready.
```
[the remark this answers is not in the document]
nesmabadr: Hmm IMO I'd rather keep it as ReadyCheck instead of StateMachine so that the purpose is clear from the naming, WDYT?

internal/manifest/readycheck/deployment_ready_check.go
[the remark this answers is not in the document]
c-pius: Okay, then let's leave it as is. But could you align the names with the values? That makes it easier follow. E.g. `	NewReplicaSetReason     = "NewReplicaSetCreated"` => `NewReplicaSetCreatedReason = "NewReplicaSetCreated"`

internal/manifest/readycheck/deployment_ready_check.go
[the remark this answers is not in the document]
c-pius: How about "StateCheck" then? :)

internal/manifest/readycheck/deployment_ready_check.go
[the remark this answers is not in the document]
nesmabadr: Sure, I just took the var names from the package as well, but the renaming makes sense

internal/manifest/readycheck/deployment_ready_check.go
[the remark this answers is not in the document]
nesmabadr: sounds good 👍

internal/manifest/readycheck/ready_check.go
```diff
@@ -1,14 +1,10 @@
-package manifest
+package readycheck
```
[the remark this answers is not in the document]
nesmabadr: Regarding the second point about scheme, I suggest postponing this till we do the Statefulset state determination because we are not yet sure that we won't need the client to list the pods.

internal/manifest/statecheck/deployment_state_check.go
```diff
@@ -0,0 +1,62 @@
...
+	return &DeploymentStateCheck{}
+}
+
+type DeploymentStateCheck struct{}
+
+func (*DeploymentStateCheck) GetState(
+	deploy *apiappsv1.Deployment,
+) (shared.State, error) {
+	deploymentState := getDeploymentState(deploy)
+	return deploymentState, nil
```
c-pius, line 32: For me personally we could move the contents from o`getDeploymentState` up into this function and avoid the additional "layer", but also fine as is.

internal/manifest/statecheck/state_check.go
```diff
@@ -0,0 +1,78 @@
...
+	*apiappsv1.Deployment
+	*apiappsv1.StatefulSet
+}
+
+func NewManagerStateCheck() *ManagerStateCheck {
+	return &ManagerStateCheck{
+		statefulSetChecker:     NewStatefulSetStateCheck(),
+		deploymentStateChecker: NewDeploymentStateCheck(),
+	}
+}
```
c-pius, line 36: Please inject these dependencies from the outside. Then we can also easily create a test file state_check.go. We can supply test stubs instead of the real dependencies. Then we can test that the correct stub is called depending on whether we are passing a deployment or statefulset to GetState. 
We can also do that as a pair programming session if you like.

c-pius (commented):
Thanks! Consider my suggestions optional, except the one regarding dependencies and testing in `internal/manifest/statecheck/state_check.go`. This I think needs to be changed and I am happy to assist in a pair programming session :)

internal/manifest/statecheck/state_check.go
```diff
@@ -0,0 +1,78 @@
...
+	apiappsv1 "k8s.io/api/apps/v1"
+	"k8s.io/cli-runtime/pkg/resource"
+
+	"github.com/kyma-project/lifecycle-manager/api/shared"
+	declarativev2 "github.com/kyma-project/lifecycle-manager/internal/declarative/v2"
+)
+
+type ManagerStateCheck struct {
+	statefulSetChecker     StatefulSetStateChecker
+	deploymentStateChecker DeploymentStateChecker
```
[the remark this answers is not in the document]
nesmabadr: oh didn't know that, thanks!

internal/manifest/statecheck/state_check.go
```diff
@@ -0,0 +1,88 @@
...
+	apiappsv1 "k8s.io/api/apps/v1"
+	"k8s.io/cli-runtime/pkg/resource"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kyma-project/lifecycle-manager/api/shared"
+)
+
+type ManagerStateCheck struct {
+	StatefulSetChecker     StatefulSetStateChecker
+	DeploymentStateChecker DeploymentStateChecker
```
c-pius, line 15: Can we make those private?

internal/manifest/statecheck/state_check.go
```diff
@@ -0,0 +1,88 @@
...
+
+const (
+	DeploymentKind  ManagerKind = "Deployment"
+	StatefulSetKind ManagerKind = "StatefulSet"
+)
+
+type Manager struct {
+	Kind ManagerKind
+	*apiappsv1.Deployment
+	*apiappsv1.StatefulSet
```
c-pius, line 36: Can we make those private?

internal/manifest/statecheck/state_check.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (m *ManagerStateCheck) GetState(ctx context.Context,
+	clnt client.Client,
+	resources []*resource.Info,
+) (shared.State, error) {
+	mgr := findManager(clnt, resources)
+	if mgr == nil {
+		return shared.StateReady, nil
+	}
```
c-pius, line 55: Just wondering, no need to address in this PR... Do you think it is reasonable that we are defaulting to Ready if we can't determine the state? My understanding is that we always expect a manager to be present. Maybe it would be a good idea to introduce another state `Unknown` that we set when we are not able to determine the state correctly. Any opinions? If you agree that we may need to do something about it I can create a separate ticket and discuss it with the team.

c-pius (commented):
Great work, thanks! Just two minor things and one question left.

internal/manifest/statecheck/state_check.go
nesmabadr: I had to export them for the test

internal/manifest/statecheck/state_check.go
nesmabadr: this was a temporary solution until the StatefulSet was introduced. But since it is now already introduced, we can even remove this

internal/manifest/statecheck/state_check.go
c-pius: If we use the NewManagerStateCheck constructor we can make them private. Or is there any reason for creating the manager directly and not via constructor?

internal/manifest/statecheck/state_check.go
c-pius: Let's discuss with the team what would be the reasonable thing here 👍🏻

internal/manifest/statecheck/state_check.go
nesmabadr: no no nevermind, I can use the constructor 👍

c-pius (dismissed):

mmitoraj (approved):

c-pius (approved):

PR #1854, opened 2024-09-11
fix: Reintroduce kyma event filter

lindnerby:
**Description**

- Re-add event filter for kymas control loop for gen and label change only. Generic events are too spammy and this has been removed accidentally.

**Related issue(s)**


internal/controller/kyma/setup.go
```diff
@@ -61,6 +61,7 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, opts ctrlruntime.Options
 	if err := ctrl.NewControllerManagedBy(mgr).For(&v1beta2.Kyma{}).
 		Named(controllerName).
 		WithOptions(opts).
+		WithEventFilter(predicates).
```
amritanshusikdar, line 64: Since we all agreed on adding changes to the respective _test.go file, do you plan on skipping it for this one?

PR #1860, opened 2024-09-12
fix: Test check pipeline

lindnerby:
**Description**

- Change output of considered test files and print test and go file change count
- Consider all files that end on _test.go only as test files

**Related issue(s)**


ruanxin (approved):
nice report

ruanxin (approved):

Issue #1506, opened 2024-04-29
feat: [Zero-Downtime] Istio-Gateway Secret Certificate Bundling


Tomasz-Smelcerz-SAP:
### Description

After https://github.com/kyma-project/lifecycle-manager/issues/2084 has been merged, we have a [safe migration path](https://github.com/kyma-project/lifecycle-manager/issues/1890) for the implementation of the istio gateway secret management.

We need to modify the existing istio gateway secret controller, to now be able to bundle root certificates together to allow SKR clients with older root certificates to authenticate with the istio gateway. A bundle is just two or more root certificates [appended together](https://github.com/kyma-project/lifecycle-manager/compare/main...Tomasz-Smelcerz-SAP:poc/zero-downtime#diff-b27f02fec7d2ab1a03a26d0ee14c1105d7b390e7d9d7466d0c8b4cc7faee1478R63).

The logic is based on the [POC](https://github.com/kyma-project/lifecycle-manager/issues/1430#issuecomment-2068796573) results, but it is simplified: The additional secret object is removed, instead the Istio Gateway secret plays the key role in the migration process. Furthermore, no change to the runtime-watcher is required because the x509 has been updated and now handles ca-bundles by default.

**Outline of actions:**
1. Change of the Root Certificate secret is detected
2. Update the ca.crt field of the istio gateway Secret with new Certificate at position 0 (old Certificate shall move back to position 1, even older Certificate now at position 2 can be deleted now)
3. Expiration of the old certificate happens in less than configured number of hours (e.g: 24)
4. Switch the tls.crt, tls.key attributes to the new certificate

For details about the migration see the attached image.

### Reasons

We need a robust, zero-downtime solution for the Watcher TLS certificate rotation

### Acceptance Criteria

- [ ] Controller is modified to store a ca-bundle in the ca.crt field.
- [ ] A flag in KLM exists to switch to legacy mode (current setup).
- [ ] Documentation is updated.
- [ ] Certificate bundling is tested

### Testing approach

Unit and E2E tests

### Attachments

![watcher-certificate-migration3](https://github.com/kyma-project/lifecycle-manager/assets/15265082/916ace82-1c43-45e4-a9ad-92d02e24a70f)

### Related Issues
https://github.com/kyma-project/lifecycle-manager/issues/1430

c-pius:
Putting this back to Unrefined as we agreed to have this one re-estimated, just so we don't forget to do that timely.

PR #1871, opened 2024-09-16
feat: Zero downtime runtime watcher tls management

### LINKED ISSUE #1507: feat: [Zero-Downtime] Runtime-Watcher TLS configuration management
### Description

Zero-Downtime: Implement the Runtime-Watcher TLS configuration renewal logic.

The logic is based on the [POC](https://github.com/kyma-project/lifecycle-manager/issues/1430#issuecomment-2068796573) results, but it is simplified: The additional secret object is removed, instead the Istio Gateway secret plays the key role in the migration process.
This is based on the following observation:
**We must adjust SKR watcher client TLS configuration if and ONLY IF the Istio Gateway TLS configuration changes**
This has an impact for the design of the zero-downtime certificate rotation solution.
The system is designed with two independent components, running asynchronously to each other:
- The first one observes rotation of the "Root" certificates in KCP and manages the Istio Gateway secret accordingly. It's not related to any particular Kyma or SKR
- The second one manages SKR watcher client TLS configuration. It generates/updates the relevant secrets in KCP and SKR. It is coupled to the reconciliation of the Kyma CR.

**Note:** This issue describes the **second** component

**Responsibility 1: Bootstrap**

1. No Watcher TLS secret exists in the KCP
2. Wait until the Istio Gateway secret is available in the KCP
3. Create Watcher TLS certificate in the KCP (using certificate CR - the Cert Manager creates the secret)

**Responsibility 2: Migration**
1. When both of the following happen:
    - Root certificate is more recent than the Watcher TLS secret in the KCP
    - Istio-Gateway secret  is more recent than the Watcher TLS secret in the KCP
3. Re-generate the Watcher TLS certificate in the KCP (already implemented but triggerred differently)

**Responsibility 3: Synchronization**

1. When any of the conditions occur:
    - Watcher TLS configuration is missing in SKR
    - Watcher TLS secret in KCP is more recent than the corresponding secret in the SKR
    - Istio-Gateway secret is more recent than secret in the SKR
2. Then generate Watcher TLS configuration secret in the SKR, taking the `tls.crt` and `tls.key` from the corresponding secret in the KCP, but `ca.crt` data from the Istio-Gateway secret

Note: Instead of 1. we can also just sync the data with every reconciliation (patch)

**Implementation Notes:**
- This logic is tightly coupled with any given SKR, so it looks it can be implemented as the part of current Kyma reconciliation loop.
- The code for Watcher certificates and secrets generation/renewal and for synchronization of these to the SKR is already implemented. It must be adjusted to account for new requirements.

### Reasons

We need a robust, zero-downtime solution for the Watcher TLS certificate rotation

### Acceptance Criteria

- [ ] Create a follow-up issue to provide a metric in the SKR that reports the client certificate used for watcher. Then in Plutono we can see how the migration process works.
- [x] Implement the solution along with necessary unit and integration tests
- [ ] Update the documentation
- [x] Manually test the rotation logic
- [x] Update the existing e2e test and add a new one if necessary.

### Testing approach

unit tests, integration tests, e2e test(s)
Existing tests:
- https://github.com/kyma-project/lifecycle-manager/blob/a0c49436f3d11d03c9a7556ec11c7c9f69d621d9/tests/e2e/ca_certificate_rotation_test.go
- https://github.com/kyma-project/lifecycle-manager/blob/a0c49436f3d11d03c9a7556ec11c7c9f69d621d9/tests/e2e/self_signed_certificate_rotation_test.go

### Attachments
watcher-certificate-migration3

### Related Issues
https://github.com/kyma-project/lifecycle-manager/issues/1430

LeelaChacha:
Closes #1507 

Changes Proposed:
1. Placeholder implementation of istio gateway secret. To be implemented in #1506 
2. Watcher TLS installation uses the CA certs from istio gateway instead of root CA secret
3. Watcher TLS migration condition changed


LeelaChacha:
It's not relevant anymore. Istio Gateway Secret is managed by a separate reconciler.

Issue #1802, opened 2024-08-27
Automate Local Test Setup Script for Lifecycle-Manager


jeremyharisch:
## Description
We need to create a script to automate the local test setup for the Lifecycle-Manager. Currently, developers have to manually follow the steps outlined in our [local test setup documentation](https://github.com/kyma-project/lifecycle-manager/blob/main/docs/developer-tutorials/local-test-setup.md). While this documentation is sufficient, the manual process is time-consuming, and some developers have already written their own scripts to expedite this process.

To streamline development and ensure consistency, we decided to create an official script that will be maintained by the team. This script will automate the setup of the local environment for testing the Lifecycle-Manager, reducing the repetitive manual steps and ensuring that the setup aligns with our documentation.

A first step towards this has been done with the [VSCode tasks](https://github.com/kyma-project/lifecycle-manager/blob/main/.vscode/tasks.json). Those proofed to be quite handy, but are VS Code specific right now. The aim is to refactor this with the following goal:

- all functionality is provided by plain scripts executable on MacOS and Linux where arguments may be passed to the script
- for each script, we provide VSCode and Goland "wrappers" calling the script
  - the wrapper may use native features of the IDE for getting the config and then passes it as arg to the script
- the scripts will also be used in the E2E test runners, thereby we ensure that we have consistent setup throughout all environments

## Reasons
- Efficiency: The manual process is time-consuming and prone to errors. A script will automate the repetitive tasks, saving time and reducing the risk of mistakes.
- Consistency: Ensuring all developers are using the same setup process will help avoid environment-specific issues during local testing.
- Maintainability: An officially maintained script will ensure that our setup process stays up-to-date with any changes in the project or its dependencies, keeping our documentation in sync.

## Acceptance Criteria

- [x] scripts cover the same functionality as in [VSCode tasks](https://github.com/kyma-project/lifecycle-manager/blob/main/.vscode/tasks.json)
- [x] an additional script should check for the following tools in defined versions and fails if not present, or warns (ideally with prompt) if present in a wrong version
    - Golang
    - k3d
    - kubectl
    - Docker
- [x] scripts must be executable on MacOS and Linux
- [x] Documentation:
    - The documentation should be updated to include the new scripts and provide instructions on how to use them.
Ensure that the script and the documentation are kept in sync with any updates.
- [x] The scripts should be tested to ensure it correctly sets up the environment as specified.
[...]

PR #1898, opened 2024-09-26
chore: Add VSCode tasks and launch config for local setup

c-pius:
**Description**

- provides VSCode tasks that aid local test setup. Execute tasks by `Shift + Cmd + P` => `Tasks: Run Task`
  - in a future step, the script for creating the clusters may be a shared one for local testing and PR E2E tests

**Related issue(s)**

- partially resolves: https://github.com/kyma-project/lifecycle-manager/issues/1802


scripts/tests/create_test_clusters.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+# Change to the directory where the script is located
+cd "$(dirname "$0")"
+
+# create SKR cluster
+if k3d cluster list | grep -q "skr"; then
```
nesmabadr, line 7: this command would also get a cluster named `skr-local` for example, so I think this should be `k3d cluster list | grep "^skr\s"` to do an exact match of the name "skr"

scripts/tests/create_test_clusters.sh
```diff
@@ -0,0 +1,51 @@
...
+        -p 10080:80@loadbalancer \
+        -p 10443:443@loadbalancer \
+        --k3s-arg --tls-san="skr.cluster.local@server:*" \
+        --image rancher/k3s:v1.28.7-k3s1 \
+        --k3s-arg --disable="traefik@server:*" \
+        --k3s-arg --tls-san="host.k3d.internal@server:*"
+fi
+
+# create KCP cluster
+if k3d cluster list | grep -q "kcp"; then
```
nesmabadr, line 20: same here

nesmabadr (approved):

Issue #1712, opened 2024-07-24
Create subfolders for `/docs/user` and `/docs/contributor`


mmitoraj:
**Description**

Lifecycle Manager, in general, is not an end-user-facing project. Most of the documents are relevant only to those who contribute to the project; however, some documents are directed to its users.

To have a clear division of the roles we could split the `/docs` folder into `/docs/user` and `/docs/contributor`.

We had a case when the SRE team could not understand why the spec of a Kyma CR was different in the KCP and SKR. These were very technical details that were not explained in the `Kyma CR` documentation as they are not important to the users of Lifecycle Manager. However, the `/docs/contributor` folder could be the right location for that kind of information.

Also, looking at the folder structure of other `kyma-project` repositories developed by other Kyma teams, especially modules, the split into `/docs/user` and /`docs/contributor` is the template convention accepted and widely used by most of them.

**TBD:** The open question is how to categorize individual documents and define their final location (user vs. contributor).

**Reasons**

More transparent project documentation.

**Assignees**

@kyma-project/technical-writers
@kyma-project/jellyfish 

**Attachments**

c-pius:
We may take this also as an opportunity to clearly defining our target groups. E.g., user: the end user of a Kyma runtime; contributor: developer or SRE (may be SAP internal or external). We also noted this here: https://github.com/kyma-project/lifecycle-manager/issues/1447

c-pius:
As part of this refactoring and addressing new personas, we may consider to get rid of docs which seem to be obsolete anyway, i.e.:

- https://github.com/kyma-project/lifecycle-manager/blob/main/docs/developer-tutorials/prepare-gcr-registry.md
- https://github.com/kyma-project/lifecycle-manager/blob/main/docs/developer-tutorials/provision-cluster-and-registry.md
- https://github.com/kyma-project/lifecycle-manager/blob/main/docs/developer-tutorials/starting-operator-with-webhooks.md

mmitoraj:
Initial evaluation: 
- [ ] [config-private-registry.md](https://github.com/kyma-project/lifecycle-manager/blob/main/docs/developer-tutorials/config-private-registry.md) - **user**
- [ ] [local-test-setup.md](https://github.com/kyma-project/lifecycle-manager/blob/main/docs/developer-tutorials/local-test-setup.md) - **user**
[...]

Issue #1447, opened 2024-04-04
Align user-tutorials and developer-tutorials


c-pius:
**Description**

With https://github.com/kyma-project/lifecycle-manager/blob/main/docs/developer-tutorials/local-test-setup.md we have a very comprehensive document on how to:

- create two k3d clusters for KCP and SKR
- deploy KLM into the KCP cluster (either from a published image or local build)
- deploy template operator ModuleTemplate into KCP
- deploy KymaCR into KCP
- enable the module in the SKR

At the same time we have https://github.com/kyma-project/lifecycle-manager/blob/main/docs/user-tutorials/01-10-control-plane-quick-start.md where I am not sure what the real purpose is, considering the document above. The context section states more or less the same outline, but the actual content is not as extensive and even a bit confusing at some points.

We need to think about aligning both documents. But first we should clarify what our expected audience is. What distinguishes a KLM "user" from a "developer". What kind of documentation do we need for either of those?

**Reasons**

We want a transparent and convenient-to-navigate documentation for KLM

**Assignees**

@kyma-project/technical-writers

**Acceptance Criteria**
- [ ] [Control Plane Quick Start](https://github.com/kyma-project/lifecycle-manager/blob/main/docs/user-tutorials/01-10-control-plane-quick-start.md) doc removed

janmedrek:
The `Control Plane Quick Start` document should be dropped, I've checked it and it is just a narrowed-down version of the developer docu that we have.

Our user documentation should rather focus on what Module Teams can do to leverage additional KLM functionalities, but we are moving towards separation of that topic. We've already deprecated the CustomStateCheck feature so CustomResourcePolicy is the only one left. We can consider adding something for "unmanaging" the modules.

mmitoraj:
We decided to leave only one of the tutorials. I consider it complete with the PR splitting the `docs` into `user` and `contributor`. Still, I would like to work more on the `/docs/user` folder docu and some unifications in the `/docs/contributor` but it is a different story and requires a separate issue. I'm closing this one having the follow-ups in mind.

PR #1905, opened 2024-09-30
docs: Create a new folder structure

mmitoraj:
**Description**

- Create a new folder structure: docs/user vs. docs/contributor
- Remove obsolete documentation
- Updated the main README.md
- Replaced API documentation with CRD documentation

**Related issue(s)**
https://github.com/kyma-project/lifecycle-manager/issues/1447
https://github.com/kyma-project/lifecycle-manager/issues/1712


docs/contributor/resources/03-moduletemplate.md
```diff
@@ -1,6 +1,14 @@
-# ModuleTemplate Custom Resource
+# ModuleTemplate
 
-The core of our modular discovery is the [ModuleTemplate custom resource (CR)](../../../api/v1beta2/moduletemplate_types.go). It is used to initialize and resolve modules.
+The `moduletemplates.operator.kyma-project.io` Custom Resource Definition (CRD) is a comprehensive specification that defines the structure and format used to configure the ModuleTemplate resource.
+
+The Manifest CR contains modules' metadata with links to their images and manifests. ModuleTemplate CR represents a module in a particular version. Based on this resource Lifecycle Manager enables or disables modules in a cluster. It contains the list of added modules and their state.
```
c-pius, line 5: I think a few things got tangled up here.

> The Manifest CR contains modules' metadata with links to their images and manifests. ModuleTemplate CR represents a module in a particular version.

Are you sure you want to say `Manifest CR`? And overall this section will only be true fully once we released the new ModuleTemplate CR changes. But for me it would be okay to add the description already as proposed.

> Based on this resource Lifecycle Manager enables or disables modules in a cluster. It contains the list of added modules and their state.

I think this rather describes the module list in spec and status of KymaCR. Here, I would rather expect something like *ModuleTemplates define the set of module that can be installed by Lifecycle Manager*.

> Manifest CR [...] ModuleTemplate CR [...]

in manifest.md and watcher.md you also introduced the abbreviation: `[...] custom resource (CR)`

docs/contributor/resources/01-kyma.md
```diff
@@ -1,4 +1,12 @@
-# Kyma Custom Resource
+# Kyma
+
+The `kymas.operator.kyma-project.io` Custom Resource Definition (CRD) is a comprehensive specification that defines the structure and format used to manage a cluster and its desired state. It contains the list of added modules and their state.
```
c-pius, line 3: Maybe strip the `is a comprehensive specification` part for all CRDs (including Kyma in user folder). It doesn't really say something I think. Just `[...] defines [...]` should be enough.

docs/contributor/resources/README.md
```diff
@@ -1,22 +1,11 @@
-# Lifecycle Manager API
+# Lifecycle Manager Resources
 
-## Overview
+The API of Lifecycle Manager is based on Kubernetes Custom Resource Definitions (CRDs), which  extend the Kubernetes API with custom additions. The CRDs allow Lifecycle Manager to manage clusters and modules. To inspect the specification of the Lifecycle Manager resources, see:
```
c-pius, line 3:
```suggestion
The API of Lifecycle Manager is based on Kubernetes Custom Resource Definitions (CRDs), which extend the Kubernetes API with custom additions. The CRDs allow Lifecycle Manager to manage clusters and modules. To inspect the specification of the Lifecycle Manager resources, see:
```

docs/user/01-10-kyma-crd.md
```diff
@@ -0,0 +1,11 @@
+# Kyma
```
c-pius, line 1: In contributors folder the CRD files don't have a `-crd` suffix. Intended?

docs/contributor/resources/04-watcher.md
```diff
@@ -0,0 +1,11 @@
+# Watcher
+
+The `watchers.operator.kyma-project.io` Custom Resource Definition (CRD) is a comprehensive specification that defines the structure and format used to configure the Watcher resource.
+
+The Watcher custom resource (CR) defines the callback functionality for synchronized remote clusters that allows lower latencies before the Control Plane detects any changes.
```
c-pius, line 5: This sentence makes me wonder if we should have a public glossary explaining terms like "synchronized remote cluster", "Control Plane", ...? We can decide on that later, but for now I would propose that we use `Kyma Runtime cluster` and `Kyma Control Plane cluster` instead as those are the terms that we also find in the architecture.md

docs/contributor/resources/03-moduletemplate.md
mmitoraj: As we agreed, let's rephrase to:

> The ModuleTemplate custom resource (CR) defines a module, in a particular version, that can be added to or deleted from the module list in the Kyma CR. Each ModuleTemplate CR represents one module.

docs/user/01-10-kyma-crd.md
mmitoraj: It was... as in the `contributor` they are in a directory called resources so we know where the files belong to and what they are about. But in the `user` it is the only doc now and there is no context. But I'm starting having second thoughts so we might want to change that later.

c-pius: But makes sense for me then. As long as there is a reasoning behind it I am not dogmatic about it :)

c-pius (dismissed):

c-pius (approved):

docs/contributor/resources/02-manifest.md
```diff
@@ -1,6 +1,14 @@
-# Manifest Custom Resource
+# Manifest
 
-The [Manifest custom resource (CR)](../../../api/v1beta2/manifest_types.go) is our internal representation of what results from the resolution of a ModuleTemplate CR in the context of a single cluster represented by a Kyma CR. Thus, a lot of configuration elements are similar or entirely equivalent to the data layer found in a ModuleTemplate CR.
+The `manifests.operator.kyma-project.io` Custom Resource Definition (CRD) defines the structure and format used to configure the Manifest resource.
+
+The Manifest custom resource (CR) represent resources that make up a module and are to be installed by Lifecycle Manager. The Manifest CR is a rendered module added to a particular cluster.
```
nataliasitko, line 5:
```suggestion
The Manifest custom resource (CR) represents resources that make up a module and are to be installed by Lifecycle Manager. The Manifest CR is a rendered module added to a particular cluster.
```

nataliasitko (approved):

Issue #1841, opened 2024-09-06
[ModuleCatalog] Add `.spec.info` to ModuleTemplate


c-pius:
### Description

In https://github.com/kyma-project/lifecycle-manager/issues/1681 it has been decided to add information about the module to the ModuleTemplate. We need to update the ModuleTemplate to include this information.

The information shall look as follows:

```yaml
spec:
  info:
    repository: <link>
    documentation: <link>
    icons:
    - name: <string>
      link: <link>
```

.spec.info is OPTIONAL (for compatibility reasons), but the individual entries are REQUIRED.

### Reasons

Enables CLI and dashboard to show additional info about the module

### Acceptance Criteria

- [x] `.spec.info` added to ModuleTemplate
- [x] Added an entry to https://github.com/kyma-project/lifecycle-manager/issues/776 that `.spec.info` must be added as required in a future major API version
- [x] Update documentation

PR #1906, opened 2024-09-30
feat: Add ModuleInfo as optional to ModuleTemplate

ruanxin:
**Description**

**Related issue(s)**
https://github.com/kyma-project/lifecycle-manager/issues/1841


api/v1beta2/moduletemplate_types.go
```diff
@@ -126,6 +126,30 @@ type ModuleTemplateSpec struct {
 
 	// CustomStateCheck is deprecated.
 	CustomStateCheck []*CustomStateCheck `json:"customStateCheck,omitempty"`
+
+	// Info is the information about the module.
+	// +optional
+	Info ModuleInfo `json:"info,omitempty"`
```
c-pius, line 132:
```suggestion
	// Info contains metadata about the module.
	// +optional
	Info ModuleInfo `json:"info,omitempty"`
```

Okay as is, but I like the way you phrased it in the docs.

api/v1beta2/moduletemplate_types.go
```diff
@@ -126,6 +126,30 @@ type ModuleTemplateSpec struct {
...
+type ModuleInfo struct {
+	// Repository is the link to the repository of the module.
+	Repository string `json:"repository"`
+
+	// Documentation is the link to the documentation of the module.
+	Documentation string `json:"documentation"`
+
+	// Icons is a list of icons associated with the module.
+	// +optional
+	Icons []ModuleIcon `json:"icons"`
```
c-pius, line 144:
```suggestion
	// Icons is a list of icons associated with the module.
	// +optional
	// +listType=map
	// +listMapKey=name
	Icons []ModuleIcon `json:"icons"`
```

To force unique names.

api/v1beta2/moduletemplate_types.go
```diff
@@ -126,6 +126,30 @@ type ModuleTemplateSpec struct {
...
+
+type ModuleInfo struct {
+	// Repository is the link to the repository of the module.
+	Repository string `json:"repository"`
+
+	// Documentation is the link to the documentation of the module.
+	Documentation string `json:"documentation"`
+
+	// Icons is a list of icons associated with the module.
+	// +optional
```
c-pius, line 143: I think in the ticket we didn't define this as optional. But I also think it makes sense to have this optional. I am just wondering if we should rather default to an empty list instead of optional? This may be easier for clients to handle as they can avoid the null checks then.

The same would then apply to the `resources` and `associatedResources` lists we introduce to in the other tickets. WDYT?

api/v1beta2/moduletemplate_types.go
```diff
@@ -126,6 +126,30 @@ type ModuleTemplateSpec struct {
...
+}
+
+type ModuleInfo struct {
+	// Repository is the link to the repository of the module.
+	Repository string `json:"repository"`
+
+	// Documentation is the link to the documentation of the module.
+	Documentation string `json:"documentation"`
+
+	// Icons is a list of icons associated with the module.
```
c-pius, line 142:
```suggestion
	// Icons is a list of icons of the module.
```

docs/contributor/resources/03-moduletemplate.md
```diff
@@ -55,6 +55,24 @@ spec:
...
+    repository: https://github.com/example/repo
+    documentation: https://docs.example.com
+    icons:
+    - name: example-icon
+      link: https://example.com/icon.png
+```
+
+- repository: The link to the repository of the module.
+- documentation: The link to the documentation of the module.
+- icons: A list of icons associated with the module, each with a name and link.
```
c-pius, line 74:
```suggestion
- icons: A list of icons of the module, each with a name and link.
```

c-pius (commented):
I think you also need to run the manifests make target to update the CRD.

api/v1beta2/moduletemplate_types.go
ruanxin: I think it's a generally common practice for those optional fields, we mark it as `+Optional`, for us, it's consistent to follow, and it's also clear to read in the CR compared to an empty entry`icons:[]`. For client tools, if they implement a solid solution, the schema of CRD should be their reference to check, it's easy to handle if some fields are declared but not presented in CR, I would not decrease the readability of CR for this.

I also asked to github Copilot, shared the same opinion
image

c-pius (approved):

docs/contributor/resources/03-moduletemplate.md
```diff
@@ -55,6 +55,24 @@ spec:
 
 If not specified, the **namespace** of the resource mentioned in **.spec.data** will be controlled by the `sync-namespace` flag; otherwise, it will be respected. All other attributes (including **.metadata.name**, **apiVersion**, and **kind**) are taken over as stated. Note that since it behaves similarly to a `template`, any subresources, such as **status**, are ignored, even if specified in the field.
 
+### **.spec.info**
+
+The `info` field contains metadata about the module, including repository, documentation, and icons. For example:
```
mmitoraj, line 60:
```suggestion
The **info** field contains module metadata, including the repository URL, documentation link, and icons. For example:
```

c-pius (dismissed):

mmitoraj (dismissed):

Issue #1842, opened 2024-09-06
[ModuleCatalog] Add `.spec.resources` to ModuleTemplate


c-pius:
### Description

In https://github.com/kyma-project/lifecycle-manager/issues/1681 it has been decided to add `resources` to ModuleTemplates. We need to update the ModuleTemplate to include this information.

The information shall look as follows:

```yaml
spec:
  resources:
  - name: <string>
    link: <link>
```

`spec.resources` is OPTIONAL (for compatibility reasons). The items in a resources entry are REQUIRED.

### Reasons

Enables CLI and dashboard to support in the manual installation of a module by downloading the needed resources and applying them to the cluster.

### Acceptance Criteria

- [x] `.spec.resources` added to ModuleTemplate
- [x] Added an entry to https://github.com/kyma-project/lifecycle-manager/issues/776 that `.spec.resources` must be added as required in a future major API version

janmedrek:
@c-pius wouldn't something like `.spec.rawManifest` make it more transparent what that section really stands for?

I am just questioning the naming, I would make sure it is self-explanatory when looking at the resource without the use of Busola.

c-pius:
@janmedrek sorry, your comment slipped through while I was on vacation. Actually, `.spec.resources` is a list of named items with a download link. Indeed, it is intended that this list contains the entry `rawManifest`. We chose to go for this "list container" instead of a direct `.spec.rawManifest` field so that we can add further resources in the future without doing additional API changes.

So it would look like:

![Image](https://github.com/user-attachments/assets/d4bd716f-0e87-4508-a23b-2aa98f0976d0)

PR #1910, opened 2024-10-01
feat: Add Resources as optional to ModuleTemplate

c-pius:
**Description**

- adds new optional `Resources` field to ModuleTemplate
- adds documentation for this list
- does NOT add tests since KLM does not make use of this field itself

**Related issue(s)**

- resolves https://github.com/kyma-project/lifecycle-manager/issues/1842


c-pius:
Docs to be added once https://github.com/kyma-project/lifecycle-manager/pull/1905 is merged:

> ### **.spec.resources**
> 
> The `resources` field is a list of additional resources of the module that can be fetched. As of now, the primary expected use case is for module teams to add a link to the raw manifest of the module.
>

api/v1beta2/moduletemplate_types.go
```diff
@@ -148,6 +154,14 @@ type ModuleTemplateList struct {
 	Items              []ModuleTemplate `json:"items"`
 }
 
+type Resource struct {
+	// Name is the name of the resource.
+	Name string `json:"name"`
```
nesmabadr, line 159: I was wondering if we want to somehow validate that if the resources array is not empty, then there should be an entry which has the name of "rawManifest"? I am not aware of any built in Kubebuilder markers that can validate this, but in general we do need to have this validation or?

c-pius: I think there would be "min items" (at least in JSON schema). But I was actually thinking about the same... The rawManifest is only necessary if the module supports manual installation which is not the case for all. But what we could do is to raise a warning in modulectl that "when missing rawManfiest in resources, the module template does not support manual installation".

c-pius: Also added this as a dicussion point to the modulectl related ticket

nesmabadr (approved):

mmitoraj (approved):

PR #1917, opened 2024-10-02
feat: Add proper categories for CRDs

a-thaler:
**Description**

- Add proper categories for telemetry CRDs according to linked ADR
- ...
- ...

**Related issue(s)**

- https://github.com/kyma-project/community/issues/912


c-pius (commented):
I think it appears a bit odd that KymaCR is listed when asking for "kyma-modules". After all, no hard opinion, but wouldn't it make sense to have:

- `kubectl get kyma` return the KymaCR
- `kubectl get kyma-modules` return all Kyma module CRs?

a-thaler:
The kyma CR contains the module config only at the moment, incl the state if the modules are synced. So I thought it might be relevant when asking for all modules. But I also have no hard oppinion here.
As a side note: `kubectl get kyma` already works as it is a CRD, so question is if we need this PR here at all to add the new category.

c-pius:
> As a side note: kubectl get kyma already works as it is a CRD, so question is if we need this PR here at all to add the new category.

yes correct. I will check again with the other jellies if there are strong opinions from there side. But from my side I would propose that we close this for now and if there is a desire or even use case in the future to include KymaCR in kyma-modules we will just re-open and merge it.

pbochynski:
> I will check again with the other jellies if there are strong opinions from there side. But from my side I would propose that we close this for now and if there is a desire or even use case in the future to include KymaCR in kyma-modules we will just re-open and merge it.

I also think we should not include kyma resource in kyma-modules category.

a-thaler:
So let's close the PR

PR #1918, opened 2024-10-02
chore: Also scan actions folder for dependabot actions upgrades

lindnerby:
**Description**

- Replace `directory` with `directories` as mentioned here:
https://docs.github.com/en/enterprise-cloud@latest/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directories
- Use workflows and actions folder

**Related issue(s)**

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,9 @@
 version: 2
 updates:
   - package-ecosystem: "github-actions"
-    directory: "/"
+    directories:
+      - "/workflows"
+      - "/actions"
     schedule:
       interval: "weekly"
     commit-message:


.github/dependabot.yml
```diff
@@ -3,7 +3,9 @@
 version: 2
 updates:
   - package-ecosystem: "github-actions"
-    directory: "/"
+    directories:
+      - "/workflows"
```
c-pius, line 7: just wondering if we shoudl keep this one at `"/"` because that's the one explicitly described in the docs.

lindnerby: Yes, we can do that. Also I think these folders need to be from root, since the documentation for` /` states it maps `/.github/workflows`

lindnerby: Pushed. PTAL

PR #1929, opened 2024-10-04
chore: Bump to latest runtime-watcher release 1.1.5 to be used in KLM…

lindnerby:
… release

**Description**

- Latest release of runtime-watcher img should be used: https://github.com/kyma-project/runtime-watcher/releases/tag/1.1.5

**Related issue(s)**


skr-webhook/resources.yaml
```diff
@@ -92,7 +92,7 @@ spec:
               value: "v1"
             - name: KCP_ADDR
               value: "kcp-base-url-invalid"
-          image: "europe-docker.pkg.dev/kyma-project/prod/runtime-watcher-skr:0.1.1"
+          image: "europe-docker.pkg.dev/kyma-project/prod/runtime-watcher:1.1.5"
```
c-pius, line 95: https://github.com/kyma-project/lifecycle-manager/blob/main/config/watcher/kustomization.yaml#L17-L21

here is where we actually set this. Should we somehow make this clear by removing this line or setting it to a placeholder value?

lindnerby: No idea how to do this quickly now since we want to release today. But maybe worth an investigation issue?

c-pius (approved):

Issue #1822, opened 2024-09-03
Make `watched-by` label more transparent


janmedrek:
### Description

The `watched-by` label is used by Runtime Watcher to query resources that should be monitored. Since we want to keep the feature and provide more transparency we would need to revisit its value.

The proposal is to switch from `watched-by: watcher` or `watched-by: lifecycle-manager` to `watched-by: kyma-runtime-watcher`.

### Reasons

We want to get rid of cryptic label messages in order to not confuse the users.

### Acceptance Criteria

- [x] Label's value switched from `watched-by: watcher`/`watched-by: lifecycle-manager` to `watched-by: kyma`.
- [x] Drop the `watched-by: module-manager` label

### Attachments

https://github.com/kyma-project/lifecycle-manager/blob/a180040ef2e6b8742a61ccc44eaaabae1dded8f5/config/watcher/operator_v1beta2_watcher.yaml#L9

https://github.com/kyma-project/lifecycle-manager/blob/9be2538f0201fe61eb7ac1f620242137f6178c74/internal/declarative/v2/default_transforms.go#L66

https://github.com/kyma-project/lifecycle-manager/blob/2f79d89435c8aba14e3043098116b04a59089853/internal/remote/skr_context.go#L212

c-pius:
I think this needs to be revisited... One STAGE we have now many other resources that are labelled with `watched-by`:

```
kubectl get all --all-namespaces -l operator.kyma-project.io/watched-by=kyma      
NAMESPACE     NAME                                   TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)    AGE
kyma-system   service/api-gateway-operator-metrics   ClusterIP   10.106.202.107   <none>        8080/TCP   43h
kyma-system   service/api-gateway-webhook-service    ClusterIP   10.105.74.52     <none>        9443/TCP   43h
kyma-system   service/btp-manager-metrics-service    ClusterIP   10.109.23.159    <none>        8080/TCP   43h
kyma-system   service/istio-operator-metrics         ClusterIP   10.109.86.88     <none>        8080/TCP   43h
kyma-system   service/warden-admission               ClusterIP   10.107.179.142   <none>        443/TCP    43h
kyma-system   service/warden-operator                ClusterIP   10.104.153.188   <none>        8443/TCP   43h

NAMESPACE     NAME                                             READY   UP-TO-DATE   AVAILABLE   AGE
[...]

PR #1945, opened 2024-10-09
chore: Remove watched-by label from manifest resources

c-pius:
**Description**

- removes the `watched-by` label from the "rawManifest" resources. Label stays only on the KymaCR in the SKR

**Related issue(s)**

- resolves https://github.com/kyma-project/lifecycle-manager/issues/1822

### DIFF
--- a/internal/declarative/v2/default_transforms.go
+++ b/internal/declarative/v2/default_transforms.go
@@ -39,11 +39,10 @@ func KymaComponentTransform(_ context.Context, obj Object, resources []*unstruct
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


c-pius:
To Reviewer: What would be a good e2e test for verifying that KymaCR has the Watched-By Label? Right now we don't have it tested. I am wondering, since labels become more important, should we add an e2e test focusing on labels, verifying expected labels on:

- KymaCR
- ModuleCR
- Resources from raw manifest
- (ValidatingWebhookConfiguration?)

ruanxin (approved):

Issue #1943, opened 2024-10-08
Split Image Build and E2E Test Workflows


c-pius:
### Description

Image build workflow now supports custom tags: https://github.com/kyma-project/test-infra/pull/11981

This allows us to refactor our pipelines to get rid of the issue wrt. to `pull_request_target`. Proposal is to have two separate workflows:

1) Triggers the image build and supplies the commit hash as custom tag
    - this workflow MUST continue to use `pull_request_target` as trigger (security requirement to get a GH token)
2) Waits for the built image with the commit hash as custom tag, once there, starts the rest of the E2E test workflow
    - this workflow MUST use `pull_request` as trigger (e2e tests and setup should run in the up to date proposed config)

### Reasons

Make our E2E tests work properly again

### Acceptance Criteria

- [x] Each E2E test run uses an up to date image of the current PR ref
- [x] Each E2E test run executes test code of the current PR ref

PR #1950, opened 2024-10-10
chore: Decouple build image job from e2e

ruanxin:
**Description**

The PR verifying: https://github.com/kyma-project/lifecycle-manager/pull/1948

**Related issue(s)**
https://github.com/kyma-project/lifecycle-manager/issues/1943

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -38,9 +38,6 @@ jobs:
               echo 'TAGS'
             } >> "$GITHUB_OUTPUT"
           fi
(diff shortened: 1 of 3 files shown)


c-pius (commented):
Should we add some documentation in the build-image.yml what image tags are built? Also considering the different triggers that we support?

c-pius (approved):
LGTM we may consider to to externalize the compute tag to a separate action that we can share between repos. But for the sake of getting this fix in fast I would say we do that in a follow-up.

Issue #1840, opened 2024-09-06
[ModuleCatalog] Add `.spec.manager` info to ModuleTemplate


c-pius:
### Description

In https://github.com/kyma-project/lifecycle-manager/issues/1681 it has been decided to add explicit information identifying a module's manager to the ModuleTemplate. We need to update the ModuleTemplate to include this information.

The information shall look as follows:

```
spec:
  manager:
    group: <group>
    version: <version>
    kind: <kind>
    namespace: <namespace>
    name: <name>
```

`.spec.manager` is OPTIONAL (for compatibility reasons), but the individual entries are REQUIRED, except Namespace which can be OPTIONAL due to CRDs not being namespaced. It should be checked whether existing K8s types and validations can be re-used.

### Reasons

Enables CLI and dashboard to determine whether a module is installed or not (manual installation case)

### Acceptance Criteria

- [x] re-used K8s types if possible
- [x] `.spec.manager` added to ModuleTemplate
- [x] Added an entry to https://github.com/kyma-project/lifecycle-manager/issues/776 that `.spec.manager` must be added as required in a future major API version

janmedrek:
Let's reconsider other options such as introducing a new resource just for CLI/UI monitoring purpose.

c-pius:
Discussed this quickly in the arch meeting today. PB (not present, only left a comment), is okay with the proposal that an arbitrary resource may be provided. If the resource is present, it is assumed that the module is installed in the cluster. We can make the namespace optional to support cluster scoped resources.

Still, the expected resource to be provided is the module manager (deployment). Should we nevertheless rename the field or keep manager?

PR #1958, opened 2024-10-14
feat: Add spec.manager

ruanxin:
**Description**

**Related issue(s)**
#1840


api/v1beta2/moduletemplate_types.go
```diff
@@ -138,11 +138,34 @@ type ModuleTemplateSpec struct {
 
 	// Info contains metadata about the module.
 	// +optional
-	Info ModuleInfo `json:"info,omitempty"`
+	Info *ModuleInfo `json:"info,omitempty"`
```
c-pius, line 141: Why is this change necessary?

api/v1beta2/moduletemplate_types.go
```diff
@@ -138,11 +138,34 @@ type ModuleTemplateSpec struct {
...
 
 	// AssociatedResources is a list of module related resources that usually must be cleaned when uninstalling a module. Informational purpose only.
 	// +optional
 	AssociatedResources []apimetav1.GroupVersionKind `json:"associatedResources,omitempty"`
+
+	// Manager contains information identifying a module's resource that indicates the readiness of the module. Typically, this should be the manager Deployment of the module.
```
c-pius, line 147:
```suggestion
	// Manager contains information for identifying a module's resource that can be used as indicator for the installation readiness of the module. Typically, this is the manager Deployment of the module. In exceptional cases, it may also be another resource. 
```

api/v1beta2/moduletemplate_types.go
```diff
@@ -138,11 +138,34 @@ type ModuleTemplateSpec struct {
...
+// Manager defines the structure for the manager field in ModuleTemplateSpec.
+type Manager struct {
+	// Group is the API group of the manager.
+	Group string `json:"group"`
+
+	// Version is the API version of the manager.
+	Version string `json:"version"`
+
+	// Kind is the kind of the manager.
+	Kind string `json:"kind"`
```
c-pius, line 161: Maybe it would be possible to re-use the K8s `GroupVersionKind` type (see also AssociatedResources field). We would need to add a marker that inlines the field: `json:",inline"`.

But I am also okay using them as is. For Namespace and Name I haven't found a proper existing type that can be re-used.

docs/contributor/resources/03-moduletemplate.md
```diff
@@ -75,6 +75,38 @@ spec:
 - documentation: The link to the documentation of the module.
 - icons: A list of icons of the module, each with a name and link.
 
+### **.spec.manager**
+
+The `manager` field provides details about a module's resource that indicates its readiness. Typically, this is the
+`Deployment` managing the module. If the resource is not namespaced, such as the module CR's `CRD`, the namespace can be
+optional.
```
c-pius, line 82:
```suggestion
The `manager` field provides information for identifying a module's resource that can be used as indicator for the installation readiness of the module. Typically, this is the manager Deployment of the module. In exceptional cases, it may also be another resource. If the resource is not namespaced, such as the module CR's `CRD`, the namespace can be optional.
```

docs/contributor/resources/03-moduletemplate.md
```diff
@@ -75,6 +75,38 @@ spec:
...
+For example, if the module has a manager `Deployment` in the `kyma-system` namespace:
+
+```yaml
+spec:
+  manager:
+    group: apps
+    version: v1
+    kind: Deployment
+    namespace: kyma-system
+    name: [module manager]
```
c-pius, line 93:
```suggestion
    name: [module manager name]
```

docs/contributor/resources/03-moduletemplate.md
```diff
@@ -75,6 +75,38 @@ spec:
...
+
+For a module without a manager, such as [cloud manager](https://github.com/kyma-project/cloud-manager):
+
+```yaml
+spec:
+  manager:
+    group: apiextensions.k8s.io
+    version: v1
+    kind: CustomResourceDefinition
+    name: [module CRD]
```
c-pius, line 104:
```suggestion
    name: [module CRD name]
```

docs/contributor/resources/03-moduletemplate.md
```diff
@@ -75,6 +75,38 @@ spec:
...
+```
+
+This field helps identify the module's manager, which is useful for determining whether a module is successfully
+installed, especially in manual installation scenarios.
```
c-pius, line 108: I think this should be obvious from the above description. If we consider this to be helpful, I would at least move it up to the rest of the description.

c-pius (commented):
I was thinking a bit about the name as well. I think staying with `manager` is okay as we also describe that this is our primary expected case. Also, I still have my doubts if Cloud Manager will even have a use case for manual installation. So I would say it is okay. If we really find that the name is bad, I think we could also rename it in a future API version or even introduce a conversion webhook replacing it with another field name.

api/v1beta2/moduletemplate_types.go
ruanxin: use pointer will make the actual code create this `ModuleInfo` a bit easier, with `nil` providing null field, and not necessarily creating an empty `ModuleInfo` struct.

docs/contributor/resources/03-moduletemplate.md
ruanxin: agree, removed

api/v1beta2/moduletemplate_types.go
ruanxin: good point, refactored

docs/contributor/resources/03-moduletemplate.md
```diff
@@ -75,6 +75,32 @@ spec:
...
+For a module without a manager, such as [cloud manager](https://github.com/kyma-project/cloud-manager):
+
+```yaml
+spec:
+  manager:
+    group: apiextensions.k8s.io
+    version: v1
+    kind: CustomResourceDefinition
+    name: [module CRD name]
+```
```
c-pius, line 103:
```suggestion
```

Thinking about it, maybe we should remove this example entirely. Just to not give the impression that this is something that should be done.

c-pius (dismissed):
Just one last thought but also LGTM as is.

docs/contributor/resources/03-moduletemplate.md
ruanxin: if we need to have this use case, I think we should mention it, for documentation purpose, it also brings value,  implicitely as hidden rule which applied for certain modules could lead to more confusion.

However, as for whether it make sense of making CRD as the manager, we can open a dicussion on this. It might be better to leave this field optional, assuming that for modules without a manager, they are considered ready by default. Until a decision is made, I believe we should keep this use case as a valid example

docs/contributor/resources/03-moduletemplate.md
```diff
@@ -75,6 +75,32 @@ spec:
 - documentation: The link to the documentation of the module.
 - icons: A list of icons of the module, each with a name and link.
 
+### **.spec.manager**
+
+The `manager` field provides information for identifying a module's resource that can be used as indicator for the installation readiness of the module. Typically, this is the manager Deployment of the module. In exceptional cases, it may also be another resource. If the resource is not namespaced, such as the module CR's `CRD`, the namespace can be optional.
```
mmitoraj, line 80:
```suggestion
The `manager` field provides information for identifying a module's resource that can indicate the module's installation readiness. Typically, the manager is the module's `Deployment` resource. In exceptional cases, it may also be another resource. The **namespace** parameter is optional if the resource is not namespaced. For example,  if the resource is the module CR's `CustomResourceDefinition`.
```

docs/contributor/resources/03-moduletemplate.md
```diff
@@ -75,6 +75,32 @@ spec:
...
+
+For example, if the module has a manager `Deployment` in the `kyma-system` namespace:
```
mmitoraj, line 82:
```suggestion
In this example, the module's manager is the `Deployment` resource in the `kyma-system` namespace.
```

docs/contributor/resources/03-moduletemplate.md
```diff
@@ -75,6 +75,32 @@ spec:
...
+spec:
+  manager:
+    group: apps
+    version: v1
+    kind: Deployment
+    namespace: kyma-system
+    name: [module manager name]
+```
+
+For a module without a manager, such as [cloud manager](https://github.com/kyma-project/cloud-manager):
```
mmitoraj, line 94:
```suggestion
In this example, the module doesn't have a manager and in the **manager** field you specifie the module's `CustomResourceDefinition`. For example, of the [Cloud Manager](https://github.com/kyma-project/cloud-manager) module:
```
I'm just not sure if it means that there is no manager or if the manager is the CRD?

docs/contributor/resources/03-moduletemplate.md
```diff
@@ -75,6 +75,32 @@ spec:
...
+spec:
+  manager:
+    group: apps
+    version: v1
+    kind: Deployment
+    namespace: kyma-system
+    name: [module manager name]
+```
+
+In this example, the module doesn't have a manager and in the **manager** field you specifie the module's `CustomResourceDefinition`. For example, of the [Cloud Manager](https://github.com/kyma-project/cloud-manager) module:
```
mmitoraj, line 94:
```suggestion
In this example, the module's manager is the module's `CustomResourceDefinition` that does not require the **namespace** parameter.
```

c-pius (approved):

mmitoraj (approved):

PR #1966, opened 2024-10-16
chore: Replace Kyma CLI with modulectl in e2e

lindnerby:
**Description**

- Replace kyma cli action with modulectl
- Replace cmd usage

**Related issue(s)**


.github/actions/deploy-template-operator-with-modulereleasemeta/action.yml
```diff
@@ -19,7 +19,7 @@ runs:
       shell: bash
       run: |
         make build-manifests
-        kyma alpha create module --module-config-file ./module-config.yaml --path . --registry localhost:5111 --insecure
+        module create --module-config-file ./module-config.yaml --registry http://localhost:5111 --insecure
```
nesmabadr, line 22: We need to modify the module-config.yaml in the template-operator to have the defaultCR and manifest from the release URLs instead of the local file paths.

.github/actions/deploy-template-operator/action.yml
```diff
@@ -19,7 +19,7 @@ runs:
       shell: bash
       run: |
         make build-manifests
-        kyma alpha create module --module-config-file ./module-config.yaml --path . --registry localhost:5111 --insecure
+        module create --module-config-file ./module-config.yaml --registry http://localhost:5111 --insecure
```
nesmabadr, line 22: I suggest that we keep the Kyma CLI for this action since we agreed that the module teams will still be using the CLI until they have their modulereleasemeta in place. Also, this action will be removed after all modules are already using the modulereleasemeta.

Also, one more thing, using the modulectl generates the new format of the ModuleTemplate (and later on a different name <modulename-moduleversion>), and this will not be similar to the real modules' ModuleTemplates on landscapes

.github/actions/get-configuration/action.yml
```diff
@@ -29,6 +32,7 @@ runs:
         echo "k8s_version=${{ github.event.inputs.k8s_version || '1.30.3' }}" >> $GITHUB_OUTPUT
         echo "istio_version=1.20.3" >> $GITHUB_OUTPUT
         echo "k3d_version=5.7.4" >> $GITHUB_OUTPUT
+        echo "modulectl_version=1.0.0" >> $GITHUB_OUTPUT
```
nesmabadr, line 35: I think we need to document somewhere that we need to bump the modulectl version here when we do a new modulectl release, otherwise it might be forgotten?

docs/contributor/04-local-test-setup.md
```diff
@@ -104,8 +104,7 @@ This setup is deployed with the following security features enabled:
    make build-manifests
    
    # create the a ModuleTemplate CR and save it to the template.yaml file
-   kyma alpha create module -p ./ --version 1.2.3 \
-   --registry k3d-registry.localhost:5111 --insecure --module-config-file ./module-config.yaml
+   modulectl create --module-config-file ./module-config.yaml --registry http://k3d-registry.localhost:5111 --insecure
```
nesmabadr, line 107: If the PR for shortening the config file flag gets merged first, please make sure to update the flag here

nesmabadr:
One more note, we need to recreate the ModuleTemplates used for the E2E tests (https://github.com/kyma-project/lifecycle-manager/tree/main/tests/moduletemplates)

docs/contributor/04-local-test-setup.md
lindnerby: We are using released version here. Changes to the modulectl won't affect KLM e2e until we release a new version and use it the modulectl action.

.github/actions/get-configuration/action.yml
lindnerby: I guess if we want to use new features of it we need to bump it anyways?

nesmabadr: yes, but with the Kyma CLI we used the latest version, so it is a new thing to update the modulectl version for the E2E tests

lindnerby: Alright. Any proposals where to document this?

mmitoraj (approved):

.github/actions/get-configuration/action.yml
nesmabadr: Is the modulectl release documented somewhere? Maybe we can add it there?

lindnerby: I don't we have documents about modulectl yet.

.github/actions/deploy-template-operator/action.yml
nesmabadr: As agreed, we will keep using the modulectl here to remove the Kyma-cli actions, and the ModuleTemplates generated by kyma CLI (the ones in tests/moduletemplates) are already used in some tests, so we still cover cases with the old OCM layers format

.github/actions/deploy-template-operator-with-modulereleasemeta/action.yml
nesmabadr: This will be done when we bump the modulectl version here since we already deprecated the use of local file paths

nesmabadr (approved):

PR #1721, opened 2024-07-25
refactor: Remove Redundancy and duplication, check for same naming pattern on Kustomize

jeremyharisch:
**Description**

Changes proposed in this pull request:

- Removing the term `manager` from the kustomize setup, since it is duplicated. I.e. `klm-controller-manager-role` which reads as `kyma-lifecycle-manager-controller-manager-role`, or in default profile `lifecycle-manager-manager-role`. Use only `controller` instead
- Make sure everything has `klm-` prefix in control-plane setup
- Align the resources to the following format `<prefix>-<resource type>-<details>`
    - `klm-leader-election-role` &#8594; `klm-controller-role-leader-election`
    - `klm-manager-controler-role` &#8594; `klm-controller-role` 
    - ...
- Removing manifest-role and use kubebuilder rbac generator as done for other resoruces - to get rid of roles created by ahdn use use kubebuilder framwork as much as possible
- Adapt E2E Test

<details>
  <summary>Old resources</summary>
  
  ### default
```yaml
kcp-system
---
kymas.operator.kyma-project.io
---
manifests.operator.kyma-project.io
---
moduletemplates.operator.kyma-project.io
---
watchers.operator.kyma-project.io
---
lifecycle-manager-controller-manager
---
lifecycle-manager-leader-election-role
---
lifecycle-manager-manager-role
---
lifecycle-manager-manager-role-crd
---
lifecycle-manager-manager-role-manifest
---
lifecycle-manager-metrics-reader
---
lifecycle-manager-leader-election-rolebinding
---
lifecycle-manager-manager-rolebinding
---
lifecycle-manager-manager-rolebinding-crd
---
lifecycle-manager-manager-rolebinding-manifest
---
lifecycle-manager-metrics-rolebinding
---
lifecycle-manager-metrics-service
---
lifecycle-manager-webhook-service
---
lifecycle-manager-controller-manager
---
lifecycle-manager-serving-cert
---
lifecycle-manager-selfsigned-issuer
---
lifecycle-manager-validating-webhook-configuration
```

### control-plane profile

```
kymas.operator.kyma-project.io
---
manifests.operator.kyma-project.io
---
moduletemplates.operator.kyma-project.io
---
watchers.operator.kyma-project.io
---
klm-controller-manager
---
klm-leader-election-role
---
klm-manager-role
---
klm-manager-role-crd
---
klm-manager-role-manifest
---
klm-metrics-reader
---
klm-manager-rolebinding-istio-system
---
klm-leader-election-rolebinding
[...]

PR #2004, opened 2024-11-03
feat: Introduce Dedicated Istio Gateway Secret

### LINKED ISSUE #1890: feat: [Zero-Downtime] - safe migration
### Description

In order to introduce the [zero-downtime](https://github.com/kyma-project/lifecycle-manager/issues/1430) procedure, we need a safe migration path.
By "safe migration path" I understand a setup, where a "revert" to the old behavior is as simple as possible - in case there is a bug in the new solution or it is not working as expected for whatever reason.
In particular, the revert should be as simple as switching some Lifecycle-Manager flags (runtime arguments) - the less, the better.

#### An Idea for how this could work:
The current solution uses root certificate secret as the Istio-Gateway secret directly.
Once the root cert is rotated, the LM code deletes the client secrets (that are based on the "previous" root), causing cert-manager to renew them.
The new ("zero-downtime") solution uses a dedicated secret for the Istio-Gateway, managed entirely by Lifecycle-Manager.
This new secret of course has a different name from the "root" secret, that is still managed and rotated by the cert-manager.
The dedicated secret decouples Istio-Gateway from changes to the root certificate - it is Lifecycle-Manager that decides when to propagate the changes.
How can we revert from the new solution to the old one?
If the secret name used for Istio-Gateway is different in the old (current) and new (future) solution, then in order to switch back we would have to change the Helm Charts, or at least the entry in the `values.yaml`. In addition we would probably need to deploy a different LM version - the one with the "old" logic. But then we're reverting the Lifecycle-Manager version, along with all the other features, security fixes etc.

To improve the situation, we should:
- have a LM version that is capable of running either the old or the new code - depending on some feature flag.
- the Istio Gateway secret name should be **the same** in both scenarios, to avoid manual tweaks in the Helm charts.

The first requirement is relatively easy - we just need to extract the relevant cert-management logic to a component and then provide two different implementations: the old one (current) and the new one. And we need a flag to decide which component should be actually used at runtime.

The second requirement is more tricky. In order to make it work, we should change the Lifecycle-Manager in the following way (it's just an idea, maybe it can be done in a simpler way):
- The root certificate is no longer directly used to configure Istio Gateway
- We introduce our own, managed secret that is a copy of the Istio-Gatewy secret.
- There is a new "agent" that actively syncs the root secret to the copy.
- Minimal changes in the code are required - The reconciliation logic no longer inspects the timestamps on the root secret, it inspects the created copy instead
- Kustomize configuration of the Istio Gateway (and Helm charts) are changed so that the new secret is used instead of the "root" one. 

By introducing this solution ("safe-migration"), we achieve the following:
- Both the "safe-migration" and "final" solutions use **the same secret name** for the Istio-Gateway. Hence, when reverting, there's no need to change anything in the Helm charts.
- The revert can be accomplished by switching a single boolean flag on the Lifecycle-Manager, like: `--cert-mangament-legacy=true`, assuming no additional configuration flags are required.
- Don't worry about the "copy" or the new "agent" - in the final solution these also exists. And this issue introduces just a temporary solution. Once the final solution works, we'll remove the support for the "old" logic entirely.

Implementation notes for the syncing "agent" - for lack of a better name, let's call it: "Istio Gateway Secret Manager"
- What we really need is something that watches the root secret and acts upon changes on this object
- So it can be a single goroutine, a mini-controller, etc. Maybe even something else - a little research is required.
- See that although the secret name is the same, the data in the secret differs. The "safe-migration" solution makes IstioGatewaySecret just a copy of the root secret, while the final solution uses a **modified** (newRoot+oldRoot) values for the caBundle in the IstioGatewaySecret.

See also: https://github.com/kyma-project/lifecycle-manager/issues/1506
### Reasons

Safe migration path - ability to revert **fast** and with minimal risk of doing it wrong - in case of troubles.
### Implementation notes:
- introduce a new mechanism for copying the secret. Can be a single goroutine with a watcher/informer. It should have a low-frequency polling in case k8s event is lost (AFAIK k8s events are not 100% reliable)
- modify the condition upon which the skr watcher client secrets are removed: https://github.com/kyma-project/lifecycle-manager/blob/f0ac9bb8aa0c8770409b1123e2c4fbf4d75980f4/pkg/watcher/certificate_manager.go#L299
- if necessary, consider adding some annotation on the Istio Gateway Secret (IGS) object to make "rotation detection" in the code easy.

### Acceptance Criteria

- [x] Double-check the proposed solution if it's not missing anything important
- [ ] Verify with security team if the LM-managed secret is OK (we need this anyway for the "new solution")
- [x] introduce a secret manager component that does a copy
- [x] change the [condition for watcher client secret removal](https://github.com/kyma-project/lifecycle-manager/blob/main/pkg/watcher/certificate_manager.go#L299) so that it involves updates to the Istio Gataway Secret (which should be an up-to-date copy of the Root Secret)

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

### Related Issues:
https://github.com/kyma-project/lifecycle-manager/issues/1430

LeelaChacha:
Closes #1890


pkg/zerodw/gateway.go
```diff
@@ -0,0 +1,137 @@
+package zerodw
```
ruanxin, line 1: please rename this package to something readable, I don't know what "dw" means

config/watcher/gateway.yaml
```diff
@@ -20,5 +20,5 @@ spec:
         number: 443
         protocol: HTTPS
       tls:
-        credentialName: klm-watcher
+        credentialName: istio-gateway-secret
```
ruanxin, line 23: Can you shortly explain why we have to change the credentailName here? also in this PR, you name it [gateway-secret](https://github.com/kyma-project/lifecycle-manager/pull/1871/files#diff-076f5cb5879121781dc0b5aaf99c461555cbab02b46a9b13b94482242f0aa985R23), and we previously have issue regarding alignment of resource name, https://github.com/kyma-project/lifecycle-manager/pull/1721#issuecomment-2256185803, please also have a reference to that decision for this renaming.

ruanxin: If it make sense to have a short meeting to sync the decision, please don't hesitate bring it to team, you can sync this with @Tomasz-Smelcerz-SAP first.

LeelaChacha: would `klm-istio-gateway` be more fitting then?

pkg/watcher/certificate_manager.go
```diff
@@ -318,6 +292,14 @@ func (c *CertificateManager) RemoveSecretAfterCARotated(ctx context.Context, kym
 	return nil
 }
 
-func certificateRenewalTimePassed(certStatus certmanagerv1.CertificateStatus) bool {
-	return certStatus.RenewalTime.Before(&(apimetav1.Time{Time: time.Now()}))
+func isGatewaySecretNewerThanWatcherCert(gatewaySecret *apicorev1.Secret, watcherSecret *apicorev1.Secret) bool {
```
ruanxin, line 295: I would like to see a unit test for this function

pkg/watcher/skr_webhook_manifest_manager.go
```diff
@@ -87,9 +85,17 @@ func (m *SKRWebhookManifestManager) Install(ctx context.Context, kyma *v1beta2.K
 		return fmt.Errorf("failed to get skrContext: %w", err)
 	}
 
+	_, err = getCertificateSecret(ctx, m.kcpClient, client.ObjectKey{
```
ruanxin, line 88: you can save a few lines if put this into if condition as `if _, err := getCertificateSecret(...); err != nil`

pkg/watcher/certificate_manager.go
LeelaChacha: good point!

pkg/watcher/skr_webhook_manifest_manager.go
```diff
@@ -237,13 +246,24 @@ func (m *SKRWebhookManifestManager) getUnstructuredResourcesConfig(ctx context.C
...
 		skrWatcherImage: m.config.SkrWatcherImage,
-		caCert:          tlsSecret.Data[caCertKey],
+		caCert:          gatewaySecret.Data[caCertKey],
 		tlsCert:         tlsSecret.Data[tlsCertKey],
 		tlsKey:          tlsSecret.Data[tlsPrivateKeyKey],
 		remoteNs:        remoteNs,
 	}, nil
 }
 
+func getCertificateSecret(ctx context.Context, clnt client.Client, objKey client.ObjectKey) (*apicorev1.Secret, error) {
```
ruanxin, line 256: all of objKey are: `Namespace: m.certificateConfig.IstioNamespace, Name: gatewaysecret.GatewaySecretName,`, right? why not design this function belongs to `CertificateManager` as `func (c *CertificateManager) getCertificateSecret(ctx context.Context)`

ruanxin: I expected a function comment to explain the purpose of this secret.

cmd/main.go
```diff
@@ -201,12 +204,21 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
 	go cleanupStoredVersions(flagVar.DropCrdStoredVersionMap, mgr, setupLog)
 	go scheduleMetricsCleanup(kymaMetrics, flagVar.MetricsCleanupIntervalInMinutes, mgr, setupLog)
 
+	setupIstioGatewaySecretRotation(config, kcpClient, setupLog)
```
ruanxin, line 207: I would prefer put the goruntine declaration here so that aligns with above method and has a clear view that all of them are go rountine.

pkg/gatewaysecret/secret.go
```diff
@@ -0,0 +1,61 @@
...
+	err := sm.kcpClient.Get(ctx, objKey, secret)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get secret %s: %w", objKey.Name, err)
+	}
+
+	return secret, nil
+}
+
+// isNotFound returns true if the error is a NotFound error.
+func isNotFound(err error) bool {
```
ruanxin, line 33: we have utility isNotFound https://github.com/kyma-project/lifecycle-manager/blob/0d91106f187024f74f65846cf78222d8ad471079/pkg/util/error.go#L15

reuse that please

pkg/gatewaysecret/handler.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (gsh *GatewaySecretHandler) handleExisting(rootSecret *apicorev1.Secret, gwSecret *apicorev1.Secret) error {
+	caCert := certmanagerv1.Certificate{}
+	if err := gsh.kcpClient.Get(context.Background(),
+		client.ObjectKey{Namespace: istioNamespace, Name: kcpCACertName},
+		&caCert); err != nil {
+		return fmt.Errorf("failed to get CA certificate: %w", err)
+	}
+
+	if gwSecretLastModifiedAtValue, ok := gwSecret.Annotations[LastModifiedAtAnnotation]; ok {
```
ruanxin, line 65: I would expect this get last modified timestamp is a function from `secretManager`

pkg/gatewaysecret/handler.go
```diff
@@ -0,0 +1,137 @@
...
+
+const (
+	GatewaySecretName = "istio-gateway-secret"
+	kcpRootSecretName = "klm-watcher"
+	kcpCACertName     = "klm-watcher-serving"
+	istioNamespace    = flags.DefaultIstioNamespace
+)
+
+type GatewaySecretHandler struct {
+	*secretManager
```
ruanxin, line 28: Why need this wrapper struct only for `secretManager`, can you just use secretManager do all the thing?

pkg/gatewaysecret/handler.go
```diff
@@ -0,0 +1,137 @@
...
+		Data: map[string][]byte{
+			"tls.crt": rootSecret.Data["tls.crt"],
+			"tls.key": rootSecret.Data["tls.key"],
+			"ca.crt":  rootSecret.Data["ca.crt"],
+		},
+	}
+	return gwSecret
+}
+
+func WatchChangesOnRootCertificate(clientset *kubernetes.Clientset, gatewaySecretHandler *GatewaySecretHandler,
```
ruanxin, line 105: I expect the scenario you designed for this function fully tested in an dedicated e2e test, for example, the root cert get updated, test your watched one should get new secret synced, right?

pkg/gatewaysecret/handler.go
```diff
@@ -0,0 +1,137 @@
...
+
+		switch event.Type {
+		case watch.Added:
+			fallthrough
+		case watch.Modified:
+			err := gatewaySecretHandler.ManageGatewaySecret(rootCASecret)
+			if err != nil {
+				log.Error(err, "unable to manage istio gateway secret")
+			}
+		case watch.Deleted:
```
ruanxin, line 127: if it's deleted, we don't clean up the secret?

tests/e2e/ca_certificate_rotation_test.go
```diff
@@ -34,17 +33,21 @@ var _ = Describe("CA Certificate Rotation", Ordered, func() {
 			Namespace: RemoteNamespace,
 		}
 		It("Then KCP TLS Certificate is removed", func() {
-			timeNow := &apimetav1.Time{Time: time.Now()}
-			expectedLogMessage := "CA Certificate was rotated, removing certificate"
+			var err error
+			namespacedSecretName := types.NamespacedName{
+				Name:      watcher.ResolveTLSCertName(kyma.Name),
+				Namespace: "istio-system",
```
ruanxin, line 39: think about put this `istio-system` as a const and shared within e2e test.

pkg/gatewaysecret/handler.go
LeelaChacha: thats is an undefined state. it should never be deleted AFAIK. Even if there is a workflow in the cefrt manager to delete it, it would be added again, that's when we'll tigger the update again. Note that secret being deleted doesn't mean that the copied gateway secret is invalid. It expires only at the `notAfter` date

pkg/gatewaysecret/handler.go
LeelaChacha: but that is already being tested in the ca rotation test. The root cert is set to rotate every 1 minute and we check if the skr cert is rotated. so we are are also checking the rotation of the gateway by proxy. I would suggest that we explicitly check the gateway secret in same test, rather than a dedicated test. WDZT? @ruanxin

pkg/gatewaysecret/handler.go
ruanxin: then please write a code comment here to explain this case.

pkg/gatewaysecret/handler.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (gsh *GatewaySecretHandler) handleNonExisting(rootSecret *apicorev1.Secret) error {
+	gwSecret := gsh.newGatewaySecret(rootSecret)
+	return gsh.create(context.Background(), gwSecret)
+}
+
+func (gsh *GatewaySecretHandler) handleExisting(rootSecret *apicorev1.Secret, gwSecret *apicorev1.Secret) error {
+	caCert := certmanagerv1.Certificate{}
+	if err := gsh.kcpClient.Get(context.Background(),
```
ruanxin, line 59: we should not use Background context, check the function comment, this context will never cancelled.

config/watcher/gateway.yaml
ruanxin: I would vote for `klm-istio-gateway`

pkg/watcher/skr_webhook_manifest_manager.go
LeelaChacha: This function is used in 5 places. Sometimes for the gateway secret and sometimes for the watcher secret. Hence was made generic

config/watcher/gateway.yaml
LeelaChacha: I'll use that for now, but let's discuss it briefly after a daily.

pkg/watcher/skr_webhook_manifest_manager.go
LeelaChacha: makes sense!

pkg/watcher/certificate_manager_test.go
```diff
@@ -0,0 +1,109 @@
...
+					ObjectMeta: apimetav1.ObjectMeta{
+						Annotations: map[string]string{
+							"lastModifiedAt": "2024-11-01T00:00:00Z",
+						},
+					},
+				},
+				watcherSecret: &apicorev1.Secret{
+					ObjectMeta: apimetav1.ObjectMeta{
+						CreationTimestamp: apimetav1.Time{
+							Time: time.Date(2024, 11, 0o1, 0, 0, 5, 0, time.UTC),
```
ruanxin, line 36: why use `0o1`? can't be `0`?

tests/e2e/rbac_privileges_test.go
```diff
@@ -178,7 +178,7 @@ var _ = Describe("RBAC Privileges", func() {
 				{
 					APIGroups: []string{""},
 					Resources: []string{"secrets"},
-					Verbs:     []string{"list", "watch", "create", "delete"},
+					Verbs:     []string{"list", "watch", "create", "delete", "update"},
```
ruanxin, line 181: please also update this permission in `management-plane-charts` lifecycle-manager/files/roles/klm-watcher-certmanager-role.yaml before release

cmd/main.go
ruanxin: what I mean is just like this 
```
	go cleanupStoredVersions(flagVar.DropCrdStoredVersionMap, mgr, setupLog)
	go scheduleMetricsCleanup(kymaMetrics, flagVar.MetricsCleanupIntervalInMinutes, mgr, setupLog)

	go setupIstioGatewaySecretRotation(config, kcpClient, setupLog)
}

func setupIstioGatewaySecretRotation(config *rest.Config, kcpClient *remote.ConfigAndClient, setupLog logr.Logger) {
	kcpClientset := kubernetes.NewForConfigOrDie(config)
	gatewaySecretHandler := gatewaysecret.NewGatewaySecretHandler(kcpClient)
	gatewaysecret.WatchChangesOnRootCertificate(kcpClientset, gatewaySecretHandler, setupLog)
}
```

tests/e2e/suite_test.go
```diff
@@ -32,6 +32,7 @@ import (
 const (
 	kcpConfigEnvVar, skrConfigEnvVar = "KCP_KUBECONFIG", "SKR_KUBECONFIG"
 	clientQPS, clientBurst           = 1000, 2000
+	istioNamespace                   = "istio-system"
```
ruanxin, line 35: actually I have a better place to declaire namespace used for tests, https://github.com/kyma-project/lifecycle-manager/pull/2030/files#diff-5dab9c6f73d5f7fde3503baa47c3f8ef0528addf1143961dac081556181d7ebbR15-R16, if you PR is merged after that one, please update accordingly, otherwise, do this in another PR.

pkg/testutils/watcher.go
```diff
@@ -104,3 +151,30 @@ func GetCACertificate(ctx context.Context, namespacedCertName types.NamespacedNa
...
+	if err != nil {
+		return fmt.Errorf("failed to get gateway secret %w", err)
+	}
+
+	currentTime, err := gatewaysecret.GetValidLastModifiedAt(gwSecret)
+	if err != nil {
+		return fmt.Errorf("failed to get last modified time %w", err)
+	}
+
+	if currentTime != oldTime {
```
ruanxin, line 176: I think to only compare different is not solid enough, currentTime should larger than oldTime.

config/manager/kustomization.yaml
```diff
@@ -10,5 +10,5 @@ generatorOptions:
 
 images:
 - name: controller
-  newName: europe-docker.pkg.dev/kyma-project/prod/lifecycle-manager
-  newTag: latest
+  newName: k3d-kcp-registry.localhost:5000/lifecycle-manager
```
Tomasz-Smelcerz-SAP, line 13: looks like a leftover from local testing?

pkg/watcher/skr_webhook_manifest_manager.go
```diff
@@ -87,9 +85,13 @@ func (m *SKRWebhookManifestManager) Install(ctx context.Context, kyma *v1beta2.K
 		return fmt.Errorf("failed to get skrContext: %w", err)
 	}
 
+	if _, err = gatewaysecret.GetGatewaySecret(ctx, m.kcpClient); err != nil {
```
Tomasz-Smelcerz-SAP, line 88: Isn't that redundant? Later on, this value is required anyway so we make another call like this.

pkg/watcher/certificate_manager.go
```diff
@@ -279,35 +277,17 @@ func (e *CertificateNotReadyError) Error() string {
...
 func (c *CertificateManager) RemoveSecretAfterCARotated(ctx context.Context, kymaObjKey client.ObjectKey) error {
-	caCertificateStatus, err := c.GetCACertificateStatus(ctx)
+	gatewaySecret, err := gatewaysecret.GetGatewaySecret(ctx, c.kcpClient)
 	if err != nil {
-		return fmt.Errorf("error while fetching CA Certificate: %w", err)
+		return err
 	}
-
-	certSecret, err := c.getCertificateSecret(ctx)
+	watcherCert, err := c.getCertificateSecret(ctx)
```
Tomasz-Smelcerz-SAP, line 285: I think it should be renamed to `watcherSecret`, as it is a secret object.

pkg/watcher/certificate_manager.go
```diff
@@ -279,35 +277,17 @@ func (e *CertificateNotReadyError) Error() string {
...
-		c.caCertCache.SetCACertToCache(caCert)
-		return caCert.Status, nil
-	}
-
-	return cachedCertStatus, nil
-}
-
 func (c *CertificateManager) RemoveSecretAfterCARotated(ctx context.Context, kymaObjKey client.ObjectKey) error {
-	caCertificateStatus, err := c.GetCACertificateStatus(ctx)
+	gatewaySecret, err := gatewaysecret.GetGatewaySecret(ctx, c.kcpClient)
```
Tomasz-Smelcerz-SAP, line 281: This object will be fetched very frequently. In the previous implementation we used an explicit cache. Is this call cached properly?

pkg/gatewaysecret/handler.go
```diff
@@ -0,0 +1,182 @@
+package gatewaysecret
```
Tomasz-Smelcerz-SAP, line 1: There are no unit tests in this package.
As we try to increase our code quality, it would be nice to provide some tests here.
A hint for what can be tested: I'd include the secret-watching logic, all time comparisons here and also that the "lastModifiedAt" annotation is always present on the secret.
Remember that the unit tests should not just be written to "test a scenario": They should be written with an intention to document the expected behavior of the unit (in our case, the unit is usually a single function/method)

pkg/gatewaysecret/handler.go
```diff
@@ -0,0 +1,182 @@
...
+	return time.Time{}, errCouldNotGetLastModifiedAt
+}
+
+func WatchChangesOnRootCertificate(clientset *kubernetes.Clientset, gatewaySecretHandler *GatewaySecretHandler,
+	log logr.Logger,
+) {
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+
+	secretWatch, err := clientset.CoreV1().Secrets(istioNamespace).Watch(ctx, apimetav1.ListOptions{
```
Tomasz-Smelcerz-SAP, line 151: In this point a strong compile-time dependency on k8s library code is introduced.
It makes it impossible to properly unit-test our logic contained in the `for event := ...` loop.

pkg/watcher/certificate_manager.go
LeelaChacha: all secret objects are [cached by the Kyma reconciler](https://github.com/kyma-project/lifecycle-manager/blob/39caf9bb6be863211bde44c17b392cd296bfabb2/internal/controller/kyma/setup.go#L67)

pkg/watcher/skr_webhook_manifest_manager.go
LeelaChacha: Yes, it was a part of the issue that the watcher secret is only created if the gateway secret is ready.

Later on, we fetch the gateway secret to check if the watcher secret needs to be updated; if an error occurs there, we don't update the watcher secret, but it will be created.
Therefore, stopping the watcher's secret from being created altogether when the gateway is not ready makes sense and is not redundant.

pkg/gatewaysecret/handler.go
LeelaChacha: I have added extensive unit tests for all the functions in `handler.go`. The struct now has a dependency injection to make it more testable. PTAL since I am not sure if my testing and mocking approach is a standard one.

pkg/watcher/skr_webhook_manifest_manager.go
Tomasz-Smelcerz-SAP: ```Later on, we fetch the gateway secret to check if the watcher secret needs to be updated;```
I still don't get it. We fetch the secret in two places in this file: line 88 and 235. This places are quite "close" to each other with respect to actual execution. Can't we use the object fetched in line 88 also in line 235?
I mean, if there is a justified likelyhood that this secret somehow changes between lines 88 and 235, then it's fine to fetch it again.
Otherwise it is a code smell: Somehow the current code structure makes it hard for us to fetch the object just once and then use it.
If you don't know how to fix it in the current codebase (Or you don't want to do it in this PR), I'd suggest to create an issue dedicated to refactoring of this "double lookup".

pkg/gatewaysecret/handler_test.go
```diff
@@ -0,0 +1,351 @@
...
+						NotBefore: &apimetav1.Time{
+							Time: time.Date(2024, 11, 1, 0, 0, 5, 0, time.UTC),
+						},
+					},
+				},
+			},
+			want: true,
+		},
+		{
+			name: "gateway secret is newer but has parsing error for lastModifiedAt",
```
Tomasz-Smelcerz-SAP, line 150: When there is a parsing error, we don't know if it's newer. Maybe `gateway secret has parsing error for lastModifiedAt` is enough. But that is a very important test case, it documents that when there is a problem with `lastModifiedAt`, we consider this as "requires update", which makes sense, as the update should "fix" the invalid condition.

pkg/gatewaysecret/handler.go
```diff
@@ -0,0 +1,142 @@
...
+		return nil
+	}
+
+	return gsh.Update(ctx, gwSecret)
+}
+
+func CopyRootSecretDataIntoGatewaySecret(gwSecret *apicorev1.Secret, caCert certmanagerv1.Certificate,
+	rootSecret *apicorev1.Secret,
+) bool {
+	if !GatewaySecretRequiresUpdate(gwSecret, caCert) {
```
Tomasz-Smelcerz-SAP, line 75: Following the "Single Responsibility Principle" I don't think putting this condition here is a good choice.
First, it complicates the unit test: You have to setup some timestamps just to test if "copy" works...
Second, it complicates the signature of this function, `caCert` is used only to test if copying is required.
I would move this check to the place where `CopyRootSecretDataIntoGatewaySecret` is invoked (i.e. the body of the `handleExisting` method)

pkg/gatewaysecret/handler_test.go
```diff
@@ -0,0 +1,351 @@
...
+		},
+		Data: map[string][]byte{
+			"tls.crt": []byte(oldTLSCertValue),
+			"tls.key": []byte(oldTLSKeyValue),
+			"ca.crt":  []byte(oldCACertValue),
+		},
+	}
+
+	// Newer than gateway secret
+	caCert := certmanagerv1.Certificate{
```
Tomasz-Smelcerz-SAP, line 198: I think the `CopyRootSecretDataIntoGatewaySecret` function should be testable without having to fiddle with timestamps. Please see my other comment.

pkg/gatewaysecret/handler_test.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+	createFunc := func(ctx context.Context, gwSecret *apicorev1.Secret) error {
+		require.Equal(t, newTLSCertValue, string(gwSecret.Data["tls.crt"]))
+		require.Equal(t, newTLSKeyValue, string(gwSecret.Data["tls.key"]))
+		require.Equal(t, newCACertValue, string(gwSecret.Data["ca.crt"]))
+
+		return nil
+	}
+	mockManager := MockSecretManager{findGatewaySecretFunc: findGatewaySecretFunc, createFunc: createFunc}
+	handler := gatewaysecret.GatewaySecretHandler{SecretManager: mockManager}
```
Tomasz-Smelcerz-SAP, line 262: I think this test is too complex to set up; it indicates a "too much coupling" condition.
Take a look at the event loop. You don't need entire `*GatewaySecretHandler` instance there. You use just one function. The event loop test should just test if this function is called with correct arguments for certain event types, and that the error is correrctly reported.
In that case you only need to mock one function - the one that is invoked in the event loop.

config/manager/kustomization.yaml
Tomasz-Smelcerz-SAP: Once more :)

Tomasz-Smelcerz-SAP (approved):

Tomasz-Smelcerz-SAP (approved):

PR #2008, opened 2024-11-04
chore: Log cluster creation command to console

### LINKED ISSUE #1493: Modify k3d GH action to print the command used to create the cluster
### Description

In the new GH action we compose the [k3d creation command](https://github.com/kyma-project/lifecycle-manager/blob/main/.github/actions/create-k3d-cluster/action.yml) with additional arguments and then execute it. For transparency, we should also log the full command executed.

### Reasons

More easily see what has been created in case of issues

### Acceptance Criteria

- [x] when the Create k3d cluster command is executed, we log the exact command used to create it

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

amritanshusikdar:
**Description**
Log the command used to create the k3d clusters to the console for better verbosity.

**Related issue(s)**
Resolves #1493

### DIFF
--- a/.github/actions/create-k3d-cluster/action.yml
+++ b/.github/actions/create-k3d-cluster/action.yml
@@ -29,6 +29,7 @@ runs:
         create_command+="--k3s-arg --disable=traefik@server:* "
         create_command+="--k3s-arg --tls-san=host.k3d.internal@server:* "
 
-        echo "Creating `k3d` clusters using command: ${create_command}"
+        echo "Creating clusters using command:"
+        echo "${create_command}"
 
         eval ${create_command}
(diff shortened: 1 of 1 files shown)


.github/actions/create-k3d-cluster/action.yml
```diff
@@ -29,4 +29,6 @@ runs:
         create_command+="--k3s-arg --disable=traefik@server:* "
         create_command+="--k3s-arg --tls-san=host.k3d.internal@server:* "
 
+        echo "Creating `k3d` clusters using command: ${create_command}"
```
lindnerby, line 32:
```suggestion
        echo "Creating clusters using command:"
        echo "${create_command}"
```

lindnerby (approved):

Issue #1851, opened 2024-09-09
[ModuleCatalog] Extend Catalog sync to handle both `ModuleTemplate` and `ModuleReleaseMeta` CRs


jeremyharisch:
### Note
The original issue was about creating a dedicated sync controller.
For now though, we decided to just add the implementation to the current codebase with necessary refactorings.
I am leaving the original issue unchanged (below) for reference; this will be extracted to a separate issue after the first version of the extended CatalogSync code is implemented and released.

Current Acceptance Criteria:

- [x] Extend the module catalog sync with support for ModuleReleaseMeta CRs
- [ ] Update the documentation
- [x] Extend the test "CRDs sync to SKR and annotations updated in KCP kyma" to cover ModuleReleaseMeta CRs

### Original Description:
The synchronization process of `ModuleTemplate` CRs to all SKR clusters is currently embedded within the Kyma reconcile loop. We have decided to expand this functionality to include `ModuleReleaseMeta` CRs and to relocate these synchronization tasks to a new, dedicated controller. This approach is intended to enhance performance by offloading the sync operations from the main Kyma reconcile loop.

Detailed decision ticket: https://github.com/kyma-project/lifecycle-manager/issues/1815

### Current Implementation:
The existing synchronization of `ModuleTemplate` CRs is integrated into the Kyma reconcile loop, detailed [here](https://github.com/kyma-project/lifecycle-manager/blob/main/internal/controller/kyma/controller.go#L327).

### Objective:
- **Develop a Dedicated Sync Controller**:
   - This new controller will handle the synchronization of both `ModuleTemplate` and `ModuleReleaseMeta` CRs from the KCP cluster to all SKR clusters.
   - It will manage the lifecycle of these CRs—handling creation, updates, and deletions—and will ensure that all corresponding Kyma CRs are requeued accordingly.

- **Performance Optimization**:
   - By isolating the synchronization process, we aim to reduce the overhead in the Kyma reconcile loop, which currently dedicates approximately 20% of its time to sync operations.

- **Testing and Validation**:
   - Comprehensive E2E and integration testing will be conducted to ensure the new controller operates effectively and integrates smoothly with existing systems.
   - Adjustments will be made to existing tests that are affected by this change.

### Benefits of Separation of Concerns:
- **Enhanced Performance**: Removing the sync tasks from the Kyma controller allows for more focused and efficient reconciliation processes.
[...]

PR #2014, opened 2024-11-05
feat: Modulereleasemeta catalog sync

Tomasz-Smelcerz-SAP:
**Description**

Extend existing remote catalog sync logic with support for `ModuleReleaseMeta` objects and corresponding CRD.

**Note for reviewer**: The catalog sync code in the `internal/remote` package requires refactoring to remove code duplication and improve it's internal structure. This is a **work in progress** that started already with PR: #2002 
For now, to add support for syncing `ModuleReleaseMeta`, I tried to avoid big refactoring as much as possible.
This package is already complex and adding massive refactoring on top of a new feature means that the PR would be very hard to review (and the risk of introducing errors would be significant)
This is why there's so much code duplication at the moment - it's easier to understand the new sync logic if it just re-uses the already tried approach.
This code will be refactored later in a dedicated refactoring issue: https://github.com/kyma-project/lifecycle-manager/issues/2015

- Extend the catalog sync logic in the `internal/remote` package with support for `ModuleReleaseMeta` objects
- Use the new functionality in the Kyma reconciliation loop.

**Related issue(s)**
#1851


internal/remote/modulereleasemeta_syncer.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+// SyncToSKR first lists all currently available ModuleReleaseMetas in the Runtime.
+// If there is a NoMatchError, it will attempt to install the CRD but only if there are available crs to copy.
+// It will use a 2 stage process:
+// 1. All ModuleReleaseMeta that have to be created based on the ModuleReleaseMetas existing in the Control Plane.
+// 2. All ModuleReleaseMeta that have to be removed as they are not existing in the Control Plane.
+// It uses Server-Side-Apply Patches to optimize the turnaround required.
+func (mts *moduleReleaseMetaSyncer) SyncToSKR(ctx context.Context, kyma types.NamespacedName, kcpModuleReleases []v1beta2.ModuleReleaseMeta) error {
```
nesmabadr, line 53: The kyma parameter is not used here

internal/remote/moduletemplate_syncer.go
```diff
@@ -13,42 +13,44 @@ import (
...
+// SyncToSKR first lists all currently available moduleTemplates in the Runtime.
 // If there is a NoMatchError, it will attempt to install the CRD but only if there are available crs to copy.
 // It will use a 2 stage process:
-// 1. All ModuleTemplates that either have to be created based on the given Control Plane Templates
-// 2. All ModuleTemplates that have to be removed as they were deleted form the Control Plane Templates
+// 1. All ModuleTemplates that have to be created based on the ModuleTemplates existing in the  Control Plane.
+// 2. All ModuleTemplates that have to be removed as they are not existing in the Control Plane.
 // It uses Server-Side-Apply Patches to optimize the turnaround required.
-func (mts *syncer) SyncToSKR(ctx context.Context, kyma types.NamespacedName, kcpModules []v1beta2.ModuleTemplate) error {
+func (mts *moduleTemplateSyncer) SyncToSKR(ctx context.Context, kyma types.NamespacedName, kcpModules []v1beta2.ModuleTemplate) error {
```
nesmabadr, line 53: same, there is no need for the kyma parameter

internal/remote/crd.go
```diff
@@ -0,0 +1,37 @@
+package remote
+
+import (
+	"errors"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+)
+
+func crdReady(crd *apiextensionsv1.CustomResourceDefinition) bool {
```
nesmabadr, line 10: There is another file named crd_upgrade. I think we can move these functions there since they all deal with the upgrade of the crd to the remote cluster

internal/remote/modulereleasemeta_syncworker.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	if len(errs) != 0 {
+		errs = append(errs, errModuleReleaseMetaCleanup)
+		return errors.Join(errs...)
+	}
+	return nil
+}
+
+func createModuleReleaseMetaCRDInRuntime(ctx context.Context, kcpClient client.Client, skrClient client.Client) error {
```
nesmabadr, line 111: This function shares the same logic with createModuleTemplateCRDInRuntime, so we can move it to the crd.go and just pass the crdType (ModuleTemplateKind or ModuleReleaseMetaKind).

internal/remote/modulereleasemeta_syncer_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+	// Create a fake client with the SKR objects
+	scheme, err := v1beta2.SchemeBuilder.Build()
+	require.NoError(t, err)
+	skrClient := fake.NewClientBuilder().
+		WithObjects(&mrmSKR2, &mrmSKR3, &mrmSKR4).
+		WithScheme(scheme).
+		Build()
+
+	// onSyncConcurrentlyFn "pretends" to be the moduleReleaseMetaConcurrentWorker.SyncConcurrently
```
nesmabadr, line 38: IMO there is no need for these comments

internal/remote/modulereleasemeta_syncer.go
Tomasz-Smelcerz-SAP: Good catch! Leftover from intermediate refactoring steps.

internal/remote/crd.go
Tomasz-Smelcerz-SAP: I wanted to leave it separately on purpose, to make the scope of refactoring more visible.
It is going to be moved elsewhere anyway when implementing #2015.
For now I can move it into `crd_upgrade.go`, no problem.

nesmabadr (approved):

nesmabadr (approved):

nesmabadr (approved):

PR #2017, opened 2024-11-06
feat: Bump modulectl version in E2E tests

### LINKED ISSUE #2010: Bump the version of modulectl used in E2E
### Description

A new version (1.1.0) of modulectl was released which has all the required changes for ModuleCatalog. We need to bump the version used in the E2E tests.

**Acceptance Criteria:**

- [x] Bump modulectl in E2E tests to 1.1.0
- [x] Adapt the new changes to the moduleconfig files used.
- [x] Serve Manifest and DefaultCR via web server
- [x] Confirm that E2E tests are also succeeding with this in the internal charts repository

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

nesmabadr:
**Description**

- Bump modulectl version in E2E tests
- Create a local http server to populate the manifest field

**Related issue(s)**
Resolves #2010


.github/actions/deploy-template-operator-with-modulereleasemeta/action.yml
```diff
@@ -114,16 +117,17 @@ runs:
...
         echo "name: kyma-project.io/module/template-operator
-        channel: regular
         version: 1.0.1
         security: sec-scanners-config.yaml
-        defaultCR: ./config/samples/default-sample-cr.yaml
-        manifest: template-operator.yaml
-        annotations:
-          operator.kyma-project.io/doc-url: https://kyma-project.io" >> warning-module-config.yaml
-        modulectl create --module-config-file ./warning-module-config.yaml --registry http://localhost:5111 --insecure
+        defaultCR: https://github.com/kyma-project/template-operator/releases/download/1.0.1/default-sample-cr.yaml
```
ruanxin, line 122: I expect the defaultCR also coming from localhost which aligned with manifest data, to have consistency here.

scripts/tests/https_server.py
```diff
@@ -0,0 +1,35 @@
...
+    # Create an SSL context
+    context = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
+    context.load_cert_chain(certfile=cert_file_path, keyfile=key_file_path)
+
+    # Configure and start the HTTPS server
+    httpd = http.server.HTTPServer(('localhost', port), http.server.SimpleHTTPRequestHandler)
+    httpd.socket = context.wrap_socket(httpd.socket, server_side=True)
+
+    print(f"Serving on https://localhost:{port} from {directory_to_serve}")
+    while True:
```
ruanxin, line 27: I didn't expect that python http module doesn't support https, basically what I assume is just using one line of code to start the server.
```
$ python3 -m http.server 8080
```
But since we need to write a python script for configure certificate, I actually would write with Golang, so that we only use one language and lower future maintenance efforts.
Here is a reference article for golang, the main function is quite simple:
https://eli.thegreenplace.net/2021/go-https-servers-with-tls/ 

Btw, this `while True` is suspicion, it has `server.serve_forever()` based on this article. https://realpython.com/python-http-server/#encrypt-the-connection-with-https

I'm sorry it's kind of waste efforts here, but if it's indeed requires introducing code from a different language, it's better bring it into discussion first, of course, you can still bring this into team discuss if insist to keep this python script.

nesmabadr: Yeah I actually agree with using Golang. I was thinking about it, but since it was easy to do the python server, I just went for it, so will try to do it in Golang and see how it goes

mmitoraj (approved):

ruanxin (approved):

PR #2023, opened 2024-11-11
feat: Add Kyma Manifest in Error state for plutono dashboard 

ruanxin:
**Description**

resolve the previously closed issue due to can't filter out Timestamp in table view, we should use `Instant` here to only query latest metrics.

**Related issue(s)**
https://github.com/kyma-project/lifecycle-manager/issues/1528

### DIFF
--- a/config/grafana/status.json
+++ b/config/grafana/status.json
@@ -75,7 +75,7 @@
         "text": {},
         "textMode": "auto"
       },
-      "pluginVersion": "7.5.31",
+      "pluginVersion": "7.5.34",
       "targets": [
         {
           "exemplar": true,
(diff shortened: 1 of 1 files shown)


config/grafana/status.json
```diff
@@ -88,6 +88,86 @@
...
+      },
+      "id": 12,
+      "options": {
+        "showHeader": true
+      },
+      "pluginVersion": "7.5.34",
+      "targets": [
+        {
+          "exemplar": true,
+          "expr": "lifecycle_mgr_module_state{state=\"Error\"} > 0",
```
nesmabadr, line 133: I tested this on DEV but it is not showing any entries although there are modules in Error there

nesmabadr: As discussed, the issue is that we no longer update the module status if the remote kyma is unreachable. Another issue will be created to address this

nesmabadr (approved):

PR #2033, opened 2024-11-18
feat: Add mandatory-module label to mandatory ModuleTemplate

### LINKED ISSUE #1198: [Mandatory Modules] Introduce mandatory module Label in kyma reconciler
### Description

With https://github.com/kyma-project/lifecycle-manager/issues/1052 Mandatory Modules have been introduced. To simplify the lookup logic, we could introduce new labels for mandatory Modules. Those labels could be added to the `ModuleTemplates` during lookup and be requeued, thus every time a ModuleTemplate gets created/updated on cluster and the mandatory spec field is st to true, the label will be added.

### Reasons

- Simplify lookup logic in KLM.
- Easier to filter ModuleTemplates when using kubectl

### Acceptance Criteria

- [x] Define new label for mandatory modules: operator.kyma-project.io/mandatory-module: true / false
- [x] When ModuleTemplate gets created/updated
    - [x] Add new defined label if `spec.mandatory` is set to `true`
    - [x] Reove new defined label if `spec.mandatory` is set to `false`
- [x] Simplify `GetMandatory` in lookup pkg by filtering with a label

### Feature Testing

_No response_

### Testing approach

* E2E Testing

### Attachments

_No response_

nesmabadr:
**Description**

- Add mandatory-module label to mandatory ModuleTemplate
- Add tests

**Related issue(s)**
Resolves #1198


pkg/testutils/moduletemplate.go
```diff
@@ -79,6 +80,31 @@ func UpdateModuleTemplateSpec(ctx context.Context,
...
+	if moduleTemplate == nil {
+		return fmt.Errorf("module template not found, %s", moduleName)
+	}
+
+	if moduleTemplate.Labels[key] != value {
+		return fmt.Errorf("label %s:%s not found", key, value)
+	}
+	return nil
+}
+
```
c-pius, line 107: I am okay with this as is, but one idea to simplify may be to use the lookup by label already. I.e., get all ModuleTemplates with `operator.kyma-project.io/mandatory-module: true`, assert that the len of returned ModuleTemplates is 1 and the entry has the expected module name.

pkg/testutils/builder/moduletemplate.go
```diff
@@ -34,6 +34,9 @@ func NewModuleTemplateBuilder() ModuleTemplateBuilder {
 			ObjectMeta: apimetav1.ObjectMeta{
 				Name:      random.Name(),
 				Namespace: apimetav1.NamespaceDefault,
+				Labels: map[string]string{
+					shared.IsMandatoryModule: shared.EnableLabelValue,
+				},
```
c-pius, line 39: I am a bit confused by this change. Isn't this applying the mandatory module label to all tests?

pkg/templatelookup/mandatory.go
```diff
@@ -4,23 +4,27 @@ import (
...
 	error,
 ) {
-	moduleTemplateList := &v1beta2.ModuleTemplateList{}
-	if err := kymaClient.List(ctx, moduleTemplateList); err != nil {
-		return nil, fmt.Errorf("could not list ModuleTemplates: %w", err)
+	mandatoryModuleTemplateList := &v1beta2.ModuleTemplateList{}
+	labelSelector := k8slabels.SelectorFromSet(k8slabels.Set{shared.IsMandatoryModule: shared.EnableLabelValue})
+	if err := kymaClient.List(ctx, mandatoryModuleTemplateList,
+		&client.ListOptions{LabelSelector: labelSelector}); err != nil {
+		return nil, fmt.Errorf("could not list mandatory ModuleTemplates: %w", err)
```
c-pius, line 22: Maybe we could also add a field selector `client.MatchingFields{"metadata.deletionTimestamp": ""}`. Then we could get rid of filtering in the for loop bellow.

c-pius (commented):
Maybe I missed the discussion while I was on vacation, but I am honestly not sure what we gain from this. The lookup for mandatory modules is maybe slightly simpler as we filter directly by label, but we still need the for loop to map the list to a map. So not much of a simplification or performance gain.
But this comes at the price of rather complex handling in the kyma controller to ensure this label. Maybe I missing the obvious reason why we want to use introduce this label?

pkg/testutils/moduletemplate.go
nesmabadr: the GetMandatory already uses the label for the lookup

pkg/testutils/builder/moduletemplate.go
nesmabadr: yes but that's for the integration tests in which some don't have the kyma controller initialized. also, there is another followup issue to this to introduce the label using modulectl create command, so anyways the label should be there before the kyma controller

nesmabadr: ah thanks! I meant to have it only for the mandatory modules integration tests.

pkg/testutils/moduletemplate.go
c-pius: Sorry, seems like I wasn't fully awake yet when reviewing this

pkg/templatelookup/mandatory.go
c-pius: Maybe this could work? 

```go
// GetMandatory returns ModuleTemplates TOs (Transfer Objects) which are marked are mandatory modules.
func GetMandatory(ctx context.Context, kymaClient client.Reader) (ModuleTemplatesByModuleName,
	error,
) {
	mandatoryModuleTemplateList := &v1beta2.ModuleTemplateList{}
	labelSelector := k8slabels.SelectorFromSet(k8slabels.Set{shared.IsMandatoryModule: shared.EnableLabelValue})
	fieldSeleector := fields.OneTermEqualSelector("metadata.deletionTimestamp", "")
	if err := kymaClient.List(ctx, mandatoryModuleTemplateList,
		&client.ListOptions{LabelSelector: labelSelector, FieldSelector: fieldSeleector}); err != nil {
		return nil, fmt.Errorf("could not list mandatory ModuleTemplates: %w", err)
	}

	mandatoryModules := make(map[string]*ModuleTemplateInfo)
	for _, moduleTemplate := range mandatoryModuleTemplateList.Items {
		mandatoryModules[moduleTemplate.Name] = &ModuleTemplateInfo{
			ModuleTemplate: &moduleTemplate,
			Err:            nil,
		}
	}
	return mandatoryModules, nil
}

```

nesmabadr: no no never mind I think I was doing some other thing that made it not working. I am trying it now with the correct value

c-pius (approved):

PR #2075, opened 2024-11-29
fix: Vulnerabilities in Istio Gateway Secret Rotation

c-pius:
**Description**

- tries to fix vulnerabilities reported by checkmarx

**Related issue(s)**

### DIFF
--- a/pkg/gatewaysecret/handler.go
+++ b/pkg/gatewaysecret/handler.go
@@ -172,7 +172,7 @@ func (gsh *GatewaySecretHandler) StartRootCertificateWatch(clientset *kubernetes
 	})
 	if err != nil {
 		log.Error(err, "unable to start watching root certificate")
-		panic(err)
+		return
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/gatewaysecret/handler.go
```diff
@@ -172,7 +172,7 @@ func (gsh *GatewaySecretHandler) StartRootCertificateWatch(clientset *kubernetes
 	})
 	if err != nil {
 		log.Error(err, "unable to start watching root certificate")
-		panic(err)
+		return
```
c-pius, line 175: to reviewer: Checkmarx was complaining that the panic is not properly handled. Since this is a go routine, I think the panic would anyway just abort the startup of the watch. Therefore we can also just return here since the error is logged and the go routine will "gracefully" stop.

c-pius: If it is critical, then we need to exit in a way that is also terminating the main thread. Don't know out of my head how to best do that, Xin suggested a channel which may be okay. Maybe you can find a more elegant way (maybe your bootstrap).

PR #2078, opened 2024-11-29
fix: Handle Race Condition for Istio Gateway Secret

LeelaChacha:
Fix for https://github.com/kyma-project/lifecycle-manager/pull/2004

### DIFF
--- a/pkg/gatewaysecret/handler.go
+++ b/pkg/gatewaysecret/handler.go
@@ -6,14 +6,13 @@ import (
 	"fmt"
 	"time"
 
-	"k8s.io/client-go/rest"
-
 	"github.com/go-logr/logr"
 	apicorev1 "k8s.io/api/core/v1"
 	apimetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/kyma-project/lifecycle-manager/pkg/util"
@@ -93,10 +92,9 @@ func GetValidLastModifiedAt(secret *apicorev1.Secret) (time.Time, error) {
 }
 
 func (h *Handler) findGatewaySecret(ctx context.Context) (*apicorev1.Secret, error) {
-	secret := &apicorev1.Secret{}
-	var err error
-	if secret, err = h.kcpClientset.CoreV1().Secrets(istioNamespace).Get(ctx, gatewaySecretName,
-		apimetav1.GetOptions{}); err != nil {
+	secret, err := h.kcpClientset.CoreV1().Secrets(istioNamespace).Get(ctx, gatewaySecretName,
+		apimetav1.GetOptions{})
+	if err != nil {
 		return nil, fmt.Errorf("failed to get secret %s: %w", gatewaySecretName, err)
 	}
 	return secret, nil
@@ -104,7 +102,8 @@ func (h *Handler) findGatewaySecret(ctx context.Context) (*apicorev1.Secret, err
 
 func (h *Handler) Create(ctx context.Context, secret *apicorev1.Secret) error {
 	h.updateLastModifiedAt(secret)
-	if _, err := h.kcpClientset.CoreV1().Secrets(istioNamespace).Create(ctx, secret, apimetav1.CreateOptions{}); err != nil {
+	if _, err := h.kcpClientset.CoreV1().Secrets(istioNamespace).Create(ctx, secret,
+		apimetav1.CreateOptions{}); err != nil {
 		return fmt.Errorf("failed to create secret %s: %w", secret.Name, err)
 	}
 	return nil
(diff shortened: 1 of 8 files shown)


c-pius:
FYI: we also had some checkmarx related issues: https://github.com/kyma-project/lifecycle-manager/pull/2075
I think we can close my PR in favor of yours, would you mind double checking the fixes though? Maybe you will need to change the passing of the configs and the panics in the same way.

LeelaChacha:
> the panics

WDYT about my comment on the panic? Should the goroutine exit with a bootstrap error?

pkg/gatewaysecret/handler.go
```diff
@@ -26,48 +26,57 @@ const (
...
+var (
+	errCouldNotGetLastModifiedAt = errors.New("getting lastModifiedAt time failed")
+	errExpectedOneRootCASecret   = errors.New("expected exactly one root CA secret")
+)
 
-type GatewaySecretHandler struct {
-	kcpClient client.Client
+type Handler struct {
+	kcpClient    client.Client
+	kcpClientset *kubernetes.Clientset
```
ruanxin, line 36: I think it make more sense only use `Clientset` in this case, then this package is decoupled with `controller-runtime`

pkg/gatewaysecret/handler.go
```diff
@@ -161,21 +170,44 @@ func GetGatewaySecret(ctx context.Context, clnt client.Client) (*apicorev1.Secre
 	return secret, nil
 }
 
-func (gsh *GatewaySecretHandler) StartRootCertificateWatch(clientset *kubernetes.Clientset,
-	log logr.Logger,
-) {
+func (h *Handler) StartRootCertificateWatch() {
 	ctx, cancel := context.WithCancel(context.TODO())
```
ruanxin, line 174: Please don't use TODO context, in this case, use background

pkg/gatewaysecret/handler.go
```diff
@@ -161,21 +170,44 @@ func GetGatewaySecret(ctx context.Context, clnt client.Client) (*apicorev1.Secre
...
 	defer cancel()
 
-	secretWatch, err := clientset.CoreV1().Secrets(istioNamespace).Watch(ctx, apimetav1.ListOptions{
+	h.handleAlreadyCreatedRootCertificate(ctx)
+	h.handleNewRootCertificates(ctx)
+}
+
+func (h *Handler) handleAlreadyCreatedRootCertificate(ctx context.Context) {
+	rootCASecrets, err := h.kcpClientset.CoreV1().Secrets(istioNamespace).List(ctx, apimetav1.ListOptions{
```
ruanxin, line 182: Why use `.List` but not `.Get`?

pkg/gatewaysecret/handler.go
```diff
@@ -161,21 +170,44 @@ func GetGatewaySecret(ctx context.Context, clnt client.Client) (*apicorev1.Secre
...
+}
+
+func (h *Handler) handleNewRootCertificates(ctx context.Context) {
+	secretWatch, err := h.kcpClientset.CoreV1().Secrets(istioNamespace).Watch(ctx, apimetav1.ListOptions{
 		FieldSelector: fields.OneTermEqualSelector(apimetav1.ObjectNameField, kcpRootSecretName).String(),
 	})
 	if err != nil {
-		log.Error(err, "unable to start watching root certificate")
+		h.log.Error(err, "unable to start watching root certificate")
 		panic(err)
```
ruanxin, line 207: if you panic in goruntine, it just crashed this goruntine itself, but in your case, if can't watching root certificate, the whole lifecycle manager should not started, right?

In this case, you need to send error back to go main goruntine and panic there. So you should use channel here.

Another option is not design whole `StartRootCertificateWatch` as goruntine, instead, before `WatchEvents`, you can still running in the main goruntine, you only need to make `WatchEvents` as a dedicated thread, it can simplify your panic handling.

pkg/gatewaysecret/handler.go
```diff
@@ -161,21 +170,44 @@ func GetGatewaySecret(ctx context.Context, clnt client.Client) (*apicorev1.Secre
...
+		FieldSelector: fields.OneTermEqualSelector(apimetav1.ObjectNameField, kcpRootSecretName).String(),
+	})
+	if err != nil {
+		h.log.Error(err, "unable to list root certificate")
+		panic(err)
+	}
+	if len(rootCASecrets.Items) != 1 {
+		h.log.Error(errExpectedOneRootCASecret, errExpectedOneRootCASecret.Error(),
+			"found", len(rootCASecrets.Items))
+		panic(fmt.Errorf("%w: found %d", errExpectedOneRootCASecret, len(rootCASecrets.Items)))
```
ruanxin, line 192: same here, if you panic in a goruntine, the main thread is still running. https://github.com/kyma-project/lifecycle-manager/pull/2078/files#r1865020057

pkg/gatewaysecret/handler.go
LeelaChacha: would have needed to add an extra rbac role, that's why :)

ruanxin: then we should add missing `get` rules, have `list` but without `get` it make no sense.

LeelaChacha: ok

pkg/gatewaysecret/handler.go
ruanxin: actually you can design another struct to handle `GetRootCACertificate`, `RequiresUpdate`, then you can even decouple certmanger

Issue #2024, opened 2024-11-11
Kyma Modules State not updated when the remote cluster is unreachable


nesmabadr:
### Description

When the remote cluster is removed or unreachable, the kyma modules state in the Kyma.Status remain Ready, although it should be in an Error state. 

There are several kymas which have this issue on DEV, for example: 926fb10a-925d-4c55-863d-a8a2da2ffb56.

### Acceptance Criteria
- [ ] Status should be set to error if the cluster is not reachable
- [ ] Introduce E2E test for this scenario: https://github.com/kyma-project/lifecycle-manager/blob/2342670c9be72df6b33c4cd6ce0d12ccbb65b371/tests/e2e/deprovision_test.go#L78

### Steps to reproduce

- Delete the remote cluster
- check the state of any enabled module in the Kyma.Status, it is in Ready state.

### Environment Type

Managed

### Environment Info

Kubernetes Version: x.y.z
Docker Version: x.y.z
modulectl Version: x.y.z
If used - k3d/k3s/minikube versions

c-pius:
Can verify that this worked on DEV for the mentioned cluster. For STAGE, we don't have a existing Kyma that had this issue. Provoking this error may be possible, but even if we change the kubeconfig secret to point to a non-existing host, I think we would also need to restart KLM because the endpoint is cached. Could we ask SRE to temporarily shutdown the whole SKR cluster?

PR #2093, opened 2024-12-05
fix: Modules should be in error state when SKR cluster is deleted

Tomasz-Smelcerz-SAP:
**Description**

I decided to just set the state of all modules to "error" on SKR context initialization error, instead of reading the status of each module from the related `Manifest` CR.
I think it is correct action in all cases.
It would be incorrect if there's a permanent condition that prevents SKR context initialization (cluster deleted? wrong secret?) AND at the same time the manifest for a module would be NOT in an error state.
A simpler explanation: this approach would be incorrect if the Kyma reconciler cannot access SKR, but Manifest reconciler can (or it doesn't need to, but it sets a status different than "error")
Because such a situation is very unlikely, I decided to go with simpler solution to preserve resources (memory/cpu/network)

- Set the state of all modules to "error" on SKR context initialization error
- Add e2e test for the "cluster deleted" scenario

**Related issue(s)**
Fixes: #2024

### DIFF
--- a/internal/controller/kyma/controller.go
+++ b/internal/controller/kyma/controller.go
@@ -137,6 +137,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		if err != nil {
 			r.SkrContextFactory.InvalidateCache(kyma.GetNamespacedName())
 			r.Metrics.RecordRequeueReason(metrics.SyncContextRetrieval, queue.UnexpectedRequeue)
+			setModuleStatusesToError(kyma, util.NestedErrorMessage(err))
 			return r.requeueWithError(ctx, kyma, err)
 		}
 	}
@@ -672,4 +673,14 @@ func needUpdateForMandatoryModuleLabel(moduleTemplate v1beta2.ModuleTemplate) bo
 	}
 
 	return false
+}
+
+func setModuleStatusesToError(kyma *v1beta2.Kyma, optMessage string) {
+	moduleStatuses := kyma.Status.Modules
+	for i := range moduleStatuses {
+		moduleStatuses[i].State = shared.StateError
+		if optMessage != "" {
+			moduleStatuses[i].Message = optMessage
+		}
+	}
 }
--- a/internal/controller/kyma/controller.go
+++ b/internal/controller/kyma/controller.go
@@ -676,12 +676,12 @@ func needUpdateForMandatoryModuleLabel(moduleTemplate v1beta2.ModuleTemplate) bo
 	return false
 }
 
-func setModuleStatusesToError(kyma *v1beta2.Kyma, optMessage string) {
+func setModuleStatusesToError(kyma *v1beta2.Kyma, message string) {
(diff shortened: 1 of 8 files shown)


internal/controller/kyma/controller.go
```diff
@@ -673,3 +675,13 @@ func needUpdateForMandatoryModuleLabel(moduleTemplate v1beta2.ModuleTemplate) bo
 
 	return false
 }
+
+func setModuleStatusesToError(kyma *v1beta2.Kyma, optMessage string) {
```
c-pius, line 679: `opt` in `optMessage` indicating that it is optional? I am okay with it but have not seen this convention in the repo before... For me it would also be okay to just go with `message` instead.

pkg/module/sync/runner.go
```diff
@@ -246,14 +246,14 @@ func (r *Runner) setupModule(module *common.Module, kyma *v1beta2.Kyma) error {
...
 	DeleteNoLongerExistingModuleStatus(ctx, kyma, r.getModule, kymaMetrics.RemoveModuleStateMetrics,
 		moduleMetrics.RemoveModuleCRWarningCondition)
 }
 
-func updateModuleStatusFromExistingModules(
-	modules common.Modules,
+func UpdateModuleStatusFromExistingModules(
 	kyma *v1beta2.Kyma,
+	modules common.Modules,
 ) {
```
c-pius, line 257: Why has this function been made public?

tests/e2e/skr_connection_lost_test.go
```diff
@@ -0,0 +1,73 @@
...
+		It("When SKR Cluster is removed", func() {
+			cmd := exec.Command("sh", "../../scripts/tests/remove_skr_host_from_coredns.sh")
+			out, err := cmd.CombinedOutput()
+			Expect(err).NotTo(HaveOccurred())
+			GinkgoWriter.Printf(string(out))
+			cmd = exec.Command("k3d", "cluster", "rm", "skr")
+			out, err = cmd.CombinedOutput()
+			Expect(err).NotTo(HaveOccurred())
+			GinkgoWriter.Printf(string(out))
+		})
```
c-pius, line 57: It may be a bit more concise to just run `k3d cluster stop skr`. I used this to verify locally and it also worked. But also no prob leave this as is.

pkg/module/sync/runner.go
Tomasz-Smelcerz-SAP: My mistake, refactoring leftover. Fixed

internal/controller/kyma/controller.go
Tomasz-Smelcerz-SAP: Yes. It constantly makes me uncomfortable how we are unable to idiomatically express optional values in golang.
You are right about the convention though. Renamed.

c-pius: > It constantly makes me uncomfortable how we are unable to idiomatically express optional values in golang

Agree

c-pius (approved):

Issue #2084, opened 2024-12-02
Redesign sync istio Gateway Secret as a dedicate controller


ruanxin:
### Description

The current implementation of watching the Istio gateway secret as a goroutine has flaws. If the goroutine crashes, the Kyma Lifecycle Manager (KLM) loses a critical feature to keep the Istio gateway secret up-to-date, and it's difficult to observe this in a goroutine.

https://github.com/kyma-project/lifecycle-manager/pull/2078/files#diff-22c8411fd113828cf386f8346a3c7e6a0e273579459deeceaabc14ea7ff217ecR194

## Solution:
Redesign this feature as a dedicated controller. This approach relies on controller-runtime to watch root secret changes and ensures the sync happens at regular intervals.

### Reasons

To have a reliable way to sync istio gateway secret.

### Acceptance Criteria

- [ ] Implement sync istio gateway secret as controller
- [ ] remove `startCAWatch` and related code https://github.com/kyma-project/lifecycle-manager/pull/2078/files#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R206

PR #2095, opened 2024-12-05
feat: Istio gateway secret controller

lindnerby:
**Description**

- Extract separate controller for istio gateway secret rotation
- Remove go routine usage
- Introduce mockery config

**Related issue(s)**
* https://github.com/kyma-project/lifecycle-manager/issues/2084


cmd/main.go
```diff
@@ -194,6 +195,7 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
 	if flagVar.EnablePurgeFinalizer {
 		setupPurgeReconciler(mgr, skrContextProvider, eventRecorder, flagVar, options, setupLog)
 	}
+	setupIstioReconciler(mgr, flagVar, options, setupLog)
```
nesmabadr, line 198: should we move this to the enablekcpwatcher if condition?

cmd/main.go
```diff
@@ -486,3 +488,12 @@ func setupMandatoryModuleDeletionReconciler(mgr ctrl.Manager,
 		os.Exit(bootstrapFailedExitCode)
 	}
 }
+
+func setupIstioReconciler(mgr ctrl.Manager, flagVar *flags.FlagVar, options ctrlruntime.Options, setupLog logr.Logger) {
+	options.MaxConcurrentReconciles = flagVar.MaxConcurrentKymaReconciles
```
nesmabadr, line 493: what about using the watcher reconciles and requeue intervals as well for this controller?

internal/controller/istio/setup.go
```diff
@@ -0,0 +1,51 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	apicorev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	ctrlruntime "sigs.k8s.io/controller-runtime/pkg/controller"
+)
+
+const (
+	controllerName    = "istio-controller"
+	kcpRootSecretName = "klm-watcher" //nolint:gosec // gatewaySecretName is not a credential
```
nesmabadr, line 16: we need to read this from the flags

cmd/main.go
```diff
@@ -50,6 +49,7 @@ import (
 	"github.com/kyma-project/lifecycle-manager/api/shared"
 	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
 	"github.com/kyma-project/lifecycle-manager/internal"
+	"github.com/kyma-project/lifecycle-manager/internal/controller/istio"
```
lindnerby, line 52: How do we like the controller name?

cmd/main.go
```diff
@@ -486,3 +477,20 @@ func setupMandatoryModuleDeletionReconciler(mgr ctrl.Manager,
...
+
+func setupIstioReconciler(mgr ctrl.Manager, flagVar *flags.FlagVar, options ctrlruntime.Options, setupLog logr.Logger) {
+	options.MaxConcurrentReconciles = flagVar.MaxConcurrentKymaReconciles
+
+	handler := gatewaysecret.NewGatewaySecretHandler(mgr.GetConfig(), setupLog)
+	requeueIntervals := queue.RequeueIntervals{
+		Success: flagVar.WatcherRequeueSuccessInterval,
+		Busy:    flags.DefaultKymaRequeueBusyInterval,
+		Error:   flags.DefaultKymaRequeueErrInterval,
+		Warning: flags.DefaultKymaRequeueWarningInterval,
```
lindnerby, line 489: Do we need explicit intervals for the controller?

internal/controller/istio/controller.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+}
+
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := logf.FromContext(ctx)
+	logger.V(log.DebugLevel).Info("reconcile istio gateway secret")
+
+	secret := &apicorev1.Secret{}
+	if err := r.Get(ctx, req.NamespacedName, secret); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get istio gateway secret: %w", err)
```
lindnerby, line 37: Do we need to create the secret if not found? Should be CM's responsibility?

internal/controller/istio/setup.go
```diff
@@ -0,0 +1,54 @@
...
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return isRootSecret(e.ObjectNew)
+		},
+
+		// TODO probably not needed
+		//DeleteFunc: func(e event.DeleteEvent) bool {
+		//	return isRootSecret(e.Object)
+		//},
+		//GenericFunc: func(e event.GenericEvent) bool {
```
lindnerby, line 35: Let's check generic events as well if we need them

unit-test-coverage.yaml
```diff
@@ -17,4 +18,4 @@ packages:
   internal/remote: 20.2
   internal/util/collections: 86
   pkg/templatelookup: 77.1
-  pkg/gatewaysecret: 22.6
+  pkg/gatewaysecret: 100
```
lindnerby, line 21: 💯

internal/controller/istiogatewaysecret/setup.go
```diff
@@ -0,0 +1,81 @@
...
+			if gwSecretLastModifiedAt, err := time.Parse(time.RFC3339, gwSecretLastModifiedAtValue); err == nil {
+				return gwSecretLastModifiedAt, nil
+			}
+		}
+		return time.Time{}, errCouldNotGetLastModifiedAt
+	}
+	handler := gatewaysecret.NewGatewaySecretHandler(clnt, parseLastModifiedFunc)
+
+	var getSecretFunc GetterFunc = func(ctx context.Context, name types.NamespacedName) (*apicorev1.Secret, error) {
+		secret := &apicorev1.Secret{}
```
nesmabadr, line 45: so why are we not using the client here so that it is clnt.GetGatewaysecret? and then we only use the request namespaced name to check first if it is the actual rootsecret name and namespace since anyways it should be them because we have the secret predicate? I mean we shouldn't be fetching the secret if the name or namespace are not those of the root secret in the reconciler if it somehow enters the reconciliation loop despite of the predicates

internal/controller/istiogatewaysecret/controller_test.go
```diff
@@ -0,0 +1,124 @@
...
+	request := ctrl.Request{NamespacedName: types.NamespacedName{Name: secretName, Namespace: secretNamespace}}
+
+	// ACT
+	// ASSERT
+	_, _ = reconciler.Reconcile(context.TODO(), request)
+}
+
+func TestReconcile_WhenGetSecretFuncReturnsSecret_HandlerManageGatewaySecretIsCalled(t *testing.T) {
+	// ARRANGE
+	secret := &apicorev1.Secret{}
```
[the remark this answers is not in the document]
lindnerby: The previous one is asserting that the GetterFunc is called with the values of the request.
This one is asserting that the Handler impl is called and one time only.

I try to convey the different testcases by descriptive test names.
Let me know if they should be changed to something better!

internal/controller/istiogatewaysecret/setup.go
```diff
@@ -0,0 +1,81 @@
...
+
+const (
+	controllerName    = "istio-controller"
+	kcpRootSecretName = "klm-watcher"
+)
+
+var errCouldNotGetLastModifiedAt = errors.New("getting lastModifiedAt time failed")
+
+func SetupReconciler(mgr ctrl.Manager, flagVar *flags.FlagVar, options ctrlruntime.Options) error {
+	options.MaxConcurrentReconciles = flagVar.MaxConcurrentKymaReconciles
```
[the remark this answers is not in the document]
lindnerby: Yes 😄 Thank!

internal/controller/istiogatewaysecret/setup.go
lindnerby: Handler and Reconciler are not sharing the same dependency and the `GetterFunc` here might also be obsolete, due to the investigation I plan considering if we can use this other Reconciler interface:
https://github.com/kubernetes-sigs/controller-runtime/blob/aea2e32a936584b06ae6f7992f856fe7292b0297/pkg/reconcile/reconcile.go#L131

For the sake of encapsulation I want this dependency to be separately tested in `controller_test` and the scope of the Reconciler type.

internal/controller/istiogatewaysecret/controller_test.go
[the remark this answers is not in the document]
nesmabadr: yes but I mean it is the same setup so it could be just one test for both? The test names convey the purpose so I am fine with keeping them separate. 
But in this case do we actually need the following lines in the previous test?

	request := ctrl.Request{NamespacedName: types.NamespacedName{Name: secretName, Namespace: secretNamespace}}

	// ACT
	// ASSERT
	_, _ = reconciler.Reconcile(context.TODO(), request)

internal/controller/istiogatewaysecret/setup.go
nesmabadr: I am not sure I understand this. This is just the setup of the reconciler so why not just fetch the secret that the reconciler is intended for? what I mean is something like this:
var getSecretFunc GetterFunc = func(ctx context.Context, name types.NamespacedName) (*apicorev1.Secret, error) {
		if name.Name != shared.GatewaySecretName || name.Namespace != shared.IstioNamespace {
			return nil, fmt.Errorf("invalid secret name %s or namespace %s", name.Name, name.Namespace)
		}

		secret, err := clnt.GetGatewaySecret(ctx)
		if err != nil {
			return nil, fmt.Errorf("failed to get root gateway secret %w", err)
		}

		return secret, nil
	}

lindnerby: Because the root secret is different from gateway secret 👍

internal/controller/istiogatewaysecret/controller_test.go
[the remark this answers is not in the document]
lindnerby: Hm yes you are right:I changed it to just return nil instead of a Secret and moved the request more in focus. I think it's better now 👍

nesmabadr (approved):

Issue #2059, opened 2024-11-25
[Module Catalog] Support `internal` and `beta` flags for updated module metadata


c-pius:
### Description

Current ModuleTemplates support `operator.kyma-project.io/internal`, `operator.kyma-project.io/beta`, and `operator.kyma-project.io/sync` flags (see also [docs](https://github.com/kyma-project/lifecycle-manager/blob/main/docs/contributor/resources/03-moduletemplate.md#operatorkyma-projectio-labels)).

With our change to the split setup with ModuleTemplates and ModuleReleaseMeta, this doesn't cover the full story anymore. E.g., if a beta ModuleTemplate is synced to a beta SKR, how can it be enabled? Right now, we would need to reference it in a ModuleReleaseMeta which is synced to all clusters, not only beta ones.

The simple idea is to also introduce `beta` and `internal` flags to `ModuleReleaseMeta.spec`. Dashboard and CLI should then only display those modules depending on the presence of the same flags in the KymaCR.

In addition, we introduce `spec.internal` and `spec.beta` on the ModuleReleaseMeta level with the same behavior as for ModuleTemplate. 

### Reasons

Right now we have undefined behavior / an incomplete story for that case.

### Acceptance Criteria

- [x] Given ModuleReleaseMeta with `.spec.internal`
  - [x] only sync ModuleReleaseMeta to SKRs where KymaCR is marked with `operator.kyma-project.io/internal`
  - [x] only sync ModuleTemplates of the same module to SKRs where KymaCR is marked with `operator.kyma-project.io/internal`
- [x] Given ModuleReleaseMeta with `.spec.beta`
  - [x] only sync ModuleReleaseMeta to SKRs where KymaCR is marked with `operator.kyma-project.io/beta`
  - [x] only sync ModuleTemplates of the same module to SKRs where KymaCR is marked with `operator.kyma-project.io/beta`

c-pius:
@mrCherry97 we are planning to move the `internal` / `beta` flags to the `.spec` of the new ModuleReleaseMeta resource. When true, the ModuleReleaseMeta and related ModuleTemplates will only be synced to those SKRs where the same is true. Does that work for you or any concerns?

c-pius:
Yes exactly. The current behavior doesn't change. It is just that the indicator whether a module is in `beta` moves from the ModuleTemplate to the ModuleReleaseMeta.

mrCherry97:
[...]

Issue #2090, opened 2024-12-04
Remove the `operator.kyma-project.io/sync` label


janmedrek:
**Description**

With the single-cluster mode, there is no reason to keep the label, it should always be synced.

**Acceptance Criteria**
- [x] Remove the `operator.kyma-project.io/sync` label from the Kyma CR
- [x] adjust the docs
- [x] @janmedrek: clarify if we need to keep it for the ModuleTemplates
- [x] [Additional, after clarification] Drop the label for ModuleTemplates and finally from KLM completely

PR #2097, opened 2024-12-05
feat: Support `internal` and `beta` flags in new module metadata

c-pius:
~~🚧 to be verified if we can also extend our integration tests~~ will be done as a separate PR
**Description**

- introduces `beta` and `internal` flags to ModuleReleaseMeta
- respects the `beta` and `internal` flags when syncing ModuleReleaseMeta and its referenced modules to the SKR
- keeps the sync of ModuleTemplates using old format in it's current way
  - can be removed once we migrated fully
- determination of what to sync fully covered in unit tests
- E2E test for syncing module release meta extended with the new cases

**Related issue(s)**

- resolves https://github.com/kyma-project/lifecycle-manager/issues/2059

### DIFF
--- a/internal/remote/remote_catalog.go
+++ b/internal/remote/remote_catalog.go
@@ -77,25 +77,25 @@ func newRemoteCatalog(kcpClient client.Client, skrContextFactory SkrContextProvi
 }
 
 func (c *RemoteCatalog) SyncModuleCatalog(ctx context.Context, kyma *v1beta2.Kyma) error {
-	moduleReleaseMetas := &[]v1beta2.ModuleReleaseMeta{}
-	if err := c.GetModuleReleaseMetasToSync(ctx, moduleReleaseMetas, kyma); err != nil {
+	moduleReleaseMetas := []v1beta2.ModuleReleaseMeta{}
+	if err := c.GetModuleReleaseMetasToSync(ctx, &moduleReleaseMetas, kyma); err != nil {
 		return err
 	}
 
-	moduleTemplates := &[]v1beta2.ModuleTemplate{}
-	if err := c.GetModuleTemplatesToSync(ctx, moduleTemplates, moduleReleaseMetas); err != nil {
+	moduleTemplates := []v1beta2.ModuleTemplate{}
+	if err := c.GetModuleTemplatesToSync(ctx, &moduleTemplates, &moduleReleaseMetas); err != nil {
 		return err
 	}
 
 	// https://github.com/kyma-project/lifecycle-manager/issues/2096
 	// Remove this block after the migration to the new ModuleTemplate format is completed.
-	oldModuleTemplate := &[]v1beta2.ModuleTemplate{}
-	if err := c.GetOldModuleTemplatesToSync(ctx, oldModuleTemplate, kyma); err != nil {
+	oldModuleTemplate := []v1beta2.ModuleTemplate{}
+	if err := c.GetOldModuleTemplatesToSync(ctx, &oldModuleTemplate, kyma); err != nil {
 		return err
 	}
-	*moduleTemplates = append(*moduleTemplates, *oldModuleTemplate...)
+	moduleTemplates = append(moduleTemplates, oldModuleTemplate...)
 
-	return c.sync(ctx, kyma.GetNamespacedName(), moduleTemplates, moduleReleaseMetas)
+	return c.sync(ctx, kyma.GetNamespacedName(), &moduleTemplates, &moduleReleaseMetas)
 }
 
 func (c *RemoteCatalog) sync(
--- a/internal/remote/remote_catalog.go
+++ b/internal/remote/remote_catalog.go
@@ -168,6 +168,13 @@ func (c *RemoteCatalog) GetModuleTemplatesToSync(
 		return nil, fmt.Errorf("failed to list ModuleTemplates: %w", err)
 	}
 
+	return c.FilterModuleTemplatesToSync(moduleTemplateList.Items, moduleReleaseMetas), nil
+}
+
+func (c *RemoteCatalog) FilterModuleTemplatesToSync(
+	moduleTemplates []v1beta2.ModuleTemplate,
+	moduleReleaseMetas []v1beta2.ModuleReleaseMeta,
+) []v1beta2.ModuleTemplate {
 	moduleTemplatesToSync := map[string]bool{}
 	for _, moduleReleaseMeta := range moduleReleaseMetas {
 		for _, channel := range moduleReleaseMeta.Spec.Channels {
@@ -175,8 +182,8 @@ func (c *RemoteCatalog) GetModuleTemplatesToSync(
 		}
 	}
 
-	moduleTemplates := []v1beta2.ModuleTemplate{}
-	for _, moduleTemplate := range moduleTemplateList.Items {
+	filteredModuleTemplates := []v1beta2.ModuleTemplate{}
+	for _, moduleTemplate := range moduleTemplates {
 		if moduleTemplate.IsMandatory() {
 			continue
 		}
@@ -186,11 +193,11 @@ func (c *RemoteCatalog) GetModuleTemplatesToSync(
 		}
 
 		if _, found := moduleTemplatesToSync[moduleTemplate.Name]; found {
-			moduleTemplates = append(moduleTemplates, moduleTemplate)
+			filteredModuleTemplates = append(filteredModuleTemplates, moduleTemplate)
 		}
 	}
 
-	return moduleTemplates, nil
+	return filteredModuleTemplates
 }
 
 // https://github.com/kyma-project/lifecycle-manager/issues/2096
--- a/internal/remote/remote_catalog.go
+++ b/internal/remote/remote_catalog.go
@@ -87,7 +87,7 @@ func (c *RemoteCatalog) SyncModuleCatalog(ctx context.Context, kyma *v1beta2.Kym
 		return err
 	}
 
-	// TODO: https://github.com/kyma-project/lifecycle-manager/issues/2096
+	// https://github.com/kyma-project/lifecycle-manager/issues/2096
 	// Remove this block after the migration to the new ModuleTemplate format is completed.
 	oldModuleTemplate, err := c.GetOldModuleTemplatesToSync(ctx, kyma)
 	if err != nil {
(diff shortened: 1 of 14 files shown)


internal/remote/remote_catalog.go
```diff
@@ -102,9 +124,96 @@ func (c *RemoteCatalog) Delete(
...
+	}
+
+	for _, moduleTemplate := range moduleTemplateList.Items {
+		if moduleTemplate.IsMandatory() {
+			continue
+		}
+
+		if moduleTemplate.HasSyncDisabled() {
+			continue
+		}
```
c-pius, line 186: to reviewer: the mandatory check shouldn't be necessary in theory as mandatory modules should not be referenced in ModuleReleaseMetas. I still kept this here to be on the save side.

Regarding the HasSyncDisabled... This will be removed in a follow-up ticket, so I decided to still consider it until then: https://github.com/kyma-project/lifecycle-manager/issues/2090

pkg/testutils/moduletemplate.go
```diff
@@ -48,6 +48,22 @@ func ModuleTemplateExists(ctx context.Context,
 	return nil
 }
 
+func ModuleTemplateExistsByName(ctx context.Context,
```
c-pius, line 51: to reviewer: I was unable to re-use the existing `ModuleTemplateExists` function. It seems as if this existing function relies on the ModuleReleaseMeta to find the module template (which I find weird). In my case, I want to verify that ModuleTemplates are not synced anymore when not referenced in a ModuleReleaseMeta. Hence, I was unable to use the existing function.

I think this needs a cleanup but I would propose to do this in a follow-up as this PR is already quite big.

internal/remote/remote_catalog.go
```diff
@@ -76,22 +76,44 @@ func newRemoteCatalog(kcpClient client.Client, skrContextFactory SkrContextProvi
...
+
+	return c.sync(ctx, kyma.GetNamespacedName(), &moduleTemplates, &moduleReleaseMetas)
+}
+
+func (c *RemoteCatalog) sync(
 	ctx context.Context,
 	kyma types.NamespacedName,
-	kcpModules []v1beta2.ModuleTemplate,
-	kcpModuleReleaseMeta []v1beta2.ModuleReleaseMeta,
+	kcpModules *[]v1beta2.ModuleTemplate,
```
Tomasz-Smelcerz-SAP, line 104: The slice pointer is not necessary here

internal/remote/remote_catalog.go
```diff
@@ -102,9 +124,96 @@ func (c *RemoteCatalog) Delete(
...
 
 	moduleTemplates := c.moduleTemplateSyncAPIFactoryFn(c.kcpClient, skrContext.Client, &c.settings)
 	return moduleTemplates.DeleteAllManaged(ctx)
 }
+
+// GetModuleReleaseMetasToSync returns a list of ModuleReleaseMetas that should be synced to the SKR.
+// A ModuleReleaseMeta that is Beta or Internal is synced only if the Kyma is also Beta or Internal.
+func (c *RemoteCatalog) GetModuleReleaseMetasToSync(
+	ctx context.Context,
+	moduleReleaseMetas *[]v1beta2.ModuleReleaseMeta,
```
Tomasz-Smelcerz-SAP, line 138: I think that `moduleReleaseMetas` should be a return value instead of parameter. Currently it's feels very C-language style, where it's a common idiom to return values via pointer arguments - because C lacks multiple return types.
I know that this is how k8s client `.List()` works and honestly I don't understand why they did it this way, maybe something to avoid unnecessary list (slice) allocations. I think our use case is much simpler and we should not return values via parameters.

internal/remote/remote_catalog.go
```diff
@@ -102,9 +124,96 @@ func (c *RemoteCatalog) Delete(
...
+	return nil
+}
+
+// GetModuleTemplatesToSync returns a list of ModuleTemplates that should be synced to the SKR.
+// A ModuleTemplate is synced if it is not mandatory and does not have sync disabled. In addition,
+// it must be referenced by a ModuleReleaseMeta that is synced.
+func (c *RemoteCatalog) GetModuleTemplatesToSync(
+	ctx context.Context,
+	moduleTemplates *[]v1beta2.ModuleTemplate,
+	moduleReleaseMetas *[]v1beta2.ModuleReleaseMeta,
```
Tomasz-Smelcerz-SAP, line 165: 'moduleReleaseMetas' doesn't have to be a slice pointer here. It makes the code less readable and implies that the function requires a pointer - maybe it's re-allocating the target object? It's not the case here.

internal/remote/remote_catalog.go
```diff
@@ -102,9 +124,96 @@ func (c *RemoteCatalog) Delete(
...
+
+	return nil
+}
+
+// GetModuleTemplatesToSync returns a list of ModuleTemplates that should be synced to the SKR.
+// A ModuleTemplate is synced if it is not mandatory and does not have sync disabled. In addition,
+// it must be referenced by a ModuleReleaseMeta that is synced.
+func (c *RemoteCatalog) GetModuleTemplatesToSync(
+	ctx context.Context,
+	moduleTemplates *[]v1beta2.ModuleTemplate,
```
Tomasz-Smelcerz-SAP, line 164: I'd make `moduleTemplates` a proper return value of this function, see my comment above.

internal/remote/remote_catalog.go
```diff
@@ -102,9 +124,96 @@ func (c *RemoteCatalog) Delete(
...
+	}
+
+	moduleTemplatesToSync := map[string]bool{}
+	for _, moduleReleaseMeta := range *moduleReleaseMetas {
+		for _, channel := range moduleReleaseMeta.Spec.Channels {
+			moduleTemplatesToSync[fmt.Sprintf("%s-%s", moduleReleaseMeta.Spec.ModuleName, channel.Version)] = true
+		}
+	}
+
+	for _, moduleTemplate := range moduleTemplateList.Items {
```
Tomasz-Smelcerz-SAP, line 179: I'd vote for separation of concerns here - The filtering logic should be in it's own little function/method.
In this way it can be tested without having to mock k8s client. At this moment it's not possible.

internal/remote/remote_catalog.go
```diff
@@ -76,22 +76,44 @@ func newRemoteCatalog(kcpClient client.Client, skrContextFactory SkrContextProvi
...
+		return err
+	}
+
+	// https://github.com/kyma-project/lifecycle-manager/issues/2096
+	// Remove this block after the migration to the new ModuleTemplate format is completed.
+	oldModuleTemplate := []v1beta2.ModuleTemplate{}
+	if err := c.GetOldModuleTemplatesToSync(ctx, &oldModuleTemplate, kyma); err != nil {
+		return err
+	}
+	moduleTemplates = append(moduleTemplates, oldModuleTemplate...)
```
Tomasz-Smelcerz-SAP, line 96: It is not clear to me if duplicates are possible here. Maybe it is, as the fetching functions already filter the data so that the lists are mutually different, but it's very hard to tell just by looking at the source code.
I understand that this is temporary, Maybe wrapping this in the `filterDuplicates` function - also temporarily - could help?

Tomasz-Smelcerz-SAP (changes requested):

internal/remote/remote_catalog.go
c-pius: I factored it out to `FilterModuleTemplatesToSync`.

> At this moment it's not possible.

If you meant testing the filtering logic, actually it was already fully tested and the setup with fake client rather simple. If you look into the test code, you will see that we just setup a list of ModuleTemplates and wrap it in the fake client.

internal/remote/remote_catalog.go
c-pius: The lists are mutually exclusive. Even if there were duplicate entries, I it wouldn't harm because we would just apply it twice to the SKR (first apply creates it, second apply leaves it unchanged). If you prefer I can still introduce the duplicate check.

pkg/testutils/moduletemplate.go
Tomasz-Smelcerz-SAP: Can you provide the list of such findings as a comment in the issue? I am pretty sure we'll forget about it.

tests/e2e/modulereleasemeta_sync_test.go
```diff
@@ -130,6 +210,18 @@ var _ = Describe("ModuleReleaseMeta Sync", Ordered, func() {
 				WithContext(ctx).
 				WithArguments(module.Name, RemoteNamespace, skrClient).
 				Should(Equal(ErrNotFound))
+
+			By("Then the Template Operator v2 ModuleTemplate exists in the KCP Cluster")
```
Tomasz-Smelcerz-SAP, line 214: Shouldn't it be like:
```suggestion
			By("And the Template Operator v2 ModuleTemplate exists in the KCP Cluster")
```

internal/remote/remote_catalog.go
```diff
@@ -102,9 +124,103 @@ func (c *RemoteCatalog) Delete(
...
+	for _, moduleTemplate := range moduleTemplates {
+		if moduleTemplate.IsMandatory() {
+			continue
+		}
+
+		if moduleTemplate.HasSyncDisabled() {
+			continue
+		}
+
+		if _, found := moduleTemplatesToSync[moduleTemplate.Name]; found {
```
Tomasz-Smelcerz-SAP, line 195: I am thinking if this is safe. Here we're matching the module based on it's `metadata.Name`. Wouldn't it be more robust to use module's `spec.ModuleName + "-" + spec.Version`? I know these two should be in-sync, but what if they're not?
What do you think?

c-pius: good point, better save than sorry. Adapted here: https://github.com/kyma-project/lifecycle-manager/pull/2097/commits/dfbc32e57d623b0b9d3e64ca12796efce1ca85c1

tests/e2e/modulereleasemeta_sync_test.go
c-pius: My intention was to indicate some grouping where we group together ModuleReleaseMeta assertions and ModuleTempalte assertions:

```
- When the ModuleReleaseMeta is deleted for the Template Operator Module
  - Then the ModuleReleaseMeta no longer exists on the KCP Cluster
  - And the ModuleReleaseMeta no longer exists on the SKR Cluster
  - Then the Template Operator v2 ModuleTemplate exists in the KCP Cluster
  - And Template Operator v2 ModuleTemplate no longer exists on the SKR Cluster
```

For me I found this easier to comprehend. But if you think this is a bad idea I can also change it to an And.

Tomasz-Smelcerz-SAP (approved):

PR #2111, opened 2024-12-11
feat: Prevent unallowed internal/beta module installation

c-pius:
**Description**

- Prevents the installation of beta or internal module for non-beta or non-internal Kymas
  - previously, installation was possible even though the metadata was not synced to the SKR
  - for old module templates we had this covered, but for new module templates this didn't work
- covered by unit tests
- covered by integration tests
- covered by e2e tests

**Related issue(s)**

- resolves https://github.com/kyma-project/lifecycle-manager/issues/2059


tests/integration/controller/kcp/module_installation_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+			if shouldHaveInstallation {
+				Eventually(expectInstallation, Timeout, Interval).WithArguments(kymaName, moduleName).Should(Succeed())
+			} else {
+				// we use Consistently here as the installation may require multiple reconciliation runs to be installed
+				// otherwise, we may get a false positive after the first reconciliation where no module is installed yet, but in a consecutive run it will be
+				Consistently(expectNoInstallation, Timeout, Interval).WithArguments(kymaName, moduleName).Should(Succeed())
+			}
+		},
+		Entry("Given Module{Beta: false, Internal: false}; Kyma{Beta: false, Internal: false}; Expect Installation: true", false, false, false, false, true),
```
Tomasz-Smelcerz-SAP, line 55: There are 16 possible flag combinations, but only 11 cases here. It's hard to see which ones are skipped and why. Perhaps it's worth to provide all 16 - if some make no sense, then maybe we can comment it out so that it's clear why it makes no sense.
What  do you think?

c-pius: Added the missing test cases.

Tomasz-Smelcerz-SAP (approved):

Issue #1638, opened 2024-06-20
Rename `AvailableModule`


c-pius:
### Description

`Kyma` struct has the method [`FetchAvailableModules()`](https://github.com/kyma-project/lifecycle-manager/blob/2fed199d47f5166a26c11cd524d5671052f53aa0/pkg/templatelookup/availableModules.go#L38C49-L38C64) returning an `AvailableModule` struct which basically is a `Module` accompanied by an `enabled` flag. It basically lists all from KymaCR Spec and Status and adds the info if enabled (in Spec) or disabled but still there (not in Spec, but still in Status).

Personally, I find the term "Available" highly confusing/misleading here. I would have expected something related to the modules available in the module catalog.

We need to refactor this and come up a proper name or avoid the explicit naming altogether.

### Reasons

Confusing naming makes it hard to understand and reason about the code.

### Acceptance Criteria

- [x] Consider renaming only or refactor the `AvailableModule` type holistically to make it more explicitly understandable and self-explanatory 
- [x] If refactoring is considered make it somehow explicit to indicate usage as DTO
- [x] #2140

medmes:
As a follow up ticket: #2140 , this ticket include the proposed improvement of refactoring of the `templatelookup` package and it's dependency.

PR #2120, opened 2024-12-13
refactor available module

medmes:
**Description**

- ...
- ...
- ...

**Related issue(s)**

#1638


go.mod
```diff
@@ -36,6 +36,7 @@ require (
 require (
 	github.com/go-co-op/gocron v1.37.0
 	github.com/kyma-project/template-operator/api v0.0.0-20240404131948-52c84f14e73c
+	github.com/mandelsoft/goutils v0.0.0-20241005173814-114fa825bbdc
```
ruanxin, line 39: please be careful what is imported, this is not what we want.

pkg/templatelookup/modulestatusinfo.go
```diff
@@ -13,52 +13,52 @@ var (
...
-func (a AvailableModule) installedwithVersionInStatus() bool {
-	return !a.Enabled && shared.NoneChannel.Equals(a.Channel) && a.Version != ""
+func (a ModuleStatusInfo) installedwithVersionInStatus() bool {
+	return !a.IsEnabled && shared.NoneChannel.Equals(a.Channel) && a.Version != ""
 }
 
-// FindAvailableModules returns a list of AvailableModule objects based on the Kyma CR Spec and Status.
-func FindAvailableModules(kyma *v1beta2.Kyma) []AvailableModule {
+// FetchModuleStatusInfo returns a list of ModuleStatusInfo objects based on the Kyma CR Spec and Status.
+func FetchModuleStatusInfo(kyma *v1beta2.Kyma) []ModuleStatusInfo {
```
ruanxin, line 38: But this is not all coming from module status, it contains the information from kyma.spec.modules

ruanxin (changes requested):
besides what's in the comment, please also give a summary of third AC, even if there is no alternative idea.

> If good new ideas can be gained here to improve the template_lookup pkg please create new issues for it!

go.mod
medmes: It shouldn t be added this line!

pkg/templatelookup/modulestatusinfo.go
c-pius: I see the point that `ModuleStatusInfo` may be confused with the `ModuleTemplate.status` sub-resource. How about we just name it `ModuleInfo` instead? Specific enough to say that it is information about a module, but generic enough so that we can place all information we want in it?

pkg/templatelookup/modulestatusinfo.go
```diff
@@ -13,52 +13,52 @@ var (
...
 // installedwithVersionInStatus returns true if the Module installed using a specific version (instead of a channel) is reported in Status.
-func (a AvailableModule) installedwithVersionInStatus() bool {
-	return !a.Enabled && shared.NoneChannel.Equals(a.Channel) && a.Version != ""
+func (a ModuleStatusInfo) installedwithVersionInStatus() bool {
+	return !a.IsEnabled && shared.NoneChannel.Equals(a.Channel) && a.Version != ""
 }
 
-// FindAvailableModules returns a list of AvailableModule objects based on the Kyma CR Spec and Status.
-func FindAvailableModules(kyma *v1beta2.Kyma) []AvailableModule {
+// FetchModuleStatusInfo returns a list of ModuleStatusInfo objects based on the Kyma CR Spec and Status.
```
c-pius, line 37: I think here it would also make sense to explain in more detail what this does. E.g., 

```go
// FetchModuleStatusInfo returns a list of [...] objects containing information about modules referenced by the Kyma CR.
// This includes modules that are enabled in `.spec.modules[]` and modules that are not enabled in `.spec.modules[]` but still contain an entry in `.status.modules[]`.
```

internal/manifest/parser/template_to_module.go
```diff
@@ -67,27 +67,27 @@ func (p *Parser) GenerateMandatoryModulesFromTemplates(ctx context.Context,
...
 	template *templatelookup.ModuleTemplateInfo, modules common.Modules,
 ) common.Modules {
 	if template.Err != nil && !errors.Is(template.Err, templatelookup.ErrTemplateNotAllowed) {
 		modules = append(modules, &common.Module{
 			ModuleName:  module.Name,
 			Template:    template,
-			Enabled:     module.Enabled,
-			IsUnmanaged: module.Unmanaged,
+			Enabled:     module.IsEnabled,
+			IsUnmanaged: module.IsUnmanaged,
```
c-pius, line 91: Thanks for starting to align these! Personally, I would vote for the other way round though (just `Enabled` and `Unmanaged`). We typically use `IsSomething` for methods, and just `Something` for the attribute.

See e.g., here https://github.com/kyma-project/lifecycle-manager/blob/31dca4c4202dc903b04199c232d397bf9abd6ddc/api/v1beta2/modulereleasemeta_types.go#L71-L77 or here https://github.com/kyma-project/lifecycle-manager/blob/31dca4c4202dc903b04199c232d397bf9abd6ddc/api/v1beta2/kyma_types.go#L391-L399

I know that this one `IsUnmanaged` occurrence is coming out of common.Module. If it is not a big deal we could include change htis here. If so, let's rather do a separate PR.


func (kyma *Kyma) IsBeta() bool {
	beta, found := kyma.Labels[shared.BetaLabel]
	return found && shared.IsEnabled(beta)
}

c-pius (commented):
For me it would be fine now. Just one minor fix in a comment, and I would propose to create a follow-up to extend the test cases.

ruanxin (approved):

PR #2129, opened 2024-12-16
fix: Flaky remote_sync_test

c-pius:
**Description**

- fixes the Flaky remote_sync_test integration test
- makes the re-try on create module templates more robust
- in total, 5 consecutive executions succeeded now (previously we had roughly a 50/50 chance)

**Related issue(s)**

### DIFF
--- a/pkg/testutils/moduletemplate.go
+++ b/pkg/testutils/moduletemplate.go
@@ -15,7 +15,9 @@ import (
 	"github.com/kyma-project/lifecycle-manager/pkg/util"
 )
 
-func CreateModuleTemplate(ctx context.Context, clnt client.Client, moduleTemplate *v1beta2.ModuleTemplate) error {
+func CreateModuleTemplate(ctx context.Context,
+	clnt client.Client,
+	moduleTemplate *v1beta2.ModuleTemplate) error {
 	moduleTemplate.SetResourceVersion("") // must be reset to enable retries
 	if err := clnt.Create(ctx, moduleTemplate); client.IgnoreAlreadyExists(err) != nil {
 		return fmt.Errorf("creating ModuleTemplate failed: %w", err)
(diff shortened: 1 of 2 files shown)


pkg/testutils/moduletemplate.go
```diff
@@ -27,7 +36,7 @@ func GetModuleTemplate(ctx context.Context,
 	}
 
 	moduleReleaseMeta, err := GetModuleReleaseMeta(ctx, module.Name, namespace, clnt)
-	if client.IgnoreNotFound(err) != nil {
+	if !meta.IsNoMatchError(err) && client.IgnoreNotFound(err) != nil {
 		return nil, fmt.Errorf("failed to get ModuleReleaseMeta: %w", err)
 	}
```
c-pius, line 41: to reviewer: this is what caused the trouble. When the CRD was not synced to the SKR cluster, this returned and error. In turn, the retries of creating the ModuleTemplates were failing because of the set `resourceVersion` and we ended in an endless loop.
It worked out when the other parallel tests including ModuleReleaseMeta had synced the CRD to the SKR beforehand.

medmes: Can we use this reset of the `resourceVersion` during the tearDown of the tests. that the next test suite will start fresh? or it's just for this specific case?

pkg/testutils/moduletemplate.go
```diff
@@ -14,6 +15,14 @@ import (
 	"github.com/kyma-project/lifecycle-manager/pkg/util"
 )
 
+func CreateModuleTemplate(ctx context.Context, clnt client.Client, moduleTemplate *v1beta2.ModuleTemplate) error {
```
medmes, line 18: func CreateModuleTemplate(ctx context.Context, client client.Client, moduleTemplate *v1beta2.ModuleTemplate) error {

medmes (approved):

PR #2139, opened 2024-12-18
feat(Maintenance Windows): Add skipMaintenanceWindows to Kyma Spec

### LINKED ISSUE #2100: [Maintenance Windows] Add `.spec.skipMaintenanceWindows` to Kyma
### Description

To support the use of [Maintenance Windows](https://github.com/kyma-project/kyma/issues/18611) during module upgrades, we need to provide Kyma user to opt-in the usage of maintenance windows. The user can decide by adding to he SKR Kyma spec.skipMaintenanceWindows, that all module upgrades that require a downtime are done during the defined maintenance windows by policy or not.
If the user decides to opt-out of the MW usage, module upgrades will happen as soon as KLM picks up the upgrade scenario from ModuleReleaseMetas in KCP.

We include this as an optional in the Kyma:
```
spec:
 skipMaintenanceWindows: <bool>
```
Since the information is optional, the default assumed is false. Meaning that by default users have their module version upgrades tied to the Maintenance Windows.

### Reasons

We need to harmonize Kyma related upgrade windows with BTP defined major update windows, so that if module upgrades require downtimes, the upgrade is scheduled within the BTP major upgrade window for the given region and plan.

### Acceptance Criteria

- [ ] `.spec.skipMaintenanceWindows` added to Kyma
- [x] Update documentation

### Feature Testing

Unit tests, End-to-End tests

### Testing approach

_No response_

### Attachments

_No response_

nesmabadr:
**Description**

- Add skipMaintenanceWindows to Kyma Spec
- Add documentation

**Related issue(s)**
Resolves #2100


docs/contributor/resources/01-kyma.md
```diff
@@ -19,6 +19,10 @@ It can be used to only listen to ModuleTemplate CRs provided under the same cont
 
 ## Configuration
 
+### **.spec.skipMaintenanceWindows**
+
+The skipMaintenanceWindows is used to indicate whether the module upgrades that require downtime should bypass the defined Maintenance Windows. If it is set to true, the module upgrade will happen as soon as there is a new module version in the Kyma Control Plane.
```
mmitoraj, line 24:
```suggestion
Use the **skipMaintenanceWindows** parameter to indicate whether the module upgrades that require downtime should bypass the defined Maintenance Windows. If it is set to `true`, the module upgrade will happen as soon as a new module version is released in the Kyma Control Plane. 
```

Tomasz-Smelcerz-SAP (approved):

PR #2145, opened 2024-12-20
refactor: Use Shared Rate Limiter Code

### LINKED ISSUE #2026: [Research] make rate limiting works as expected
The requeue metrics we observed in plutono dashboard is `manifest_sync_resources - unexpected`, which is requeued by following code in manifest reconciler, and should be guarded by rate limiting.

https://github.com/kyma-project/lifecycle-manager/blob/aaf6b0b0c618efafd89e0b4434982dd06e369c85/internal/declarative/v2/reconciler.go#L539-L542

The rate limiter we configured for manifest reconciler:
https://github.com/kyma-project/lifecycle-manager/blob/401ee8304608d740a83f6370b8488b96a39f154a/internal/rate_limiter.go#L11-L19

The parameter configured for the production landscape
```
- --failure-max-delay=30s
- --failure-base-delay=5s
- --rate-limiter-burst=2000
- --rate-limiter-frequency=1000
```
Based on the configuration, we should expected error item requeued with 5s interval, then increase exponentially to maximum 30s, however, it seems not the fact, the error item will requeued immediately.

**AC**
- [ ] research and find the root cause why the current setup is not working, in case it's a bug internally from controller-runtime, propose an alternative approach.
- [ ] use shared rate limiter code for manifest and kyma reconcociler, they currently use same configuration.
https://github.com/kyma-project/lifecycle-manager/blob/ed7ea6355ffe1d4067b42ece462b055376ec5933/cmd/main.go#L282 

**Attachements**
https://danielmangum.com/posts/controller-runtime-client-go-rate-limiting/

LeelaChacha:
Closes #2026


LeelaChacha:
Reasons for adding a map for intended requeue errors:
1. Not much overhead since it constant time complexity and it is the second condition in the if block, so if the error is nil, it won't waste cycles anyways.
2. Can be extended in the future with other errors
3. No function signatures had to be changed (SyncResources function can continue just returning an error, which makes sense instead of a boolean and an error)

internal/declarative/v2/reconciler.go
```diff
@@ -560,6 +560,15 @@ func (r *Reconciler) finishReconcile(ctx context.Context, manifest *v1beta2.Mani
 	return ctrl.Result{RequeueAfter: requeueAfter}, nil
 }
 
+func isIntendedRequeueError(err error) bool {
```
ruanxin, line 563: this idea is not bad, however, by introducing this, it means you have to make reconciler logic aware of lower-level error detail, like here you have to make reconciler knows `skrresources.ErrWarningResourceSyncStateDiff`, which is a typical case violate Dependency inversion principle: High-level modules should not import anything from low-level modules, but depends on abstractions.

The bad thing of existing code is, it makes `ErrWarningResourceSyncStateDiff` public, which makes possible to import here, but originally it's only designed for skrresources package internally, it should refactored as `errWarningResourceSyncStateDiff`.

Also to be aware, this anti pattern appears in existing code somewhere as well, we should not following it and introduce more.

btw, let's address this refactoring in another PR, but this one is only for use shared rate limiter code as your PR title indicated.

ruanxin (approved):

PR #2158, opened 2025-01-03
test: Labelling SKR resources

### LINKED ISSUE #1949: Implement an E2E Test verifying the expected labels on KymaCR
### Description

We currently don't have an E2E test that verifies that the `watched-by` label is present with the expected value `kyma` on the SKR KymaCR. We should either find an existing E2E test that can be extended with this, or we create an entire E2E test that is about metadata and verifies expected labels on multiple resources.

This also needs to consider that we sometimes already have such verification in E2E tests. E.g., https://github.com/kyma-project/lifecycle-manager/pull/1913 includes a verification that `managed-by` label is present on all raw manifest resources and tests that those are removed once the module is unmanaged.

Ideally, we could even think about this as a "sanity check" E2E test where the checks can be re-used in the more detailed E2E tests.

### Reasons

Some labels are significant to our implementations (e.g., the `watched-by` label). We need to make sure that those are consistent as expected.

### Acceptance Criteria

- [x] `watched-by` label verified in E2E tests
- [x] `managed-by` label verified in E2E tests

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

c-pius:
**Description**

- introduces labelling test that verifies if:
  - watched-by is present on KymaCR
  - managed-by is present on KLM-managed SKR resources
  - istio and warden labels are present on kyma-system namespace
  - managed-by is present on KLM-managed module manifest resources
  - managed-by is present on module default CR
- removes redundant label checks in other e2e tests where not relevant for the test itself

**Related issue(s)**

- resolves #1949


medmes (approved):
LGTM!

PR #2171, opened 2025-01-13
feat: Read Maintenance Window Policies

### LINKED ISSUE #2163: [Maintenance Windows] Read maintenance window policies
### Description

Split out of https://github.com/kyma-project/lifecycle-manager/issues/2101

KLM needs to read the maintenance window policies. It was agreed that for now it is sufficient to read it on startup (no continuous watch on the policy). Until it is rolled out to production, the policy config map and volume mount are only added to the local test kustomization target. Therefore, KLM should not error out when not finding the policy file. Instead, it should work with an empty policy (no maintenance windows). With this, KLM should remain functional even without maintenance window policy config map mounted.

### AC

- [ ] kustomization for local setup
  - [ ] creates config map for a default maintenance window policy
  - [ ] mounts the config map
- [ ] on startup, KLM reads the mounted file and resolves the maintenance window policy via the resolver module
  - [ ] read maintenance window policies cached
  - [ ] if the file is not found, it defaults to empty maintenance window policy (no maintenane windows)
- [ ] existing e2e tests remain functional
- [ ] new e2e test verifies that that the maintenance window policy can be read and resolved, and the default in case there is no file
  - [ ] focus on a minimal config, does not need to cover the full library (see https://github.com/kyma-project/lifecycle-manager/issues/2163#issuecomment-2579710911)
  - [ ] create a follow-up ticket to test the resolver more detailed and include test cases for this (it may also require input from SRE to come up with reasonable testdata as this is rather domain specific)
- [ ] unit tests verify the proper delegation of the maintenance window policy resolving via the resolver module

nesmabadr:
**Description**

- Add maintenance-config config map to local setup and mount the volume to the manager container in the klm deployment

**Related issue(s)**
Resolves #2163


lindnerby:
As discussed: For now we will keep the usage of resolver logic like it is here with reduced capabilities of testing and accept the low coverage. Within the follow-ups of [#2189](https://github.com/kyma-project/lifecycle-manager/issues/2189) & [#2190](https://github.com/kyma-project/lifecycle-manager/issues/2190) we will move the module to a separate public repo and replace the usage here to be able to reach max test coverage.

lindnerby (approved):

Issue #1085, opened 2023-11-24
Refactor testutils pkg


nesmabadr:
**Description:**

The testutils package has lots of files. We need to add subpackages or refactor the package so that common files are grouped into sub packages. Also, there still some test files which include their util functions within the files, they need to be moved to the testutils package.

**ACs:**
- [ ] Consider only utility functions from unit and E2E tests (not env tests)
- [ ] Refactor the testutils pkg (Create Subpackages)
- [ ] Move util functions from test files to the respective testutils pkg/file

Tomasz-Smelcerz-SAP:
I have a problem with this one.
I think the entire "testutil" package is a bad design choice.
The packages should have a purpose. The "code used in tests" is not a good purpose. If we don't have any valid reason to group the code according to it's role/purpose, then I'd say the code should be "close" to the place where it's actually used.
"How close" the code should be isn't precisely defined, but at the same time we don't have too many options there.


Considering that, I'd like to refine this ticket to have a more clear understanding of what we want to achieve.

My ideas:

- The first step would be to find out which code from "testutils" package is actually used in business code. If it's used in the business code, then it's not just a "testutil code" - it's a code that should have it's dedicated package.
- All that is left after the first step is a code that's really only used in tests. Then maybe we should move it closer to the tests, instead of keeping it in an artificial package?

Available options for location of the utility code for the tests:
- In the same file where the test is defined [We don't want this as this makes tests less readable]
- In the same package where the test is defined, but in a separate file [Leads to many files in a test package, harder to quickly identify which files are related to one another]
- In a sub-package of the package where the test is defined [Number of files doesn't change, but it's easier to identify related functionalities]

Of course there may be some corner cases like cross-referencing functions for different tests that cause problems when trying to move them into separate packages because of circular dependencies.
Such corner-cases must be handled case-by-case.

c-pius:
Agreed to:

- code that is used in business logic must be in a regular package
- code that is specific to one test should be close to the test (may be same file or file next to the test file)
- code that is common to multiple tests should be in a `commontestutils` package (may involve sub-packages, depending on how much there is left and how they can be grouped reasonably)

PR #2173, opened 2025-01-13
refactor pkg/testutils package

Tomasz-Smelcerz-SAP:
**Description**

Splitting the `pkg/testutils` package: first steps
I focused to move the test functions `closer` to the tests that are using them.
I tried to not modify any code - it's just moving of files and necessary package imports changes.

- moved some files into new package: `tests/e2e/commontestutils`
- moved some files into new package: `tests/integration/commontestutils`

**Related issue(s)**
See also: #1085


c-pius (commented):
it looks good to me, but I am uncertain about the left follow-ups. https://github.com/kyma-project/lifecycle-manager/issues/2174 is about specific helpers for that test, fair enough.

But what do we do with the rest remaining in `pkg/testutils`? If closing the origin ticket here, I think we still need to lay out a plan for how to continue with that. At least, we should put it to another location, e.g. `tests/commontestutils`

Tomasz-Smelcerz-SAP:
fair point. Personally I think that solving #2174 will tell us more about desired structures of packages for all the tests, not just this one. I chose `watcher_test.go` just to narrow the scope and avoid insanely huge PR.
I focused mostly on e2e test.
My investigation lead me to the following conclusion:
- We probably need to put every e2e test in it's own package (#2174 will tell)
- We probably need also to distribute the "helper" code into sub-packages - not necessarily the same as the packages containing the test scenarios (may be the same, may be different, may be some mix)

These two observations come mostly from the fact that current package structure results in the huge, unstructured "bags" of test code and this violates some well established software engineering practices: Like "Law of Demeter", "information hiding", "high cohesion" just to name the few.
So it's natural to think about splitting this into sub-packages. Because there is more than one way to do it, I decided we need a playground for trying out different approaches: #2174

c-pius (approved):
Okay, sounds like a good proposal to me. It is just be important to me that we can include explicitly in the PoC ticket that this should also include a proposal on the way forward with the `pkg/testutils` package.

Tomasz-Smelcerz-SAP:
> Okay, sounds like a good proposal to me. It is just be important to me that we can include explicitly in the PoC ticket that this should also include a proposal on the way forward with the `pkg/testutils` package.

I've updated the #2174 . Now it's quite a big PoC :)

Issue #2101, opened 2024-12-09
[Maintenance Windows] Skip module upgrades during maintenance window


lindnerby:
Depends on: https://github.com/kyma-project/lifecycle-manager/issues/2163

Once KLM can read maintenance window policies and have them cached, it needs to respect these for modules requiring downtime. Overall, the following cases need to be covered:

- [x] `Module.RequiresDowntime: FALSE`, `Kyma.UseMaintenanceWindows: FALSE` => module updated immediately
- [x] `Module.RequiresDowntime: TRUE`, `Kyma.UseMaintenanceWindows: FALSE` => module updated immediately
- [x] `Module.RequiresDowntime: TRUE`, `Kyma.UseMaintenanceWindows: TRUE` => module updated within next window, in the meantime reconciled in its currently installed version
- [x] `Module.RequiresDowntime: FALSE`, `Kyma.UseMaintenanceWindows: TRUE` => module updated immediately

### AC

- [x] Implemented logic for above cases
- [x] logic unit tested
- [ ] Cases E2E tested, see concept evaluation here: https://github.com/kyma-project/lifecycle-manager/issues/2164

### Related issue
* https://github.com/kyma-project/kyma/issues/18611

c-pius:
I was wondering if it is possible to split this into multiple issues:

- first ticket to make KLM aware of the maintenance windows config (volume mount, kustomize patch, ...)
- second ticket to investigate how we can do the testing (e.g., for e2e testing we need to find out about the granularity of the maintenance windows that can be maintained (minute level), and how to properly configure a test environment to test with it (can we predict the system time of the executed tests or do we need to temper with the system clock, ...))
- third ticket to implement the actual maintenance window logic and testing

Regarding the feature flag, I am wondering if this is needed. Existing E2E tests don't use module templates with `requiresDowntime`, so nothing changes for those, even when maintenance windows are maintained. For the new logic and new tests, starting point will be reading a file from the system. If the file is not found, we just treat it as no policies maintained instead of erroring out. Once we roll this out to production, we can change the latter part to requiring the file and erroring out if not found.

nesmabadr:
For the E2E test, we can add the maintenance windows component to the watcher_local_test Kustomization profile by using the command `kustomize edit add component ../maintenance_windows` from the watcher_local_test path in the workflow

c-pius:
[...]

PR #2187, opened 2025-01-16
test: Add Concept for Maintenance Window with Module Downtime E2E Test

### LINKED ISSUE #2164: [Maintenance Windows] Investigate maintenance window testing approach
### Description

Split out of https://github.com/kyma-project/lifecycle-manager/issues/2101

There needs to be an e2e test that verifies the module lifecycle when downtime is required. As it appears non-trivial to setup a test environment for the time-sensitive maintenance window, it should be investigated first how this can be done in e2e tests. E.g., can we arbitrarly modify the system clock so that we can easily alternate between being within our outside a maintenance window?

See https://github.com/kyma-project/lifecycle-manager/issues/2101 for cases that need to be covered by the test.

### AC

- [ ] proposed concept how to create such E2E test
- [ ] updated the AC in https://github.com/kyma-project/lifecycle-manager/issues/2101 accordingly to implement this test

LeelaChacha:
**Description**

- E2E test concept for Maintenance Window tests with module downtime required

**Related issue(s)**
Closes #2164


LeelaChacha:
Note: This test is supposed to fail. It should be green when [this issue](https://github.com/kyma-project/lifecycle-manager/issues/2101) is resolved.

tests/e2e/Makefile
```diff
@@ -171,3 +171,6 @@ modulereleasemeta-not-allowed-installation:
 
 labelling:
 	go test -timeout 20m -ginkgo.v -ginkgo.focus "Labelling SKR resources"
+
+modulereleasemeta-maintenance-window-with-module-downtime:
+	go test -timeout 20m -ginkgo.v -ginkgo.focus "Maintenance Window With ModuleReleaseMeta and Module Downtime"
```
c-pius, line 176:
```suggestion
	go test -timeout 20m -ginkgo.v -ginkgo.focus "Maintenance Window With ModuleReleaseMeta and Module Downtime"

```

.github/actions/deploy-template-operator-with-modulereleasemeta/action.yml
```diff
@@ -42,13 +42,22 @@ runs:
...
         yq eval '(. | select(.kind == "Deployment") | .metadata.name) = "${{ env.ModuleDeploymentNameInNewerVersion }}"' -i template-operator.yaml
         ./deploy_moduletemplate.sh ${{ env.ModuleName }} ${{ env.NewerVersion }}
+    - name: Create and apply Template Operator ModuleTemplate with ModuleDeploymentNameInNewerVersion with downtime
+      working-directory: template-operator
+      if: ${{ matrix.e2e-test == 'modulereleasemeta-maintenance-window-with-module-downtime' }}
+      shell: bash
+      run: |
+        make build-manifests
+        yq eval '(. | select(.kind == "Deployment") | .metadata.name) = "${{ env.ModuleDeploymentNameInNewerVersion }}"' -i template-operator.yaml
+        ./deploy_moduletemplate.sh ${{ env.ModuleName }} ${{ env.NewerVersion }} true false true true
```
c-pius, line 60: Maybe it would be easier to understand in the future if we consolidate this with `Create and apply Template Operator ModuleTemplate with ModuleDeploymentNameInNewerVersion` and in there we do a check:

```sh
if matrix.e2e-test == 'modulereleasemeta-maintenance-window-with-module-downtime
  REQUIRE_DOWNTIME = true
else
  REQUIRE_DOWNTIME = false

./deploy_moduletemplate.sh ${{ env.ModuleName }} ${{ env.NewerVersion }} true false true $REQUIRE_DOWNTIME
```

c-pius (commented):
Nice work!

.github/actions/deploy-template-operator-with-modulereleasemeta/action.yml
LeelaChacha: makes sense. but I would then also like to put the other bools in variables. So its clear what the trues and falses mean. Sounds good?

c-pius: Yes, sounds good 👍🏻

c-pius: I think I also created a ticket earlier to make the script inputs named arguments for the same reason :)

c-pius (approved):

PR #2196, opened 2025-01-20
feat: Implement MaintenanceWindow determination logic

c-pius:
**Description**

- adds the logic to determine if a maintenance window is required and active
- passes this implementation to the template lookup where it will be used
  - it is not used yet, just passed
  - in the next PR, we will make use of this new functionality

**Related issue(s)**

- https://github.com/kyma-project/lifecycle-manager/issues/2101


tests/integration/controller/kyma/suite_test.go
```diff
@@ -134,6 +137,7 @@ var _ = BeforeSuite(func() {
 	kcpClient = mgr.GetClient()
 	testEventRec := event.NewRecorderWrapper(mgr.GetEventRecorderFor(shared.OperatorName))
 	testSkrContextFactory := testskrcontext.NewSingleClusterFactory(kcpClient, mgr.GetConfig(), testEventRec)
+	maintenanceWindowHandler, _ := maintenancewindows.InitializeMaintenanceWindow(logr, "policy", "/etc/maintenance-policy")
```
LeelaChacha, line 140: policiesDirectory and policyName need to be interchanged. Or perhaps it could be made clear that an empty policy will be used.

LeelaChacha, line 140: here as well.

cmd/main.go
```diff
@@ -193,14 +194,16 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
 	mandatoryModulesMetrics := metrics.NewMandatoryModulesMetrics()
 
 	// The maintenance windows policy should be passed to the reconciler to be resolved: https://github.com/kyma-project/lifecycle-manager/issues/2101
-	_, err = maintenancewindows.InitializeMaintenanceWindowsPolicy(setupLog, maintenanceWindowPoliciesDirectory,
+	maintenanceWindow, err := maintenancewindows.InitializeMaintenanceWindow(setupLog,
+		maintenanceWindowPoliciesDirectory,
```
LeelaChacha, line 198: just a suggestion but maintenance window is a bit vague. maintenanceWindowHandler (like in the integration tests) is more expressive. WDYT?

cmd/main.go
```diff
@@ -193,14 +194,16 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
 	mandatoryModulesMetrics := metrics.NewMandatoryModulesMetrics()
 
 	// The maintenance windows policy should be passed to the reconciler to be resolved: https://github.com/kyma-project/lifecycle-manager/issues/2101
-	_, err = maintenancewindows.InitializeMaintenanceWindowsPolicy(setupLog, maintenanceWindowPoliciesDirectory,
+	maintenanceWindow, err := maintenancewindows.InitializeMaintenanceWindow(setupLog,
```
LeelaChacha, line 197: remove comment please

LeelaChacha (changes requested):

cmd/main.go
c-pius: I actually removed it intentionally (the integration test part was just an occurrence that I missed :)). I am super open to discuss this but `handler` (similar to `service`) doesn't really add any benefit in my opinion as it is still super generic. I think that `MaintenanceWindow#IsActive` or `MaintenanceWindow#IsRequired` actually speaks for itself without Handler or Service addition. But I admit that I also am guilty of adding such generic suffixes often, and I try to avoid it now.

Maybe @lindnerby also has an opinion on it?

This is also a related pretty funny read: http://steve-yegge.blogspot.com/2006/03/execution-in-kingdom-of-nouns.html

lindnerby: Since InitializeMaintenanceWindow is returning a MaintenanceWindow, so the variable maintenanceWindow is fine imo. I agree that Handler and Service are overused in general, just like Manager, and would try to use something more specific and descriptive. Funny article btw. :)

LeelaChacha: Alright, I agree with your points. Just thought that the name maintenanceWindow suggests that there is just one. But not really that opinionated on it. I‘mma approve

Issue #2083, opened 2024-12-02
Investigate need for `kyma-system` cache options


c-pius:
### Description

Our cache options include Kyma and ModuleTemplate resources in `kyma-system` namespace, see:

- https://github.com/kyma-project/lifecycle-manager/blob/9491cf4b96ee99d959af783cee3cd4bc2098b0f9/internal/cache_options.go#L50
- https://github.com/kyma-project/lifecycle-manager/blob/9491cf4b96ee99d959af783cee3cd4bc2098b0f9/internal/cache_options.go#L56

This in turn requires us to have extra permissions for `kyma-system` namespace: https://github.com/kyma-project/lifecycle-manager/blob/9491cf4b96ee99d959af783cee3cd4bc2098b0f9/config/rbac/namespace_bindings/skr_role.yaml

We need to investigate if this indeed enables us to cache resources that we fetch from the SKR, or if this would be caching resources in `kyma-system` namespace on KCP which don't exist. In fact, all SKR resources are fetched via a client that is configured using the kubeconfig secret of the SKR. That's why we also don't need extra permissions set for the SKR. If it indeed works, we should:

a) clearly document why those entries are in the cache and what the additional permissions are needed for
b) include ModuleReleaseMeta in this configuration as it should be treated the same way as ModuleTemplates

If it turns out that this doesn't work as expected, we should remove this unnecessary cache options config and permissions (also including management plane charts).

Thanks @nesmabadr for supporting the discovery of this topic!

### Reasons

Keep the project clean.

### Acceptance Criteria

- [x] found whether the cache options for resources in the SKR works as expected
- [ ] if not:
  - [x] removed config from cache options
  - [x] removed skr_role.yaml
  - [x] removed skr_role from management-plane-charts (including the automatic fetching)
- [ ] if so:
  - [ ] added ModuleReleaseMeta to the configuration
  - [ ] clearly documented:
    - [ ] why we have those entries in cache options
    - [ ] why we have skr_role.yaml

medmes:
# Summary of Findings and Suggestions

There is no built-in way to programmatically confirm whether a resource read was fetched from the cache or directly from the API server. In client-go, the behavior depends on how the resource is accessed:

- Using **mgr.GetClient()** follows the default client-go behavior, where cache management is handled internally.

- Using **mgr.GetAPIReader()** fetches the resource directly from the API server, bypassing the cache.

[...]

PR #2198, opened 2025-01-21
feat: Removed skr permission  and cache config for remote kyma & ModuleTemp…

medmes:
**Description**
After investigation , When using the default client configuration(which is the normal behaviour we use) there's no possibility to verify if the resource is retrieved by infomrer queue. or from a direct API read.
The only way, is to roll out the resources cache configuration and test it by observing metrics on Stage.

- Removed cache config for Kyma and ModuleTemplate resource.
- Remove unnessery role, role_binding manifests for skr permission.

**Related issue(s)**
#2083


internal/cache_options.go
```diff
@@ -47,14 +47,12 @@ func (c *KcpCacheOptions) GetCacheOptions() cache.Options {
 			},
 			&v1beta2.Kyma{}: {
 				Namespaces: map[string]cache.Config{
-					c.remoteNamespace: {},
-					c.kcpNamespace:    {},
+					c.kcpNamespace: {},
```
lindnerby, line 50: The field `remoteNamespace string` can be removed from `KcpCacheOptions` as well, right?

lindnerby (changes requested):

PR #2199, opened 2025-01-21
fix: Fix the syncing of ModuleReleaseMeta Beta and Internal values when edited on the SKR Cluster

nesmabadr:
**Description**

- Remove the omitEmpty, because this makes the serialization omit this field if it's of zero-value (false in case of boolean). It is replaced with optional
- Add E2E test to cover this functionality

**Related issue(s)**


pkg/testutils/modulereleasemeta.go
```diff
@@ -147,3 +153,33 @@ func UpdateAllModuleReleaseMetaChannelVersions(ctx context.Context, client clien
...
+
+func ModuleReleaseMetaInternalValueIsCorrect(ctx context.Context, client client.Client, namespace, name string,
+	expectedValue bool,
+) error {
+	meta := &v1beta2.ModuleReleaseMeta{}
+	if err := client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, meta); err != nil {
+		return err
+	}
+
+	if meta.Spec.Beta != expectedValue {
```
c-pius, line 180: This must be `meta.Spec.Internal`. Btw. We would also have `IsInternal` / `IsBeta` helpers on the MRM

tests/e2e/modulereleasemeta_sync_test.go
```diff
@@ -59,6 +59,45 @@ var _ = Describe("ModuleReleaseMeta Sync", Ordered, func() {
...
+		})
+
+		It("When Beta value is set to true on the SKR Cluster", func() {
+			Eventually(SetModuleReleaseMetaBeta).
+				WithContext(ctx).
+				WithArguments(true, module.Name, RemoteNamespace, skrClient).
+				Should(Succeed())
+		})
+
+		It("Then Beta value is reverted back to its value on the KCP Cluster", func() {
```
c-pius, line 82:
```suggestion
		It("Then Beta value is reverted back to its value from the KCP Cluster", func() {
```

Maybe this makes it a bit clearer. I was first reading it like we check something on the KCP cluster.

tests/e2e/modulereleasemeta_sync_test.go
```diff
@@ -59,6 +59,45 @@ var _ = Describe("ModuleReleaseMeta Sync", Ordered, func() {
...
+		})
+
+		It("When Internal value is set to true on the SKR Cluster", func() {
+			Eventually(SetModuleReleaseMetaInternal).
+				WithContext(ctx).
+				WithArguments(true, module.Name, RemoteNamespace, skrClient).
+				Should(Succeed())
+		})
+
+		It("Then Internal value is reverted back to its value on the KCP Cluster", func() {
```
c-pius, line 96:
```suggestion
		It("Then Internal value is reverted back to its value from the KCP Cluster", func() {
```

c-pius (changes requested):

c-pius (approved):

PR #2200, opened 2025-01-21
feat: Refactor template lookup to use strategy pattern

c-pius:
**Description**

- refactors the template lookup to use a strategy pattern for looking up the module template info
  - previously we had lot's of branching depending on the module metadata (e.g., old or new data, installed via version or channel)
  - now we have dedicated strategies for each of those cases
    - each strategy determines itself whether it applies or not
    - if it applies, it can look up the module template info based for the case it covers
  - the aim is to make it easier to adapt the cases in the future (add, remove, refine)
 
**:bulb: Note to reviewer**

- this PR is huge (apologies), the actual changes are not that much though
- focus on the changes in `templatelookup`
  - the existing functionality has been moved from `regular.go` to the strategies in `moduletempalteinfolookup`
  - the business logic is just copied, no new functionality has been added
- the rest is boilerplate for setting up the strategies and adapting the changed method signatures in tests

**Related issue(s)**

- this is preparation for https://github.com/kyma-project/lifecycle-manager/issues/2101


pkg/templatelookup/moduletemplateinfolookup/by_version_test.go
```diff
@@ -0,0 +1,106 @@
+package moduletemplateinfolookup_test
+
```
lindnerby, line 2: Also here, some branches deserve to be covered:
* When filterTemplatesByVersion returns an error due to failing to list module templates
* When filterTemplatesByVersion returns an error due to more than one template candidate
* When filterTemplatesByVersion returns an error due to no templates found
* When filterTemplatesByVersion returns an error due to the template being marked as mandatory

pkg/templatelookup/regular_test.go
```diff
@@ -893,18 +835,23 @@ func TestTemplateLookup_GetRegularTemplates_WhenModuleTemplateNotFound(t *testin
...
 		t.Run(testCase.name, func(t *testing.T) {
 			givenTemplateList := &v1beta2.ModuleTemplateList{}
-			lookup := templatelookup.NewTemplateLookup(NewFakeModuleTemplateReader(*givenTemplateList,
-				v1beta2.ModuleReleaseMetaList{}),
+			reader := NewFakeModuleTemplateReader(*givenTemplateList,
+				v1beta2.ModuleReleaseMetaList{})
+			lookup := templatelookup.NewTemplateLookup(reader,
 				provider.NewCachedDescriptorProvider(),
-				maintenanceWindowStub{})
+				moduletemplateinfolookup.NewAggregatedModuleTemplateInfoLookupStrategy([]moduletemplateinfolookup.ModuleTemplateInfoLookupStrategy{
```
lindnerby, line 850: Since these dependencies are now abstracted via interface, the unit tests of `TemplateLookup` should work on mock implementations.

pkg/templatelookup/regular_test.go
```diff
@@ -1139,13 +1091,53 @@ func TestTemplateNameMatch(t *testing.T) {
...
+func executeGetRegularTemplatesTestCases(t *testing.T,
+	testCases getRegularTemplatesTestCases,
+	availableModuleTemplates v1beta2.ModuleTemplateList,
+	availableModuleReleaseMetas v1beta2.ModuleReleaseMetaList,
+	moduleToInstall v1beta2.Module,
+) {
+	t.Helper()
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			reader := NewFakeModuleTemplateReader(availableModuleTemplates, availableModuleReleaseMetas)
```
lindnerby, line 1118: We don't need constructor functions for mock impls. Also can be renamed to `MockReader` or `MockModuleTemplateReader`

pkg/module/sync/runner.go
```diff
@@ -275,7 +276,7 @@ func generateModuleStatus(module *common.Module, existStatus *v1beta2.ModuleStat
 		newModuleStatus.Message = module.Template.Err.Error()
```
lindnerby, line 276: Can we provide unit tests for `generateModuleStatus`? Would make refactoring safer.
Also possible as follow-up I guess.

lindnerby (changes requested):
Nice job! Just left some mostly testing related suggestions 🙌

pkg/templatelookup/moduletemplateinfolookup/by_module_release_meta_test.go
```diff
@@ -0,0 +1,80 @@
+package moduletemplateinfolookup_test
+
```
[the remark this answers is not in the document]
c-pius: > Also there is no handling of GetChannelVersionForModule returns empty string and no error. Is this a valid scenario?

The Channel has a validation for minLength: 3. So calling it with "" will return an ErrChannelNotFound. If you prefer we can still add the validation that input channel and read channels have are not empty. But not strictly necessary I think.

pkg/templatelookup/moduletemplateinfolookup/by_version.go
```diff
@@ -0,0 +1,82 @@
...
+	var filteredTemplates []*v1beta2.ModuleTemplate
+	for _, template := range templateList.Items {
+		if TemplateNameMatch(&template,
+			name) && shared.NoneChannel.Equals(template.Spec.Channel) && template.Spec.Version == version {
+			filteredTemplates = append(filteredTemplates, &template)
+			continue
+		}
+	}
+	if len(filteredTemplates) > 1 {
+		return nil, newMoreThanOneTemplateCandidateErr(name, templateList.Items)
```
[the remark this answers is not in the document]
c-pius: I would say that we then should also be consistent in testing the other helper funcs from common.go. I have added this here: https://github.com/kyma-project/lifecycle-manager/issues/2202

I don't want to appear lazy here, but I think it is quite a big task to do good testing of the overall package and will also need to refactor parts of the internal workings. Therefore I would propose to tackle this as a focused topic with a dedicated ticket.

pkg/templatelookup/moduletemplateinfolookup/by_version_test.go
c-pius: https://github.com/kyma-project/lifecycle-manager/issues/2202

pkg/module/sync/runner.go
c-pius: Yes, also makes sense. But again I would like to source this out as this is not the focus of this PR.
https://github.com/kyma-project/lifecycle-manager/issues/2205

lindnerby (approved):

Issue #2165, opened 2025-01-07
[Maintenance Windows] Align on open questions with SRE wrt. maintenance windows


c-pius:
### Description

Once the maintenance window feature is ready, we can roll it out to production. Before we do so, we should align on a couple of decisions. 

Regarding the handling of missing policies: In its current implementation, we don't fail on startup of lifecycle manager if we don't find a maintenance window policy. What happens instead is that if we determine that we require a maintenance window for a module update, we will:

- update if we find an active one
- wait without changing the module state if we find one but it is in the future
- wait with setting the module to error state if we don't have a policy file
- wait with setting the module to error state if the matching logic doesn't find a matching window
  - the error is coming from the library already

**Timbeboxed to 1 day**

### AC

- [x] ~align whether we want to enable it for "old" module templates~ - we've agreed to not support the feature for old MTs, if someone wants the Maintenance Windows they should migrate
  - [x] as of now only enabled for new module templates that are accompanied by ModuleReleaseMeta
  - [x] enabling it for old module templates should be a one liner, we require additional e2e testing though which may be a bit cumbersome to setup
  - [x] even if we enable it, we would require the Neighbors publishing pipeline to be updated (Kyma CLI does not support the `requiresDowntime` field)
- [x] align on configuration values that we use for the window
  - [x] the library provided by SRE offers some configuration options, we should discuss with SRE which options to use https://github.com/kyma-project/lifecycle-manager/blob/main/maintenancewindows/resolver/maintwindow.go#L41-L65
  - [x] use ongoing windows: currently set to true as we require active ongoing windows for the reconciliation
  - [x] min duration of ongoing window: currently set to 20 mins, we need to discuss what a reasonable time frame is (maybe we should also ask SRE what the typical window size is)
  - [x] the rest is currently defaulted
  - [x] consult SRE on which behaviour should be applied in case of missing configs/policies
  - [x] how frequently are policies changing (is it sufficient to read at startup or should we put a watch on it)
- [x] decided whether missing maintenance window policy config file is continued to be defaulted or if KLM shall error out
  - [ ] management-plane-chart config updated to include volume mount for maintenance window policy

c-pius:
Questions to be clarified:

- [ ] how frequently is the maintenance window configuration changing?
  - as of now, we only read it on startup on KLM
  - if it changes rather frequently, we may need to read it more often
- [ ] what are typical durations of a maintenance window?
[...]

PR #2206, opened 2025-01-23
feat: Skip module updates requiring maintenance windows when not active

c-pius:
**Description**

- Implements a decorator implementing the maintenance window logic
- once it looked up the module template via it's decorated instance, it:
  - checks if a maintenance window is required
  - if so, checks if it is active
- if it is not active or an error happens, it sets an error and the template to nil
  - as a consequence, the manifest will not be updated and we continue to reconcile its current version
  - additionally, we display a message in the kyma module status
  - this is the same behavior as we have already for the version downgrade

**Related issue(s)**

- resolves: https://github.com/kyma-project/lifecycle-manager/issues/2101


pkg/module/sync/runner.go
```diff
@@ -350,6 +330,41 @@ func generateModuleStatus(module *common.Module, existStatus *v1beta2.ModuleStat
...
+		}
+	case errors.Is(module.Template.Err, moduletemplateinfolookup.ErrWaitingForNextMaintenanceWindow):
+		newModuleStatus := existStatus.DeepCopy()
+		newModuleStatus.Message = module.Template.Err.Error()
+		return *newModuleStatus
+	case errors.Is(module.Template.Err, moduletemplateinfolookup.ErrFailedToDetermineIfMaintenanceWindowIsActive):
+		newModuleStatus := existStatus.DeepCopy()
+		newModuleStatus.Message = module.Template.Err.Error()
+		newModuleStatus.State = shared.StateError
+		return *newModuleStatus
```
c-pius, line 356: Note to reviewer: these are the two now conditions that have been added. The rest here is just moving the existing conditions to the new helper func. If not moving to helper func, linter complains that the func is too long.

lindnerby (dismissed):
lgtm

cmd/main.go
```diff
@@ -291,7 +296,8 @@ func setupKymaReconciler(mgr ctrl.Manager, descriptorProvider *provider.CachedDe
 	moduleTemplateInfoLookupStrategies := moduletemplateinfolookup.NewModuleTemplateInfoLookupStrategies([]moduletemplateinfolookup.ModuleTemplateInfoLookupStrategy{
 		moduletemplateinfolookup.NewByVersionStrategy(mgr.GetClient()),
 		moduletemplateinfolookup.NewByChannelStrategy(mgr.GetClient()),
-		moduletemplateinfolookup.NewByModuleReleaseMetaStrategy(mgr.GetClient()),
+		moduletemplateinfolookup.NewWithMaintenanceWindowDecorator(maintenanceWindow,
+			moduletemplateinfolookup.NewByModuleReleaseMetaStrategy(mgr.GetClient())),
```
c-pius, line 300: I have only enabled the maintenance window decorator for the new module metadata. We can discuss in the daily if we want to enable it for old module metadata as well. If so, we need to invest some more time to also e2e test these scenarios. I have moved the decision to the rollout issue: https://github.com/kyma-project/lifecycle-manager/issues/2165

lindnerby (approved):

PR #2207, opened 2025-01-24
chore: RBAC cleanup

lindnerby:
**Description**

- Remove kubebuilder markers for RBAC creation from various controllers, since it will always create ClusterRole
- Restructure & cleanup RBAC resources
- Change manager ClusterRole to kcp-system namespaced Role
- Remove duplicate Roles
- Additional cleanup

**Related issue(s)**


config/watcher_local_test/kustomization.yaml
```diff
@@ -1,90 +1,79 @@
...
-# When deployed on KCP it will be set by the default kustomize.
+resources:
+  - namespaces.yaml
+  - ../control-plane
+components:
+  - ../istio
+  - ../maintenance_windows
 patches:
-- path: patches/adjust_resources_for_local_setup.yaml
-- patch: |-
```
c-pius, line 25: Possibility to unified approach.

config/rbac/manager_role.yaml
```diff
@@ -0,0 +1,188 @@
...
+    resources:
+      - configmaps
+    verbs:
+      - create
+      - delete
+      - get
+      - list
+      - patch
+      - update
+      - watch
```
c-pius, line 19: Do we really need all verbs for config maps?

config/rbac/manager_role.yaml
```diff
@@ -0,0 +1,188 @@
...
+      - ""
+    resources:
+      - secrets
+    verbs:
+      - create
+      - delete
+      - get
+      - list
+      - update
+      - watch
```
c-pius, line 40: Do we really need all verbs on secrets? We already had a test PR with reduced permissions where E2E tests worked: https://github.com/kyma-project/lifecycle-manager/pull/2081

config/rbac/manager_role.yaml
```diff
@@ -0,0 +1,188 @@
...
+    resources:
+      - kymas
+    verbs:
+      - create
+      - delete
+      - get
+      - list
+      - patch
+      - update
+      - watch
```
c-pius, line 97: Are create and delete really needed? Should only be needed for SKR where the permissions are independent from these here.

c-pius: I am also wondering if we could reduce this to metadata even. Status is a different sub-resource and spec we shouldn't be writing on KCP.

config/rbac/manager_role.yaml
```diff
@@ -0,0 +1,188 @@
...
+    resources:
+      - watchers
+    verbs:
+      - create
+      - delete
+      - get
+      - list
+      - patch
+      - update
+      - watch
```
c-pius, line 174: Is the watcher CR really created and deleted by lifecycle-manager? I thought this is coming from argo.

config/control-plane/kustomization.yaml
```diff
@@ -1,42 +1,31 @@
...
   - ../grafana
-
 patches:
-  - patch: |-
+  - path: patches/deployment_resources.yaml
+  - path: patches/unique_deployment_webhook_patch.yaml
+  - path: patches/unique_certificate_name.yaml # Override certificate name to ensure a unique CM Cert when run with other kubebuilder operators
+  - target:
+      kind: Deployment
+    patch: |-
```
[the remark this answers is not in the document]
lindnerby: I also assume that the different kind of patch mechanism, have different kinds of feature sets. The inline patch here e.g. supports list changes, etc. but not sure.
Like the thoughts about naming conventions. We should move both of these topics to a dedicated issue.
Stub here: https://github.com/kyma-project/lifecycle-manager/issues/2281

config/control-plane/kustomization.yaml
```diff
@@ -64,64 +53,56 @@ patches:
...
-      kind: ConfigMap
-      name: dashboard-(overview|status|watcher|mandatory-modules)
-      version: v1
-  - path: patches/unique_manager_webhook_patch.yaml
-  - path: patches/adjust_resources_in_deployment.yaml
-    # We override the certificate name to ensure that Cert-Manager uses a unique cert in conjunction with other
-    # kubebuilder operators.
-  - path: patches/unique_certificate_name.yaml
-
 # Note: Now as the 'patchesJson6902' is deprecated, the direct use of the 'PatchTransformer' is the only way to change a resource namespace to something different from the value configured by the global namespace transformer.
```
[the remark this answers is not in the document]
lindnerby: No idea to be honest, there is still a PatchTransformer used below.

PROJECT
```diff
@@ -107,13 +107,4 @@ resources:
   webhooks:
     conversion: true
     webhookVersion: v1
-- api:
```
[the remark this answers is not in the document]
lindnerby: We don't need to, it would have have informational purpose now.
If you run `kubebuilder init` or `kubebuilder create api` you can make use of the resources part to scaffold things.

config/watcher_local_test/namespaces.yaml
```diff
@@ -6,4 +7,4 @@ metadata:
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: kyma-system
\ No newline at end of file
+  name: kyma-system
```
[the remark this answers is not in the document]
lindnerby: Great catch! Did not see this!

config/rbac/manager_role.yaml
lindnerby: Probably this was before the introduction of istiogatewaysecret controller. We create/get/update secrets there. I guess we can omit delete from rbac role here though. 👍

config/rbac/manager_role.yaml
lindnerby: Probably not. I reduced it to "read" policy rules.

config/rbac/manager_role.yaml
lindnerby: True, I removed the the create and delete!

PR #2216, opened 2025-01-27
fix: Memory Leak in transport.tlsTransportCache

### LINKED ISSUE #1714: Linear growth of the memory consumption
### Description

The LM Pod memory consumption grows linearly. This may indicate a memory "leak" or some ever-growing cache.
Requires investigation

**Timebox: 3 days**

### Steps to reproduce

- Take a look at the collected LM memory usage metrics

### Environment Type

Managed

### Environment Info

Kubernetes Version: x.y.z
Docker Version: x.y.z
Kyma CLI Version: x.y.z

### Attachments

The "Drops" in the graph are related to the new LM instance (version update or redeployment). It's not GC-related memory release. Every color is a different LM instance.

image

LeelaChacha:
#### The cause of the memory leak:
These are the SKR Kubeconfig rotation intervals:

> DEV: rotation interval is 1 h / max. lifetime: 1:40 h
> STAGE: rotation interval is 7:12 h / max. lifetime: 12h
> PROD: rotation interval is 14:24 h / max. lifetime: 24h

Due to the certificate rotation, the [`tlcCacheKey` that the transport package calculates](https://github.com/kubernetes/kubernetes/blob/2bda5dd8c794d933145938affca30f012986e8fa/staging/src/k8s.io/client-go/transport/cache.go#L163) for the tlsConfigs of the SKR Clients was changing, hence after rotation, the cache would grow, with older tlsConfigs never being evicted.

#### Workaround:
There is no option to disable the caching in the transport package. A workaround is setting the restConfig's `Proxy` field to `http.ProxyFromEnvironment`. This field is nil in our case, and the transport package assigns it the value [`http.ProxyFromEnvironment` by default](https://github.com/kubernetes/kubernetes/blob/2bda5dd8c794d933145938affca30f012986e8fa/staging/src/k8s.io/client-go/transport/cache.go#L129) anyway. However, before the default assignment, [the `tlsCacheKey` is generated](https://github.com/kubernetes/kubernetes/blob/2bda5dd8c794d933145938affca30f012986e8fa/staging/src/k8s.io/client-go/transport/cache.go#L76) and here, if the `Proxy` is set, the package [does not cache](https://github.com/kubernetes/kubernetes/blob/2bda5dd8c794d933145938affca30f012986e8fa/staging/src/k8s.io/client-go/transport/cache.go#L160) the config. So, we effectively disable the caching with a NoOp.

---

Closes #1714


LeelaChacha:
**Notes for reviewer:**

1. These are the SKR Kubeconfig rotation interval:

> DEV:  rotation interval is 1 h  / max. lifetime: 1:40 h
> STAGE: rotation interval is 7:12 h  /  max. lifetime: 12h
> PROD: rotation interval is 14:24 h  /  max. lifetime: 24h

Due to the certificate rotation, the tlcCacheKey that the transport package calculates for the tlsConfigs of the SKR Clients was changing, hence after rotation, the cache would grow, with older tlsConfigs never being evicted.

2. To test, simulate this behaviour in your cluster.
3. You can use the following method to check the memory consumption of the HTTPClientFor method in each manifest reconcile loop:

```golang
printFunctionMemoryConsumption("HTTPClientFor", "inuse_objects")

...

func printFunctionMemoryConsumption(functionName string, valueType string) {
	valueTypesToIndexMap := map[string]int{"alloc_objects": 0, "alloc_space": 1, "inuse_objects": 2, "inuse_space": 3}
	valueTypeIndex := valueTypesToIndexMap[valueType]

	var buf bytes.Buffer
	_ = pprof.WriteHeapProfile(&buf)
	prof, _ := profile.Parse(&buf)

	count := 0
	for _, sample := range prof.Sample {
		for _, loc := range sample.Location {
			for _, line := range loc.Line {
				if line.Function != nil && strings.Contains(line.Function.Name, functionName) {
					count += int(sample.Value[valueTypeIndex])
				}
			}
		}
	}
	fmt.Printf("-- %s for function %s: %d\n", valueType, functionName, count)
}
```

Alternatively, you can use the pprof server in the KLM using the flag `--pprof`
4. The pprof dumps seem to have an object/space threshold, so in order to see values for the HTTPClientFor or other methods, more than a few Manifests are required.

c-pius (approved):

c-pius (approved):

Issue #2089, opened 2024-12-03
Trigger warning for PRs likely requiring updates to management-plane-charts


c-pius:
### Description

Find a way that notifies us when PRs to lifecycle-manager are likely requiring changes to management-plane-charts. For instance, changes to the configuration or the E2E test files (we used to have a github action that was comparing the make manifests" outcome between main head and PR). If there are changes, it is likely that we also need to modify something in management-plane-charts and the developer should double check.

Document what is expected from the assignee in case the above is triggered. E.g., extend the AC of the original issue to account for the changes in management-plane-charts, create a PR that is preparing and testing the required change and can serve as a base for the release manager to create the new release.

### Reasons

Prevent delayed releases due to the release manager finding out too late that changes to the charts are needed.

### Acceptance Criteria

**Timeboxed to 1 day**
- [x] approach identifying critical changes aligned and implemented
- [x] documentation available stating what should be done if so

medmes:
## Proposal

### 1. Check Config Changes
**File:** `.github/workflows/check-config-changes.yaml`  
**Purpose:**  
Detects changes in configuration files (e.g. under `config/` and `e2e/`) on pull requests. If changes are detected, a warning comment is posted and a `configs-changed` label is applied.

**Key Steps:**
- **Get changed files:** Lists and filters files changed in the PR.
- **Evaluate changes:** Sets an output (`config_changed`) via `GITHUB_OUTPUT`.
- **Notify:** Posts a comment and applies a label if changes exist.

---

### 2. Check Manifests Changes
**File:** `.github/workflows/check-manifests.yaml`  
**Purpose:**  
Generates manifests for both the PR and the main branch, compares them, and flags differences. If differences are detected, a warning comment and an `outdated-manifests` label are applied; otherwise, the label is safely removed.

**Key Steps:**
- **Generate & Cache:** Runs manifest generation on both the PR and main branches.
- **Compare:** Compares the cached manifests and sets an `outdated_manifests` output.
- **Notify & Manage Label:** Posts a comment/adds a label if differences exist; otherwise, removes the label if present.

---
### Changes on management-plane-chart
- Refer to lifecycle-manager branch and changes needed files configs 
-- kustomize into helm charts

PR #2219, opened 2025-01-30
Trigger warning for PRs likely requiring updates to management-plane-charts

medmes:
**Description**

This PR introduce a Github wokflow for detecting changes on the config directory which needs to be reflected into management-plane-chart repository.

- Github Workflow.

**Related issue(s)**
#2089

### DIFF
--- a/.github/workflows/check-configs-changes.yaml
+++ b/.github/workflows/check-configs-changes.yaml
@@ -1,0 +1,66 @@
+name: "Check for Configs Changes"
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  check-config-changes:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 2  # Ensure we can compare with the base branch
+
+      - name: Get list of changed files
+        id: changed-files
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const { data: files } = await github.rest.pulls.listFiles({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.payload.pull_request.number,
+            });
+
+            const configFiles = files.filter(file => file.filename.startsWith('config/'));
+
+            core.setOutput('configFiles', configFiles.map(file => file.filename).join(','));
+
+      - name: Evaluate Changes
+        run: |
+          echo "Changed config files:"
+          echo "${{ steps.changed-files.outputs.configFiles }}" | tr ',' '\n'
+          
+          if [ "${{ steps.changed-files.outputs.configFiles }}" != "" ]; then
+            echo "⚠️ Config directory changes detected! Please review."
+            echo "config_changed=true" >> $GITHUB_ENV
+          else
+            echo "✅ No changes in config directory."
+            echo "config_changed=false" >> $GITHUB_ENV
+          fi
+
+      - name: Add PR Comment if Config Changes Detected
+        if: env.config_changed == 'true'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.payload.pull_request.number,
+              body: "⚠️ **Config directory changes detected!** Please review the changes before merging."
+            });
+
+      - name: Add Label to PR if Config Changes Detected
+        if: env.config_changed == 'true'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            github.rest.issues.addLabels({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.payload.pull_request.number,
+              labels: ["config-change"]
+            });


c-pius (changes requested):
Overall looks good, just some general suggestions:

- I think I would split this into two workflows, one verifying that manifests are up to date, one checking if there are relevant changes
- regarding the config changes, I think it would be more robust if we work on the actual output based on the configuration
  - so instead of checking if there are changes on the config files, I would suggest that we generate the KLM manifests from PR and main and compare if there are changes
  - we already did this in an earlier implementation that can be used as a base again: https://github.com/kyma-project/lifecycle-manager/pull/1358
- the previous point will cover the configuration part, but the ticket also says that the e2e test files are relevant. Here I don't think that we have output to work on, so an approach where we just look if there is a diff in the files should be sufficient. The GH `actions` and `scripts` part should be the most relevant. Also, on top level the `scripts/tests` part should be included in the check.
- I personally prefer consistency in approaches whenever possible. Here, we have multiple cases where we share info between different steps. First, we use the `outputs` of the previous steps, later, we use env variables. I would suggest to consistently use `outputs` in this case.

.github/workflows/check-configs-changes.yml
```diff
@@ -0,0 +1,70 @@
+name: "Check Config Changes"
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
```
c-pius, line 5: Labeled and unlabeled were needed on the previous approach but not here. Let's remove them. And since opened, synchronize and reopened are the default types, I think we should omit it entirely.

.github/workflows/check-manifests-changes.yml
```diff
@@ -0,0 +1,128 @@
+name: "Check if Manifests are Up-to-Date"
+
+env:
+  PR_CACHE_KEY: pr-manifests-${{ github.run_id }}-${{ github.run_attempt }}
+  MAIN_CACHE_KEY: main-manifests-${{ github.run_id }}-${{ github.run_attempt }}
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
```
c-pius, line 9: See above, I would remove the default types

.github/workflows/check-configs-changes.yml
```diff
@@ -0,0 +1,70 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
+
+      - name: Install Dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y make
```
c-pius, line 19: Are you sure we need to install it explicitly? https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#installed-apt-packages

Also looks like we are not even using `make` in this action? Is it a leftover from testing?

.github/workflows/check-configs-changes.yml
```diff
@@ -0,0 +1,70 @@
...
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const { data: files } = await github.rest.pulls.listFiles({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.payload.pull_request.number,
+            });
+            const configFiles = files.filter(file => file.filename.startsWith('config/'));
+            core.setOutput('configFiles', configFiles.map(file => file.filename).join(','));
```
c-pius, line 32: This is still targeting the `config/` folder. These are the sources for generating the manifest output via `make dry-run-control-plane`. As we already check the output in the other action, I don't think it makes sense to check the sources here. Eventually, we only care about the output, not the sources.

So this one should be replaced with the other files that we care about, but don't have an output. These are:

- `.github/actions` (if we change an action, we should make sure it still works in the other repos where we import these)
- `.github/workflows` (if we change a workflow, we should check whether we also need to change the same in management-plane-charts)
- `./tests/scripts` (those are used by the actions, so again we should check if it still works in the other repos)
- `versions.yaml` (we should check if our tooling version changes still work in the other repos)

The workflow then also needs to be renamed. Maybe something like "check-pipeline-changes`

c-pius: The rest of this workflow looks good. Maybe update the descriptions to be more general and not only about `/config` files

.github/workflows/check-manifests-changes.yml
```diff
@@ -0,0 +1,128 @@
...
+              repo: context.repo.repo,
+              issue_number: context.payload.pull_request.number,
+              body: "❌ **Detected diff in manifests!** Run 'make manifests' and commit changes."
+            });
+            github.rest.issues.addLabels({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.payload.pull_request.number,
+              labels: ["outdated-manifests"]
+            });
```
c-pius, line 99: The check above (compare-manifests) doesn't determine if the manifests are outdated. It determines if there is a diff between main and PR which may be an intended thing. Please adjust the descriptions and variable names accordingly.

.github/workflows/check-manifests-changes.yml
```diff
@@ -0,0 +1,128 @@
...
+      - name: Add PR Comment if Manifests Are Outdated
+        if: steps.compare-manifests.outputs.outdated_manifests == 'true'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.payload.pull_request.number,
+              body: "❌ **Detected diff in manifests!** Run 'make manifests' and commit changes."
```
c-pius, line 92: This is also not entirely true anymore. See my other comment for this block. I would propose that we ignore the `make manifests` case for now and add that with a separate PR.

c-pius (changes requested):

.github/workflows/check-pipeline-changes.yml
```diff
@@ -0,0 +1,74 @@
...
+          echo "${{ steps.changed-files.outputs.pipelineFiles }}" | tr ',' '\n'
+          if [ -n "${{ steps.changed-files.outputs.pipelineFiles }}" ]; then
+            echo "⚠️ Pipeline-related changes detected!"
+            echo "pipeline_changed=true" >> $GITHUB_OUTPUT
+          else
+            echo "✅ No pipeline-related changes detected."
+            echo "pipeline_changed=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Add Warning if Pipeline Changes Detected
```
c-pius, line 52:
```suggestion
      - name: Add Warning If Pipeline Changes Detected
```

.github/workflows/check-pipeline-changes.yml
```diff
@@ -0,0 +1,74 @@
...
+        if: steps.eval-changes.outputs.pipeline_changed == 'true'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.payload.pull_request.number,
+              body: "⚠️ **Pipeline-related file changes detected!** Please review if related updates (e.g. manifest generation or workflow adjustments) are required."
```
c-pius, line 61:
```suggestion
              body: "⚠️ **Pipeline-related file changed!** Please make sure to check if changes are needed in related repositories like management-plane-charts, runtime-watchter, etc.."
```

.github/workflows/check-pipeline-changes.yml
```diff
@@ -0,0 +1,74 @@
...
+        if: steps.eval-changes.outputs.pipeline_changed == 'true'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            github.rest.issues.addLabels({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.payload.pull_request.number, 
+              labels: ["pipeline-changed"]
+            });
```
c-pius, line 74: Optional: as in the other workflow, you could also remove the label again if `pipeline_changed == false`

.github/workflows/check-manifests-changes.yml
```diff
@@ -0,0 +1,128 @@
...
+      - name: Add PR Comment if Manifest Differences Detected
+        if: steps.compare-manifests.outputs.manifests_diff_detected == 'true'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.payload.pull_request.number,
+              body: "❌ **Manifest output differences detected!** Please verify if these changes are intended or if further action is required."
```
c-pius, line 92:
```suggestion
              body: "❌ **Manifests created with 'make dry-run-control-plane' changed!** Please make sure to check if changes are needed in related repositories like management-plane-charts, runtime-watchter, etc.."
```

c-pius (commented):
Just some small things left, good work!

c-pius (approved):

PR #2233, opened 2025-02-05
fix: Nil pointer when unmanaging module with `Ignore` CustomResourcePolicy

### LINKED ISSUE #2225: Nil Pointer Reference in Manifest Reconciler if
### Description

If a module has `customResourcePolicy = Ignore` and is set with the unmanaged annotation, the Manifest Reconciler will try to remove the managed-by on the defaultCR and the synced resources.

The defaultCR does not exist and causes a null pointer exception [here](https://github.com/kyma-project/lifecycle-manager/blob/3903f6aa36cfde59a12951727482e57cdeaf7132/internal/manifest/modulecr/client.go#L34).

### Steps to reproduce

Deploy template-operator with `customResourcePolicy: Ignore` and `managed: False`

### Environment Type

Managed

### AC

- [x] When `customResourcePolicy: Ignore`
  - [x] And `managed` is set to `False`
  - [x] Then no exception is thrown
- [x] Consider to add an e2e test overall testing the `customResourcePolicy: Ignore` case
   - [ ] if too extensive for this bug fix, create a follow-up ticket

### Environment Info

Kubernetes Version: x.y.z
Docker Version: x.y.z
modulectl Version: x.y.z

### Attachments

_No response_

c-pius:
**Description**

- fixes the nil pointer exception
- refactors the label removal service
  - it should be e2e responsible for label removal from all objects
  - therefore, the check if needed for defaultCR and respective lookup moved from reconciler into the service
  - this way the removal can now be completely unit tested
- extends the unmanaging e2e test to include the case wehere CustomResourcePolicy is Ignore

**Related issue(s)**

- closes #2225

### DIFF
--- a/internal/manifest/labelsremoval/labels_removal.go
+++ b/internal/manifest/labelsremoval/labels_removal.go
@@ -2,6 +2,7 @@ package labelsremoval
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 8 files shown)


internal/manifest/labelsremoval/labels_removal.go
```diff
@@ -54,19 +61,33 @@ func HandleLabelsRemovalFromResources(ctx context.Context, manifestCR *v1beta2.M
...
+	return nil
+}
+
+func removeFromDefaultCR(ctx context.Context,
+	manifest *v1beta2.Manifest,
+	skrClient client.Client,
+) error {
+	if manifest.Spec.Resource == nil {
 		return nil
 	}
```
c-pius, line 78: this will prevent the nil pointer

internal/manifest/modulecr/client.go
```diff
@@ -30,6 +30,10 @@ func NewClient(client client.Client) *Client {
 func (c *Client) GetCR(ctx context.Context, manifest *v1beta2.Manifest) (*unstructured.Unstructured,
 	error,
 ) {
+	if manifest.Spec.Resource == nil {
+		return nil, nil
+	}
```
c-pius, line 35: Additional safety net in case somebody else starts using this

internal/manifest/modulecr/client.go
```diff
@@ -30,6 +33,10 @@ func NewClient(client client.Client) *Client {
 func (c *Client) GetCR(ctx context.Context, manifest *v1beta2.Manifest) (*unstructured.Unstructured,
 	error,
 ) {
+	if manifest.Spec.Resource == nil {
+		return nil, ErrNoResourceDefined
+	}
```
c-pius, line 38: additional safety net in case somewhere else we start using this func

lindnerby (approved):
lgtm

Issue #2221, opened 2025-01-30
[Maintenance Windows] Write metric and configure alerting when unable to get config at startup


c-pius:
### Description

Write a metric at startup of KLM indicating whether the maintenance window was initialized with or without error.

See https://github.com/kyma-project/lifecycle-manager/blob/main/config/grafana/overview.json for existing metrics and alerts.

See Prometheus rules file under `files/misc `folder: `prometheus-rules.yaml` in the **management-plane-chart** repsoitory.
### Reasons

Follow up from https://github.com/kyma-project/lifecycle-manager/issues/2165

Not being able to read the config should not prevent the startup of KLM but we should be aware as soon as possible that there may be config issue so that we can react before this becomes an issue with the module.

### Acceptance Criteria

- [x] When maintenance window is initialized with error
  - [x] Then metric is written indicating that there is an error
- [x] When maintenance window is initialized without error
  - [x] Then metric is reset indicating that there is NO error
- [x] Show the metric in the "Lifecylce Manager Overview" dashboard
- [x] Alert configured to notify us if the metric indicates an error for more than 10 mins
- [x] Extend the KLM metrics E2E test with this case
- [x] Evaluate on DEV if this is working (confirm if alerting is available on DEV)

LeelaChacha:
PR to add alerting: `management-plane-charts/pull/2704`

LeelaChacha:
Canary Verification:
- Metric works on STAGE
- Dashboard PR could not be verified because not part of this release

PR #2238, opened 2025-02-05
feat: Maintenance window metric

ruanxin:
**Description**
add maintenance window metric

**Related issue(s)**
https://github.com/kyma-project/lifecycle-manager/issues/2221


.github/actions/deploy-lifecycle-manager-e2e/action.yml
```diff
@@ -151,13 +151,6 @@ runs:
         klm_image_repo: ${{ inputs.klm_image_repo }}
     - name: Expose Metrics Endpoint
       working-directory: lifecycle-manager
-      if: ${{ matrix.e2e-test == 'kyma-metrics' ||
```
ruanxin, line 154: To Reviewer:

I expose metrics endpoint for all e2e tests, it's not bring too much value to restrict it to certain tests.

LeelaChacha (changes requested):

LeelaChacha:
WDYT about returning an error in case the policy file is not found? Currently, we return nil for error in that case. I would expect the read_success metric to be false for this case.

ruanxin:
>WDYT about returning an error in case the policy file is not found? Currently, we return nil for error in that case. I would expect the read_success metric to be false for this case.

I agree, this case should not return `nil`

LeelaChacha (approved):
LGTM

PR #2240, opened 2025-02-07
chore: Use flag for min maintwindow size

### LINKED ISSUE #2222: [Maintenance Windows] Configure and mount maintenance windows
### Description

Update the local deployment and management plane charts configurations to mount the maintenance window config. Additionally, make the MinWindowSize configurable via flag.

### Reasons

Follow-up from https://github.com/kyma-project/lifecycle-manager/issues/2165

### Acceptance Criteria

- [x] maintenance window config mounted by default in `watcher_local_test` kustomization
- [x] maintenance window config mounted in management-plane-charts
  - [x] Config maps are already on DEV and STAGE
  - [x] ~SRE replied that they are planning to roll it out to PROD as well next week (starting Feb 3rd). Let's confirm shortly before we create the PR if it is there so that we don't need to make it optional/configurable fia flag.~ => policy is there now on PROD
- [x] `MinWindowSize` made a flag for easier re-configuration
  - [x] the remaining options are hard coded / defaulted

> Note: E2E test for not mounting case is included in https://github.com/kyma-project/lifecycle-manager/issues/2221

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

LeelaChacha:
**Description**

- Introduce KLM flag `min-maintenance-window-size`
- Mounts MWs for local testing

**Related issue(s)**

Closes #2222

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -208,7 +208,6 @@ func initMaintenanceWindow(minWindowSize time.Duration, logger logr.Logger) main
 	maintenanceWindow, err := maintenancewindows.InitializeMaintenanceWindow(logger,
 		maintenanceWindowPoliciesDirectory,
 		maintenanceWindowPolicyName,
-		true,
 		minWindowSize)
 	if err != nil {
 		maintenanceWindowsMetrics.RecordConfigReadSuccess(false)
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -80,7 +80,6 @@ const (
 
 	maintenanceWindowPolicyName        = "policy"
 	maintenanceWindowPoliciesDirectory = "/etc/maintenance-policy"
-	minMaintenanceWindowSize           = 20 * time.Minute
 )
 
 var (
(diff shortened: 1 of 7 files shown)


cmd/main.go
```diff
@@ -204,15 +203,13 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
...
 
-func initMaintenanceWindow(logger logr.Logger) maintenancewindows.MaintenanceWindow {
+func initMaintenanceWindow(minWindowSize time.Duration, logger logr.Logger) maintenancewindows.MaintenanceWindow {
 	maintenanceWindowsMetrics := metrics.NewMaintenanceWindowMetrics()
 	maintenanceWindow, err := maintenancewindows.InitializeMaintenanceWindow(logger,
 		maintenanceWindowPoliciesDirectory,
 		maintenanceWindowPolicyName,
-		// align the configuration values before rollout
-		// https://github.com/kyma-project/lifecycle-manager/issues/2165
 		true,
```
c-pius, line 211: I am wondering if we should remove this. After clarifying with SRE, I think we don't have a use case where we would change this to false. So we could remove it as arg here and just put the constant into the MaintenanceWindow. WDYT?

LeelaChacha: makes sense to me

PR #2247, opened 2025-02-11
feat: Add Maintenance Window Config Read Status to Monitoring Dashboard

### LINKED ISSUE #2221: [Maintenance Windows] Write metric and configure alerting when unable to get config at startup

LeelaChacha:
**Description**

- `maintenance_window_config_read_success`  metric added to KLM Overview dashboard

**Related issue(s)**

Closes #2221


config/grafana/overview.json
```diff
@@ -1977,6 +1977,100 @@
...
+        "y": 110
+      },
+      "id": 84,
+      "panels": [],
+      "title": "Maintenance Windows",
+      "type": "row"
+    },
+    {
+      "datasource": null,
+      "description": "This indicates if the KLM was successful in reading the maintenance policy file from the mounted config map.",
```
medmes, line 1997: `"description": "Indicates whether the KLM successfully read the maintenance policy file from the mounted ConfigMap.",`

config/grafana/overview.json
```diff
@@ -1977,6 +1977,100 @@
...
+          "expr": "maintenance_window_config_read_success",
+          "format": "time_series",
+          "interval": "",
+          "legendFormat": "{{pod}}",
+          "refId": "A"
+        }
+      ],
+      "timeFrom": null,
+      "timeShift": null,
+      "title": "Maintenance Window Config Read Status",
```
medmes, line 2072: Maybe more formal? as it's a title.
  
`"title": "Maintenance Window Config Read Status",`

medmes (changes requested):
Looks good for me , just cosmetic changes!
Thanks!

config/grafana/overview.json
LeelaChacha: I took inspiration from how the other panels are named. This seemed the most straightforward way to express what the panel visualizes. Do you have any other suggestions?

medmes: Not really but maybe:
"title": "Maintenance Window Configuration Read Status",

LeelaChacha: works for me

medmes (approved):

PR #2251, opened 2025-02-13
feat: Unexepected field managers reporting

### LINKED ISSUE #2175: Detect unexpected field managers
As aligned in https://github.com/kyma-project/community/issues/953, we eventually want to revert changes to resources we manage by unexpected field managers.
Before actually reverting the changes, we agreed to first implement a detection of unexpected field managers. We want to collect data about:

- what objects/resources have unexpected field managers
- what part of the object they manage
  - if this part is exclusive to the unexpected manager

The approach for how to detect the unexpected managers can be derived from the ADR (see Option 2). To keep it simple in the beginning without the need to rename managers, we can first start with an allow list instead of the naming pattern. This allow list should include:

- the values KLM uses (values are differing between the reconcile loops)
- the values K8s uses (e.g., k3d or kube-controller-manager values for the status)

### Reason

See context of linked ADR.

### AC

- [ ] determined approach how to collect the needed data
  - e.g., where to collect them (existing control loops or separate, e.g., via Job)
  - e.g., where to write them (metrics, local file that can be extracted, ...)
- [ ] data collection is happening on an opt-in basis
  - e.g., when a certain label is set to Kyma/Manifest
- [ ] collected data includes
  - all objects managed by lifecycle-manager
  - indicator whether the object as other field managers (except the known ones like from k3d/k8s)
  - name(s) of the unexpected field manager(s)
  - fields that the unexpected field manager manages
  - indicator whether lifecycle-manager is also managing those

Tomasz-Smelcerz-SAP:
**Description**

The idea behind this PR is as follows:
In order to create a robust "unknown managers cleanup" process, we first must collect the data.
This PR allows to collect this data in the logging stream for easy analysis.
Once it starts working, we'll be able to see what kind of "unknown managers" are there in the environments. Then we can add logic to fix these "unknown" managers. During that process, the amount of logs for "unknown managers" should decrease.
After a few iterations we should achieve the state where (almost) no logs for "unknown" managers are produced.
At that time we should make the code introduced here more "production-ready", for example, replace env vars with proper flags etc.
For now it's a bit ad-hoc because it's not clear yet which parts of it will remain in the final solution.
It is also not clear if we want to use the same mechanism for other reconciliation processes. If that is the case, then the entire thing should be extracted to a dedicated package.

Added detection for "unknown" managers for the Manifest-related resources (resources defined by a Kyma module):

- There is a default list of names of "known" managers. It can be reconfigured via env var. All other names are considered "unknown"
- If an "unknown" manager is detected on some managed object, the entire list of `ManagedFieldsEntry` is collected for that object. The advantage of this is that all available data is collected for the purpose of analysis.
- All the collected data is logged at the end of Manifest Synchronization process. It may contain data for more than one resource in the SKR.
- To avoid overloading of the Cloud Logging, the collected data is compressed and then base64-encoded. It can be easily decoded using standard CLI tools (gzip + base64)
- To prevent overloading Cloud Logging, the logging frequency can be configured. By default, a log entry for any given Manifest is created only every 5 minutes, starting from the previous collection event. Any data collected within this interval is discarded. The 'frequency' value can be adjusted using the environment variable.
- By default, all Manifests are subject to the detection/collection process. If necessary, a label name can be configured using env var. If it's configured, only the Manifests that have the label configured are processed.

**Related issue(s)**

Resolves #2175

### DIFF
--- a/internal/manifest/skrresources/collector.go
+++ b/internal/manifest/skrresources/collector.go
@@ -51,9 +51,9 @@ type LogCollectorEntry struct {
 	ManagedFields   []apimetav1.ManagedFieldsEntry `json:"managedFields"`
 }
 
-// Implements ManagedFieldsCollector interface, emits the collected data to the log stream.
+// Implements skrresources.ManagedFieldsCollector interface, emits the collected data to the log stream.
 // The collector is thread-safe.
-// The collector is frequency-limited to prevent emitting entries for the same objectKey multiple times in a short period.
+// The collector is frequency-limited to prevent emitting entries for the same objectKey multiple times in a short time.
 type LogCollector struct {
 	objectKey        string
 	frequencyLimiter *ttlcache.Cache[string, bool]
@@ -178,23 +178,13 @@ func getFrequencyLimiterTTL() int {
 }
 
 func newFrequencyLimiter() *ttlcache.Cache[string, bool] {
-	cache := ttlcache.New[string, bool](ttlcache.WithTTL[string, bool](time.Duration(getFrequencyLimiterTTL()) * time.Second))
+	cache := ttlcache.New(ttlcache.WithTTL[string, bool](time.Duration(getFrequencyLimiterTTL()) * time.Second))
 	go cache.Start()
 	return cache
 }
 
 func splitBySemicolons(value string) []string {
 	return strings.Split(value, ";")
-}
-
-// Implements ManagedFieldsCollector interface, does nothing.
-type nopCollector struct{}
-
-func (c nopCollector) Collect(ctx context.Context, obj client.Object) {
-}
-
-func (c nopCollector) Emit(ctx context.Context) error {
-	return nil
 }
 
 func getManagedFieldsAnalysisLabel() string {
--- a/internal/manifest/skrresources/collector.go
+++ b/internal/manifest/skrresources/collector.go
@@ -1,0 +1,185 @@
+package skrresources
+
+import (
+	"bytes"
+	"compress/gzip"
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"os"
+	"regexp"
+	"slices"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/jellydator/ttlcache/v3"
+	apimetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/kyma-project/lifecycle-manager/api/shared"
+	"github.com/kyma-project/lifecycle-manager/internal"
+	"github.com/kyma-project/lifecycle-manager/internal/manifest/manifestclient"
+)
+
+const (
+	knownManagersDefault = string(manifestclient.DefaultFieldOwner) + ";" +
+		shared.OperatorName + ";" +
+		"k3s" // Applied in k3s environments.
+	knownManagersEnvVar = "KLM_EXPERIMENTAL_KNOWN_MANAGERS"
+	knownManagersRegexp = `^[a-zA-Z][a-zA-Z0-9.:_/-]{1,127}$`
+
+	frequencyCacheTTLDefault = 60 * 5 // 5 minutes
+	frequencyCacheTTLEnvVar  = "KLM_EXPERIMENTAL_FREQUENCY_CACHE_TTL"
+	frequencyCacheTTLRegexp  = `^[1-9][0-9]{1,3}$`
+
+	managedFieldsAnalysisLabelEnvVar = "KLM_EXPERIMENTAL_MANAGED_FIELDS_ANALYSIS_LABEL"
+)
+
+var (
+	allowedManagers = getAllowedManagers() //nolint:gochecknoglobals // list of managers is a global configuration
+	emitCache       = newEmitCache()       //nolint:gochecknoglobals // singleton cache is used to prevent emitting the same log multiple times in a short period
+)
+
+type LogCollectorEntry struct {
+	ObjectName      string                         `json:"objectName"`
+	ObjectNamespace string                         `json:"objectNamespace"`
+	ObjectGVK       string                         `json:"objectGvk"`
+	ManagedFields   []apimetav1.ManagedFieldsEntry `json:"managedFields"`
+}
+
+// Implements ManagedFieldsCollector interface, emits the colloected data to the log stream.
+type LogCollector struct {
+	key     string
+	owner   client.FieldOwner
+	entries []LogCollectorEntry
+}
+
+func NewLogCollector(key string, owner client.FieldOwner) *LogCollector {
+	return &LogCollector{
+		key:     key,
+		owner:   owner,
+		entries: []LogCollectorEntry{},
+	}
+}
+
+func (c *LogCollector) Collect(ctx context.Context, remoteObj client.Object) {
+	managedFields := remoteObj.GetManagedFields()
+	for _, mf := range managedFields {
+		if isUnknownManager(mf.Manager) {
+			newEntry := LogCollectorEntry{
+				ObjectName:      remoteObj.GetName(),
+				ObjectNamespace: remoteObj.GetNamespace(),
+				ObjectGVK:       remoteObj.GetObjectKind().GroupVersionKind().String(),
+				ManagedFields:   slices.Clone(remoteObj.GetManagedFields()),
+			}
+			c.entries = append(c.entries, newEntry)
+			return
+		}
+	}
+}
+
+func (c *LogCollector) Emit(ctx context.Context) error {
+	if len(c.entries) > 0 {
+		if emitCache.Has(c.key) {
+			logger := logf.FromContext(ctx, "owner", c.owner)
+			logger.V(internal.TraceLogLevel).Info("Unknown managers detection skipped (frequency)")
+			return nil
+		}
+		emitCache.Set(c.key, true, ttlcache.DefaultTTL)
+
+		jsonSer, err := json.MarshalIndent(c.entries, "", "  ")
+		if err != nil {
+			return fmt.Errorf("failed to serialize managed field data: %w", err)
+		}
+		logData, err := compressAndBase64(jsonSer)
+		if err != nil {
+			return err
+		}
+
+		logger := logf.FromContext(ctx, "owner", c.owner)
+		logger.V(internal.TraceLogLevel).Info("Unknown managers detected", "base64gzip", logData)
+	}
+	return nil
+}
+
+// compressAndBase64 compresses the input byte slice using gzip and encodes it to base64 so that it can be logged as a string.
+func compressAndBase64(in []byte) (string, error) {
+	var buf bytes.Buffer
+	archive := gzip.NewWriter(&buf)
+
+	_, err := archive.Write(in)
+	if err != nil {
+		return "", fmt.Errorf("failed to write to gzip archive: %w", err)
+	}
+
+	if err := archive.Close(); err != nil {
+		return "", fmt.Errorf("failed to close gzip archive: %w", err)
+	}
+
+	return base64.StdEncoding.EncodeToString(buf.Bytes()), nil
+}
+
+func isUnknownManager(manager string) bool {
+	return !slices.Contains(allowedManagers, manager)
+}
+
+// allowedManagers returns either a list configured in the KLM_RECONCILECONFIG_KNOWN_MANAGERS environment variable or the default list.
+// The values must be separated by semicolons and are case-sensitive!
+func getAllowedManagers() []string {
+	configured := os.Getenv(knownManagersEnvVar)
+	if configured == "" {
+		return splitBySemicolons(knownManagersDefault)
+	} else {
+		rxp := regexp.MustCompile(knownManagersRegexp)
+		configuredValues := splitBySemicolons(configured)
+		res := []string{}
+		for _, name := range configuredValues {
+			if rxp.MatchString(name) {
+				res = append(res, name)
+			}
+		}
+		return res
+	}
+}
+
+func getCacheTTL() int {
+	var res int = frequencyCacheTTLDefault
+
+	if configured := os.Getenv(frequencyCacheTTLEnvVar); configured != "" {
+		rxp := regexp.MustCompile(frequencyCacheTTLRegexp)
+		if rxp.MatchString(configured) {
+			if parsed, err := strconv.Atoi(configured); err == nil {
+				res = parsed
+			}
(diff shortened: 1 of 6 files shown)


Tomasz-Smelcerz-SAP:
Example log entry produced when an unknown manager is detected:
```
{"level":"LEVEL(-3)","date":"2025-02-13T09:27:38.309308619Z","caller":"skrresources/collector.go:101","msg":"Unknown managers detected","context":{"controller":"manifest","controllerGroup":"operator.kyma-project.io","controllerKind":"Manifest","Manifest":{"name":"kyma-sample-template-operator-2235966007","namespace":"kcp-system"},"namespace":"kcp-system","name":"kyma-sample-template-operator-2235966007","reconcileID":"60b99204-670c-42a4-af5a-7701b2d5ad7e","owner":"declarative.kyma-project.io/applier","base64gzip":"H4sIAAAAAAAA/+xXQW/bNhS++1cIOkepnHTIKmCHoN1yGFAYnZfD6h6eqWeXM0VyJOVOCPrfB1KSo1gkZbdAsALJJYb4ke/xex/5Pn6cJcnDLEmSJBXrv5GY91BhWiSpwUoyMJgJiQqMUBkR3CjBGKqsAg5bVOnF8UwtgQSm60YbrJ5OudvvLBik1K/284vkd8rLX96hZKKpkJse3IYrf6PISp0WyUf3uU98AFF2tRIJAwWG7vFy11SQSSVssEsqXoGUjB4SbzNxCVLB7dxbKVkzHAVJ71HpbrhLdAgwtOXrKr/6Kcuvsnm+zH8u8pvi9Zu/hriNS37ZSIdut3I/HyPu52kx2Jj9XlRooAQDRyNuDDgXxuWvR8MO0BdgxEXHarZuMoVEcEJtaUuqCQNaOS4fvl6csaD4wt1ybuKTeUfLpJuCwRpZIGOQ8nJXr1FxNKjtykRUUnCriEBO4ykSlMnE5uxNPLLi2cYssKV0U2iJxFcfhZJRAtqTiJ2FDIkRPq7TTdEfIj9NQVUkpyijA1nSiGFH5JW4gZoZd+SB8k4LoxVGxMZLm5xW3nGY2UTYEP3d2GEXwaR2xcMq5VDhKi1W/V2ySr8G8EmSXnqF9bhHtfVVfICgFWwxDmF0jxy1Xiix9mlgAP1sjLxDEwE5mATzORKzRwkVqET7F5xtt8WpocDeIYPmD3urlDEe2nCoqCiHYC82zBNvW1aESbujUPGTgwAOOlkIZVZp8TrP8/xilUoljCCCOXEs3y5iwkimxJEcaXIxRfaZZCiEkr6oZiKqI0qLWhGM6sIewopGtdOWU9bTDFVYCeVpK5PZdvn+U6N+vlTOpFMjqRU1zVvBDf4b05S9HhkTXxaK7inDLf6qCbDegoXq6Pl6akM6LTVLcc1v9XvBPwgRakWBAGpPCd4SImreO2hfBVxPVxXlbrd3CggupnQcsSCz41+HkH5nvLvWQef7pyyt0/gu63u9zN8UVzfF9c3/w/qWh5fEkeFQuKfaq7eoxTNg6nEom8keKIM1ww9Rt2drTIP5+rtG25pMI1tvcttHCjWhYPNxzkybpQKuXRJLGtOpxbaaiOMq1DrsZNqGpDum/YfnwGrgwLS6GZ+LCaYWSmwVak359oWrideYWNsrDMs75I9XgkfA1lw0cY3H3zs1P/Go1I7O8gnIf/UNrhFdr/uubm+ajq0T78b2IZRhSc2PdkmG3p7x9+NzP5m+2YbPn9OGxz1ul8I3GKfT/Mv39H73/9PM/vr0XwAAAP//PXHmUtcTAAA="}}
```

Tomasz-Smelcerz-SAP:
How to extract the payload from the log entry: Copy the payload from the log entry, then use it in the command line like this:
echo -n "H4sIAAAAAAAA/+xXQW/bNhS+[ ... ]AP//PXHmUtcTAAA=" | base64 -d | gzip -d | less

internal/manifest/skrresources/collector.go
```diff
@@ -0,0 +1,189 @@
...
+	key     string
+	owner   client.FieldOwner
+	entries []LogCollectorEntry
+}
+
+func NewLogCollector(key string, owner client.FieldOwner) *LogCollector {
+	return &LogCollector{
+		key:     key,
+		owner:   owner,
+		entries: []LogCollectorEntry{},
```
c-pius, line 64: Should we make this a map instead so that we have the latest entry per name, namespace, gvk per emit?

internal/manifest/skrresources/collector.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func (c *LogCollector) Emit(ctx context.Context) error {
+	if len(c.entries) > 0 {
+		if emitCache.Has(c.key) {
+			logger := logf.FromContext(ctx, "owner", c.owner)
+			logger.V(internal.TraceLogLevel).Info("Unknown managers detection skipped (frequency)")
+			return nil
+		}
+		emitCache.Set(c.key, true, ttlcache.DefaultTTL)
```
c-pius, line 91: If I understand this implementation correctly, the cache just holds a token to determine if we should emit or not. The actual entries that are emitted are not part of the cache. I like the implementation, I am just wondering if we should name it something different than "cache".

BTW: maybe it should also be injected from outside instead of the global var

internal/manifest/skrresources/collector.go
```diff
@@ -0,0 +1,189 @@
...
+
+	frequencyCacheTTLDefault = 60 * 5 // 5 minutes
+	frequencyCacheTTLEnvVar  = "KLM_EXPERIMENTAL_FREQUENCY_CACHE_TTL"
+	frequencyCacheTTLRegexp  = `^[0-9]{1,4}$`
+
+	managedFieldsAnalysisLabelEnvVar = "KLM_EXPERIMENTAL_MANAGED_FIELDS_ANALYSIS_LABEL"
+)
+
+var (
+	allowedManagers = getAllowedManagers() //nolint:gochecknoglobals // list of managers is a global configuration
```
c-pius, line 42: This could be made a property of the collector to avoid the global var.

internal/manifest/skrresources/collector.go
```diff
@@ -0,0 +1,189 @@
...
+		configuredValues := splitBySemicolons(configured)
+		res := []string{}
+		for _, name := range configuredValues {
+			if rxp.MatchString(name) {
+				res = append(res, name)
+			}
+		}
+		return res
+	}
+}
```
c-pius, line 146: As we expect the defaults to be extended rather soon anyway, I don't know if it makes sense to keep them separate. To keep it simpler, we could just rely on the env var entirely.

c-pius (commented):
I think I would refactor the collector slightly. We already initialize it in the context of a manifest (in `SyncResources`). The manifest could already be passed in the constructor and internally the collector decides if it is enabled or not. In any case, we pass the initialized instance to the `ConcurrentSSA`. When `ConcurrentSSA` calls it, it always has an actual instance of the collector but within the collector it is decided based on its initialization if it is enabled or not (private field within the collector).

With this, we:

- don't need the `if-based` initialization in `SyncResources`
- don't need the `managedFieldsCollector()` helper method in `ConccurrentSSA`
- don't need the `nopCollector`
  - in the tests we can pass a manifest stub leading to a not enabled collector

Or as alternative, if we want to keep the explicit distinctino between enabled and not-enabled collection, we should keep the `nopCollector`, but explicitly inject it in `SyncResources` and the tests. Then we can also get rid of the `managedFieldsCollector()` helper func.

internal/manifest/skrresources/collector.go
```diff
@@ -0,0 +1,189 @@
...
+		if rxp.MatchString(configured) {
+			if parsed, err := strconv.Atoi(configured); err == nil {
+				res = parsed
+			}
+		}
+	}
+
+	if res < 1 {
+		res = 1
+	}
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: My goal was not to prevent negative values, but to ensure it's not zero. OK, I can achieve the same with a more complicated regexp.

internal/manifest/skrresources/collector.go
Tomasz-Smelcerz-SAP: I don't mind renaming, please suggest another, better name :)

As for the "global var": This cache must be a singleton. We can avoid the variable in this package if we provide this instance from the bootstrap code in `main.go`. But I am not sure if this "cache/[better name here]" will even stay in the final code.
From my point of view it exists only until we know exactly how to properly "fix" the managed fields.
This is why in this PR I tried to have all the code contained in a separate package, minimize changes to existing code as much as possible. Of course we can move it to the bootstrap code. Let's discuss it.

BTW: Contrary to what the linter suggests, these are NOT global variables, they are package-level vars. And they are safe to be used by multiple goroutines. So most of the arguments for "why globals are bad" do not apply here.

internal/manifest/skrresources/collector.go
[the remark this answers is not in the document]
c-pius: I think the regex should not bee too complicated: `^[1-9][0-9]{0,3}$`. But also okay to leave it as is if you prefer

internal/manifest/skrresources/collector.go
c-pius: > As for the "global var": This cache must be a singleton. We can avoid the variable in this package if we provide this instance from the bootstrap code in main.go. But I am not sure if this "cache/[better name here]" will even stay in the final code.
From my point of view it exists only until we know exactly how to properly "fix" the managed fields.
This is why in this PR I tried to have all the code contained in a separate package, minimize changes to existing code as much as possible. Of course we can move it to the bootstrap code. Let's discuss it.

Valid argumentation, let's keep it as is

> I don't mind renaming, please suggest another, better name :)

Maybe something like `waitTokenPool`? `waitTokenPool.Has(c.key) => wait` would read fluent to me. But probably super subjective. We can also keep it but maybe be more explicit in the comment:

```
// Caches the key of the manifest with a TTL.
// While the key exists in the cache, the logs are not emitted for this manifest
// to prevent emitting the same log multiple times in a short period.
emitCache       = newEmitCache()       //nolint:gochecknoglobals // package singleton
```

internal/manifest/skrresources/collector.go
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: done.

Tomasz-Smelcerz-SAP:
> I think I would refactor the collector slightly. We already initialize it in the context of a manifest (in `SyncResources`). The manifest could already be passed in the constructor and internally the collector decides if it is enabled or not. [...]

The reason I did it like that is that I wanted this Collector to be generic. Note there's no dependency on Manifest type in the collector. In that way it can be used to collect managed fields from any k8s object.
Passing the Manifest "into" the collector means it's no longer generic.
I will try to look for some other solution.

internal/manifest/skrresources/collector.go
c-pius: Misunderstood the use of the logger. I had the assumption that we pile up logs for the exact same logs before emitting all of them together. We clarified that this is not the case.

internal/manifest/skrresources/collector.go
Tomasz-Smelcerz-SAP: Agreed to rename the thing to `frequencyLimiter`

internal/manifest/skrresources/collector.go
Tomasz-Smelcerz-SAP: Let's start with a simple approach of having the defaults in the code.
Once we'll have to extend the environment we can then release another version with the defaults in the code removed entirely.

internal/manifest/skrresources/sync.go
```diff
@@ -59,3 +64,15 @@ func HasDiff(oldResources []shared.Resource, newResources []shared.Resource) boo
...
+// The detection is enabled by default, but can be controlled by setting a specific label on the manifest CR.
+func managedFieldsAnalysisEnabledFor(obj *v1beta2.Manifest) bool {
+	detectionLabelName := getManagedFieldsAnalysisLabel()
+	if detectionLabelName == "" {
+		return true
+	}
+
+	_, found := obj.GetLabels()[detectionLabelName]
+	return found
+}
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Done.

internal/manifest/skrresources/ssa_test.go
```diff
@@ -54,7 +54,7 @@ func TestConcurrentSSA(t *testing.T) {
 		t.Run(
 			testCase.name, func(t *testing.T) {
 				t.Parallel()
-				ssa := skrresources.ConcurrentSSA(testCase.ssa.clnt, testCase.ssa.owner)
+				ssa := skrresources.ConcurrentSSA(testCase.ssa.clnt, testCase.ssa.owner, nil)
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Done.

internal/manifest/skrresources/manifestcollector.go
```diff
@@ -0,0 +1,60 @@
...
+// isCollectionEnabled checks if managed fields detection is enabled for the given manifest.
+// The detection is enabled by default, but can be controlled by setting a specific label on the manifest CR.
+func isManifestCollectionEnabled(obj *v1beta2.Manifest) bool {
+	if obj == nil {
+		return false
+	}
+
+	detectionLabelName := getManagedFieldsAnalysisLabel()
+	if detectionLabelName == "" {
+		return true
```
c-pius, line 55: Just wondering, should we rather turn it off by default? Otherwise we will directly start to collect the info for all manifests on the given landscape. Its probably better if we just start with collecting it selectively?

Tomasz-Smelcerz-SAP: You got a point. What I wanted was some way to easily collect the data from all manifests. But I agree, that making it default is probably too much.
How about this:
- disabled when empty string
- enabled selectively when the configured label exists
- enabled for all when the configured label exists with a specific value (e.g: configuredLabelName=ALL)

What do you think?

c-pius: This would work if it was an ENV var, but since this is a label on a single manifest it doesn't make sense to add an `ALL` entry?

I actually don't think we will want to collect it for all anyway. We want to collect it for a relatively small number (e.g., 10 clusters) so we can detect patterns (e.g., KEB always putting a label on kyma-system namespace). So it should be fine by selective only I think.

Tomasz-Smelcerz-SAP: 👍

c-pius (approved):

PR #2259, opened 2025-02-17
feat: Transform Kyma.Status before syncing to SKR

### LINKED ISSUE #2232: SKR `KymaCR.status.modules[].template` to show `kyma-system` not `kcp-system` namespace
### Description

`KymaCR.status.modules` shows metadata about the module including Manifest and ModuleTemplate. While Manifest only exists in KCP, ModuleTemplate exists in KCP and SKR. When we sync this status to the KymaCR on the SKR, we should consider showing the SKR's `kyma-system` namespace instead of KCPs `kcp-system`.

Additionally, it needs to be evaluated if the Manifest part should be omitted on the SKR as this is a KCP resource which doesn't exist.

### Reasons

Issue raised by @kyma-project/hasselhoff (seemingly supported by @kyma-project/otters as well)

Users may take this information as a reference to lookup further info about the module, but can't find it as the namespace differs.

### Acceptance Criteria

- [x] KCP `KymaCR.status.modules[].template.metadata.namespace` shows `kcp-system`
- [x] SKR `KymaCR.status.modules[].template.metadata.namespace` shows `kyma-system`
- [x] SKR `KymaCR.status.modules[].manifest` removed
- [x] KCP `KymaCR.status` stays unchanged

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

c-pius:
**Description**

- sync `Kyma.Status.Modules[]` with
  - `Template.Namespace = kyma-system`
  - `Manifest = nil` 
- opted to not provide an e2e test for this as this is nicely unit-testable (see also the follow-up)

❗ this will be followed-up by a refactoring and increase of test coverage in https://github.com/kyma-project/lifecycle-manager/pull/2257

**Related issue(s)**

- resolves #2232


internal/remote/skr_context_test.go
```diff
@@ -1,21 +1,21 @@
-package remote_test
+package remote //nolint:testpackage // testing package internals
```
nesmabadr, line 1: do we need to have these no lints from now? or just export the functions like we do with the rest til we have that discussion about the unit testing approach?

c-pius: This is exactly what I would like to discuss in the meeting. We need to align how we want to go forward with this. For now, I set the linting exception as an example here and I want to use it as a basis for the discussion next week.

nesmabadr (approved):

PR #2270, opened 2025-02-24
chore: Bump Golang to 1.24.0

c-pius:
**Description**

- Bumps Golang to 1.24.0

**Related issue(s)**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.23.6-alpine as builder
+FROM golang:1.24.0-alpine as builder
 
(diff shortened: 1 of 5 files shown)


c-pius:
This needs to wait as:

> Error: can't load config: the Go language version (go1.23) used to build golangci-lint is lower than the targeted Go version (1.24.0)
  Failed executing command with error: can't load config: the Go language version (go1.23) used to build golangci-lint is lower than the targeted Go version (1.24.0)

Golangci-lint has not updated main with 1.24.0 yet: https://github.com/golangci/golangci-lint/blob/master/go.mod#L3

PR #2273, opened 2025-02-25
test: Increase moduletemplateinfolookup testcoverage

### LINKED ISSUE #2202: Improve testability and unit test coverage of `moduletemplateinfolookup`
### Description

Improve the testability of the code in `moduletemplateinfolookup` and cover it with more unit tests. This should at least include: 

- "by_channel_strategy": 
  - Test for Lookup when no templates are found
  - Test for Lookup when more than one template is found
  - Test for Lookup when the template has no channel
  - Test for Lookup when the template is marked as mandatory
- "by_version_strategy":
  - When filterTemplatesByVersion returns an error due to failing to list module templates
  - When filterTemplatesByVersion returns an error due to more than one template candidate
  - When filterTemplatesByVersion returns an error due to no templates found
  - When filterTemplatesByVersion returns an error due to the template being marked as mandatory
- "module_release_meta_strategy":
  - Test for Lookup when GetChannelVersionForModule returns an error
  - Test for Lookup when getTemplateByVersion returns an error
- helper functions in "common"

### Reasons

Follow-up from: https://github.com/kyma-project/lifecycle-manager/pull/2200

We refactored the template lookup and already increased test coverage of the copied code. We should further strive to increase the test coverage.

### Acceptance Criteria

- [ ] tests mentioned above covered

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

nesmabadr:
**Description**

- Increase moduletemplateinfolookup testcoverage

**Related issue(s)**
Resolves #2202


pkg/templatelookup/moduletemplateinfolookup/by_version_strategy_test.go
```diff
@@ -69,6 +69,87 @@ func Test_ByVersion_Strategy_Lookup_ReturnsModuleTemplateInfo(t *testing.T) {
 	assert.Equal(t, moduleTemplate.Spec.Channel, moduleTemplateInfo.ModuleTemplate.Spec.Channel)
 }
 
+func Test_ByVersion_Strategy_Lookup_WhenMoreThanOneModuleTemplateFound(t *testing.T) {
```
c-pius, line 72: This one should also be possible easily with a client stub returning an error:
image

c-pius (commented):
Overall looks good! Just some hints for tests cases that should be easy to add and will get us close to 100%


With this one, I am actually not sure... If I understand the code correctly, this cannot happen because if those two things don't match, the filterTemplatesByChannel func won't return a template.

image

c-pius (approved):

PR #2284, opened 2025-02-27
chore: Add yamllint pipeline

lindnerby:
**Description**

- Adds pipeline with GHA for yaml linting & config
- Add make target using same config
- Fix issues

**Related issue(s)**


.github/workflows/lint-yaml.yml
```diff
@@ -0,0 +1,20 @@
+name: Lint yaml
+on:
+  pull_request:
+    branches:
+      - main
+      - feat/**
+  workflow_dispatch:
```
c-pius, line 7:
```suggestion
  workflow_dispatch: { }
```

Maybe to indicate that it is intentionally empty?

Makefile
```diff
@@ -176,3 +176,7 @@ lint: ## Run golangci-lint against code.
 	$(LOCALBIN)/golangci-lint run --verbose -c .golangci.yaml
 	pushd api; $(LOCALBIN)/golangci-lint run --verbose -c ../.golangci.yaml; popd
 	pushd maintenancewindows; $(LOCALBIN)/golangci-lint run --verbose -c ../.golangci.yaml; popd
+
+.PHONY: lint-yaml
+lint-yaml: ## Run yamllint against repository. Assumes yamllint is installed. Install via 'brew install yamllint' or 'pip install yamllint'.
+	yamllint -c .yamllint.yml --no-warnings  .
```
c-pius, line 182: Shall we add it to the regular `lint` as a sub-target?

lindnerby: Good idea 👌

.github/workflows/lint-yaml.yml
lindnerby: I don't have strong feelings here. We can either make it mandatory to indicate explicitly that it's empty and the type.
Let's keep that for an eventual follow-up when team members what this defined. The config option is there now, we can extend it as needed.

c-pius (approved):

PR #2323, opened 2025-03-13
feat: Remove single cluster mode

ruanxin:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/api/v1beta2/condition_messages.go
+++ b/api/v1beta2/condition_messages.go
@@ -64,11 +64,10 @@ func GenerateMessage(conditionType KymaConditionType, status apimetav1.Condition
 }
 
 // GetRequiredConditionTypes returns all required ConditionTypes for a KymaCR.
-func GetRequiredConditionTypes(syncEnabled, watcherEnabled bool) []KymaConditionType {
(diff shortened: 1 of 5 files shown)


medmes:
What about this Int test?  After removing the kcp mode. Is the test title confusing?
tests/integration/controller/kcp/controller_test.go:9

amritanshusikdar:
> What about this Int test? After removing the kcp mode. Is the test title confusing? [tests/integration/controller/kcp/controller_test.go:9](https://github.com/kyma-project/lifecycle-manager/blob/429f5254fd07b035ddc166a58e4b56abdba93895/tests/integration/controller/kcp/controller_test.go#L11-L10)

This just checks if the manager field is present. Since the test is inside `/kcp` directory, its fine.

medmes (commented):
Overall, Looks good for me , there were some parts that needed to be refactored but it will be out of the scope of the ticket and with the new code architecture. I'd better think about it later!

medmes (approved):

amritanshusikdar:
management-plane-charts PR: `management-plane-charts/pull/3032`
(as per the last AC)

PR #2356, opened 2025-03-25
chore: Increase test coverage for generating kyma.status.modules

lindnerby:
**Description**

- Extract to service
- Add tests

**Related issue(s)**


internal/controller/kyma/controller.go
```diff
@@ -59,19 +60,24 @@ const (
 	patchStatusError  event.Reason = "PatchStatus"
 )
 
+type ModuleStatusService interface {
```
ruanxin, line 63: I suggest before we have a concrete plan to introduce Service layer, let's be cautious while naming some components as service, the scope of this `ModuleStatus` seems too limited to justify it as a service, I would rather name it as `ModuleStatusHandler` or something small scoped.

lindnerby: Sure, renamed it to handler 👍

ruanxin:
To reviewer: pls check my updated commit https://github.com/kyma-project/lifecycle-manager/pull/2356/commits/40c471ff8970cd83e2fbaffa480743e83ec9a30c

internal/service/kyma/status/modules/handler.go
```diff
@@ -0,0 +1,129 @@
...
+) error {
+	// This nil pointer check is for defensive programming and should never occur in a production environment.
+	if kyma == nil {
+		return errNilKyma
+	}
+
+	moduleStatusMap := kyma.GetModuleStatusMap()
+	for _, module := range modules {
+		moduleStatus, exists := moduleStatusMap[module.ModuleName]
+		// Even defensive error occurs, we are not blocking module status to be updated.
```
c-pius, line 56:
```suggestion
		// Even if defensive error occurs, we are not blocking module status to be updated.
```

c-pius (approved):

Issue #2235, opened 2025-02-05
Propagate BTP-related labels to the SKR Kyma CR


janmedrek:
### Description

KLM should propagate the following labels to the Kyma CR in the SKR:
- `kyma-project.io/global-account-id`
- `kyma-project.io/subaccount-id`

### Reasons

We want to provide Busola with information to display for their Users.

### Acceptance Criteria

- [x] `kyma-project.io/global-account-id` label synced to the SKR
- [x] `kyma-project.io/subaccount-id` label synced to the SKR

### Testing approach

* E2E

janmedrek:
@a-thaler, sorry for the miscommunication, the comment was not ignored, what we missed was the feedback regarding the next steps we take.

We have a separate ticket where we group all the topics for the next API version upgrade: https://github.com/kyma-project/lifecycle-manager/issues/776

We will discuss putting those fields directly into the API, if we have an agreement, it would be beneficial to keep it as a part of the API, we will include it in a separate ticket.

c-pius:
Labels are present on STAGE: 

![Image](https://github.com/user-attachments/assets/6e04c1d7-fa25-48b2-8f2d-701c4fc22565)

Kiemes:
Thank you for the labels. This helps already in automation processes.

How about dashboard users?
Do you plan to expose those labels also on the Kyma dashboard? This would help us a lot to keep track on which browser tab reflects which Kyma landscape. We are dealing with about 20 different landscapes. It's a bit dangerous to do modifications on the wrong landscape if that is not visible all the time on the top of the dashboard, regardless on which menu the user is.
For that I think the GA and SA **names** would be really helpful.
Do you want me to create a separate issue for this?

c-pius:
@mrCherry97 could you chime in on Tom's request from Dashboard perspective?

PR #2369, opened 2025-03-26
feat: Propagate Global Account and SubAccount Labels

amritanshusikdar:
**Description**

- Propagate the label `kyma-project.io/global-account-id` from KCP to SKR
- Propagate the label `kyma-project.io/subaccount-id` from KCP to SKR

**Related issue(s)**
Resolves: #2235

### DIFF
--- a/internal/remote/skr_context.go
+++ b/internal/remote/skr_context.go
@@ -234,12 +234,14 @@ func (s *SkrContext) getRemoteKyma(ctx context.Context) (*v1beta2.Kyma, error) {
 	return skrKyma, nil
 }
 
+// TODO: change here
 // syncWatcherLabelsAnnotations adds required labels and annotations to the skrKyma.
 // It returns true if any of the labels or annotations were changed.
 func syncWatcherLabelsAnnotations(kcpKyma, skrKyma *v1beta2.Kyma) bool {
 	labels, labelsChanged := collections.MergeMaps(skrKyma.Labels, map[string]string{
 		shared.WatchedByLabel: shared.WatchedByLabelValue,
 		shared.ManagedBy:      shared.ManagedByLabelValue,
+		"customLabel":         "customValue",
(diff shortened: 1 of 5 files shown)


internal/remote/skr_context.go
```diff
@@ -237,10 +237,22 @@ func (s *SkrContext) getRemoteKyma(ctx context.Context) (*v1beta2.Kyma, error) {
 // syncWatcherLabelsAnnotations adds required labels and annotations to the skrKyma.
 // It returns true if any of the labels or annotations were changed.
 func syncWatcherLabelsAnnotations(kcpKyma, skrKyma *v1beta2.Kyma) bool {
-	labels, labelsChanged := collections.MergeMaps(skrKyma.Labels, map[string]string{
-		shared.WatchedByLabel: shared.WatchedByLabelValue,
-		shared.ManagedBy:      shared.ManagedByLabelValue,
-	})
+	labelsMap := map[string]string{}
```
ruanxin, line 240: this function is used to sync watcher related label as the name suggests, let's not misuse it.

amritanshusikdar: Makes sense, extracted the logic into an independent function

internal/remote/skr_context.go
```diff
@@ -170,8 +170,11 @@ func (s *SkrContext) SynchronizeKymaMetadata(ctx context.Context, kcpKyma, skrKy
 		return nil
 	}
```
ruanxin, line 172: then you need to consider increasing the test coverage of this `SynchronizeKymaMetadata` since you introduced new logic.

amritanshusikdar: Added two things:
- unit tests for the newly added `syncBTPRelatedLabels()` function
- extended unit tests for `SynchronizeKymaMetadata()` function

PR #2370, opened 2025-03-26
feat: Add `--cert-management` Flag

### LINKED ISSUE #2357: [Gardener CM] Create flag to set CM mode
### Description

TBD

### Reasons

_No response_

### Acceptance Criteria

- [ ] Acceptance criterion 1
- [ ] Acceptance criterion 2

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

amritanshusikdar:
**Description**

- added `--cert-management` flag
- added unit testing for this new flag

**Related issue(s)**
Resolves #2357

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -158,7 +158,7 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
 	var skrWebhookManager *watcher.SKRWebhookManifestManager
 	var options ctrlruntime.Options
 	if flagVar.EnableKcpWatcher {
-		if skrWebhookManager, err = createSkrWebhookManager(mgr, skrContextProvider, flagVar, logger); err != nil {
+		if skrWebhookManager, err = createSkrWebhookManager(mgr, skrContextProvider, flagVar); err != nil {
 			logger.Error(err, "failed to create skr webhook manager")
 			os.Exit(bootstrapFailedExitCode)
 		}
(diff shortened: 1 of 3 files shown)


cmd/main.go
```diff
@@ -358,6 +358,9 @@ func createSkrWebhookManager(mgr ctrl.Manager, skrContextFactory remote.SkrConte
 		SkrWebhookMemoryLimits: flagVar.WatcherResourceLimitsMemory,
 		RemoteSyncNamespace:    flagVar.RemoteSyncNamespace,
 	}
+	if flagVar.CertificateManagement == "gardener" {
+		setupLog.Info("[SETUP] Request to use Gardener for certificate management is not supported yet")
```
ruanxin, line 362: This log doesn't make too much sense, we also have to change the context after the implementation done. As the AC suggest, just log gardener is enabled is sufficient, or just skip log entirely. In general, we should not relays on logs to observe which CertificateManagement to use, maybe we can consider creating metrics for this in the future.

amritanshusikdar: Agreed, it doesn't make much sense to just randomly log something when its going to be removed later on anyway. **So for now replacing the logging message with a simple TODO message.**
Another reason to do this is: in order to get access to the `logger`, the function declaration has to be modified, which is too much for a temporary placeholder.

cmd/main.go
```diff
@@ -358,6 +358,8 @@ func createSkrWebhookManager(mgr ctrl.Manager, skrContextFactory remote.SkrConte
 		SkrWebhookMemoryLimits: flagVar.WatcherResourceLimitsMemory,
 		RemoteSyncNamespace:    flagVar.RemoteSyncNamespace,
 	}
+	// To be implemented: Gardener Certificate Management
```
ruanxin, line 361: This comment is unclear—why should the implementation go here but not elsewhere?

if you have some condition like
```
if flagVar.CertificateManagement == 'gardener' {
 // TODO: xxx
}
```
then adding a comment here to indicate that the implementation logic belongs in this block is useful. However, having only a vague comment without context is unhelpful and cause confusion.

amritanshusikdar: I did add exactly as suggested in your example above. But the linter complained about having TODOs, FIXMEs, etc. I'll just remove the comment totally. Anyway during implementation the entire thing will be dealt with from start to finish.

ruanxin: then you can add some linter ignore comments to bypass linter complain. since it's a temporary solution, we can accept it.

ruanxin (approved):

PR #2371, opened 2025-03-26
chore: Update watcher certificate manifest

### LINKED ISSUE #2347: [Gardener CM] Make `klm-watcher-selfsigned` `ClusterIssuer` a `Issuer`
### Description

We should change the cert-manager ClusterIssuer to Issuer. While this is not strictly necessary, it aligns with the future Gardener CM setup that we run in parallel.

### Reasons

Follow-up from https://github.com/kyma-project/lifecycle-manager/issues/2312. Gardener CM doesn't support ClusterIssuers.

### AC

- [x] `klm-watcher-selfsigned` `ClusterIssuer` changed to `Issuer`
- [ ] `klm-watcher-selfsigned` issuer in `kcp-system` namespace
- [x] Roles updated accordingly
- [x] Cache options updated accordingly if necessary
- [x] Verified that watch mechanism still works

amritanshusikdar:
**Related issue(s)**
Resolves #2347


amritanshusikdar:
The second AC requires the `klm-watcher-selfsigned` to be in the `kcp-system` namespace. This is causing a so-called "snowball" / "butterfly" effect, i.e., a lot of things require changing since the setup in `main.go` also has a lot of hardcoded dependencies to `istio-system` namespace. Such a huge change is outside the scope of this issue, hence the 2nd AC is skipped.

amritanshusikdar:
Another thing to keep in mind is: the `management-plane-charts` need to be updated as well, as the templating there still relies on ClusterIssuer.

config/watcher/certificate.yaml
```diff
@@ -2,9 +2,10 @@
 # More document can be found at https://docs.cert-manager.io
 # WARNING: Targets CertManager v1.0. Check https://cert-manager.io/docs/installation/upgrading/ for breaking changes.
 apiVersion: cert-manager.io/v1
-kind: ClusterIssuer
+kind: Issuer
 metadata:
-  name: watcher-selfsigned
+  name: watcher-root
```
nesmabadr, line 7: Why is the certificate renamed?

amritanshusikdar: This is the `Issuer`. It has to be given a different name, otherwise it'll clash with the `Issuer` (at the bottom of the file) already with the exact same name in the exact same namespace.

The `Certificate` however, has no change in name. Just the pointer in the `.spec` to the root issuer has been changed.

nesmabadr: Yes I mean what is the use of the one at the end of the file?

nesmabadr (approved):

Issue #2358, opened 2025-03-25
[Gardener CM] Create Gardener CertificateManager implementation


lindnerby:
### Description

In order to support Gardener cert-management, we need to create a specific CertificateManager that can handle Gardener custom resources (Issuer, Certificate. etc.) to replace the currently used [CertificateManager](https://github.com/kyma-project/lifecycle-manager/blob/8a50da348ce99dead00aa6372074299d7bbc7479/pkg/watcher/certificate_manager.go) which works on `github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1` types.

If the flag `--cert-managment` is set to `gardener`, the current usage of CertificateManager is replaced by a specialized GardenerCertificateManager, that works on `github.com/gardener/cert-management/pkg/apis/cert/v1alpha1` types, in the setup of main composition root.

See [this PoC](https://github.com/kyma-project/lifecycle-manager/pull/2325/files#diff-3a2a38e72239c4d8fa7b8f569c4f54ee1e488a84af7468fec01e9b6efbc93248R9) for more implementation details. Since the resources of course differ, we need to adapt `SKRWebhookManager` and choose the correct CertificateManager.

### Acceptance Criteria

- [x] Introduce GardenerCertificateManager
- [x] Set flag changes the used implementation during setup
- [x] No manifest or chart changes, these will be introduced when using the impl via E2E test and kustomize target introduction: #2360

### Feature Testing

Unit tests

c-pius:
> If the flag --cert-managment is set to gardener, the current usage of CertificateManager is replaced by a specialized GardenerCertificateManager, that works on github.com/gardener/cert-management/pkg/apis/cert/v1alpha1 types, in the setup of main composition root.

Opted for keeping one CertificateManager as there is some shared logic but injecting two different implementations of CertificateClient which are either for Certmanager or Gardener.

PR #2382, opened 2025-04-03
chore: Refactor CertificateManager

c-pius:
**Description**

- refactoring that prepares the introduction of Gardener CM
  - doesn't change our current certificate setup/handling
- introduces `Certificate` (certificate.go) as the actual client implementation for the certificate
  - as of now for the existing cert-manager implementation
  - separate PR will introduce the same for Gardener CM
- `CertificateManager` (certificate_manager.go) is the implementation of the cert handling
  - works with interface abstraction to `Certificate` which allows it to be replaced by Gardener CM in the next PR
- Cert-related logic in `SkrWebhookManifestManager` moved to `CertificateManager` 

**Related issue(s)**

- closes https://github.com/kyma-project/lifecycle-manager/issues/2358


internal/pkg/flags/flags.go
```diff
@@ -209,8 +211,6 @@ func DefineFlagVar() *FlagVar {
 		" from finalizer removal. Example: 'ingressroutetcps.traefik.containo.us,*.helm.cattle.io'.")
 	flag.StringVar(&flagVar.RemoteSyncNamespace, "sync-namespace", DefaultRemoteSyncNamespace,
 		"Name of the namespace for syncing remote Kyma and module catalog")
-	flag.StringVar(&flagVar.CaCertName, "ca-cert-name", DefaultCaCertName,
-		"Name of the CA Certificate in Istio Namespace which is used to sign SKR Certificates")
```
c-pius, line 213: This flag was already unused

tests/integration/watcher/certificate_test.go
```diff
@@ -33,34 +36,20 @@ var _ = Describe("Create Watcher Certificates", Ordered, func() {
...
-				ObjectMeta: apimetav1.ObjectMeta{
-					Name:        "test-kyma-2",
-					Namespace:   "testcase-2",
-					Annotations: map[string]string{watcher.DomainAnnotation: "example.domain.com"},
-				},
-			},
-			issuer:         nil,
-			wantCreateErr:  true,
-			wantNewCertErr: false,
-		},
```
c-pius, line 56: This test is not needed anymore because we don't lookup the Issuer by label but configure its name via flag.

cmd/main.go
```diff
@@ -70,6 +70,9 @@ import (
 	"github.com/kyma-project/lifecycle-manager/pkg/templatelookup"
 	"github.com/kyma-project/lifecycle-manager/pkg/templatelookup/moduletemplateinfolookup"
 	"github.com/kyma-project/lifecycle-manager/pkg/watcher"
+	"github.com/kyma-project/lifecycle-manager/pkg/watcher/certificate"
```
ruanxin, line 73: We better not expose watcher internal package into main.go, since we decided to go with layered structure, in general, the internal logic of `watcher` should remain encapsulated and not be exposed at the controller level.

cmd/main.go
```diff
@@ -385,12 +380,39 @@ func createSkrWebhookManager(mgr ctrl.Manager, skrContextFactory remote.SkrConte
...
+		SkrServiceName:               watcher.SkrResourceName,
+		SkrNamespace:                 flagVar.RemoteSyncNamespace,
+		CertificateNamespace:         flagVar.IstioNamespace,
+		AdditionalDNSNames:           strings.Split(flagVar.AdditionalDNSNames, ","),
+		GatewaySecretName:            shared.GatewaySecretName,
+		RenewBuffer:                  flagVar.SelfSignedCertRenewBuffer,
+		SkrCertificateNamingTemplate: flagVar.SelfSignedCertificateNamingTemplate,
+	}
+
+	certificateManager := certificate.NewCertificateManager(
```
ruanxin, line 400: based on https://github.com/kyma-project/lifecycle-manager/pull/2382/files#r2030696256, the initiation of `CertificateManager` could be inside `SKRWebhookManifestManager`, since it's an internal component.

internal/controller/kyma/controller.go
```diff
@@ -60,6 +60,12 @@ const (
 	patchStatusError  event.Reason = "PatchStatus"
 )
 
+type SKRWebhookManager interface {
+	Reconcile(ctx context.Context, kyma *v1beta2.Kyma) error
+	Remove(ctx context.Context, kyma *v1beta2.Kyma) error
+	RemoveSkrCertificate(ctx context.Context, kymaName string) error
+}
```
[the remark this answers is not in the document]
ruanxin: This interface should be designed inside `watcher` package instead of in the controller.

internal/controller/kyma/controller.go
```diff
@@ -45,7 +45,7 @@ import (
 	"github.com/kyma-project/lifecycle-manager/pkg/status"
 	"github.com/kyma-project/lifecycle-manager/pkg/templatelookup"
 	"github.com/kyma-project/lifecycle-manager/pkg/util"
-	"github.com/kyma-project/lifecycle-manager/pkg/watcher"
+	"github.com/kyma-project/lifecycle-manager/pkg/watcher/certificate"
```
ruanxin, line 48: Although it's a valid usage to depend on sub package `watcher/certificate`, since it's in controller layer, I want to have a clear dependency here, which means the controller only depends on business layer (Service layer), and Watcher should be the public interface to controller level, the certificate remain encapsulated within watcher packge.

internal/controller/kyma/controller.go
```diff
@@ -354,9 +360,9 @@ func (r *Reconciler) handleProcessingState(ctx context.Context, kyma *v1beta2.Ky
 
 	if r.WatcherEnabled() {
 		errGroup.Go(func() error {
-			if err := r.SKRWebhookManager.Install(ctx, kyma); err != nil {
+			if err := r.SKRWebhookManager.Reconcile(ctx, kyma); err != nil {
 				r.Metrics.RecordRequeueReason(metrics.SkrWebhookResourcesInstallation, queue.UnexpectedRequeue)
-				if errors.Is(err, &watcher.CertificateNotReadyError{}) {
+				if errors.Is(err, certificate.ErrSkrCertificateNotReady) {
```
ruanxin, line 365: so this error need to be designed within `watcher` package, but not in `certificate`.

pkg/testutils/kyma.go
```diff
@@ -50,7 +49,7 @@ func NewKymaWithNamespaceName(name, namespace, channel string) *v1beta2.Kyma {
 	return builder.NewKymaBuilder().
 		WithNamePrefix(name).
 		WithNamespace(namespace).
-		WithAnnotation(watcher.DomainAnnotation, "example.domain.com").
+		WithAnnotation(shared.SKRDomainAnnotation, "example.domain.com").
```
ruanxin, line 52: maybe also rename to `SkrDomainAnnotation` to have consistency.

pkg/watcher/certificate/certificate_manager.go
```diff
@@ -0,0 +1,236 @@
...
+)
+
+var (
+	ErrDomainAnnotationEmpty   = errors.New("domain annotation is empty")
+	ErrDomainAnnotationMissing = errors.New("domain annotation is missing")
+	ErrSkrCertificateNotReady  = errors.New("SKR certificate not ready")
+)
+
+//nolint:gochecknoglobals // this is const config
+var serviceSuffixes = []string{
```
ruanxin, line 25: do we really need this as package level value?

cmd/main.go
c-pius: In my understanding, the main.go serves as our composition root (see articles sent by Ben, [#1](https://blog.ploeh.dk/2011/07/28/CompositionRoot/), [#2](https://medium.com/@cfryerdev/dependency-injection-composition-root-418a1bb19130)).

> the internal logic of watcher should remain encapsulated and not be exposed at the controller level

I would argue here that the logic is still encapsulated in the package we depend on. We just resolve the dependency graph here which is completely fine in the composition root (main.go). The "controller level" are the actual controller implementations. They should completely depend on abstractions. The only point where we map the abstractions to the concrete implementations is the composition root.

> based on https://github.com/kyma-project/lifecycle-manager/pull/2382/files#r2030696256, the initiation of CertificateManager could be inside SKRWebhookManifestManager, since it's an internal component.

If we would do that, then we would be violating the [dependency inversion principle](https://medium.com/@inzuael/solid-dependency-inversion-principle-part-5-f5bec43ab22e). The SKRWebhookManifestManager only depends on an interface abstraction of CertificateManager. If we would instantiate the actual implementation within there, then SKRWebhookManifestManager (high-level module), depends on CertificateManager (low-level module).

cmd/main.go
c-pius: See above comment on the import.

internal/controller/kyma/controller.go
c-pius: This is an interesting case. Let's first align on the other dependency-related comments and then get back to this.

pkg/testutils/kyma.go
c-pius: https://github.com/kyma-project/lifecycle-manager/pull/2382/commits/541044693eff4e4e48d95750d9eff72b8bf008e4

internal/controller/kyma/controller.go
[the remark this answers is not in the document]
ruanxin: I really think some pattern we should introduce carefully. I get the point of defining interface in the consumer side, but in complex project, especially have layered structure,  how to practice this pattern is open to debate.

First, just because an interface is defined by the consumer doesn’t necessarily mean it should reside in the package where it’s used. In this case, the SKRWebhookManager is currently only used by the Kyma controller. But imagine that, in the future, another controller or service in the same layer starts depending on it. How would we handle that? Following the pattern as suggested would require us to define the interface in each place it’s used, which would be impractical and lead to duplicated interfaces across the codebase. Eventually, we'd end up needing a shared package to hold those interfaces for reuse.

Then down to the question, if we use `watcher` package as the place to hold interface, or we follow the pattern to have a separate interface package to hold all interface, but put it inside controller is not a good idea.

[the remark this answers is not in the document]
c-pius: I would still be pro consumder-defined interfaces. In order to not block this PR going forward, how about we take this as an explicit point and example for discussion in the arch round. Let's take a decision there how we want to handle it within our projects. If the outcome is to change it we can do that in a follow-up PR. Would that be okay for you?

pkg/watcher/certificate/certificate_manager.go
ruanxin: In general, I think we should avoid using global variables as much as possible, also, a string slice is not a const. I would not bother to keep this variable inside that function, if you think it's too hidden or unclear, we could also define a function to provide those values, something like:

```
func getServiceSuffix() []string {
	return []string{"svc.cluster.local", "svc"}
}
```

cmd/main.go
c-pius: Let me try to address the individual points.

> I think our main difference is how to treat CertificateManager. From my perspective, also reflected by the package construction watcher/certificate, I would argue the CertificateManager is a submodule that belongs to SkrWebhookManifestManager, from this point of view, coupling CertificateManager with SkrWebhookManifestManager is not a bad idea.

From my perspective this is a logical grouping. `certificate/Certmanager` logically belongs to the area of `watcher`. That's why I think it makes sense to put it there. However, this doesn't mean that `watcher/SkrWebhookManifestManger` should have a direct dependency to it's sub-moudles. It should only depend on the interface it defines.

> It lowers the complexity of testing controllers which depends on SkrWebhookManifestManager, since we only mock its behavior you defined here https://github.com/kyma-project/lifecycle-manager/pull/2382/files#diff-4d414901850a616d4e06a120615896904639aef683f0a4d8860abf65c76179d4R63-R67 is enough, and don't care about CertificateManager detail. This means in the testing, you don't need to provide a mocked CertificateManager while testing kyma controller, instead, CertificateManager is only tested against SkrWebhookManifestManager.

This is already the case. `KymaController` is not aware of `CertificateManager` and if we had tests for it we would only mock the `SkrWebhookManifestManager`. No need to involve `CertificateManager`. The only reason why `KymaController` currently has a dependency to `pkg/watcher/certificate` is the `ErrSkrCertificateNotReady` error. This is a flaw and I agree that this needs to be removed but I would just like to park that fix until we have reached an overall agreement (see [comment](https://github.com/kyma-project/lifecycle-manager/pull/2382#discussion_r2031093395)). 

> Back to the composition root pattern, certainly, if you compose CertificateManager in the main, it brings more flexibility while you design unit testing, you may mock different CertificateManager behavior to test how it affects kyma controller. However, this added flexibility comes at the cost of increased complexity in your tests, which may not be necessary depending on your specific use cases.

Not sure if I understand this point properly. Could you please elaborate? Assuming it is the same point as above: The testing of `KymaController` does not need knowledge about `CertificateManager`. `KymaController` only depends on the `SKRWebhookManager` interface where we will inject a mock. No need for mocking of transitive dependencies.

> Another concern is regarding the layered structure, my ultimate goal as I mentioned in previous session, the controller layer should only depends on service, and SkrWebhookManifestManager can be a good candidate as a Service, which composed on the root, however, if you also want to compose CertificateManager, we need to first agree it also treat as service in the future, and will packaged under service or service/certificate.

I am all pro the layered architecture and I don't see how this contradicts this goal. `CertificateManager` is a service. It's just that I think it is totally valid to logically group services together (see first point of this comment). If you prefer, we can of course move `pkg/watcher/certificate` up to `pkg/certificate`. Still, I think there should not be a direct dependency.

> But it doesn't mean you must inject in the NewSKRWebhookManifestManager, you can also create a new constructer only for testing, passing mocked certificateManager as an argument or even just directly init a SkrWebhookManifestManager struct while writing test.

I honestly don't really get the benefit out of it. The only thing I can think of is that we have a bit less code in the composition root. But this is the place where such things should actually happen. And that comes at the cost of adding the dependency to the downstream service. In my opinion this is worse as this makes it prone to cyclic dependencies.

And just for clarification on, maybe we also have a misunderstanding here. The composition root in `main.go` for me is not the controller-level. It is the initialization level. The controller-level will be kept clean of lower-level dependencies.

ruanxin (dismissed):

cmd/main.go
c-pius: We have decided to move the discussion and decision to https://github.com/kyma-project/lifecycle-manager/pull/2393 and https://github.com/kyma-project/lifecycle-manager/pull/2395. We may refactor this again depending on the decision taken there.

internal/controller/kyma/controller.go
c-pius: Moved the error to `watcher`: fa24b93 See also comment below

internal/controller/kyma/controller.go
c-pius: Follow-up: https://github.com/kyma-project/lifecycle-manager/issues/2396

ruanxin (approved):

Issue #2360, opened 2025-03-25
[Gardener CM] Create E2E tests and kustomize config for GCM


lindnerby:
### Description

To finalise and test Gardener cert-manager support, we need E2E tests and a new kustomize component to setup the resources in the test cluster.

### Reasons

We should port watcher-related e2e tests to GCM to ensure everything is still working after adding GCM support.

### Acceptance Criteria

- [x] new kustomize component should be created for Gardener CM
  - [x] consider moving required RBACs to the kustomization target config folder itself (also for classic cert-manager setup)
  - [x] for required resources refer to [PoC PR](https://github.com/kyma-project/lifecycle-manager/pull/2325)
- [x] `create_test_clusters.sh` script is adapted or new script is created to setup test clusters with Gardener CM
- [x] new makefile target for gcm cluster setup (`local-watcher-test-gcm`)
- [x] E2E tests are duplicated for GCM:
  - [x] `watcher-enqueue` to `gcm-watcher-enqueue`
  - [x] `legacy-istio-gateway-secret-rotation` to `gcm-legacy-istio-gateway-secret-rotation`
  - [x] `self-signed-certificate-rotation` to `gcm-self-signed-certificate-rotation`
  - [x] `modulereleasemeta-watch-trigger` to `gcm-modulereleasemeta-watch-trigger`

PR #2395, opened 2025-04-08
docs: ADR 003 - Dependencies

c-pius:
**Description**

Changes proposed in this pull request:

:wip:

**Related issue(s)**

PR #2397, opened 2025-04-09
feat: Support Gardener certificate-management

c-pius:
**Description**

- Introduces the implementation for Gardener certificate management
- kustomization and e2e tests are not yet adapted (will be done in https://github.com/kyma-project/lifecycle-manager/issues/2360)

**Related issue(s)**


.golangci.yaml
```diff
@@ -55,6 +55,10 @@ linters-settings:
     alias:
       - pkg: github.com/kyma-project/template-operator/api/v1alpha1
         alias: templatev1alpha1
+      - pkg: github.com/gardener/cert-management/pkg/apis/cert/v1alpha1
```
c-pius, line 58: I would like to align these in a separte follow-up pull request to:

- `gcertv1alpha1`, `gcertclientv1alpha1`
- `cmcertv1`, `cmcertclientv1`

cmd/main.go
```diff
@@ -122,8 +119,12 @@ func main() {
 		go pprofStartServer(flagVar.PprofAddr, flagVar.PprofServerTimeout, setupLog)
 	}
 
-	cacheOptions := internal.GetCacheOptions(flagVar.IsKymaManaged, flagVar.IstioNamespace,
-		flagVar.IstioGatewayNamespace)
+	cacheOptions := setup.SetupCacheOptions(flagVar.IsKymaManaged,
```
c-pius, line 122: Proposal to not have the full setup in main (see also below `skrWebhookManager = setup.SetupSkrWebhookManager(mgr, skrContextProvider, flagVar, logger)`) 

However, a concrete decision how we will go forward with this will be taken in https://github.com/kyma-project/lifecycle-manager/pull/2395

internal/pkg/flags/flags.go
```diff
@@ -85,8 +89,11 @@ var (
 //nolint:funlen // defines all program flags
 func DefineFlagVar() *FlagVar {
 	flagVar := new(FlagVar)
-	flag.StringVar(&flagVar.CertificateManagement, "cert-management", "", "Configures which certificate management"+
-		" system to use. Default is CNCF open source cert-manager. Accepted value: `gardener`")
+	flag.StringVar(&flagVar.CertificateManagement, "cert-management", certmanagerv1.SchemeGroupVersion.String(),
+		fmt.Sprintf("Configures which certificate management system to use. Accepted values: '%s', '%s'. Default: '%s'",
+			certmanagerv1.SchemeGroupVersion.String(),
+			gcertv1alpha1.SchemeGroupVersion.String(),
+			certmanagerv1.SchemeGroupVersion.String()))
```
c-pius, line 96: Proposing to use the SchemeGroupVersion instead as this is already defined with the package we are using and this also aligns with the value that is configured in management-plane-charts when switching between the two solutions.

internal/setup/cache_options.go
```diff
@@ -69,31 +94,32 @@ func (c *KcpCacheOptions) GetCacheOptions() cache.Options {
 					c.kcpNamespace: {},
 				},
 			},
-			&certmanagerv1.Issuer{}: {
-				Namespaces: map[string]cache.Config{
-					c.kcpNamespace:   {},
-					c.istioNamespace: {},
-				},
-			},
```
c-pius, line 77: Issuer has been removed entirely since it is not needed anylonger. The Certificate value is provided depending on the cert-management solution in use. See getCertManagementCacheObjects

cmd/main.go
ruanxin: nice, this is way better than before👍  I'm still thinking about how to handle the dependency compliance with the single-level abstraction principle. Basically it comes back to the previous discussion: if we treat CertificateManager as internal component of SKRWebhookManifestManager, is it necessary to expose it to other level only for construction, which violate the SLA.

Let's keep your solution, I will propose some options in the meeting for us to discuss.

internal/setup/cache_options.go
```diff
@@ -1,15 +1,40 @@
...
 	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+	"github.com/kyma-project/lifecycle-manager/internal/common"
+	"github.com/kyma-project/lifecycle-manager/pkg/watcher/certificate/certmanager"
+	"github.com/kyma-project/lifecycle-manager/pkg/watcher/certificate/gardener"
 )
 
+func SetupCacheOptions(isKymaManaged bool,
+	istioNamespace string,
+	kcpNamespace string,
+	certificateManagement string, setupLog logr.Logger,
```
LeelaChacha, line 23: Would it make sense to have a custom type `type CertificateManagement string` here? would make the use of this string universal in the project.

LeelaChacha: We decided that there is no clean solution when using a custom type since the const values for the two types is dynamic

LeelaChacha (approved):
Chuck Norris Thumbs Up

PR #2403, opened 2025-04-14
feat: Add New Script to Setup Test Clusters with GCM

LeelaChacha:
**Description**

- new script that setups the E2E test clusters with GCM using helm chart

**Related issue(s)**

Part of #2360


scripts/tests/create_test_clusters_gcm.sh
```diff
@@ -0,0 +1,113 @@
...
+        --k3s-arg --tls-san="host.k3d.internal@server:*" \
+        --k3s-arg --tls-san="skr.cluster.local@server:*"
+  
+  kubectl config use-context k3d-kcp
+
+  # install istio
+  istioctl install --set profile=demo -y
+
+  # install gardener cert-manager
+  helm install cert-controller-manager \
```
LeelaChacha, line 89: For the review: 
1. This helm chart does not include the DNSEntry CRD from GCM, which is why the controller will exit. Install [this](https://github.com/gardener/cert-management/blob/master/examples/11-dns.gardener.cloud_dnsentries.yaml) in KCP to prevent crashloopbackoff. This will be added to the GCM kustomize component as a work around.
2. Does helm need to be installed in the github runner?

versions.yaml
```diff
@@ -1,5 +1,6 @@
 # defines the versions of the tools used in the project
 certManager: "1.17.1"
+gardenerCertManager: "v0.17.5"
```
c-pius, line 3: Can we remove the `v` prefix here for consistency with the other versions?

scripts/tests/create_test_clusters_gcm.sh
```diff
@@ -0,0 +1,113 @@
...
+   oci://europe-docker.pkg.dev/gardener-project/releases/charts/cert-controller-manager \
+   --version "$GARDENER_CERT_MANAGER_VERSION" \
+   --set configuration.renewalWindow="$GARDENER_CERT_MANAGER_RENEWAL_WINDOW"
```
c-pius, line 92: It looks like this is the only change between the cert-manager and GCM setup scripts and the whole rest of the script is the same for both (except the arguments of course). Can we rather outsource the certificate managemnet setup from the setup clusters script and provide two dedicated small scripts that just apply the certificate management solution into the existing cluster? I.e.:

- `create_test_clusters.sh`
- `apply_cm_certificate_management.sh`
- `apply_gcm_certificate_management.sh`

scripts/tests/create_test_clusters_gcm.sh
c-pius: > This helm chart does not include the DNSEntry CRD from GCM, which is why the controller will exit. Install [this](https://github.com/gardener/cert-management/blob/master/examples/11-dns.gardener.cloud_dnsentries.yaml) in KCP to prevent crashloopbackoff. This will be added to the GCM kustomize component as a work around.

👍🏻 

> Does helm need to be installed in the github runner?

Should be there: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#package-management

scripts/tests/create_test_clusters_gcm.sh
LeelaChacha: I was exploring this route. But that makes the argument handling a bit cumbersome in the script. But I agree, that benefits may outweigh the maintenance of duplicate script lines we have

versions.yaml
LeelaChacha: I would even suggest adding the prefix to other versions.
Since strings with with prefix `v` are not valid (canonical) semvers, it is possible that a future release might remove the prefix. In that case, it might cause unexpected bug where a dev is trying to create a cluster with a valid semver but the script sneakily adds a prefix. Just by reading the error message it would not be immediately clear where the problem lies.

c-pius: The reason for not adding the `v` was that at some locations we require it without it, e.g. here https://github.com/kyma-project/lifecycle-manager/blob/main/.github/actions/install-istioctl/action.yml#L13

Adding it when needed seemed easier than stripping it from when not needed. Eventually I am okay with both if you have strong opinions about it.

LeelaChacha: no no. no strong opinions. I also didn't mean adding the prefix to all and stripping away when needed. I just meant that the versions.yaml file should reflect the actually version string that is being used.

c-pius: I will have to suppress my urge for consistency when looking at the file, but you have got a point about robustness which should be considered more important. Okay for me to change it, I would just like to have it as a separate PR then.

c-pius (approved):

PR #2406, opened 2025-04-14
docs: ADR 002 - Constructor Injection and Composition Functions

c-pius:
**Description**

- proposes pure DI with composition root

**TODOs**

- [ ] depending on the decision, refactor code introduced in https://github.com/kyma-project/lifecycle-manager/pull/2382


docs/contributor/adr/002-pure-di-composition-root.md
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+  "github.com/kyma-project/lifecycle-manager/internal/setup"
+  "github.com/kyma-project/lifecycle-manager/internal/controller/kyma"
+)
+
+func main() {
+  skrWebhookManager := setup.SetupSkrWebhookManager(logger, flagVar)
+
+  kyma.NewReconciler(skrWebhookManager).SetupWithManager()
```
c-pius, line 69: TODO: also put construction to setup and just call teh `SetupWithManager()`?

docs/contributor/adr/002-pure-di-composition-root.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Decision
+
+It is decided that constructor injection MUST be used as the uniform approach.
+To resolve the resulting dependency tree, the [Composition Root](https://blog.ploeh.dk/2011/07/28/CompositionRoot/) pattern is applied.
+In the absence of a proven framework for creating DI containers in Go, [Pure DI](https://stackoverflow.com/a/50724263) is applied.
+
+## Consequence
+
+Consumers MUST have their dependencies injected via constructor.
```
ruanxin, line 46: Although this topic is still under discussion, I think this ADR is trying to cover a broad topic by a narrow decision. For example, to achieve dependency injection, it's not only possible through the constructor, we have method injection, property injection, for certain case, it's even impossible injected through constructor, for example, if your class depends on different behaviour based on each operation, how can you given this dependency in the constructing time?

Here is a decision-making graph cited from book [Dependency Injection Principles, Practices, and Patterns by Steven van Deursen and Mark Seemann ](https://www.manning.com/books/dependency-injection-principles-practices-patterns#reviews).

image

What I expected is we have a decision regarding how we apply composition root with certain prerequisites or constraints.

ruanxin: 1. agreed property injection is not recommanded way.
2. That SKR client we can configured in the root, since its lifecycle aligns with the application's, so constructed in the root can be a solution. However, for some objects which have short living time, there is no reason or impossible to inject in the beginning, for example, the `context.Context`, we don't see any object hold it as a member, right? because it's lifetime is per reconcile request. You even have no way to get it from main.go, it's from `(ctx context.Context, req ctrl.Request)`.
3. Yes, that's what I want to clarify first, as I mentioned in the meeting, we still haven't decided what kind of components belongs to stable dependencies, those not necessary to expose to root. If we put this ADR before those decision, it's brings too much uncertainty, also increase the efforts for refactoring current code, for example, I just search `New` in declarative package, the components below are tightly coupled, I'm not say it's good design, but certainly where to compose them are debatable.   
https://github.com/kyma-project/lifecycle-manager/blob/d9ec6ba8d7e07cf87e43055c337294ed197b076d/internal/declarative/v2/reconciler.go#L293

https://github.com/kyma-project/lifecycle-manager/blob/d9ec6ba8d7e07cf87e43055c337294ed197b076d/internal/declarative/v2/reconciler.go#L359

docs/contributor/adr/002-constructor-injection-and-composition-functions.md
```diff
@@ -0,0 +1,95 @@
+# ADR 002 - Constructor Injection and Composition Functions
+
+## Status
+
+Accepted
+
+## Context
+
+It is agreed that [Depenency Inversion](https://medium.com/@inzuael/solid-dependency-inversion-principle-part-5-f5bec43ab22e) is a key principle for solid architecture.
```
lindnerby, line 9:
```suggestion
It is agreed that [Dependency Inversion](https://medium.com/@inzuael/solid-dependency-inversion-principle-part-5-f5bec43ab22e) is a key principle for solid architecture.
```

docs/contributor/adr/002-constructor-injection-and-composition-functions.md
```diff
@@ -0,0 +1,95 @@
...
+Once a consumer abstracted their dependencies properly, the key question is how to fill the abstraction with a concrete implementation.
+
+It has to be decided how and where to resolve depdenencies.
```
lindnerby, line 12:
```suggestion
It has to be decided how and where to resolve dependencies.
```

docs/contributor/adr/002-constructor-injection-and-composition-functions.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Decision
+
+It is decided that a consumer must have its depenencies injected via constructor.
```
lindnerby, line 16:
```suggestion
It is decided that a consumer must have its dependencies injected via constructor.
```

docs/contributor/adr/002-constructor-injection-and-composition-functions.md
```diff
@@ -0,0 +1,95 @@
...
+Constructors should follow the naming pattern `New<struct name>`.
+Dependencies of a consumer are resolved in a *composition function*.
+Composition functions live in a separate pacakge at `cmd/composition`.
+Composition functions should follow the naming pattern `Compose<struct name>`.
+Composition functions may build depenencies of the unit to compose themselves or have them injected from outside (depends on the kind of dependency).
```
lindnerby, line 21:
```suggestion
Composition functions may build dependencies of the unit to compose themselves or have them injected from outside (depends on the kind of dependency).
```

lindnerby (approved):

docs/contributor/adr/002-constructor-injection-and-composition-functions.md
```diff
@@ -0,0 +1,96 @@
...
+```go
+// /cmd/composition/service/webhook/skr_webhook_manager.go
+package webhook
+
+import (
+  "github.com/kyma-project/lifecycle-manager/internal/watcher"
+)
+
+func ComposeSkrWebhookManager(logger logr.Logger, flagVar *flags.FlagVar) *watcher.SkrWebhookManager {
+  certManager := composeCertManager(logger, flagVar)
```
ruanxin, line 45: I prefer not to use deeply nested Composition functions. For components like `certManager` as stable dependencies, it's more straightforward to construct them directly rather than introducing an extra layer of composition. Adding such levels only complicates usage unnecessarily. All Composition functions should be public and only invoked in `main`, as they are intended for components with stable dependencies. Not all consumers will need or use these composition functions.

docs/contributor/adr/002-constructor-injection-and-composition-functions.md
```diff
@@ -0,0 +1,96 @@
...
+It has to be decided how and where to resolve dependencies.
+
+## Decision
+
+It is decided that a consumer must have its dependencies injected via constructor.
+Constructors should follow the naming pattern `New<struct name>`.
+Stable dependencies of a consumer are resolved in a *composition function*.
+Composition functions live in a separate pacakge at `cmd/composition`.
+Composition functions should follow the naming pattern `Compose<struct name>`.
+Composition functions may build dependencies of the unit to compose themselves or have them injected from outside (depends on the kind of dependency).
```
ruanxin, line 21: If I'm understanding correctly, the proposal is to create a composition function for every top-level dependency? That doesn't seem strictly necessary. For a service whose dependencies are all volatile, it can just be constructed directly in the root main without needing an extra composition layer, and I would say in most cases, use Construct is enough.

docs/contributor/adr/002-constructor-injection-and-composition-functions.md
```diff
@@ -0,0 +1,96 @@
...
+package reconciler
+
+import (
+  "github.com/kyma-project/lifecycle-manager/cmd/composition/service/webhook"
+  "github.com/kyma-project/lifecycle-manager/internal/controller/kyma"
+)
+
+func ComposeKymaReconciler(logger logr.Logger, flagVar *flags.FlagVar) *kyma.Reconciler {
+  return kyma.NewReconciler(
+    webhook.ComposeSkrWebhookManager(logger, flagVar)
```
ruanxin, line 73: 1. If there is no stable dependencies dedicated for reconciler, I think we don't need such Composition function tbh.
2. this webhook should be injected from the external, but not here.

```
func main() {
  // ...
 webhook := webhook.ComposeSkrWebhookManager(logger, flagVar)
  kymaReconciler := reconciler.ComposeKymaReconciler(logger, flagVar, webhook)
  kymaReconciler.SetupWithManager(
    mgr,
    opts,
    settings,
  )
}
```

c-pius:
I am okay with the proposals by @ruanxin. @lindnerby would you mind taking a look and let me know what you think before I update the doc?

lindnerby:
> I am okay with the proposals by @ruanxin. @lindnerby would you mind taking a look and let me know what you think before I update the doc?

Yes, I am fine with the proposed additions.

c-pius:
Here would be the update based on Xin's feedback: https://github.com/kyma-project/lifecycle-manager/pull/2406/commits/24e57a40d6a3c313350f9f980b6bd46bab6233ef

ruanxin (dismissed):

lindnerby (approved):

docs/contributor/adr/002-constructor-injection-and-composition-functions.md
```diff
@@ -0,0 +1,87 @@
...
+Composition functions should follow the naming pattern `Compose<struct name>`.
+Composition functions should build stable dependencies of the unit to compose themselves and have other dependencies injected from the outside.
+Composition functions do NOT return errors but instead log them and exit the startup process.
+From `main`, composition functions are called and the dependency tree is resovled.
+
+## Consequence
+
+Consumers do not build or know the concrete implementations of their abstracted dependencies.
+The `main` function is kept free from resolving the detailed dependency tree.
+The detailed dependency tree is resolved via the compositions functions under `cmd/composition`.
```
mmitoraj, line 29:
```suggestion
* Consumers do not build or know the concrete implementations of their abstracted dependencies.
* The `main` function is kept free from resolving the detailed dependency tree.
* The detailed dependency tree is resolved via the composition functions under `cmd/composition`.
```

docs/contributor/adr/002-constructor-injection-and-composition-functions.md
```diff
@@ -0,0 +1,87 @@
...
+## Decision
+
+It is decided that a consumer must have its dependencies injected via constructor.
+Constructors should follow the naming pattern `New<struct name>`.
+Stable dependencies of a consumer are resolved in a *composition function*.
+Composition functions live in a separate pacakge at `cmd/composition`.
+Composition functions should follow the naming pattern `Compose<struct name>`.
+Composition functions should build stable dependencies of the unit to compose themselves and have other dependencies injected from the outside.
+Composition functions do NOT return errors but instead log them and exit the startup process.
+From `main`, composition functions are called and the dependency tree is resovled.
```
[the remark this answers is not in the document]
c-pius: function in this case. But there is also a branch called main. I updated the suggestion accordingly.

mmitoraj (approved):

PR #2408, opened 2025-04-15
feat: Add Kustomize Component for GCM

LeelaChacha:
**Description**

- makes Kustomize Component for open-source CM independent:
  - now certmanager component is not a part of the control-plane Kustomization and is independent
    - prefixes are pre-set in resources
    - commonLabels are pre-set in resources
    - kcp-system namespace is pre-set in resources
- adds new independent Kustomize Component for GCM: `config/gardener-certmanager`
  - contains the issuers, certificates, and rbac stuff ported from the independent open-source cert-manager component
- new make target `local-deploy-with-watcher-gcm` added

**Related issue(s)**

Related to #2360


Makefile
```diff
@@ -129,7 +129,12 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 .PHONY: local-deploy-with-watcher
 local-deploy-with-watcher: generate kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/watcher_local_test | kubectl apply -f -
+	$(KUSTOMIZE) build config/watcher_local_test_with_certmanager | kubectl apply -f -
+
+.PHONY: local-deploy-with-watcher-gcm
+local-deploy-with-watcher-gcm: generate kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
```
ruanxin, line 135: I think maybe it's time we add some description to replace current comments (`## Deploy controller to the K8s cluster specified in ~/.kube/config.`) for those two commands, it's hard to tell from the command what's the difference and the purpose.

config/certmanager/certificate_watcher.yaml
```diff
@@ -4,15 +4,27 @@
...
   namespace: istio-system
 spec:
   selfSigned: { }
 ---
 apiVersion: cert-manager.io/v1
 kind: Certificate
 metadata:
-  name: watcher-serving
+  labels:
+    app.kubernetes.io/created-by: argo-cd
```
ruanxin, line 22: this label should be defined in the `commonLabels`, which configured in the root Kustomization
https://github.com/kyma-project/lifecycle-manager/blob/2b4b9a82f9501b4d47fba0ef1f4a32ee3e1fc8dd/config/control-plane/kustomization.yaml#L5-L9

config/gardener-certmanager/kustomization.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+resources:
+  - crd/dns.gardener.cloud_dnsentries.yaml # needed for GCM to work since this is not included in the helm chart for GCM
```
ruanxin, line 4: I didn't get it, why we need to this 3rd party CRD, shouldn't GCM provide this by itself if it's installed in the KCP in the future?

it not included in the helm chart on purpose or it's their design oversight? I mean it's anyway not a good idea provide by the consumer, if we go down that path, how can we ensure compatibility between the CRD and GCM, What happens if a newer GCM version no longer supports the current CRD version, How would we handle upgrades safely, those question should be clarified first.

Even if we have to do it, providing the CRD as a static local file is not an acceptable solution, we lose visibility and traceability of the source.

config/certmanager/certmanager_role_binding.yaml
```diff
@@ -1,25 +1,39 @@
...
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/name: lifecycle-manager
+    app.kubernetes.io/part-of: kcp
+  name: klm-controller-manager-certmanager
   namespace: istio-system
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: Role
-  name: controller-manager-certmanager
+  name: klm-controller-manager-certmanager
```
ruanxin, line 15: Should this prefix be offered by `namePrefix: klm-` in the kustomization? We shouldn't not add them manually.

config/control-plane/kustomization.yaml
```diff
@@ -18,7 +18,6 @@ components:
   - ../watcher
   - ../webhook
   - ../grafana
-  - ../certmanager
```
ruanxin, line 21: maybe because I'm not following the whole issue, why we decide to remove certmanager here? We added it in the management control plane later?

config/watcher_local_test/kustomization.yaml
```diff
@@ -4,19 +4,59 @@
...
   - namespace.yaml
-  - ../control-plane
+  - ../manager
 components:
+  - ../crd
+  - ../rbac
   - ../istio
+  - ../watcher
+  - ../webhook
+  - ../grafana
```
ruanxin, line 27: same here, https://github.com/kyma-project/lifecycle-manager/pull/2408/files#r2048452101

config/watcher_local_test_gcm/kustomization.yaml
```diff
@@ -0,0 +1,147 @@
...
+  app.kubernetes.io/part-of: kcp
+  app.kubernetes.io/managed-by: kustomize
+images:
+  - name: europe-docker.pkg.dev/kyma-project/prod/lifecycle-manager
+generatorOptions:
+  disableNameSuffixHash: true
+resources:
+  - namespace.yaml
+  - ../manager
+  - crd/dns.gardener.cloud_dnsentries.yaml # needed for GCM to work since this is not included in the helm chart for GCM
```
ruanxin, line 17: can we get this crd publicly in github repository? Then you can use remote address directly, it also allowing you to reference to a specific tag or commit https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md#examples

config/certmanager/certificate_watcher.yaml
```diff
@@ -28,18 +28,18 @@ spec:
     algorithm: RSA
     size: 4096
   issuerRef:
-    name: klm-watcher-root
+    name: watcher-root
```
ruanxin, line 31: good we can drop this prefix now 👍

config/gardener-certmanager/certificate_watcher.yaml
```diff
@@ -0,0 +1,41 @@
...
+  isCA: true
+  commonName: klm-watcher-selfsigned-ca
+  secretName: klm-watcher # this secret will not be prefixed, since it's not managed by kustomize
+  secretLabels:
+    operator.kyma-project.io/managed-by: "lifecycle-manager"
+  privateKey:
+    algorithm: RSA
+    size: 4096
+  issuerRef:
+    name: klm-watcher-root
```
ruanxin, line 26: just wondering is there anyway also provide issuerRef with that pefix? Not relevant to this PR, but could be a followup, we have more place manually added `klm`, not easy to maintain.

LeelaChacha: It does not work like open-source CM setup because the Issuer and Certificate are in different namespaces. And we have to constrain our GCM Issuer to the default namespace due to limitations.

ruanxin: what I mean is using that `transformers`, you just need to specify the path for the resources which require adding prefix, just tried, it's working.
```
transformers:
  - |-
    apiVersion: builtin
    kind: PrefixSuffixTransformer
    metadata:
      name: add-klm-prefix-to-resources
    prefix: klm-
    fieldSpecs:
    - path: subjects/name
      kind: RoleBinding
    - path: subjects/name
      kind: ClusterRoleBinding
    - path: spec/issuerRef/name
      kind: Certificate
```

ruanxin: btw, there are many other place have `klm-` prefix manually introduced, please double check.

https://github.com/kyma-project/lifecycle-manager/pull/2408/files#diff-e716703499082283a0b8b8ce52aff216abe1f43bdac55bf7676feab8e4424c06R40

https://github.com/kyma-project/lifecycle-manager/pull/2408/files#diff-e716703499082283a0b8b8ce52aff216abe1f43bdac55bf7676feab8e4424c06R18-R19

https://github.com/kyma-project/lifecycle-manager/pull/2408/files#diff-e716703499082283a0b8b8ce52aff216abe1f43bdac55bf7676feab8e4424c06R18-R19

Also I'm a bit confused about this comment `  secretName: klm-watcher # this secret will not be prefixed, since it's not managed by kustomize`, what does it mean will not be prefixed and actually it's prefixed?

ruanxin (approved):

PR #2420, opened 2025-04-22
feat: E2E Tests with GCM

### LINKED ISSUE #2360: [Gardener CM] Create E2E tests and kustomize config for GCM

LeelaChacha:
**Description**

- added new GitHub Actions workflow: `test-e2e-with-gcm.yml`
- added new action for setting up test clusters with GCM
- adapted action for staging of KLM for E2E tests

**Related issue(s)**

Closes #2360


LeelaChacha:
the tests are failing which is to be expected.

.github/actions/deploy-lifecycle-manager/action.yml
```diff
@@ -14,11 +14,16 @@ runs:
...
+        else
+          E2E_MAKE_TARGET=local-deploy-with-watcher
+        fi
         maxRetry=5
         for retry in $(seq 1 $maxRetry)
         do
-          if make local-deploy-with-watcher IMG=europe-docker.pkg.dev/kyma-project/${{ inputs.klm_image_repo }}/lifecycle-manager:${{ inputs.klm_version_tag }}; then
-            kubectl wait pods -n kcp-system -l app.kubernetes.io/name=lifecycle-manager --for condition=Ready --timeout=90s
+          if make "$E2E_MAKE_TARGET" IMG=europe-docker.pkg.dev/kyma-project/${{ inputs.klm_image_repo }}/lifecycle-manager:${{ inputs.klm_version_tag }}; then
+            kubectl wait pods -n kcp-system -l app.kubernetes.io/name=lifecycle-manager --for condition=Ready --timeout=180s
```
lindnerby, line 26: What's the reason behind doubling the timeout?

.github/actions/setup-test-clusters-gcm/action.yml
```diff
@@ -0,0 +1,27 @@
...
+inputs:
+  k8s_version:
+    description: The version of k8s to use. For example, 1.28.7
+    required: true
+  gardener_cert_manager_version:
+    description: The version of gardener cert-manager to deploy. For example, 0.17.5.
+    required: true
+  gardener_cert_manager_renewal_window:
+    description: The duration before the certificate expiration when the renewal should be triggered. Default is 720h.
+    required: false
```
lindnerby, line 12: For the sake of e2e tests this is required isn't it?

.github/actions/deploy-lifecycle-manager-e2e/action.yml
```diff
@@ -10,12 +10,20 @@ inputs:
 runs:
   using: composite
   steps:
+    - name: Set Kustomize Directory
+      shell: bash
+      run: |
+        if [[ "${E2E_USE_GARDENER_CERT_MANAGER}" == "1" ]]; then
```
lindnerby, line 16: I don't think we need to be that explicit and check against `"1"` 😄 
`if [[ -n "${E2E_USE_GARDENER_CERT_MANAGER}" ]]; then` should be enough for the control flow here (set vs unset)

Same for other occurrences in this PR.

lindnerby:
> the tests are failing which is to be expected.

As a result of the tests, shouldn't the klm pod come up already or are still missing some gcm code?

.github/actions/setup-test-clusters-gcm/action.yml
LeelaChacha: The script has a default. But I will make it required

.github/actions/deploy-lifecycle-manager/action.yml
LeelaChacha: was used for testing purposes, since the GCM requires a CRD that is not included in its helm chart, which was supplied in the kustomization. It would go into backoff and need longer to run because of the delay in between, consequently, klm would need longer. Now I reduced the timeout again and made the script that deploys GCM's chart also supply this CRD

LeelaChacha:
> As a result of the tests, shouldn't the klm pod come up already? Or are we still missing some gcm code?

We were missing the addition of the GCM types to the schema. I added that in my PR now.

lindnerby (approved):

Issue #2359, opened 2025-03-25
[Gardener CM] Make istio gateway rotation handler cert type agnostic


lindnerby:
### Description

With the Gardener cert-manager support being introduced, we need to make the istio gateway rotation cert-manager agnostic. That means getting the validity of the certificate resource will have to change. Currently, we get the validity of the cert from the Certificate's status subresource:
```
apiVersion: cert-manager.io/v1
kind: Certificate
...
status:
  notAfter: "2025-05-25T08:44:09Z"
  notBefore: "2025-02-24T08:44:09Z"
```
For the Gardener cert-manager Certificate, we don't have a clear spec to reliably get the validity. Refer to https://github.com/kyma-project/lifecycle-manager/issues/2312#issuecomment-2743249074 for more details.
Therefore, we need to get the certificate validity by decoding the pem from the cert itself.

### Reasons

Istio gateway secret rotation should work regardless of the type of cert-manager used.

### Acceptance Criteria

- [x] abstraction for getting certificate validity is introduced
  - [x] composed based on the flag introduced in #2357
  - [x] implementation for the open-source cert-manager Certificate uses the current approach
  - [x] implementation for the gardener cert-manager Certificate throws a NotImplementedError
    - [x] create a follow-up issue for actual implementation (read from status when [feature](https://github.com/gardener/cert-management/issues/454) added OTHERWISE decoding the PEM in secret and get the validity)
- [x] [legacy](https://github.com/kyma-project/lifecycle-manager/tree/main/internal/gatewaysecret/legacy) and [ca-bundle](https://github.com/kyma-project/lifecycle-manager/tree/main/internal/gatewaysecret/cabundle) handlers use this abstraction
- [x] unit tests are created

### Feature Testing

Unit tests

janmedrek:
Waiting for clarification from the Gardener CM's side

LeelaChacha:
Blocked by feature request to the Gardener CM team: https://github.com/gardener/cert-management/issues/454

c-pius:
Follow-ups:

- https://github.com/kyma-project/lifecycle-manager/issues/2423

PR #2421, opened 2025-04-22
chore: Make Gateway Cert Rotation Type Agnostic

c-pius:
**Description**

- cert rotation can now work with either a cert-manager or gardener certificate management implementation of certificate
- the gardener certificate management certificate implementation does not yet support the new `GetValidity` method, this will be implemented as separate ticket.

**Related issue(s)**

- closes https://github.com/kyma-project/lifecycle-manager/issues/2359


.mockery.yaml
```diff
@@ -1,11 +1,11 @@
...
-filename: "mock_{{.InterfaceName}}_test.go"
+outpkg: "testutils"
+filename: "mock_{{.InterfaceName}}.go"
 inpackage: true
 issue-845-fix: true
 packages:
-  github.com/kyma-project/lifecycle-manager/pkg/gatewaysecret:
+  github.com/kyma-project/lifecycle-manager/internal/gatewaysecret:
     interfaces:
       Client:
```
c-pius, line 11: Adapted this config to produce the MockClient at the location we had it in already (this was out of sync). We also need to consider aligning our expectations. Right now, the mock client is not stripped as part of testing code, it should be though.

internal/setup/certificate.go
```diff
@@ -0,0 +1,42 @@
...
+		flagVar.IstioNamespace,
+		config,
+	)
+	if err != nil {
+		setupLog.Error(err, "unable to initialize Gardener certificate management")
+		os.Exit(bootstrapFailedExitCode)
+	}
+
+	return certClient
+}
```
c-pius, line 42: These are outsourced from `setup/watcher.go` for re-use

.mockery.yaml
c-pius: https://github.com/kyma-project/lifecycle-manager/issues/2422

internal/setup/gatewaysecret.go
```diff
@@ -0,0 +1,44 @@
...
+	gatewaysecretclient "github.com/kyma-project/lifecycle-manager/internal/gatewaysecret/client"
+	"github.com/kyma-project/lifecycle-manager/internal/pkg/flags"
+	"github.com/kyma-project/lifecycle-manager/pkg/watcher/certificate"
+)
+
+//nolint:ireturn // chosen implementation shall be abstracted
+func SetupCertInterface(kcpClient client.Client,
+	flagVar *flags.FlagVar,
+	setupLog logr.Logger,
+) gatewaysecretclient.CertificateInterface {
```
c-pius, line 21: I am not super happy with how we set things up overall. I would like to go forward with this as a pragmatic approach but we need to get clarity on the topic: https://github.com/kyma-project/lifecycle-manager/pull/2406

pkg/watcher/certificate/gardener/certificate.go
```diff
@@ -154,3 +157,10 @@ func (c *CertificateClient) GetRenewalTime(ctx context.Context,
 
 	return expirationDate.Add(-c.config.RenewBefore), nil
 }
+
+func (c *CertificateClient) GetValidity(ctx context.Context,
```
LeelaChacha, line 161: just nitpicking here cause it is a quick fix: the parameter names should be replaced with _

LeelaChacha (changes requested):

LeelaChacha (approved):
LGTM

PR #2434, opened 2025-04-24
chore: Make WatcherCR `spec.labelsToWatch` optional

### LINKED ISSUE #2341: [SKR Watch] Make WatcherCR `spec.labelsToWatch` optional
### Description

As of now, it seems to be mandatory to provide a label in the WatcherCR `spec.labelsToWatch`. As consequence, a watched object must also have a label present in order to be watched. We should consider to make this label optional.

### Reasons

Follow-up from https://github.com/kyma-project/runtime-watcher/issues/488. In the Framefrog use case, the resource to watch is completely customer managed. To be watched, Framefrogs would need to define a label and make their module manager apply this label to each resource the customer creates. This may be simplified.

### AC

- [x] ~Investigated if indeed at least one label in WatcherCR `spec.labelsToWatch` is required~ Currently atleast one label in WatcherCR `spec.labelsToWatch` is required
  - [x] @LeelaChacha may be able to confirm this directly
- [ ] ~~Investigate/Discussed whether this can be made optional~~
  - consider aligning with Framefrogs if it would help them or if also okay as is
- [ ] ~~Created follow-up tickets for implementation if necessary~~
- [ ] consulted literature to understand how this change can be done from api compoatibility perspecitve
- [ ] implemented the change accordingly
- [ ] labelsToWatch can be empty and processing does not fail.

amritanshusikdar:
**Description**
- Makes watcherCR `spec.labelsToWatch` optional

**Related issue(s)**
Resolves #2341

### DIFF
--- a/api/v1beta2/watcher_types.go
+++ b/api/v1beta2/watcher_types.go
@@ -29,7 +29,8 @@ type WatcherSpec struct {
 	ServiceInfo Service `json:"serviceInfo"`
 
 	// LabelsToWatch describes the labels that should be watched
-	LabelsToWatch map[string]string `json:"labelsToWatch"`
+	// +optional
+	LabelsToWatch map[string]string `json:"labelsToWatch,omitempty"`
 
 	// ResourceToWatch is the GroupVersionResource of the resource that should be watched.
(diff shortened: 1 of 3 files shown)


LeelaChacha (approved):
LGTM

Issue #2212, opened 2025-01-27
Reinvestigate Orphaned Manifests in the STAGE landscape


LeelaChacha:
### Description

It was found in the investigation of #2138 that some manifest CRs on stage are neither in the kyma.spec.modules nor kyma.status.modules in the Kyma CR linked to them. Furthermore, these do not have a deletion timestamp and are being reconciled. (These are also not mandatory modules)

(For Refinement:)
We need to reinvestigate if more manifests have popped up since then. Here is a script that lists all the orphaned manifests:
```bash
#!/bin/bash

manifests=$(kubectl get manifest -n kcp-system -o json | jq -r '.items[] | "\(.metadata.name) \(.metadata.deletionTimestamp // "false") \(.metadata.labels["operator.kyma-project.io/mandatory-module"] // "false")"')

kyma_manifests=$(kubectl get kyma -n kcp-system -o json | jq -r 'if .items then .items[].status.modules[]?.manifest.metadata.name else empty end' | sort | uniq)

while IFS=' ' read -r manifest deletion_timestamp mandatory_module; do
  if [ "$mandatory_module" = "true" ]; then
    continue
  fi

  if [ "$deletion_timestamp" != "false" ]; then
    continue
  fi

  if ! echo "$kyma_manifests" | grep -q "$manifest"; then
    echo "Manifest $manifest is orphaned."
  fi
done <<< "$manifests"
```

#### NOTE: no orphaned manifests on dev or prod were found. If no new ones found on stage, then it is probably not worth investigating.

---
If more manifests than [the last time](https://github.com/kyma-project/lifecycle-manager/issues/2138#issuecomment-2611916011) have appeared. We should find out the cause of this behaviour.
### Proposal:
If no huge performance hit is expected, the Manifest Reconciler should check if the Manifest being reconciled is orphaned (not listed in the Kyma spec and status). The Kyma CRs should be cached for this purpose. Furthermore, a metric should be created that counts orphaned manifests. An alert can then be triggered if a new manifest is orphaned. Then, fresh logs will be available to diagnose this anomaly.

Consider creating a cron-job for checking the orphaned resources, it could create a report containing all the orphaned resources.

### AC:
- [ ] Evaluate the available options (reconciler vs cron-job vs whatever could fit here better)
- [ ] Manifest Reconciler/cron-job checks the corresponding Kyma to ensure the manifest is referenced.
- [ ] A new Counter metric is created and increased if an orphaned manifest is detected.
[...]

PR #2442, opened 2025-04-29
feat: Detect orphaned manifests

Tomasz-Smelcerz-SAP:
**Description**

Implement orphaned manifest detection as described in: #2212 

- Detection of orphaned Kyma manifest (Parent Kyma cannot be found)
- If orphaned, put the Manifest in the error state and do not process it
- emit a new metric: `manifest_orphaned`
- return an error from reconciliation

**Related issue(s)**
#2212


config/manager/kustomization.yaml
```diff
@@ -3,9 +3,9 @@ kind: Kustomization
 generatorOptions:
   disableNameSuffixHash: true
 resources:
-  - manager.yaml
-  - metrics_service.yaml
+- manager.yaml
+- metrics_service.yaml
 images:
-  - name: controller
```
LeelaChacha, line 9: please remove from final PR

internal/service/manifest/orphan/detection.go
```diff
@@ -0,0 +1,75 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+	"github.com/kyma-project/lifecycle-manager/pkg/util"
+)
+
+var ErrOrphanedManifest = errors.New("orphaned manifest detected")
+
+type APIClient interface {
```
lindnerby, line 14: Rename this to Repository please.

unit-test-coverage.yaml
```diff
@@ -19,9 +19,11 @@ packages:
   internal/maintenancewindows: 87
   internal/pkg/resources: 91
   internal/remote: 29
+  internal/repository/manifest/orphan: 100
   internal/service/kyma/status/modules: 78
   internal/service/kyma/status/modules/generator: 100
   internal/service/kyma/status/modules/generator/fromerror: 100
+  internal/service/manifest/orphan: 95.8
```
lindnerby, line 26: Let's push this to 💯 !

internal/service/manifest/orphan/detection.go
```diff
@@ -0,0 +1,75 @@
...
+func (s *DetectionService) DetectOrphanedManifest(ctx context.Context, manifest *v1beta2.Manifest) error {
+	if manifest.IsMandatoryModule() {
+		// Mandatory modules are not refereced by any Kyma object so cannot be orphaned
+		return nil
+	}
+
+	if manifest.GetDeletionTimestamp() != nil {
+		// If the manifest is being deleted, we don't check for orphaned status (as it should be eventually deleted)
+		return nil
+	}
```
lindnerby, line 37: Can you wrap all the skip cases into one func please

internal/declarative/v2/reconciler.go
```diff
@@ -133,6 +141,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{RequeueAfter: r.Success}, nil
 	}
 
+	err = r.orphanDetectionService.DetectOrphanedManifest(ctx, manifest)
+	if err != nil {
+		if errors.Is(err, orphan.ErrOrphanedManifest) && !isManifestRecentlyCreated(manifest.CreationTimestamp.Time) {
```
lindnerby, line 146: This can also be part of the detection service!!

So you can even skip the cyclomatic complexity linter suppression ;)

internal/declarative/v2/reconciler.go
```diff
@@ -210,6 +228,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	return r.finishReconcile(ctx, manifest, metrics.ManifestReconcileFinished, manifestStatus, nil)
 }
 
+func isManifestRecentlyCreated(manifestCreationTimestamp time.Time) bool {
```
lindnerby, line 231: Move to service as well

internal/repository/manifest/orphan/client.go
```diff
@@ -0,0 +1,29 @@
+package orphan
```
lindnerby, line 1: The repository impl does not know of the use case it is used in "orphaned manifest etc."
You can just put this repo impl to `internal/repository/kyma/client.go`, since it is actually a kyma repository implementation.

internal/service/manifest/orphan/detection_test.go
```diff
@@ -0,0 +1,245 @@
+package orphan_test
```
lindnerby, line 1: To reach 100% Copilot is suggesting to add:

1. Error when manifest.GetKymaName() fails in getParentKyma: Add a test case where manifest.GetKymaName() returns an error. This will cover the error path in the getParentKyma method.
Error when util.IsNotFound(err) is false in getParentKyma:


2. Add a test case where s.client.GetKyma returns an error that is not a "not found" error, and ensure the correct error is returned.
Edge case for isManifestReferencedInKymaStatus:


3. Add a test case where kyma.Status.Modules contains a nil entry to ensure the function handles it gracefully.

internal/declarative/v2/reconciler.go
lindnerby: Basically everything except setting the status should be part of the service.

internal/service/manifest/orphan/detection_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+	err := service.DetectOrphanedManifest(t.Context(), manifest)
+
+	require.Error(t, err)
+	require.ErrorIs(t, err, orphan.ErrOrphanedManifest)
+}
+
+func TestDetectionService_DetectOrphanedManifest_WhenClientReturnsError_ReturnError(t *testing.T) {
+	service := orphan.NewDetectionService(&clientStubGenericError{})
+	manifest := &v1beta2.Manifest{
```
lindnerby, line 76: For these cases where we just want to create a default manifest with kyma reference, maybe a test helper func?

internal/pkg/metrics/manifest.go
```diff
@@ -33,6 +33,7 @@ const (
 	ManifestReconcileFinished            ManifestRequeueReason = "manifest_reconcile_finished"
 	ManifestUnmanagedUpdate              ManifestRequeueReason = "manifest_unmanaged_update"
 	ManifestResourcesLabelRemoval        ManifestRequeueReason = "manifest_labels_removal"
+	ManifestOrphaned                     ManifestRequeueReason = "manifest_orphaned"
```
lindnerby, line 36: Since we were talking about the cleanup yesterday:

Just setting this metric should be fine, we are calling cleanup from the metrics pkg:
`internal/pkg/metrics/manifest.go`

lindnerby (changes requested):

internal/service/manifest/orphan/detection_test.go
LeelaChacha: 2. is already accounted for in test "TestDetectionService_DetectOrphanedManifest_WhenClientReturnsError_ReturnError"
3. is already accounted for in "TestDetectionService_DetectOrphanedManifest_WhenModuleStatusManifestNil_ReturnOrphanedManifestError"

internal/declarative/v2/reconciler.go
LeelaChacha: cyclomatic complexity existed before the check for manifest creation timestamp. other parts of the function will have to be refined to get the complexity down, since the if block for the error is essential to setting the status

lindnerby (approved):

PR #2456, opened 2025-05-08
chore: Align Certificate to ADR02, ADR03, ADR04

c-pius:
**Description**

- aligns the Certificate Management part to ADRs 02, 03, 04
  - note that ADR02 is not fully there yet, just renamed the composition part, will also need to do some restructuring in a follow-up pr
- naming proposals in this PR:
  - service `CertificateManager` without `Service` suffix, it should be clear from the import that it is a service?
  - repository `Certificate` without `Repository` suffix, it should be clear from the import that it is a repository?
    - also avoided `Client` suffix
  - client.Client in `Certificate`: named `kcp` as the type and layer (repository) indicates that it is a client against the cluster and chose name `kcp` to signify that it is targeting KCP
    - if the same repository is later used also against SKR, then we may rename it to client. Then it should always be clear: called kcp => KCP repository, skr => SKR repository, client => can be used as a repository for both, KCP and SKR

**Related issue(s)**


c-pius:
Closing in favor of https://github.com/kyma-project/lifecycle-manager/pull/2604

Issue #2445, opened 2025-05-05
[Gardener CM] Deploy and verify KLM on US50 Canary Landscape


c-pius:
### Description

Deploy KLM configured with Gardener CM on the US50 Canary Landscape. Verify that it is working as expected. Contact from SRE side is Benjamin who may help with access and permissions to the landscape, maybe even the deploying.

Clarify if the management-plane-charts repo is the origin for the deploy config. If so, check what needs to be adapted (e.g. flags to deploy it).

### Reasons

Kyma in restricted markets, KLM is currently the bottle neck: `/kyma/restricted-markets/issues/28#issuecomment-11672557`

### Acceptance Criteria

- [x] KLM deployed on US50 Canary Landscape
- [ ] verified KLM functions properly with Gardener CM
- [x] adapted the deployment config / documented the config
- [x] document any significant findings related to the deployment and functionalities w/r to Gardener CM

c-pius:
## Notes

The distinction whether to deploy with Gardener CM or cert-manager is done using the `apiVersions`. This approach was just continued as SRE introduced it when starting with Gardener CM.

When running the `helm template`, it retrieves the supported API versions from the cluster it connects to. If it includes `cert-manager.io/v1` KLM is deployed for `cert-manager`. If not, it deploys for `cert.gardener.cloud`.

To simulate that in the local env, there is a `/work/.apiversions` file. This file is regenerated when executing the `make codegen` target and includes `cert-manager.io/v1`. To render the chart for `cert.gardener.cloud`, remove the `cert-manager.io/v1` entry from `/work/.apiversions` before rendering the chart with `make template`.

c-pius:
We saw an issue that "SKR cert duration must not be longer than the root cert duration". Re-configured the cert duration which resolved this issue. Kyma is in ready now saying that webhook is synced.

Will receive access to the SKR cluster by PK tomorrow to verify that events reach KCP.

PR #2458, opened 2025-05-14
chore: Trim cert-related roles

c-pius:
**Description**

- trims obsolete roles from certificate management
  - removes `Issuer` in `kcp-system` and `istio-system` namesapces
  - removes `Certificate` in `kcp-system` namespace
- the need for permissions wrt. `Issuer` was removed here:
  - https://github.com/kyma-project/lifecycle-manager/commit/852c0a5446c32ff3dcea0d40b76d99c6df212669#diff-3a2a38e72239c4d8fa7b8f569c4f54ee1e488a84af7468fec01e9b6efbc93248 (see `L154 of `pkg/watcher/certificate_manager.go`)
- I don't know why we in addition introduced permissions wrt. `Certificate` in `kcp-system` namespace
  - on a quick scan through the codebase I couldn't confirm the need for it, also the tests are green without it => assume it is indeed not necessary

**Related issue(s)**

- discovered while working on https://github.com/kyma-project/lifecycle-manager/issues/2445


LeelaChacha:
> I don't know why we in addition introduced permissions wrt. Certificate in kcp-system namespace

Also, my investigation showed no need for this role. Only the webhook-serving ca cert is in the kcp-system and klm should have no need to CRUD that resource anyway.
The redundant roles were ported as-is from the certmanager kustomization component to gardener without verifying/questioning the need.

config/certmanager/certmanager_role.yaml
```diff
@@ -11,54 +11,20 @@ rules:
     resources:
       - secrets
     verbs:
-      - list
       - watch
+      - list
```
c-pius, line 15: FYI: list is needed as watch doesn't seem to work without list

config/certmanager/certmanager_role.yaml
```diff
@@ -11,54 +11,20 @@ rules:
...
-  - apiGroups:
-      - cert-manager.io
-    resources:
-      - certificates
-    verbs:
-      - create
-      - delete
       - get
-      - list
+      - create
```
c-pius, line 28: FYI: create needed as it seems the `patch` of a non-existing object (upsert) is not working without `create`

tests/e2e/rbac_privileges_test.go
```diff
@@ -195,17 +176,12 @@ var _ = Describe("RBAC Privileges", func() {
...
-					Verbs:     []string{"patch", "list", "watch", "get", "create", "delete"},
-				},
-				{
-					APIGroups: []string{"cert-manager.io"},
-					Resources: []string{"issuers"},
-					Verbs:     []string{"list", "watch"},
+					Verbs:     []string{"watch", "list", "get", "create", "patch", "delete"},
 				},
 			}
 			Expect(GetRoleBindingRolePolicyRules(ctx,
```
lindnerby, line 187: Can we change the list asserts in this e2e test from `.To(Equal(*))` to `.Should(ConsistOf(*))`?
So for `crdRoleRules`, `leaderElectionRoleRules`, `klmManagerCertManagerRoleRules`, etc.

This should eliminate the need for items to be in the same order, just checking item equality.

lindnerby (changes requested):

lindnerby (approved):

Issue #2224, opened 2025-01-30
Replace "omitEmpty" in CRDs with // +optional Kubebuilder marker


nesmabadr:
### Description

As a followup to this [bug](https://github.com/kyma-project/lifecycle-manager/pull/2199), we need to check the fields marked with omitEmpty in the CRDs, because if a user edits a field which Is marked with “omitEmpty” in the SKR Cluster, and their value in the KCP cluster is null or zero-valued, the field value does not get synced from the KCP. We should replace the omitEmpty with the Kubebuilder “// +optional” marker. 

### Acceptance Criteria

- [x] Replace omitEmpty with //+optional in Kyma, ModuleTemplate and ModuleReleaseMeta CRs
- [x] Make sure that the fields get sync correctly from the KCP even if they are null or empty.

### Feature Testing

End-to-End tests

medmes:
After we had a sync, As follow we suggested to limit the changes only for booleans fields for now as they re the one making this sync issue. 
As follow up for the narrowed fix for the booleans fields: 
https://github.com/kyma-project/lifecycle-manager/issues/2496

c-pius:
Verified: "skipMaintenanceWindows" not shown as false value in Kyma.spec. Set "requiresDowntime" to true on SKR MT which was reverted by KLM.

PR #2478, opened 2025-05-23
fix: Make +optional in the API definition and required on the runtime

medmes:
**Description**

- Remove all necessary omitempty tag and // +optional to Kyma, ModuleTemplate fields
- e2e Tests covering empty and nil synchronisation

**Related issue(s)**
#2224


c-pius (approved):
For later reference... There are still occurrences for `v1beta1` API. But since we don't use it anymore we left it as is.

Issue #2460, opened 2025-05-15
Introduce a new flag `oci-registry-cred-secret` for KLM to allow configure a name of the Secret containing the credential.


janmedrek:
**Description**
There should be a new flag introduced for selecting the Secret for registry access.
It's assumed that this Secret can be found in `kcp-system`

**Acceptance Criteria**
- [x] Introduce the flag
- [x] Make sure the secret is watched (cached)
- [ ] ~~Add E2E testing for various secret-watch-related scenarios~~ Moved to separated [issue](https://github.com/kyma-project/lifecycle-manager/issues/2541)
- [x] Adapt current KeyChainProvider so that it reads credential selector from flag instead of image spec. https://github.com/kyma-project/lifecycle-manager/blob/2ce8b53dff38ba82ce3fdac904424fa40099b009/internal/manifest/keychain_provider.go#L22
- [x] Update the document https://github.com/kyma-project/lifecycle-manager/blob/main/docs/contributor/03-config-private-registry.md

ebensom:
In the meanthime while this is being implemented, I'm setting up the ModuleTemplates in US50 Canary manually using the old oci-reg-cred label approach.

Issue #1944, opened 2024-10-09
[Restricted Markets] Make credential selector as klm flag


ruanxin:
At the moment, to access private registry, KLM designed it in this way, it expected a selector label `oci-registry-cred` as ocm resources label. It works now because during module creation, all ocm content will be persisted locally in the ModuleTemplate including this label. However, in the restricted market, the ocm becomes the single source of truth, if the credential label persisted inside, it becomes chicken egg problem. We need to have a new solution for this case.

### Proposal
The credential selector can be a configurable flag in KLM, then for different restricted market, the support team can configure it dedicated.

### Reasons

Support kyma module deployment in other landscapes.

### Feature Testing
New E2E
Refactor Integration tests

### Testing approach

have an integration test verify the provided flag can fetch credential secret and verify the content correctly.  adapt this existing test if necessary https://github.com/kyma-project/lifecycle-manager/blob/b1fc52d3dd3b167420bc8beaed2b7df07f4c9ffa/tests/integration/controller/manifest/keychain_test.go#L21

c-pius:
Closing as the sub-tasks were all implemented

PR #2480, opened 2025-05-28
feat: Support private OCI registry via credentials Secret

lindnerby:
**Description**

- Remove all label-based lookup for OCI registry credentials
- Introduces KLM flag to provide a name for Secret containing private OCI registry creds
- Introduce keychain lookup for provided Secret
- Adapt integration test
- Introduce first E2E test with private registry

**Related issue(s)**


Tomasz-Smelcerz-SAP (commented):
A question: Previously we've used a list of secrets matching the configured selector.
Now, we have just one secret, configured by name.
Is it really how it should be implemented?
I know that #2460 mentions "secret name", but it's parent issue: #1944 describes flag for configuring secret(s) selector...

tests/e2e/oci_registry_credentials_secret.go
Tomasz-Smelcerz-SAP, line 1: I think this empty file is not required?

tests/integration/controller/manifest/keychain_test.go
```diff
@@ -1,87 +0,0 @@
...
-					"digest", CredSecretLabelValue)
-				keychain, err := ocmextensions.GetAuthnKeychain(ctx,
-					imageSpecWithCredSelect.CredSecretSelector, kcpClient)
-				Expect(err).ToNot(HaveOccurred())
-				dig := &TestRegistry{target: repo, registry: repo}
-				authenticator, err := keychain.Resolve(dig)
-				Expect(err).ToNot(HaveOccurred())
-				authConfig, err := authenticator.Authorization()
-				Expect(err).ToNot(HaveOccurred())
-				Expect(authConfig.Username).To(Equal("test_user"))
```
Tomasz-Smelcerz-SAP, line 40: Is there a replacement for this test? While it uses selector, it confirms that the resulting authConfig is correct.

tests/e2e/oci_registry_credentials_secret.go
lindnerby: Yes. As explained above the E2E testing must be done in a [separate issue](https://github.com/kyma-project/lifecycle-manager/issues/2541) because modulectl does not support pushing template-operator during module creation to a private registry yet. So it's blocked [by that](https://github.com/kyma-project/modulectl/issues/259).

Would you argue to delete this stub file?

lindnerby:
> A question: Previously we've used a list of secrets matching the configured selector. Now, we have just one secret, configured by name. Is it really how it should be implemented? I know that #2460 mentions "secret name", but it's parent issue: #1944 describes flag for configuring secret(s) selector...

Good point! Will clarify.

tests/integration/controller/manifest/keychain_test.go
lindnerby: Good catch! Seems I removed this accidentally. Will introduce a replacement 👌

lindnerby:
> > A question: Previously we've used a list of secrets matching the configured selector. Now, we have just one secret, configured by name. Is it really how it should be implemented? I know that #2460 mentions "secret name", but it's parent issue: #1944 describes flag for configuring secret(s) selector...
> 
> Good point! Will clarify.

As discussed: Previously the idea was to eventually have multiple registries for different modules. Since this got dropped on registry host is suffient.

unit-test-coverage.yaml
```diff
@@ -7,7 +7,6 @@ packages:
   internal/gatewaysecret/cabundle: 100
   internal/gatewaysecret/legacy: 100
   internal/istio: 63
-  internal/manifest: 15
```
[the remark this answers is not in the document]
lindnerby: Can we leave that for the manifest refactoring. Moving client to a new package will also result in many changes. Rather want to wrap up this PR.

docs/contributor/03-config-private-registry.md
```diff
@@ -16,56 +16,30 @@ Before you proceed, prepare your registry credentials. Check also how to deal wi
...
-1. Create a docker-registry Secret manifest. Run:
+1. Create a Docker Registry Secret resource definition:
 
    ```sh
    kubectl create secret docker-registry [secret name] --docker-server=[your oci registry host] --docker-username=[username] --docker-password=[password/token]  --dry-run=client -oyaml > registry_cred_secret.yaml
    ```
 
-2. Add the following labels to the Secret so that it can be configured later in the ModuleTemplate custom resource (CR) as a label selector.
-
+2. Adapt the Deployment of lifecycle-manager (KLM). Use the flag `--oci-registry-cred-secret` together with the value of your Secret name, to ensure that the private OCI registry is used. For example, if your Secret is named `my-private-oci-reg-creds`, your lifecycle-manager Deployment should contain the following container argument:
```
mmitoraj, line 27:
```suggestion
2. Adapt the Deployment of Lifecycle Manager (KLM). Use the flag `--oci-registry-cred-secret` together with the value of your Secret name to ensure that the private OCI registry is used. For example, if your Secret is named `my-private-oci-reg-creds`, your Lifecycle Manager Deployment must contain the following container argument:
```

docs/contributor/03-config-private-registry.md
```diff
@@ -16,56 +16,30 @@ Before you proceed, prepare your registry credentials. Check also how to deal wi
...
-3. Deploy the Secret on the same cluster where the ModuleTemplate CR is to be located. For example, if the ModuleTemplate CR is on the SKR cluster, then the Secret should be deployed to the SKR. Otherwise, it should be deployed to the KCP cluster.
-
-### Generate a ModuleTemplate CR with the `oci-registry-cred` Label
-
-The `oci-registry-cred` label in a ModuleTemplate CR allows Lifecycle Manager to parse the Secret label selector and propagate it to the Manifest CR so that Lifecycle Manager knows which credentials Secret to look up.
-
-To support the ModuleTemplate CR with the `oci-registry-cred` label, use modulectl with the `registry-cred-selector` flag for creating a module command.
-
-For example, you can run the following command to push your module image and generate a ModuleTemplate CR with the `oci-registry-cred` label:
+3. Deploy the Secret on the same cluster where KLM is deployed, so it should usually be deployed to the `kcp-system` namespace on the KCP cluster:s
```
mmitoraj, line 41:
```suggestion
3. Deploy the Secret in the same cluster where KLM is deployed. Usually, it is the `kcp-system` namespace in the KCP cluster.
```

docs/contributor/03-config-private-registry.md
```diff
@@ -16,56 +16,30 @@ Before you proceed, prepare your registry credentials. Check also how to deal wi
 
 ## Procedure
 
-### Prepare a docker-registry Secret Manifest
+### Prepare a Docker Registry Secret
```
mmitoraj, line 19:
```suggestion
```
The procedure is named in the main title, so I think we can skip this heading.

unit-test-coverage.yaml
Tomasz-Smelcerz-SAP, line 1: please add coverage for the new package: `internal/manifest/keychainprovider`

tests/e2e/oci_registry_credentials_secret.go
```diff
@@ -0,0 +1 @@
+package e2e
```
Tomasz-Smelcerz-SAP, line 1: Is the secret defined in `pkg/test_samples/auth_secret.yaml` a sample secret for this test? If so, maybe it's worth adding a comment here that the secret should be used in this test.

lindnerby: No, in the E2E scenario the secret needs to be created for the docker registry used in the k3d setup with something similar like:
```
kubectl create secret docker-registry my-private-registry-auth \
  --docker-server=http://my-private-registry:5000 \
  --docker-username=myuser \
  --docker-password=mypassword \
  --docker-email=dummy@example.com \
  -n kcp-system 
``` etc..
and then applied to the cluster, etc..

mmitoraj (approved):

PR #2510, opened 2025-06-11
feat: Re-introduce omitempty for SkipMaintenanceWindows

c-pius:
**Description**

- re-introduce omitempty for SkipMaintenanceWindows
  - the field is not synced by KLM anyway and this way the "UI" is kept lean

**Related issue(s)**


lindnerby:
“f the field has `omitempty` in its JSON tag, and is not a required pointer or slice, it’s treated as optional even without `+optional`. So no effect on the CRD or eventual validation.

PR #2512, opened 2025-06-12
fix: Add Network policy with Ingress Rule for runtime-watcher metrics

### LINKED ISSUE #2511: Add a NetworkPolicy to exclude the port for the runtime-watcher metrics
### Description

The runtime-watcher metrics are no longer being exposed after introducing the NetworkPolicies. We need to add a rule to allow the runtime-watcher metrics.

Here is the PR that introduced the network policies: https://github.com/kyma-project/lifecycle-manager/pull/2470

### Steps to reproduce

Setup Lifecycle manager with watcher, and try to access the metrics endpoint.

### Environment Type

Managed

### Environment Info

Kubernetes Version: x.y.z
Docker Version: x.y.z
modulectl Version: x.y.z

### Attachments

_No response_

nesmabadr:
**Description**

- Add Network policy with Ingress Rule for runtime-watcher metrics to allow exposing the runtime-watcher metrics

**Related issue(s)**
Resolves #2511


skr-webhook/resources.yaml
```diff
@@ -145,6 +145,24 @@ spec:
...
+metadata:
+  labels:
+    app: skr-webhook
+  name: kyma-project.io--metrics-to-watcher
+  namespace: default
+spec:
+  ingress:
+    - ports:
+        - port: 2112
+          protocol: TCP
```
lindnerby, line 157:
```suggestion
          protocol: TCP
    - from:
        - {}  # allow from all sources
```

lindnerby (commented):
Don't know how metrics endpoint is scraped, but if another pod is collecting it, this might be needed.

lindnerby (approved):

PR #2535, opened 2025-06-24
feat: Support module beta and internal sync

### LINKED ISSUE #2521: Support module beta and internal sync
**Description**
Both `internal`, `beta` are related to module versions. That means they should be kept as labels in ModuleTemplates.
 
 **ACs**
- [x] Only Kyma instances explicitly marked with `internal` or `beta` are allowed to sync the corresponding `ModuleTemplate` and `ModuleReleaseMeta`.
- [x] The existing logic for syncing `ModuleTemplate` needs to change
- [x] When syncing `ModuleReleaseMeta`, if a Kyma instance is not marked with `internal` or `beta`, any channel entries containing `internal` or `beta` must be excluded, so a modified version of MRM should be synced SKR without internal/beta entries. 

**Tests**
- E2E
- Also check integration tests

nesmabadr:
**Description**

- Adjust ModuleReleaseMeta sync logic to sync only MRMs which have entries whose Mts should be synced
- Adjust ModuleTemplates sync logic
- Adapt tests

**Related issue(s)**
Resolves #2521


internal/remote/remote_catalog.go
```diff
@@ -112,52 +122,60 @@ func (c *RemoteCatalog) Delete(
...
 	if err := c.kcpClient.List(ctx, moduleReleaseMetaList); err != nil {
 		return nil, fmt.Errorf("failed to list ModuleReleaseMetas: %w", err)
 	}
 
 	moduleReleaseMetas := []v1beta2.ModuleReleaseMeta{}
+
 	for _, moduleReleaseMeta := range moduleReleaseMetaList.Items {
-		if IsAllowedModuleReleaseMeta(moduleReleaseMeta, kyma) {
-			moduleReleaseMetas = append(moduleReleaseMetas, moduleReleaseMeta)
+		mrm := moduleReleaseMeta
```
LeelaChacha, line 139: why not use allowedChannels := []v1beta2.ChannelVersionAssignment{}. the mrm object is confusing

pkg/templatelookup/regular.go
```diff
@@ -129,16 +123,6 @@ func validateTemplateModeWithoutModuleReleaseMeta(template ModuleTemplateInfo, k
 	return template
 }
 
-func validateTemplateModeWithModuleReleaseMeta(template ModuleTemplateInfo, kyma *v1beta2.Kyma,
```
LeelaChacha, line 132: could you briefly explain why this validation is not needed anymore?

nesmabadr: because we now rely on the internal and beta labels directly from the ModuleTemplate itself not the ModuleReleaseMeta. Previously we had to check that the internal/beta fields on ModuleReleaseMeta against those in Kyma

PR #2543, opened 2025-06-27
feat: Block module deletion until all Module CRs are removed

### LINKED ISSUE #2343: Block module deletion until all ModuleCRs are
Do NOT refine until ADR is finally accepted.

### Description

When a module is deleted, KLM shall stale the deletion until all Module CRs are gone from the SKR. Only then, KLM shall continue with deletion of the module.
This should be applied for both, `CustomResourcePolicy: CreateAndDelete` and `CustomResourcePolicy: Ignore`.

### Reasons

Consequence of https://github.com/kyma-project/community/issues/972. In the current implementation of waiting for the default Module CR only, there are some edge cases that lead to a deadlock in the deletion.

### AC

It should be determined what are the significant flows to test. Not sure if all possible flows require and explicit E2E test.

- [ ] Given Module enabled with `CustomResourcePolicy: CreateAndDelete`, `[one|two|three]` Module CRs exist, module deletion will be blocked by another custom resource
  - [ ] When the module is disabled
    - [ ] Then the Manifest is marked for deletion
    - [ ] Then the default ModuleCR is marked for deletion
    - [ ] Then the Manifest continues to exist
    - [ ] Then module resources including module manager continue to exist
  - [ ] When the blocking CR is removed
    - [ ] Then the default ModuleCR is deleted
    - [ ] Then the Manifest continues to exist
    - [ ] Then module resources including module manager continue to exist
  - [ ] When the other ModuleCRs are deleted by the user
    - [ ] Then the module resource including module manager are deleted
    - [ ] Then the Manifest is deleted
- [ ] Given Module enabled with `CustomResourcePolicy: Ignore`, `[one|two|three]` Module CRs exist, module deletion will be blocked by another custom resource
  - [ ] When the module is disabled
    - [ ] Then the Manifest is marked for deletion
    - [ ] Then module resources including module manager continue to exist
    - [ ] Then the Manifest continues to exist
  - [ ] When the ModuleCRs are marked for deletion by the user
    - [ ] Then the ModuleCRs continue to exist (blocked by the other CR)
    - [ ] Then the Manifest continues to exist
  - [ ] When the blocking CR is removed
    - [ ] Then the ModuleCRs are deleted
    - [ ] Then module resources including module manager are deleted
    - [ ] Then the Manifest is deleted

nesmabadr:
**Description**

- Ensure all module CRs are removed before removing the manifest finalizers
- Use manifest annotation to handle the CustomResourcePolicy instead of the spec.resource and have the spec.resource always populated to be able to fetch the GVK of the module CR.

**Related issue(s)**
Resolves #2343


internal/manifest/modulecr/client.go
```diff
@@ -74,11 +77,29 @@ func (c *Client) CheckCRDeletion(ctx context.Context, manifestCR *v1beta2.Manife
 	return resourceCR == nil, nil
 }
 
-// RemoveModuleCR deletes the module CR if available in the cluster.
+func (c *Client) CheckModuleCRsDeletion(ctx context.Context, manifestCR *v1beta2.Manifest) (bool,
```
LeelaChacha, line 80: you don't need the bool. It is always true when error is nil, and false when there is error. Plus, the naming for method returning bool should begin with "is" or "are", so consider dropping it

internal/manifest/finalizer/handler.go
```diff
@@ -76,7 +76,7 @@ func EnsureCRFinalizer(ctx context.Context, kcp client.Client, manifest *v1beta2
 }
 
 func RemoveCRFinalizer(ctx context.Context, kcp client.Client, manifest *v1beta2.Manifest) error {
-	if manifest.Spec.Resource == nil {
+	if manifest.Spec.Resource == nil || manifest.Spec.CustomResourcePolicy == v1beta2.CustomResourcePolicyIgnore {
```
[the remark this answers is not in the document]
LeelaChacha: if you decide to remove it, do a usage search and consider removing those too. This check is used at a lot of places in the code.

pkg/testutils/manifest.go
```diff
@@ -237,7 +237,7 @@ func GetManifestResource(ctx context.Context,
 	if err != nil {
 		return nil, err
 	}
-	if moduleInCluster.Spec.Resource == nil {
+	if moduleInCluster.Spec.Resource == nil || moduleInCluster.Spec.CustomResourcePolicy == v1beta2.CustomResourcePolicyIgnore {
 		return nil, ErrManifestResourceIsNil
```
LeelaChacha, line 241: consider renaming this error

tests/e2e/utils_test.go
```diff
@@ -228,6 +228,32 @@ func CheckSampleCRIsInState(ctx context.Context, name, namespace string, clnt cl
 		expectedState)
 }
 
+func CreateSampleCR(ctx context.Context, name, namespace string, clnt client.Client) error {
+	sampleCR := &unstructured.Unstructured{}
```
LeelaChacha, line 232: this is more or less duplicate of pkg/testutils/modulecr.go file's NewTestModuleCR() function. consider using that and therefore, aligning with other e2e tests

LeelaChacha (changes requested):
Let me know what you think of the suggestions

tests/e2e/utils_test.go
LeelaChacha: could you still use the function: NewTestModuleCR, pretty please?

LeelaChacha (approved):
LGTM

Issue #2583, opened 2025-07-15
Documentation: Setting Up Kyma Control Plane Lifecycle Manager


c-pius:
**Description**

Create a "Setting Up Kyma Conrol Plane Lifecycle Manager" guide. It needs to take the perspective of configuring the landscaper-based deployment of KLM. To configure KLM, so called "data objects" are used by landscaper to populate the helm chart with values. The current data object of lifecycle manager is here: `/kyma/landscape-us50-canary/blob/main/data/kcp/lifecycle-manager.yaml`. It can be extended with values from the chart in managemnet-plane-charts: `/kyma/management-plane-charts/blob/chart/lifecycle-manager/lifecycle-manager/values.yaml` following the same structure.

Important remarks:

- images (e.g. watcher) and hosts are configured by SRE already, no need to change that
- focus on configuration aspects that are likely to be relevant for the operators for tuning their landscape (e.g., adapting to varying load), e.g.:
  - number of workers
  - requeue intervals
  - certificate durations
  - ...
- do not include very static configuration, e.g.
  - resource (gateway, secret) names
  - instead, refer to the general configuration doc
  - make sure that the ones that are usually not configured by the operator are defaulted to reasonable values
    - to be clarified with SRE: where should we default, in the KLM values file in managemnet-plane-charts or in the flags Go code?

Refer to the istio documentation for an example: `/kyma/restricted-markets/blob/main/docs/01-10-installation/istio.md`

Another template is here: `https://github.com/kyma-project/community/blob/main/templates/resources/configuration.md`

**Reasons**

Operators in NS2 need to understand how they can reasonably configure KLM.

**AC**

- [x] Setting Up Kyma Conrol Plane Lifecycle Manager doc available
  - [x] describes the KLM configuration params relevant for the operator
- [x] Doc reviewed by TW
- [x] Doc reviewed by SRE

**Assignees**

@kyma-project/technical-writers

**Attachments**

c-pius:
Raw input by SRE

```
defaulting is a relatively complex topic. Everything at some level should have an in-code default, ensuring that the service is functional if no configuration is given for some parts. This can either be that stuff is disabled (example would be something like your watcher), or working with a default configuration (here example would be a logging subsystem, works perfectly fine with defaults).
 
Kubernetes-level things obviously go to the chart, as it might not have any parts in the code.
 
If there are things which is specific to a restricted-market deployment through landscaper, that default should go to the product repository.
 
[...]

PR #2594, opened 2025-07-18
docs: Move the root readme content to docs

mmitoraj:
**Description**

- As a requirement for the restricted markets documentation, we must pull the content from the `/docs` folder of the respective repositories. That's why using the root README.md is not ok. Additionally, it includes links that are not necessary and misleading to operators (e.g., Code of Conduct, Contributing, etc.). In this PR, moving the main Lifecycle Manager description from the root README.md to a newly created `/docs/README.md`

**Related issue(s)**
See also https://github.com/kyma-project/lifecycle-manager/issues/2583


README.md
```diff
@@ -5,26 +5,17 @@
...
 
-If you are a Kyma end user, see the [user documentation](./docs/user/README.md).
+If you are a Kyma end user, see the [user documentation](/docs/user/README.md).
 
 ## Development
 
-If you want to provide new features for Lifecycle Manager, develop a module, or are part of the SRE team, visit the [contributor](/docs/contributor/) folder. You will find the following documents:
+If you want to provide new features for Lifecycle Manager, develop a module, or are part of the SRE team, visit the [contributor](/docs/contributor/) and the [operator](/docs/operator/operator-index.md) folders. You will find the following documents:
 
 * [Architecture](/docs/contributor/01-architecture.md)
```
c-pius, line 20: Shouldn't all these links be moved ot the `docs/README.md` as well?

mmitoraj: My goal was to make the `docs/README.md` as link-free as possible. On the one hand, the `/docs/README.md` file could provide more guidance on the content of the docs folder. But on the other hand, we need a doc with a pure Lifecycle Manager description with as few links as possible. Let me think, and I will ping you when I come up with something so solve that

PR #2644, opened 2025-08-06
chore: Bump golangci-lint and fix linter issues

lindnerby:
**Description**

- bump version to 2.3.1
- new linter `noinlinerr` produces quite a bit of a changelog, all occurences are fixed
**Related issue(s)**


internal/controller/kyma/controller.go
```diff
@@ -471,12 +494,14 @@ func (r *Reconciler) handleDeletingState(ctx context.Context, kyma *v1beta2.Kyma
 	logger := logf.FromContext(ctx).V(log.InfoLevel)
 
 	if r.WatcherEnabled() {
-		if err := r.SKRWebhookManager.Remove(ctx, kyma); err != nil {
+		err := r.SKRWebhookManager.Remove(ctx, kyma)
```
ruanxin, line 497: Why does this inlined error handling become not recommended?

lindnerby: There are two main reasons:
* First, readability: Many parts where this construct is used with longer functions and multiple return values make the code barely readable. Draws the attention back to the function and the error handling becomes more explicit and readable.
* Second: field shadowing. Due to this construct there are often a re-declarations of the err variable, which is unnecessary.

lindnerby:
As discussed in Architecture Round, we don't want the `noinlineerr` enable and consider it bad style to enforce not using it.
Closing this and opening another for the remaining linter introduction.

PR #2657, opened 2025-08-12
feat: Decouple the controller runtime from watcher runtime - listener integrartion

medmes:
**Description**
This PR is for the second part/ consumer part of decoupling from controller-runtime library in the runtime-warcher and listener.
It tackle the breaking changes regarding consuming the right new API contract.

-  Create SKR event service which is responsible wrapping the underlying listener package.
- Delegate the SKR Event Service lifecycle management to the controller-runtime manager.
- Added type conversions from the exposed listener Custom DTO to generic Event used by controller-runtme.
- Added unit Tests
- 

**Related issue(s)**
https://github.com/kyma-project/runtime-watcher/issues/130


Tomasz-Smelcerz-SAP (changes requested):
The biggest problem I found with this PR is that it doesn't work :(
Are watcher-related tests working? These must be working first using released versions/images.
Please take a look at #2697

Issue #2427, opened 2025-04-23
Investigate why `Check If Manifests Change` GH Action does not detect Kustomize changes


nesmabadr:
### Description

The `Check If Manifests Change` sometimes does not detect that there are any changes between main and the open PR, for example when the rbac rules get changed, it still does not see any differences.

### Steps to reproduce

This PR is an [example](https://github.com/kyma-project/lifecycle-manager/pull/2404). The rbac rules got changed here but the GitHub action printed `No differences in manifest outputs detected.` [here](https://github.com/kyma-project/lifecycle-manager/actions/runs/14446328314/job/40507762902?pr=2404).

### AC

**Timebox: 1 day**
- [x] Fix the bug

### Environment Type

Managed

### Environment Info

Kubernetes Version: x.y.z
Docker Version: x.y.z
modulectl Version: x.y.z
If used - k3d/k3s/minikube versions

PR #2660, opened 2025-08-12
fix: Manifest Change workflow is flakey

lindnerby:
**Description**

The reason why the workflow does not work sometimes is still not clear, but it can be related to caching of manifest files between the jobs.
This caching is not strictly required here, as we can generate and compare files in a single job. The additional benefit is that the job becomes simpler.

- Moved everything into a single job to avoid caching of files between subsequent jobs, to eliminate the scenario of cached files "leaking" between jobs in an uncontrolled way
- Using yq to sort the YAMLs before comparing them for more meaningful diffs
- Configuring specific yq version to be used

**Related issue(s)**
* #2427

### DIFF
--- a/.github/workflows/check-manifests-changes.yml
+++ b/.github/workflows/check-manifests-changes.yml
@@ -11,9 +11,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Install yq
-        run: |
-          sudo wget -qO /usr/local/bin/yq https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64
-          sudo chmod +x /usr/local/bin/yq
+        uses: ./lifecycle-manager/.github/actions/install-yq
+        with:
+          yq_version: 4.45.1
 
       - name: Checkout main branch
         uses: actions/checkout@v4


.github/workflows/check-manifests-changes.yml
```diff
@@ -7,21 +7,38 @@ on:
...
+          fi
+
+          if [[ ! -f ./dry-run/manifests.yaml ]]; then
+            echo "❌ Manifest file not found after make command"
+            exit 1
+          fi
+
           mkdir -p ./cache/pr
-          mv ./dry-run/manifests.yaml ./cache/pr/manifests.yaml
+          cp ./dry-run/manifests.yaml ./cache/pr/manifests.yaml
```
Tomasz-Smelcerz-SAP, line 40: Why the change? Why `cp` instead of `mv`?

.github/workflows/check-manifests-changes.yml
```diff
@@ -89,51 +156,83 @@ jobs:
...
+                "like management-plane-charts, runtime-watcher, etc.",
+                "",
+                "This PR modifies the generated manifests. Please review the changes carefully."
+              ].join('\n');
+
+              await github.rest.issues.createComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: context.payload.pull_request.number,
+                body: commentBody
```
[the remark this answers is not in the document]
LeelaChacha: that makes no sense

.github/workflows/check-manifests-changes.yml
lindnerby: To preserve the original:
The workflow pattern is:
- Generate manifest → ./dry-run/manifests.yaml
- Copy to job-specific cache → ./cache/pr/manifests.yaml or ./cache/main/manifests.yaml
- Cache the job-specific directory for use in the diff job

PR #2684, opened 2025-08-25
feat: Filter Out New Mandatory MRM Format Before Catalog Sync

LeelaChacha:
**Description**

- Filter Out New Mandatory MRM Format Before Catalog Sync
  - Unit test
  - E2E test

**Related issue(s)**

### DIFF
--- a/tests/e2e/mandatory_module_test.go
+++ b/tests/e2e/mandatory_module_test.go
@@ -46,12 +46,25 @@ var _ = Describe("Mandatory Module Installation and Deletion", Ordered, func() {
 						shared.EnableLabelValue).
 					Should(Succeed())
 			})
+			By("And the Mandatory ModuleReleaseMeta is configured correctly", func() {
+				Eventually(MandatoryModuleReleaseMetaHasVersion).
+					WithContext(ctx).
+					WithArguments(kcpClient, "template-operator", "1.1.0-smoke-test").
+					Should(Succeed())
+			})
 
 			By("And the mandatory ModuleTemplate is not synchronised to the SKR cluster", func() {
 				Consistently(CheckIfExists).
 					WithContext(ctx).
-					WithArguments("template-operator-mandatory", RemoteNamespace,
+					WithArguments("template-operator", RemoteNamespace,
 						shared.OperatorGroup, "v1beta2", string(shared.ModuleTemplateKind), skrClient).
+					Should(Not(Succeed()))
+			})
+			By("And the mandatory ModuleReleaseMeta is not synchronised to the SKR cluster", func() {
+				Consistently(CheckIfExists).
+					WithContext(ctx).
+					WithArguments("template-operator", RemoteNamespace,
+						shared.OperatorGroup, "v1beta2", string(shared.ModuleReleaseMetaKind), skrClient).
 					Should(Not(Succeed()))
 			})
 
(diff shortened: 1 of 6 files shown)


tests/e2e/mandatory_module_test.go
```diff
@@ -46,14 +46,27 @@ var _ = Describe("Mandatory Module Installation and Deletion", Ordered, func() {
...
 
 			By("And the mandatory ModuleTemplate is not synchronised to the SKR cluster", func() {
 				Consistently(CheckIfExists).
 					WithContext(ctx).
-					WithArguments("template-operator-mandatory", RemoteNamespace,
+					WithArguments("template-operator", RemoteNamespace,
 						shared.OperatorGroup, "v1beta2", string(shared.ModuleTemplateKind), skrClient).
 					Should(Not(Succeed()))
 			})
+			By("And the mandatory ModuleReleaseMeta is not synchronised to the SKR cluster", func() {
```
ruanxin, line 63: why you check twice?

ruanxin (dismissed):

tests/integration/controller/kyma/kyma_test.go
```diff
@@ -80,13 +80,13 @@ var _ = Describe("Kyma enable Mandatory Module or non-existent Module Kyma.Spec.
 			By("checking the state to be Warning in KCP", func() {
```
ruanxin, line 80: you need to change the statement then, it's not Warning anymore

pkg/templatelookup/moduletemplateinfolookup/by_version_strategy.go
```diff
@@ -63,6 +63,10 @@ func (s ByVersionStrategy) filterTemplatesByVersion(ctx context.Context, name, v
 	for _, template := range templateList.Items {
 		if TemplateNameMatch(&template,
 			name) && shared.NoneChannel.Equals(template.Spec.Channel) && template.Spec.Version == version {
+			if template.Spec.Mandatory {
```
ruanxin, line 66: I doubt we need to put this check in the strategy.

All these strategies can be reused for the mandatory module as well, right? But with your change, they are limited to regular module only.

I would move this filter after the lookup, specifically in the `GetRegularTemplates` [here](https://github.com/kyma-project/lifecycle-manager/blob/fc1d2b80e07144e4e0976a00df441ebc6c96b50e/pkg/templatelookup/regular.go#L72)

Also soon we will basically just keep that by_module_release_meta_strategy, and the rest will be deprecated.

ruanxin (approved):

ruanxin (changes requested):
the state need to be clarified.

internal/service/kyma/status/modules/generator/fromerror/generator.go
```diff
@@ -46,7 +47,7 @@ func GenerateModuleStatusFromError(err error, moduleName, desiredChannel, fqdn s
 	newStatus := newDefaultErrorStatus(moduleName, desiredChannel, fqdn, err)
 
 	if errorIsTemplateNotFound(err) {
-		newStatus.State = shared.StateWarning
+		newStatus.State = shared.StateError
```
ruanxin, line 50: We should not change the state of the existing one, there is no change request to ask us to change the not found error to `Error` state.

ruanxin (approved):

Issue #2573, opened 2025-07-10
Fix Handling SKR Deprovisioning


LeelaChacha:
## Description

**Follow-up from investigation in [kyma-project/lifecycle-manager#2249](https://github.com/kyma-project/lifecycle-manager/issues/2249)**

There are two related bugs:

1. The Kyma reconciler removes all finalizers and deletes the Kyma CR even if Manifest CRs still exist, instead of waiting for all Manifest CRs to be deleted.
 https://github.com/kyma-project/lifecycle-manager/blob/8c1a35ccaf731be499b67dc7366643c21fee05a3/internal/controller/kyma/controller.go#L222-L233
2. The Manifest reconciler does not properly invalidate its client cache and therefore does not remove all finalizers in the case of mandatory modules without a default CR, especially when DNS errors or missing kubeconfig secrets occur. This can leave Manifest CRs stuck in an Error state and orphaned.
https://github.com/kyma-project/lifecycle-manager/blob/da32ce0482ead769d1e81c37c68b0df75368499b/internal/declarative/v2/reconciler.go#L173-L181
In this code, `r.renderResources(ctx, skrClient, manifest, spec)` does not return an error if defaultCR is nil. Therefore, it can be relied upon to invalidate the cache.
Since the cache entry is still valid, the following does not get executed even if the kubeconfig for the SKR has been deprovisioned:
https://github.com/kyma-project/lifecycle-manager/blob/da32ce0482ead769d1e81c37c68b0df75368499b/internal/declarative/v2/reconciler.go#L118-L122

## Steps to reproduce

Given a local E2E setup:

0. Deploy a Kyma CR and one module **without a default CR**.
1. Use the script: `scripts/tests/remove_skr_host_from_coredns.sh` and then remove the SKR cluster completely.
2. Wait until the Manifest CR is in the Error state.
3. Delete the `kyma-sample` secret in the KCP cluster and mark the Kyma CR for deletion.

**Result:**  
The Kyma reconciler removes all finalizers on the Kyma resource and the Kyma CR is deleted, even though the Manifest CR is still present and stuck in the Error state. Same result in managed environments.

**Expected:**  
The Manifest reconciler should remove all resource finalizers and the Manifest CR should be deleted. The Kyma CR should not be deleted until all Manifest CRs are gone.

## Acceptance Criteria

[...]

PR #2695, opened 2025-08-28
fix: Fix Handling SKR Deprovisioning

ruanxin:
**Description**

- extract `skrclientcache.NewService`, `skrclient.NewService`: https://github.com/kyma-project/lifecycle-manager/pull/2695/files#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R462-R475 
- refactored some declarative options to simplify code
- implement new solution to evict cache faster, check my comment: https://github.com/kyma-project/lifecycle-manager/pull/2695/files#r2315449068

**Related issue(s)**
https://github.com/kyma-project/lifecycle-manager/issues/2573


internal/declarative/v2/reconciler.go
```diff
@@ -184,6 +199,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		resources.ErrDeletionNotFinished) {
 		r.ManifestMetrics.RecordRequeueReason(metrics.ManifestPruneDiffNotFinished, queue.IntendedRequeue)
 		return ctrl.Result{Requeue: true}, nil
+	} else if util.IsConnectionRelatedError(err) {
```
ruanxin, line 202: To Reviewer:

During refinement we previously discussed the TTL cache, but that approach turned out not to be a solid solution. In this PR, I did not make any changes to the TTL cache.

As @LeelaChacha explained in the original issue, the actual root cause is that when no default CR is returned, renderResources does not throw an error. This prevents KLM from fetching SKR, meaning IsConnectionRelatedError is never triggered, and the client cache is not invalidated.

The fix introduced in this PR is to add a connection-related error check after pruneDiff. At this point, a request to SKR is guaranteed, since DeleteDiffResources is called when the manifest CR is being deleted. If SKR has already been removed, IsConnectionRelatedError will correctly return true, causing the client cache to be evicted. On the next reconcile, the expected ErrAccessSecretNotFound will then be raised.

tests/e2e/no_default_cr_test.go
```diff
@@ -57,5 +59,84 @@ var _ = Describe("Module Without Default CR", Ordered, func() {
 				WithArguments(kyma.GetName(), kyma.GetNamespace(), kcpClient, shared.StateReady).
 				Should(Succeed())
 		})
+
+		It("When a blocking finalizer is added to the Manifest CR", func() {
```
ruanxin, line 63: To Reviewer:

This is the e2e to test no default CR module gets deleted successfully when access secret get deleted.

Tomasz-Smelcerz-SAP (changes requested):
This PR is too big to be reviewed IMO.
Please split this into more PRs:
- One that contains just the minimal changes necessary to fulfill the AC criteria,
- Second one that does not change any logic/behavior, but is just a refactoring of names, moving files etc
- Third one for Options refactoring

internal/declarative/v2/client.go
Tomasz-Smelcerz-SAP, line 1: This file content's (`Client` interface) was moved into: `internal/service/skrclient/skr_client.go`

internal/declarative/v2/client_cache.go
Tomasz-Smelcerz-SAP, line 1: This file was moved (but GH doesn't detect it) to: `internal/service/skrclient/cache/client_cache.go`

internal/declarative/v2/util.go
Tomasz-Smelcerz-SAP, line 1: the `getResourceMapping` function was moved to: `internal/service/skrclient/skr_client.go`

internal/manifest/skr_client_lookup.go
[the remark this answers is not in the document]
ruanxin: this ConfigResolver is trying to use manifest CR to get restConfig from access secert.
the ConfigGetter is never being assigned currently, so it will eventually alway use 
`config, err := (&ClusterClient{DefaultClient: r.KCP.Client}).GetRESTConfig()` to get restconfig, and this being replaced by [GetAccessRestConfigByKyma](https://github.com/kyma-project/lifecycle-manager/pull/2695/files#diff-fc494532da185761341185cd16be5484fe4a2aafd4b225dd59b19f03569b837dR56)

internal/declarative/v2/reconciler.go
```diff
@@ -485,47 +516,21 @@ func pruneResource(diff []*resource.Info, resourceType string, resourceName stri
...
 		if err != nil {
 			return nil, err
 		}
-		r.AddClient(clientsCacheKey, clnt)
-	}
-
-	return clnt, nil
-}
-
-func (r *Reconciler) configClient(ctx context.Context, manifest *v1beta2.Manifest) (Client, error) {
```
ruanxin, line 511: To Reviewer:

This `configClient` has been replaced by `skrClient.ResolveClient(ctx, manifest)`, the related `ClusterInfo`, `TargetCluster` can be dropped completely.

internal/repository/secret/secret_repo_test.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	clientStub := &listClientStub{
+		list: &apicorev1.SecretList{Items: []apicorev1.Secret{*secret1, *secret2}},
+	}
+	secretRepository := secretrepository.NewRepository(clientStub, namespace)
+
+	result, err := secretRepository.List(t.Context(), k8slabels.SelectorFromSet(k8slabels.Set{"app": "test"}))
+
+	require.NoError(t, err)
+	assert.Len(t, result.Items, 2)
```
Tomasz-Smelcerz, line 46: This test looks weird to me. It looks like the intention is to find just the first object, but here we're asserting two are returned... Making assertions on what the stub returns makes no sense, as the stub is not running on production so it's behavior is irrelevant with respect to code correctness.
In my opinion we should rather extend the "called" property of the stub with additional properties that allow us to assert if the stub was called with the expected values.
That is: label selectors and namespace.
This provides us with some guarantees that the real client, in production, will receive correct arguments.
What do you think?

internal/service/accessmanager/manager.go
```diff
@@ -0,0 +1,66 @@
...
+			errors.Join(err, ErrAccessSecretNotFound))
+	}
+
+	if len(kubeConfigSecretList.Items) < 1 {
+		return nil, fmt.Errorf("secret with label %s=%s %w", shared.KymaName, kymaName, ErrAccessSecretNotFound)
+	}
+
+	if len(kubeConfigSecretList.Items) > 1 {
+		return nil, fmt.Errorf(
+			"could not safely identify the rest config source: %w", ErrMoreThanOneSecretFound)
```
Tomasz-Smelcerz, line 51: Error message is a bit cryptic. Maybe just the same message as the previous one, as these conditions are kind-of related:
`secret with label %s=%s %w` ?

internal/service/accessmanager/manager.go
```diff
@@ -0,0 +1,66 @@
...
+func (s Service) GetAccessSecretByKyma(ctx context.Context, kymaName string) (*apicorev1.Secret, error) {
+	kubeConfigSecretList, err := s.secretRepository.List(ctx,
+		k8slabels.SelectorFromSet(k8slabels.Set{shared.KymaName: kymaName}))
+	if err != nil {
+		return nil, fmt.Errorf("failed to list secrets with label %s=%s: %w", shared.KymaName, kymaName,
+			errors.Join(err, ErrAccessSecretNotFound))
+	}
+
+	if len(kubeConfigSecretList.Items) < 1 {
+		return nil, fmt.Errorf("secret with label %s=%s %w", shared.KymaName, kymaName, ErrAccessSecretNotFound)
```
Tomasz-Smelcerz, line 46:
```suggestion
		return nil, fmt.Errorf("failed to find the secret with label %s=%s %w", shared.KymaName, kymaName, ErrAccessSecretNotFound)
```

internal/service/skrclient/cache/client_cache.go
```diff
@@ -0,0 +1,66 @@
...
+func (m *Service) Size() int {
+	return m.internal.Len()
+}
+
+func getRandomTTL() time.Duration {
+	randomRange, _ := rand.Int(rand.Reader, big.NewInt(int64(ttlInSecondsUpper-ttlInSecondsLower)))
+	return time.Duration(randomRange.Int64()+int64(ttlInSecondsLower)) * time.Second
+}
+
+func (m *Service) GetCacheKey(manifest *v1beta2.Manifest) (string, bool) {
```
Tomasz-Smelcerz, line 55: This package is a very nicely isolated little abstraction, but this function breaks the encapsulation.
It strictly refers to some domain type and looks "artificial" here. No other function, for example refers to `v1beta2` or `Manifest`.
I would move this function somewhere else. As it's used only in `internal/declarative/v2/reconciler.go`, probably it should go there.

ruanxin: good point, but I would relocate it as a member function of manifest instead: https://github.com/kyma-project/lifecycle-manager/pull/2695/files#diff-c6cbfcbcd1b11c7ec09726e9db9818f50faeff22c569ab3c90fe4af843a13c01R200

internal/repository/secret/secret_repo_test.go
ruanxin: basically it's generated by copilot, I think the intention is not find just the first object, but first ensure the List return the items amount is as expected, second ensure the returned item is ordered correctly, the first item is `secret1` instead of `secret2`. So I don't think this assertion is wrong.

I agreed with you it can be extended, I asked copilot again for the refactoring, please check.

Tomasz-Smelcerz (approved):

PR #2697, opened 2025-08-29
Adjust to changes in listener/pkg/v2

Tomasz-Smelcerz-SAP:
**Description**

- ...
- ...
- ...

**Related issue(s)**


c-pius (dismissed):
LGTM. One question about the `watcherEventAdapter`. Is it temporary until things are made pretty or going to stay? Should we add the `internal/controller` pkg to the unit test coverage thresholds?

And I really appreciate your detailed PR descriptions and change summaries. Made it super easy to follow through the preceding changes in runtime-watcher leading up to this PR 👍🏻 
CC: @jacekon

Tomasz-Smelcerz-SAP:
> One question about the `watcherEventAdapter`. Is it temporary until things are made pretty or going to stay? Should we add the `internal/controller` pkg to the unit test coverage thresholds?

We can do this, yes. But it has 100% now, so it can only get worse over time :)
But seriously - I think it is sort-of temporary and awaits a bigger refactoring like the one introduced in https://github.com/kyma-project/lifecycle-manager/pull/2657
Feel free to suggest a "better place" (a dedicated package?) for this thing.

c-pius (approved):
> We can do this, yes. But it has 100% now, so it can only get worse over time :)

That's the point of the action 😄 

> But seriously - I think it is sort-of temporary and will be replaced by a bigger refactoring like the one introduced in https://github.com/kyma-project/lifecycle-manager/pull/2657
Feel free to suggest a "better place" (a dedicated package?) for this thing - until the "big refactoring" comes.

No it is okay for me already as indicated by my previous approval. Especially now since you mentioned that this will be refactored soon anyway.

Issue #995, opened 2023-10-31
Remove the fallback to use generation in ComponentDescriptorCache key


nesmabadr:
**Description**

Currently we use the module version from the annotation added in the ModuleTemplate as part of the ComponentDescriptorCache key. If the annotation is not found we fall back to using the generation instead. After all ModuleTemplates have the module version annotation, we need to remove this fallback logic.

**ACs:**

- [ ] Remove the fallback logic to use generation as part of the componentDescriptorCache key once all ModuleTemplates have the module version annotation

janmedrek:
Related to versions in the Module Catalogue: https://github.com/kyma-project/lifecycle-manager/issues/1472

c-pius:
Do we need to block this until `once all ModuleTemplates have the module version assigned in the CR ` is true?

Issue #2601, opened 2025-07-22
Read component descriptor from oci registry


ruanxin:
### Description

With #2634 and #2597 finished, KLM is able to construct ocm remote address without depending on ModuleTemplate, and fetch component descriptor remotely.

Going forward, we will get:

- OCI registry host from 
  - either the flag introduced in: #2597 OR
  - the existing secret configured via `oci-registry-credential-secret` flag
- the component name and version from ModuleReleaseMeta (after #2634 is implemented) 
  - `.spec.ocmComponentName` and `.spec.channels[].version`
  - OR for mandatory modules: `.spec.ocmComponentName` and `.spec.mandatory.version`

### DEV Note

- right now we are taking the descriptor directly from the template: https://github.com/kyma-project/lifecycle-manager/blob/e7efceb4af3833ccd7e25a5f02eba12b260c3d14/internal/descriptor/provider/provider.go#L30-L67
  - this needs to be refactored to look it up from the registry with the information described above
- `GetDescriptor` is used by regular Kyma reconciler + mandatory installation and deletion reconcilers already
  - the signature will need to be changed as it currently accepts a ModuleTemplate

### Reasons

unblock deprecating component descriptor from ModuleTemplate.

### Acceptance Criteria

- [ ] Refactor `CachedDescriptorProvider`, make sure it fetches and caches component descriptor from remote oci registry instead of from ModuleTemplate.

Tomasz-Smelcerz-SAP:
We are blocked here by: <internal-repo>/kyma/test-infra/issues/781 - we need all the ModuleReleaseMeta to have OCM version related attributes. In particular spec.mandatory.version

PR #2729, opened 2025-09-09
chore: Refactor descriptor cache key generation

ruanxin:
The pull request "chore: Refactor descriptor cache key generation" refactors the cache key generation process by moving it into the domain model and simplifying the interfaces. It also improves error handling for version scenarios and enhances test coverage. The changes include:

- Removal of the DescriptorKey type and its associated functions.
- Simplification of cache key generation by using a string format directly.
- Updates to the DescriptorCache and CachedDescriptorProvider to use the new key generation approach.
- Adjustments to tests to align with the refactored key generation logic.
- Removal of legacy code related to semantic version handling.

This PR addresses issue #995

### DIFF
--- a/unit-test-coverage.yaml
+++ b/unit-test-coverage.yaml
@@ -2,7 +2,7 @@ packages:
   internal/controller: 100
   internal/controller/istiogatewaysecret: 28
   internal/crd: 92
-  internal/descriptor/cache: 92
+  internal/descriptor/cache: 88
   internal/descriptor/provider: 66
   internal/event: 100
   internal/gatewaysecret/cabundle: 100


api/v1beta2/moduletemplate_types_test.go
```diff
@@ -11,101 +10,6 @@ import (
 	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
 )
 
-func Test_GetSemanticVersion(t *testing.T) {
```
ruanxin, line 14: To Reviewer:

This GetSemanticVersion can be dropped since the `spec.version` validation is done on the CRD level, and we are not using annotation anymore.

internal/descriptor/provider/provider.go
```diff
@@ -44,7 +45,8 @@ func (c *CachedDescriptorProvider) GetDescriptor(template *v1beta2.ModuleTemplat
 
 		return desc, nil
 	}
-	key := cache.GenerateDescriptorKey(template)
+	key := c.GenerateDescriptorKey(template.Name, template.GetVersion())
```
ruanxin, line 48: To Reviewer:

The parameter should be refactored in this issue https://github.com/kyma-project/lifecycle-manager/issues/2601, not from template, instead with ocmComponentName and ocm version

Issue #2526, opened 2025-06-17
Deprecate Component Descriptor from ModuleTemplate


ruanxin:
Created on 2024-06-17 by Xin Ruan (@ruanxin)

## Decision log

| Name               | Description            |
| ------------------ | ---------------------- |
| Title              | Deprecate Component Descriptor from ModuleTemplate |
| Due date           | 2024-07-30             |
| Status             | Proposed on 2024-06-18 |
| Decision type      | Binary                 |
| Affected decisions | -                      |

## Context
Currently, the context in `ModuleTemplate` serves two purposes:

1. Provide metadata for client tooling (kyma cli, dashboard UI)
2. Persist Component Descriptor from OCM to indicate KLM download module resources  

This approach creates issues in restricted markets where container images are localized. Because the image locations change, the original ModuleTemplate becomes invalid. This forces us to generate a new, localized ModuleTemplate from the localized OCM, adding an extra step to the process.

## Decision:
We will deprecate Component Descriptor from ModuleTemplate.

Since KLM is the only component that uses this descriptor for downloading resources, we can simplify the process by configuring KLM to read this information directly from the OCM registry. This change would make the ModuleTemplate a generic resource that doesn't require localization.

## Consequences
✅ Benefits
- Simplify module localisation efforts
  ModuleTemplate becomes a generic component without need to be localized, eliminating the need to convert OCM to ModuleTemplate.

- Reduced redundancy
  Avoids duplication of metadata between ModuleTemplate and OCM.

⚠️ Trade-offs
- For security scanning, the scanning tool scan directly from OCM component metadata, not from ModuleTemplate copy anymore.
- For KLM, the ocm registry should provided from alternative approach, but can't be relies on ModuleTemplate. #2597

PR #2736, opened 2025-09-15
feat: Read component descriptor from oci registry

Tomasz-Smelcerz-SAP:
**Description**

**Note to reviewers**: I was *really* trying to keep this PR as minimal as possible. There are many things to fix/improve/cleanup and many times I've seen that, but I decided not to, in order to *introduce as few changes as possible*.
Keep that in mind when suggesting improvements. Do we really want to make this PR *even bigger*? I would rather prefer to make a note in the sources and then create a follow-up PR to clean all the notes. Just remember that the work remaining in https://github.com/kyma-project/lifecycle-manager/issues/2526 and our other efforts (like simplifying/getting rid of `templatelookup.go`) will probably make most of these review-fixes obsolete or unnecessary.
Up to you. 

- Component Descriptors (`CD`) are now read from the OCM repository (which usually is some OCI registry)
- To read a `CD`, one must provide an OCM Component Name and a version. In our current setup, the OCM Component Name is available **only** in a `ModuleReleaseMeta` for a given Module.
- It means that with this PR the `ModuleReleaseMeta` becomes **mandatory**
- To keep this PR minimal, I did not remove code/tests related to the previous "modes" of operation, where `CD` was fetched from a `ModuleTemplate` and `ModuleReleaseMeta` was, in some cases, optional. This code requires cleanup in a separate PR. Of course when the test is failing, I had to refactor it
- An interesting case: What to do if user configures a Module for which there is no `ModuleReleaseMeta`? Currently, when user configures a module on a Kyma, for which there's no ModuleTemplate, the Kyma is put in `Warning` state. To keep it consistent, I use the same error->status resolution strategy. To be discussed.
- Code coverage for some packages (`templatelookup`) decreased. This is because some test cases were disabled. This is (mostly) a legacy code that will disappear anyway. A new method of resolving ModuleTemplates along with ModuleReleaseMeta should be introduced, but this is not a part of **this** PR. I think it is OK to temporarily "lower the bar" - we're in the middle of a bigger refactoring here. Related issues that will introduce the necessary cleanpup/changes are listed are listed [here](https://github.com/kyma-project/lifecycle-manager/issues/2526)
- There is an [open discussion](https://github.com/kyma-project/lifecycle-manager/discussions/2746) that focuses on the little detail: The OCI registry URL. Previously it was taken from the static ComponentDescriptor data embedded in the ModuleTemplate. Now we have at least two different values we can use: An explicitly configured one and the one from ComponentDescriptor, after it is successfully fetched (we still need OCI Registry URL for subsequent operations, like raw-manifest fetching). Which URL to choose? Please take a look at the discussion. I decided to rely on explicitly provided value as I had no feedback that I should chose a different strategy here.

**Related issue(s)**
https://github.com/kyma-project/lifecycle-manager/issues/2601

### DIFF
--- a/internal/imagerewrite/rewrite.go
+++ b/internal/imagerewrite/rewrite.go
@@ -37,7 +37,7 @@ type DockerImageReference struct {
 
 // Regex matches prefix@sha256:checksum, prefix:version, and prefix:version@sha256:checksum.
 var ociImagePattern = regexp.MustCompile(
-	`^(?:(?P<host>[\w.-]+(?::\d+)?(?:/[\w-]+)*)/)?` +
+	`^(?:(?P<host>[\w.-]+(?::\d+)?(?:/[\w.-]+)*)/)?` +
 		`(?P<name>[a-z0-9]+(?:[_-][a-z0-9]+)*)` +
 		`(?:` +
 		`(?::(?P<tag>[\w.\-]+))(?:@(?P<digest>sha256:[a-fA-F0-9]{64}))?|` +
--- a/internal/imagerewrite/rewrite_test.go
+++ b/internal/imagerewrite/rewrite_test.go
@@ -29,6 +29,7 @@ func TestDockerImageReference(t *testing.T) {
 			"localhost:5111/myimage:5.4.3",
 			"localhost:5111/myimage:5.4.3@sha256:f9f4a45fe9091a8e55b55b80241c522b45a66501703728d386dc4171f70af803",
 			"localhost:5111/myimage@sha256:f9f4a45fe9091a8e55b55b80241c522b45a66501703728d386dc4171f70af803",
+			"example.pkg.io/example-project/prod/external/gcr.io/kaniko-project/executor:v1.24.0",
 		})
 
 		// then
@@ -111,6 +112,12 @@ func TestDockerImageReference(t *testing.T) {
 				expectedHostAndPath: "localhost:5111",
 				expectedNameAndTag:  "myimage",
 				expectedDigest:      "sha256:f9f4a45fe9091a8e55b55b80241c522b45a66501703728d386dc4171f70af803",
+			},
+			{
+				name:                "Image with paths are allowed and dots in path",
+				expectedHostAndPath: "example.pkg.io/example-project/prod/external/gcr.io/kaniko-project",
+				expectedNameAndTag:  "executor:v1.24.0",
+				expectedDigest:      "",
 			},
 		}
 


pkg/testutils/moduletemplate.go
```diff
@@ -32,7 +31,7 @@ func GetModuleTemplate(ctx context.Context,
 	clnt client.Client,
 	module v1beta2.Module,
 	kyma *v1beta2.Kyma,
-) (*v1beta2.ModuleTemplate, error) {
+) (*templatelookup.ModuleTemplateInfo, error) {
```
ruanxin, line 34: Please don't change this return type, if return `emplatelookup.ModuleTemplateInfo`, you can't use it for ModuleTemplate update. It will cause error in `SetModuleTemplateBetaLabel`, this function is expected to return a real `ModuleTemplate`, then do `clnt.Update(ctx, moduleTemplate)`

```
SetModuleTemplateBetaLabel attempt failed: failed to update module template: no kind is registered for the type templatelookup.ModuleTemplateInfo in scheme "pkg/runtime/scheme.go:110" 
```

This function is `GetModuleTemplate`, the expected returned type is `ModuleTemplate`, create another function if you want a `ModuleTemplateInfo`

internal/descriptor/provider/provider.go
```diff
@@ -1,109 +1,90 @@
...
-		desc, ok := template.Spec.Descriptor.Object.(*types.Descriptor)
-		if !ok {
-			return nil, ErrTypeAssert
-		}
+// Convenience interface to get the OCM identity of a component from objects
+// that already have all required data.
+// Then we don't have to create intermediate variables of type ocmidentity.Component.
+type OCMIProvider interface {
+	GetOCMIdentity() (*ocmidentity.Component, error)
+}
```
c-pius: Wondering if the `ModuleTemplateInfo` could also directly implement the `ocmidentity.Component` interface (`Name()`, `Version()`). Then we wouldn't need this interface and could directly pass the `ModuleTemplateInfo`. But the function names are rather generic, so also not too sure if this is a good idea.

But if we keep the interface, could we move the type up to before the constructor? I think it is easier to read if the file is ordered like "types -> vars -> constructor -> funcs"

internal/descriptor/provider/provider.go
```diff
@@ -1,109 +1,90 @@
...
 
-		if desc == nil {
-			return nil, ErrDescriptorNil
-		}
+func (c *CachedDescriptorProvider) Add(ocmi ocmidentity.Component) error {
```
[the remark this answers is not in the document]
c-pius: Good catch actually. I am a bit confused about the usage and exact implementation of the provider though. In `regular.go`.

For the same module (ocmi), we always first add (updateCache: true), and then get (updateCache:false). Why not only doing a get that is updating the cache in the beginning and get rid of Add? Also, I think the `updateCache` bool is also misleading. I would expect that this forces the cache to update. But instead, it is only controlling if the cache is updated AFTER not finding the descriptor in the cache. Not sure what the purpose is for that?

```go
for _, moduleInfo := range FetchModuleInfo(kyma) {
       // ...
       if err := t.descriptorProvider.Add(*ocmi); err != nil {
			templateInfo.Err = fmt.Errorf("failed to get descriptor: %w", err)
			templates[moduleInfo.Name] = &templateInfo
			continue
		}
		for i := range kyma.Status.Modules {
			moduleStatus := &kyma.Status.Modules[i]
			if moduleMatch(moduleStatus, moduleInfo.Name) {
				descriptor, err := t.descriptorProvider.GetDescriptor(*ocmi)
				if err != nil {
					msg := "could not handle channel skew as descriptor from template cannot be fetched"
					templateInfo.Err = fmt.Errorf("%w: %s", ErrTemplateUpdateNotAllowed, msg)
					continue
				}
				markInvalidSkewUpdate(ctx, &templateInfo, moduleStatus, descriptor.Version)
			}
		}
```

internal/manifest/parser/template_to_module.go
```diff
@@ -203,27 +207,44 @@ func getLocalizedImagesFromDescriptor(descriptor *types.Descriptor) []string {
...
+		// instead of the one from the layer (it is the same as in the ComponentDescriptor).
+		// These two values may be different and the explicitly configured one is safer to use,
+		// as it is known to be reachable.
+		// After all, we've been able to read the ComponentDescriptor using it.
+		ociImageCopy := img.OCI{
+			Repo: ociRepoFromConfig,
+			Name: ociImage.Name,
+			Ref:  ociImage.Ref,
+			Type: ociImage.Type,
+		}
```
c-pius: Don't understand the need for it. In `imageSpec, err := layer.ConvertToImageSpec(ociRepoFromConfig)`, we already overwrite the repo part, why creating another copy where we again set the repo explicitly?

internal/repository/modulereleasemeta/repository.go
```diff
@@ -0,0 +1,32 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+)
+
+type Repository struct {
+	client.Reader
+}
+
+func NewRepository(reader client.Reader) *Repository {
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Indeed I think It's a leftover from development. I will double-check and remove it. Thanks for spotting!

internal/repository/oci/ocirepo.go
```diff
@@ -0,0 +1,121 @@
...
+
+type defaultCraneWrapper struct{}
+
+func (c *defaultCraneWrapper) Config(ref string, opt ...crane.Option) ([]byte, error) {
+	return crane.Config(ref, opt...) //nolint:wrapcheck // the crane wrapper should be transparent
+}
+
+func (c *defaultCraneWrapper) PullLayer(ref string, opt ...crane.Option) (containerregistryv1.Layer, error) {
+	return crane.PullLayer(ref, opt...) //nolint:wrapcheck // the crane wrapper should be transparent
+}
```
c-pius: Why not moving this out of here to a normal structing implemetning a public interface defined in ocirepo.go? Then we just inject it into the constructor as usually and don't need the additional `export_test.go`?

internal/service/componentdescriptor/service.go
```diff
@@ -0,0 +1,223 @@
...
+			fmt.Errorf("failed to unmarshal config data into ComponentDescriptorConfig for %s: %w",
+				commonErrMsg(ocmi), err)
+	}
+
+	if ocmArtifactConfig.ComponentDescriptorLayer == nil {
+		return nil, fmt.Errorf("%w for %s", ErrLayerNil, commonErrMsg(ocmi))
+	}
+
+	compDescLayerDigest := ocmArtifactConfig.ComponentDescriptorLayer.Digest
+	if string(compDescLayerDigest) == "" {
```
c-pius: is the type cast needed?

internal/service/componentdescriptor/service.go
```diff
@@ -0,0 +1,223 @@
...
+	ErrNotFoundInTar    = errors.New("not found in TAR archive")
+	ErrTarTooLarge      = errors.New("entry in the TAR archive is too large")
+	ErrLayerEmpty       = errors.New("layer content is empty")
+	ErrDecode           = errors.New("failed to decode component descriptor")
+)
+
+type OCIRepository interface {
+	GetConfigFile(ctx context.Context, name, tag string) ([]byte, error)
+	PullLayer(ctx context.Context, name, tag, digest string) (containerregistryv1.Layer, error)
+}
```
c-pius: How about passing the `ocmidentity.Component` directly instead of separated name and tag (version)

c-pius: Digest could also use the digest type. In the call below, the typecast should not be needed anyway.

internal/service/componentdescriptor/service.go
```diff
@@ -0,0 +1,223 @@
...
+			ErrDecode, commonErrMsg(ocmi), err)
+	}
+	return desc, nil
+}
+
+// SimplifiedLayer represents only necessary part of containerregistryv1.Layer to simplify testing.
+type SimplifiedLayer interface {
+	Digest() (containerregistryv1.Hash, error)
+	Uncompressed() (io.ReadCloser, error)
+}
```
c-pius: This should not be needed. We can have `func extractFile(ioh extractFileIOHelper, layer containerregistryv1.Layer, fileName string) ([]byte, error) {` with mock layer

```go
type mockLayer struct {
	containerregistryv1.Layer

	errOnUncompressed error
	errOnDigest       error
}
```

Needed Uncompressed() and Digest() are overriden. Rest are implicitly exposed but will not work since we didn't provide an actual Layer

internal/service/componentdescriptor/service.go
```diff
@@ -0,0 +1,223 @@
...
+	ReadAll(r io.Reader) ([]byte, error) // Reads all data from the reader
+}
+
+type defaultExtractFileIOHelper struct {
+	readAllFunc func(r io.Reader) ([]byte, error)
+}
+
+func (d *defaultExtractFileIOHelper) ReadAll(r io.Reader) ([]byte, error) {
+	return d.readAllFunc(r)
+}
```
c-pius: same comment as for the wrapper in ocirepo.go. If we don't decide to go that we, I would at least name it consistently with Wrapper or Helper because eventually it is the same thing.

pkg/templatelookup/regular.go
```diff
@@ -11,19 +11,34 @@ import (
...
+	//                                          represented by this ModuleTemplateInfo.
+}
+
+// Implements provider.OCMIProvider interface.
+func (m ModuleTemplateInfo) GetOCMIdentity() (*ocmidentity.Component, error) {
+	if m.ComponentIdentity == nil {
+		return nil, fmt.Errorf("%w for module template %s", ErrNoIdentity, m.Name)
+	}
+	return m.ComponentIdentity, nil
 }
```
c-pius: See comment in provider.go

internal/service/componentdescriptor/service.go
Tomasz-Smelcerz-SAP: > How about passing the `ocmidentity.Component` directly instead of separated name and tag (version)

I was thinking about it, but it is "polluting" OCIRepository with an abstraction that doesn't belong there (OCM is not something that OCI Repository "knows" about). We could have a dedicated type that represents <name, tag, digest> and do the conversion. Because this repository is very simple, I decided for just strings. But I am always in favor of having a dedicated type.
Should I introduce it?
Or we could use the fact that the Component type offers methods: Name() and Version() - we could accept an interface that matches these methods.

cmd/main.go
```diff
@@ -221,21 +224,45 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
...
+
+	ociRegistryHost := getOciRegistryHost(mgr.GetConfig(), flagVar, logger)
+	var insecure bool
+
+	if noSchemeRef, found := strings.CutPrefix(ociRegistryHost, "http://"); found {
+		insecure = true
+		ociRegistryHost = noSchemeRef
+	} else if noSchemeRef, found := strings.CutPrefix(ociRegistryHost, "https://"); found {
+		ociRegistryHost = noSchemeRef
+	}
```
ruanxin: this whole black can be `getOciRegistryHost`, and return the host, if insecure as `(string, bool, error)`

internal/repository/oci/ocirepo.go
```diff
@@ -0,0 +1,121 @@
...
+var (
+	ErrKeyChainNotNil   = errors.New("keychain lookup must not be nil")
+	ErrNoProtocolScheme = errors.New("hostref must not contain protocol scheme (http/https)")
+	ErrNoLeadingSlash   = errors.New("hostref must not start with a '/'")
+)
+
+// RepositoryReader provides basic support to read data from OCI repositories.
+type RepositoryReader struct {
+	keyChainLookup spec.KeyChainLookup
+	hostref        string
```
ruanxin: Each part of URL have a standard name, this is just `host`, I never heard of `hostref`.

internal/descriptor/provider/provider.go
```diff
@@ -1,109 +1,90 @@
...
-			c.DescriptorCache.Set(key, desc)
-			return nil
-		}
+		return descriptor, nil
 	}
 
-	ocmDesc, err := compdesc.Decode(
-		template.Spec.Descriptor.Raw, []compdesc.DecodeOption{compdesc.DisableValidation(true)}...,
-	)
+	descriptor, err := c.descriptorService.GetComponentDescriptor(context.Background(), ocmi)
```
ruanxin: You should not use background context here without cancelling.

// Background returns a non-nil, empty [Context]. It is never canceled, has no
// values, and has no deadline. It is typically used by the main function,
// initialization, and tests, and as the top-level Context for incoming
// requests.

the context here should have `WithCancel`, and get cancelled after function call.

```
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()
```

internal/service/componentdescriptor/extractfile_internal_test.go
```diff
@@ -0,0 +1,74 @@
+package componentdescriptor
```
ruanxin: rename to `componentdescriptor_test`

internal/service/componentdescriptor/untar_internal_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"bytes"
+	"errors"
+	"io"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestUnTar(t *testing.T) {
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Why would I do that? Nobody else wants to use that function and making it public increases package API surface for no reason.
I could extract this to a separate sub-package, but then: What's the difference between that and just making that public? I am still making public API bigger, although **nobody needs it**.
That's a bad practice, package APIs should be minimal.
BTW: Function doesn't have to be exported to be testable in Go, if in doubt, please see this [official go docs](https://pkg.go.dev/testing)
`If the test file is in the same package, it may refer to unexported identifiers within the package, as in this example:`

internal/repository/oci/ocirepo.go
```diff
@@ -0,0 +1,121 @@
...
+	insecure       bool
+	cWrapper       craneWrapper // in runtime delegates to crane package functions
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: The answer is: It uses types from crane library, is strictly tied to it. It is not generic. I can rename it, but these types will still point out to the crane library. It will be misleading then: The type name will *suggest* a generic abstraction, but the interface methods will force you to have just a single, crane-based implementation.

internal/repository/oci/ocirepo.go
Tomasz-Smelcerz-SAP: > Why not moving this out of here to a normal structing implemetning a public interface defined in ocirepo.go? Then we just inject it into the constructor

Because it is not a part of package's public API.
I consider every package as a separate little entity, kind-of "living on it's own". It has a public API and this API should be minimal, i.e: it should only give users the minimal set of features required to perform the operations that API promises to perform.
`craneWrapper` is an internal facility introduced **only** to facilitate testing, it's not a part of the user-facing contract.

Why would then users of `internal/repository/oci` be bothered with a knowledge about such things? I mean they can read the sources, but why would we pollute the package's public API (constructors) with such details?
Exposing such internals violates a few well recognized principles (more or less, these are usually related to OO, but similar concepts still apply in Go):
- Information Hiding
- Law of Demeter (Principle of Least Knowledge)
- Encapsulation

internal/descriptor/types/ocmidentity/ocmidentity.go
```diff
@@ -0,0 +1,47 @@
...
+		return nil, fmt.Errorf("invalid component version: %w", ErrValueNotProvided)
+	}
+
+	return &Component{
+		componentName:    name,
+		componentVersion: version,
+	}, nil
+}
+
+func (c *Component) Name() string {
```
lindnerby: Since these are only read methods (similar to getters) they can even have a value receiver instead of pointer to further indicate that:
`func (c Component) Name() string {`

internal/descriptor/types/ocmidentity/ocmidentity.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"errors"
+	"fmt"
+)
+
+var ErrValueNotProvided = errors.New("value not provided")
+
+// Component uniquely identifies an OCM Component.
+// See: https://ocm.software/docs/overview/important-terms/#component-identity
+type Component struct {
```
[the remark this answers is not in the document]
lindnerby: Usade and function during the code even call this differently already:
`OCMIdentity`
Would be good to align this and settle for a name, to talk about **one** thing from the beginning.

.github/actions/deploy-template-operator-with-modulereleasemeta/action.yml
```diff
@@ -79,10 +79,9 @@ runs:
         sed -i 's/k3d-private-oci-reg.localhost:5001/private-oci-reg.localhost:5000/g' ./template.yaml
         kubectl get crds
         kubectl apply -f template.yaml
-    - name: Create and apply ModuleReleaseMeta from the latest release
+    - name: Create and apply ModuleReleaseMeta from the template-operator repo
       working-directory: template-operator
       if: ${{ matrix.e2e-test == 'kyma-metrics' ||
-        matrix.e2e-test == 'non-blocking-deletion' ||
```
lindnerby: Quick info on why this testcase is not important anymore? I don't know the details.

api/v1beta2/modulereleasemeta_types.go
```diff
@@ -102,3 +104,9 @@ func (m ModuleReleaseMeta) IsBeta() bool {
 func (m ModuleReleaseMeta) IsInternal() bool {
 	return m.Spec.Internal
 }
+
+// FullOCMName returns the fully qualified OCM component name for a given module name.
+// This is used by OCM-related functionality, end-users do not have to use this format.
+func FullOCMName(moduleName string) string {
+	return shared.KymaGroup + "/module/" + moduleName
+}
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Moved.

internal/descriptor/cache/key_test.go
```diff
@@ -0,0 +1,34 @@
...
+			want:          "name:1.0.0",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ocmi := ocmidentity.MustNew(tc.moduleName, tc.moduleVersion)
+			got := cache.GenerateDescriptorKey(*ocmi)
+			assert.Equal(t, tc.want, string(got))
+		})
```
lindnerby: Just to align some things here:

- `DescriptorKey` could also get its own file if the test gets it? Or incorporate this test as well in `cache_test.go`.
- If there is only one testcase there is no need for a table driven structure. So either we come up with more testcases here or have just one plain test.

internal/repository/oci/export_test.go
```diff
@@ -0,0 +1,6 @@
+package oci
+
+// compiled only when running tests.
```
lindnerby: Why introduce this indirection and not create a separate pkg for the wrapper isolated with its tests.
And then inject the wrapper dependency for consumers as usual.

Tomasz-Smelcerz-SAP: Why introduce another package "B" just to **test** package "A"?
Packages should have purpose. In this case, package "A" - `internal/repository/oci` can "handle itself" very well, we can write all the tests we want without introducing new abstractions.
I guess "do not multiply entities beyond necessity" is still the correct approach?

internal/descriptor/provider/provider.go
```diff
@@ -1,109 +1,90 @@
...
-
-func (c *CachedDescriptorProvider) GenerateDescriptorKey(name, version string) string {
-	return fmt.Sprintf("%s:%s", name, version)
+	return c.getDescriptor(*ocmi, false)
 }
 
-func (c *CachedDescriptorProvider) Add(template *v1beta2.ModuleTemplate) error {
-	if template == nil {
-		return ErrTemplateNil
+func (c *CachedDescriptorProvider) getDescriptor(ocmi ocmidentity.Component, updateCache bool) (
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Done.

cmd/main.go
ruanxin: That’s totally fine — the review comments are meant to highlight potential issues, not necessarily to be fixed right away. If you’d prefer not to address them in this PR, feel free to create a follow-up issue instead.

At the same time, not mentioning these code smells would be my oversight. Continuously improving our existing codebase should always be our goal.

internal/descriptor/provider/provider.go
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: I am also confused by that and that's why
- I did not refactor this code over what was absolutely necessary
- I tried to keep the contract of existing functions **exactly** as it was, to avoid now (in PR review) the discussions if my changes are correct or not. I mean, you should still look and verify If the contract is unchanged (Add, Get), because that was my intention - so that they behave in the same way with respect to cache update/read behavior.

My focus in this PR was to read component descriptor from the OCI registry, not to improve module template lookup logic. And even in that "narrow scope" I made many mistakes, as you correctly pointed out in other comments.
That's why I am so against introducing any changes here, even the ones that "seem obvious". I prefer to do that in another PR, once we prove that all changes introduced so far are OK. Small steps. I don't feel confident here.

As for the `updateCache` boolean another reviewer also didn't like it, so I removed it. Now both the `Get()` and `Add()` functions contain duplicated code, but well - it's the price to be paid if you don't like ad-hoc booleans like that.
Personally I don't mind this duplication, this code will likely to be removed anyway - once we'll clean up module template lookup logic (not in this PR).

internal/service/componentdescriptor/service_test.go
```diff
@@ -0,0 +1,303 @@
...
+	m.pullLayerName = name
+	m.pullLayerTag = tag
+	m.pullLayerDigest = digest
+	if m.pullLayerError != nil {
+		return nil, m.pullLayerError
+	}
+	return m.pullLayerResult, nil
+}
+
+func wrapAsTar(t *testing.T, fileName string, data []byte) []byte {
```
ruanxin: This is the bad concesquence that you deeply coupled tar related operation into `GetComponentDescriptor`, a good unit test doesn't require such low level detail. The reason I mentioned here https://github.com/kyma-project/lifecycle-manager/pull/2736#discussion_r2435890501

internal/descriptor/types/ocmidentity/ocmidentity.go
```diff
@@ -0,0 +1,47 @@
...
+func MustNew(name, version string) *Component {
+	c, err := New(name, version)
+	if err != nil {
+		panic(err)
+	}
+	return c
+}
+
+// New is a constructor that ensures that both name and version are provided.
+func New(name, version string) (*Component, error) {
```
ruanxin: go have naming convesion to initilze struct, normally should be `New[struct name]`, so here `NewComponent` is better. Actually I also second @lindnerby suggestions, rename it to `ComponentId` or maybe better name

internal/manifest/img/parse_test.go
```diff
@@ -47,9 +52,13 @@ func TestParse(t *testing.T) {
 	for _, testCase := range tests {
 		t.Run(testCase.name, func(t *testing.T) {
 			var moduleTemplateFromFile v1beta2.ModuleTemplate
-			builder.ReadComponentDescriptorFromFile(testCase.DescriptorSourceFile,
+			builder.ReadComponentDescriptorFromFile(testCase.descriptorSourceFile,
 				&moduleTemplateFromFile)
-			descriptor, err := provider.NewCachedDescriptorProvider().GetDescriptor(&moduleTemplateFromFile)
+			ocmi, err := ocmidentity.New("kyma-project.io/module/template-operator", testCase.descriptorVersion)
+			require.NoError(t, err)
+			descriptor, err := provider.NewCachedDescriptorProvider(
```
ruanxin: I didn't get the point, for each test case, you create a FakeService specifically, then what's the meaning for that `stopped` field, each FakeService not sharing the state.

And even you need that `stopped` state, you current implementation is thread unsafe for concurrent tests, you need a mutext for update it's state.

internal/service/componentdescriptor/untar_internal_test.go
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: > The fact that something is possible doesn’t mean it’s recommended. 

Please back that statement up with some arguments. In my view **this is recommended** because **it is explicitly mentioned** in the most official, idiomatic, **definitive** reference docs for Go: it's standard library documentation. If that's **not a recommendation** to you, then what is?
A book written on how to write Java programs? Well, **Go's not Java** if you haven't figured that out already.
Go authors are perfectly aware of Java's rambling mantra "you should not test private things" and they **explicitly** state that it is not how you write tests in Go.
And, contrary to you, **I have the evidence**: [Official Go docs](https://pkg.go.dev/testing)

And, by the way, it's not only Go authors that realize the value to test the code directly, without any artificial obstacles like the one existing in Java, where you don't test private methods, because **the compiler doesn't let you to**.
Take a more modern language like Rust, or even younger - Zig. You will be surprised.
You think that these guys are under-educated, and they should "fix" their languages with the help of “Unit Testing Principles, Practices, and Patterns”? That's ridiculous.

> Testing private methods directly often leads to unmockable, tightly coupled code.

It's just your opinion, not backed up with any argument so far.
Whether the code is coupled or not has nothing to to with how you test it. You can have tightly coupled code with all the testing done via public interface. And a loosely coupled code with all unit tests defined inside the package (in Go). These two things are orthogonal.

Maybe if you really believe in what's written by these Ivory Tower Java Architects you should switch back to Java? I am programming in Go and I am not going to align to a flawed "advices" written for a flawed langague which, in my opinion (and not only mine...) Java is.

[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: > Let's take your current unit test for GetComponentDescriptor at the moment in your PR, It's actually not a unit test, more of an integration test already. You need to compress some data with tar, that's not necessary for unit testing this function. Basically, after you abstract those tar-related operations, you don't need to bother about this detail while testing GetComponentDescriptor. Do you see the value now?

This is the only thing I can agree with. It may be the case that the current logic for reading files  from Layers is too complex and it's worth extracting. But it always comes with a price: You a also increase the complexity (as the system has now more "moving parts") and you put an additional requirement on the consumers of your package: You force them to provide the additional dependency. It complicates the API of an otherwise simple package and is not always worth it.
However, I'll give it a try to see what can be done.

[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: @ruanxin I extracted all the abstractions into dedicated struct fields.
You can take a look.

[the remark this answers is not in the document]
ruanxin: I just want to quote some paragraph from book - [The art of unit testing](https://www.artofunittesting.com/) for later references:
 
> 8.2.1 Testing private or protected methods
Private or protected methods are usually private for a good reason in the developer’s
mind. Sometimes it’s to hide implementation details, so that the implementation can
change later without the end functionality changing. It could also be for security-
related or IP-related reasons (obfuscation, for example).
When you test a private method, you’re testing against a contract internal to the
system, which may well change. Internal contracts are dynamic, and they can change
when you refactor the system. When they change, your test could fail because some
internal work is being done differently, even though the overall functionality of the
system remains the same.
For testing purposes, the public contract (the overall functionality) is all that you
need to care about. Testing the functionality of private methods may lead to breaking
tests, even though the overall functionality is correct.
Think of it this way: no private method exists without a reason. Somewhere down
the line there’s a public method that ends up invoking this method, or invokes a pri-
vate method that ends up invoking the method you’re interested in. That means that any private method is usually part of a bigger unit of work, or a use case in the system,
that starts out with a public API and ends with one of the three end results: return
value, state change, or third-party call (or all three).
With this viewpoint, if you see a private method, find the public use case in the sys-
tem that will exercise it. If you test only the private method and it works, that doesn’t
mean that the rest of the system is using this private method correctly or handles the
results it provides correctly. You might have a system that works perfectly on the
inside, but all that nice inside stuff is used horribly wrong from the public APIs.
Sometimes if a private method is worth testing, it might be worth making it public,
static, or at least internal and defining a public contract against any code that uses it.
In some cases, the design may be cleaner if you put the method in a different class
altogether. We’ll look at these approaches in a moment.
Does this mean there should eventually be no private methods in the code base?
No. With TDD, you usually write tests against methods that are public, and those pub-
lic methods are later refactored into calling smaller, private methods. All the while,
the tests against the public methods continue to pass.

image

internal/descriptor/types/ocmidentity/ocmidentity.go
```diff
@@ -0,0 +1,47 @@
...
+
+// Component uniquely identifies an OCM Component.
+// See: https://ocm.software/docs/overview/important-terms/#component-identity
+type Component struct {
+	componentName    string
+	componentVersion string
+}
+
+// MustNew is a convenience constructor that panics if name or version are not provided.
+func MustNew(name, version string) *Component {
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: moved, renamed to: `MustNewComponentId`

internal/descriptor/cache/cache.go
```diff
@@ -29,6 +31,12 @@ func (d *DescriptorCache) Get(key string) *types.Descriptor {
...
 }
 
-func (d *DescriptorCache) Set(key string, value *types.Descriptor) {
+func (d *DescriptorCache) Set(key DescriptorKey, value *types.Descriptor) {
 	d.cache.Store(key, value)
 }
+
+type DescriptorKey string
+
+func GenerateDescriptorKey(ocmi ocmidentity.Component) DescriptorKey {
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: as written above, it's `ocmi ocmidentity.ComponentId`
I tried to fix all occurrences, but maybe something sneaked through my filters. Please report if so.
BTW: `ocmi` because it's shorter that `componentId` and it contains an important prefix indicating that it's `ocm`-related.

internal/service/componentdescriptor/service.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	return compdescBytes, nil
+}
+
+type extractFileIOHelper interface {
+	ReadAll(r io.Reader) ([]byte, error) // Reads all data from the reader
+}
+
+type defaultExtractFileIOHelper struct {
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Refactored. Will be injected as agreed with @ruanxin

internal/descriptor/types/ocmidentity/ocmidentity.go
Tomasz-Smelcerz-SAP: Changed to value receiver

internal/descriptor/types/ocmidentity/ocmidentity.go
Tomasz-Smelcerz-SAP: Changed to `NewComponentId`

internal/descriptor/cache/key_test.go
Tomasz-Smelcerz-SAP: Done.

internal/manifest/img/parse_test.go
Tomasz-Smelcerz-SAP: This stopped state is required to handle just one thing - testing if caching "works". There's been such existing check in one of the tests, so I wanted to keep this logic.
How does this work? CacheDescriptorProvider is now opaque. One doesn't know if the ComponentDescriptor is fetched from the cache, or from the underlying service.
By stopping the service I can test if the ComponentDescriptor is really fetched from the cache. If the cache is not used/empty, the call is propagated to a (now stopped) service, and the panic there will ensure test failure.

Your remark about concurrency may be valid, I need to double-check.

internal/descriptor/types/ocmidentity/ocmidentity.go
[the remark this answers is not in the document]
lindnerby: Thanks! What is not so intuitive is the name itself. 
`ocmId` is at least telling that it's an id. And it's short, so nice to use.

.github/actions/deploy-template-operator-with-modulereleasemeta/action.yml
ruanxin: To Review: 
this test applied MRM twice, that's the reason why the first one should be deleted https://github.com/kyma-project/lifecycle-manager/pull/2736/files#diff-328e54a8869347e6d4401cafdc31b8d28331fd651340f2ec5a954880b58e0897L85

.github/actions/deploy-template-operator-with-modulereleasemeta/action.yml
ruanxin: check this comment: https://github.com/kyma-project/lifecycle-manager/pull/2736#discussion_r2447030150

internal/descriptor/types/ocmidentity/ocmidentity.go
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: So you want me to change the variable name from `ocmi` to `ocmId`? I have no problem with that, just asking to be sure.

internal/descriptor/cache/cache.go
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Done.

internal/service/componentdescriptor/service.go
lindnerby: They are gone. 👍

internal/service/componentdescriptor/service.go
```diff
@@ -0,0 +1,223 @@
...
+	tarReader *tar.Reader
+}
+
+func (d *defaultUntarIOHelper) Next() (*tar.Header, error) {
+	return d.tarReader.Next() //nolint:wrapcheck // this helper should be transparent
+}
+
+func (d *defaultUntarIOHelper) CopyN(dst io.Writer, n int64) (int64, error) {
+	return io.CopyN(dst, d.tarReader, n) //nolint:wrapcheck // this helper should be transparent
+}
```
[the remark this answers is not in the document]
lindnerby: They are gone. 👍

internal/descriptor/provider/provider.go
```diff
@@ -1,109 +1,90 @@
...
-	}
-	key := c.GenerateDescriptorKey(template.Name, template.Spec.Version)
+func (c *CachedDescriptorProvider) GetDescriptor(ocmi ocmidentity.Component) (*types.Descriptor, error) {
+	return c.getDescriptor(ocmi, false)
+}
 
-	descriptor := c.DescriptorCache.Get(key)
-	if descriptor != nil {
-		return descriptor, nil
+func (c *CachedDescriptorProvider) GetDescriptorWithIdentity(ocp OCMIProvider) (*types.Descriptor, error) {
```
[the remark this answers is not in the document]
c-pius: See https://github.com/kyma-project/lifecycle-manager/pull/2736#discussion_r2459407675, we keep it as is for now and maybe refactor later.

internal/descriptor/types/ocmidentity/ocmidentity.go
```diff
@@ -0,0 +1,47 @@
...
+	}, nil
+}
+
+func (c *Component) Name() string {
+	return c.componentName
+}
+
+func (c *Component) Version() string {
+	return c.componentVersion
+}
```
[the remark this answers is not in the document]
c-pius: See https://github.com/kyma-project/lifecycle-manager/pull/2736#discussion_r2459407675, we keep it as is for now and maybe refactor later

internal/manifest/parser/template_to_module.go
c-pius: I indeed got confused here and didn't understand correctly on my side. Thanks for clarifying!

internal/service/componentdescriptor/extractor.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+	"io"
+
+	containerregistryv1 "github.com/google/go-containerregistry/pkg/v1"
+)
+
+// FileExtractor is responsible for extracting a specific file from a container image layer.
+type FileExtractor struct {
+	// reads all data from the provided reader
+	readAll func(reader io.Reader) ([]byte, error)
```
c-pius: If I understand it correctly, mocking this reader is the reason why this struct is tested "private". How about changing to the same pattern as we introduced for the OCI repository? I.e., make this configurable via a functional options pattern, then the test could be put into a separate `_test` package.

internal/service/componentdescriptor/untar_internal_test.go
```diff
@@ -0,0 +1,121 @@
...
```
c-pius: The `maxSize = maxDescriptorSizeBytes` path is uncovered. Not sure if worth testing, just wanted to mention it here

image

internal/service/componentdescriptor/untar.go
```diff
@@ -0,0 +1,64 @@
...
+const (
+	maxDescriptorSizeBytes = 100 * 1024 // 100KiB, our average is around 4KiB
+	tarReadChunkSize       = 10 * 1024  // 10KiB, for our average size we'll read it in one go
+)
+
+// TarExtractor is responsible for extracting named files from a tar archive.
+type TarExtractor struct {
+	next  func(tarReader *tar.Reader) (*tar.Header, error)           // wrapper over tar.Reader.Next()
+	copyN func(dst io.Writer, src io.Reader, n int64) (int64, error) // wrapper over io.CopyN()
+}
```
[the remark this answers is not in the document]
c-pius: https://github.com/Tomasz-Smelcerz-SAP/lifecycle-manager/pull/2/commits/433fa654966028bbe411b432506b2b92c65b9df8

internal/service/componentdescriptor/extractfile_internal_test.go
c-pius: Done

c-pius:
And thanks a lot @Tomasz-Smelcerz-SAP for all the effort you put into this!

tests/integration/controller/mandatorymodule/installation/controller_test.go
```diff
@@ -49,14 +48,14 @@ var _ = Describe("Mandatory Module Installation", Ordered, func() {
...
 				}
 				return nil
-			}).
+			}, Timeout, Interval).
 				Should(Succeed())
 		})
 
 		It("And Manifest CR for the Mandatory Module should be created with correct Owner Reference", func() {
-			Eventually(checkMandatoryManifestForKyma).
+			Eventually(checkMandatoryManifestForKyma, Timeout, Interval).
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Done.

Issue #2580, opened 2025-07-14
[GCM] Configure Individual Renewal Times For CA and SKR Certificates


LeelaChacha:
### Description

TODO

### Acceptance Criteria

- [ ] Acceptance criterion 1
- [ ] Acceptance criterion 2

c-pius:
Can be refined now:

- https://github.com/kyma-project/lifecycle-manager/pull/2726#event-19591595462
- https://github.com/gardener/cert-management/releases/tag/v0.18.0

PR #2737, opened 2025-09-15
feat: Configure Cert Duration and RenewBefore

c-pius:
**Description**

- Configures the certificate for GCM and cert-manager:
  - CA duration: 90d
  - CA renewBefore: 61d
  - SKR duration: 60d
  - SKR renewBefore: 30d
- Istio Gatway Secret Switch: 33d

With this configuration, it should be possible that a SKR cert never outlives the CA cert (this will raise an error in GCM). Additionally the switch in the Istio Gateway Secret is set to 33d before CA expiry. This means that the keys will be switched one day before the consecutive rotation of the cert: `90d (expiry) - 33d (time window) == 57d < 58d == 29d (first rotation) + 29d (second rotation)`.

As the configuration can be the same for GCM and cert-manager, I decided to go with the same configuration for both.

**Related issue(s)**

closes https://github.com/kyma-project/lifecycle-manager/issues/2580


tests/e2e/commontestutils/watcher.go
```diff
@@ -157,18 +157,22 @@ func GetLastModifiedTimeFromAnnotation(secret *apicorev1.Secret) (time.Time, err
...
+            "ensureRenewedAfter": null,
+            "renew": true
+        }
+    }`)
+
+	if err := kcpClient.Patch(ctx, cert, client.RawPatch(types.MergePatchType, patch)); err != nil {
+		return fmt.Errorf("failed to patch CA certificate: %w", err)
 	}
 	return nil
 }
```
c-pius, line 178: This was necessary because the update to GCM 0.18.0 has made the test flaky. We started to see the following message indicating race conditions:
```
Operation cannot be fulfilled on certificates.cert.gardener.cloud \"klm-watcher-serving\": the object has been modified; please apply your changes to the latest version and try again
```

Issue #2377, opened 2025-04-01
Investigate improvements in our handling of tools


c-pius:
### Description

We already invested in aligning the versions of our used tooling (see `versions.yaml` file). We can investigate how to improve this even further.

E.g.:

- use [Tool dependencies](https://antonz.org/go-1-24/#tool-dependencies) to keep our versions up to date
  - if so, what is the alternative for non-go tools?
- cache the tools in our pipelines to speed up test execution
  - consult with Huskies who are already doing this
- more?

### Reasons

Use up to date tooling

### Acceptance Criteria

- [ ] Evaluated Tool dependencies and took decision with the team whether to introduce it
- [ ] Evaluated caching tools in pipeline and took decision with the team whether to introduce it
- [ ] Brainstormed about further improvements and took decision with the team whether to introduce them

PR #2748, opened 2025-09-17
chore: Introduce Go Tools with `golangci-lint@2.4.0`

c-pius:
**Description**

- introduces a `tools` package following the tools [feature introduced with Go 1.24](https://tip.golang.org/doc/go1.24#tools)
  - a tool can be added by `go get -tool <tool>`, e.g. `go get -tool github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.4.0`
  - it can then be executed by `go tool <tool> name`, e.g. `go tool golangci-lint`
  - OR it can be built like `go install tool`
  - in our case, we use `go install tool` to build the binary and put it in our bin folder to use
- the benefit of this approach is that have the dependency versioned and we can have it auto updated by dependabot
- in our case, the other benefit is that we can specify the Go version to build the binary via the go.mod file
  - this enables us to adopt new Go versions quicker as for instance here, our linting fails because the binary we got from the normal Go install is compiled with their Go version which fails: https://github.com/kyma-project/lifecycle-manager/pull/2741
    - (I have already tested this but want to provide separate smaller PRs)
- note that golangci-lint is actually not recommending to install from sources, but this is what we have done with `go install` already now

**Planned Follow-Ups**

- add more of our Go-native tools to this approach, e.g., k3d, yq, ...
- use the same approach for our other projects
- fix the TODOs from the `.golangci.yaml`
  - `wsl_v5` and `noinlineerr` we already agreed to not use, the others I will check and fix

**Related issue(s)**

- this would be a first step towards https://github.com/kyma-project/lifecycle-manager/issues/2377


c-pius:
Closing in favor of using a binary install.

PR #2756, opened 2025-09-18
refactor: Remove watcher restart label

### LINKED ISSUE #1158: Remove watcher Restart-Label
### Description

A year ago, a label named `operator.kyma-project.io/pod-restart-trigger` was introduced and used to an automatic pod restart. This feature has been removed, and it looks like the label is not used anymore. Please double check if this label is still needed, if not remove.

### Reasons

Remove unneeded code.

### Acceptance Criteria

If label is unused:
- [ ] Remove `operator.kyma-project.io/pod-restart-trigger` label
- [ ] Remove usage (i.e. [here](https://github.com/kyma-project/lifecycle-manager/blob/c710461f7aa929c06f0d892f8dc251d9bda84625/pkg/watcher/skr_webhook_resources.go#L156))
- [ ] Remove label from resources.yaml https://github.com/kyma-project/lifecycle-manager/blob/ddb74a3e60149535ac8fa750c72bcd9a4d35e48c/skr-webhook/resources.yaml#L68

If label has usage:
- [ ] Update Watcher and KLM docu what this label is used for and where

### Feature Testing

_No response_

### Testing approach

Tests should still work after removing this feature

### Attachments

_No response_

LeelaChacha:
**Description**

- Remove `operator.kyma-project.io/pod-restart-trigger` label

**Related issue(s)**

Closes #1158


c-pius (commented):
Are we sure that this label is not needed? The `watcher-zero-downtime` test has been failing consistently for all three commits of this PR.

We typically see the watcher restarting when the certificate secret changes. How is this implemented right now?

c-pius:
> We typically see the watcher restarting when the certificate secret changes. How is this implemented right now?

Have just reproduced this locally. On main, the watcher restarts when the cert secret is re-created. With this branch, it doesn't.

c-pius:
Sure. I was able to re-produce it twice. On main, when observing the logs, we can see the following after re-creating the SKR secret. I was using an outdated version of runtime-watcher (1.1.20) because of the platform problem. But that shouldn't be an issue I think.

image

PR #2772, opened 2025-09-24
fix: Rate Limiting in Kyma Processing State

### LINKED ISSUE #2467: Fix Exponential Backoff in Kyma Reconciler
### Description

**Background**

We use the exponential backoff for rate limiting in the Kyma reconciler. However, the condtions for the backoff to work depend on the ctrl.Result{RequeueAfter: time.Duration} and the error by the Reconcile() method. 

According to comments in the controller runtime library, returning an error is necessary to use exponential backoff (`Result.Requeue` is deprecated):

https://github.com/kubernetes-sigs/controller-runtime/blob/v0.21.0/pkg/reconcile/reconcile.go#L109-L122

**Description**
In https://github.com/kyma-project/lifecycle-manager/issues/2411, we found that a requeue for the Kyma reconciler is not working as expected.
At the end of the processing state reconciliation, the controller requeues with a fixed requeue interval determined for that state. This causes the exponential backoff to be overwritten by a fixed value of 2s:
https://github.com/kyma-project/lifecycle-manager/blob/852c0a5446c32ff3dcea0d40b76d99c6df212669/internal/controller/kyma/controller.go#L390-L392
Furthermore, since the error returned by updateStatus() is usually nil, we requeue constantly when Kyma is in Error state

### Reasons

_No response_

### Acceptance Criteria

- [x] Error backoff for processing state in Kyma Reconciler is utilized 
  - Add a custom err that is returned when Kyma statue is error
- [x] Document this behaviour (on a CRD level - https://github.com/kyma-project/lifecycle-manager/blob/e7efceb4af3833ccd7e25a5f02eba12b260c3d14/api/v1beta1/kyma_types.go#L96)
  - Make it explicit that "error requeue interval is subject to rate limiting"
- [ ] Follow up ticket created to investigate other requeues in Kyma and all other reconcilers to check for this bug.

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

LeelaChacha:
**Description**

- if Kyma state is Error, then handleProcessingState method in Kyma Reconciler will return a new error so that the ctrl runtime requeues with rate limiting.

**Related issue(s)**

Closes #2467


lindnerby:
> Yes, we need a bigger refactoring before we can even introduce any test for the reconciler, so I didn't want to start with that in this ticket, as time was limited.
> 
> I verified that the feature is working instead: looking at metrics with one Kyma in error state (on main and on PR), and printing out reconcile start timestamps for the kyma reconciler (on main and on PR).

👍  lgtm

PR #2775, opened 2025-09-30
refactor: Remove Options from Declarative Reconciler

### LINKED ISSUE #2329: Replace declarative options with normal configuration
### Description

The current implementation of the manifest reconciler utilizes the declarative library with an ​option pattern (e.g., functional options for state checks, metrics, etc.). While this pattern was initially valuable to support external configurability, it now introduces unnecessary complexity since the library is ​no longer exposed for external use. 

https://github.com/kyma-project/lifecycle-manager/blob/2b4b9a82f9501b4d47fba0ef1f4a32ee3e1fc8dd/internal/controller/manifest/controller.go#L30-L32

https://github.com/kyma-project/lifecycle-manager/blob/b8f15be9c9faea9508abda1716709886a389e3d0/internal/declarative/v2/options.go#L39-L52

### Reasons

The abstraction reduces code readability and obscures the actual reconciliation flow for developers working within the codebase.

### Acceptance Criteria

- [x] Remove the option pattern from the declarative library’s reconciler setup.
- [x] Inline configuration logic directly into the Manifest controller constructor, explicitly passing dependencies.
- [x] The whole file options.go should be deleted, so also the DefaultOptions() builder pattern, etc. should be removed and default dependencies passed directly

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

LeelaChacha:
**Description**

- Remove Options from Declarative Reconciler

**Related issue(s)**

Closes #2329

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -482,7 +482,6 @@ func setupManifestReconciler(mgr ctrl.Manager,
 
 	kcpClient := mgr.GetClient()
 	cachedManifestParser := declarativev2.NewInMemoryCachedManifestParser(declarativev2.DefaultInMemoryParseTTL)
-	postRenderTransforms := declarativev2.GetDefaultTransforms()
 	statefulChecker := statecheck.NewStatefulSetStateCheck()
 	deploymentChecker := statecheck.NewDeploymentStateCheck()
 	customStateCheck := statecheck.NewManagerStateCheck(statefulChecker, deploymentChecker)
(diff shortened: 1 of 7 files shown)


internal/declarative/v2/options.go
```diff
@@ -1,138 +0,0 @@
...
-			shared.ManagedBy: shared.ManagedByLabelValue,
-		},
-	)
-}
-
-type Options struct {
-	client.Client
-	ManifestCache
-	ManifestParser
-	record.EventRecorder
```
LeelaChacha, line 43: To Reviewer: This was not ported/moved because it has no use

Issue #2492, opened 2025-06-03
Release api, maintanencewindow, listener as go module support semver in the existing repo


LeelaChacha:
### Description

We have the following sub-modules in the lifecycle-manager: `api`, `maintwindows`
And the `listener` submodule in the runtime-watcher repo.
To facilitate proper semver releases, we need to provide an extra tag in their repos in the format: `<module-name>/v<major.minor.patch>`.
A dedicated pipeline should be created to automate the process and avoid human errors.

### Acceptance Criteria

- [ ] Dedicated release pipeline created for each sub-module in their repos
  - basic sanity checks for proper semantic versioning
  - only tag (no release) created on the main branch when checks succeed
  - tag format: `<module-name>/v<major.minor.patch>` 
    - `v-` prefix is important!
    - `<module-name>` same as in `go.mod` file of the sub-module

c-pius:
We still need to discuss how we release the packages. The packages are consumed by ourselves and other teams which we may need to notify of new versions. We should document that this is separate from our regular lifecycle-manager image release workflow and agree on how we handle it.

@ruanxin and @c-pius create a proposal how to handle it.

PR #2791, opened 2025-10-06
chore: Update workflow names and permissions for API, KLM, and Maintenancewindows modules

ruanxin:
**Description**

fix error: 

```
Invalid workflow file: .github/workflows/create-release-api.yml#L10
The workflow is not valid. .github/workflows/create-release-api.yml (Line: 10, Col: 3): Error calling workflow 'kyma-project/lifecycle-manager/.github/workflows/create-release-go-module.yml@0669937815a0d7f3a9b5ddbb9d27a201ffe8ef4a'. The nested job 'release-listener' is requesting 'contents: write', but is only allowed 'contents: read'.
```

**Related issue(s)**
https://github.com/kyma-project/lifecycle-manager/issues/2492

### DIFF
--- a/internal/repository/watcher/certificate/gcm/certificate/certificate_repo.go
+++ b/internal/repository/watcher/certificate/gcm/certificate/certificate_repo.go
@@ -79,7 +79,6 @@ func (r *Repository) Create(ctx context.Context, name, commonName string, dnsNam
 		Spec: gcertv1alpha1.CertificateSpec{
 			CommonName:   &commonName,
 			Duration:     &apimetav1.Duration{Duration: r.certConfig.Duration},
-			RenewBefore:  &apimetav1.Duration{Duration: r.certConfig.RenewBefore},
 			DNSNames:     dnsNames,
 			SecretName:   &name,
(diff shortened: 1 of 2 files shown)


.github/workflows/create-release-api.yml
```diff
@@ -1,10 +1,12 @@
-name: "Create Release For Lifecycle Manager API Module"
+name: "Create Release For KLM API Module"
```
ruanxin, line 1: To Reviewer: shortered the name so that more distinct between API and Image in the action page.

https://github.com/kyma-project/lifecycle-manager/actions

.github/workflows/create-release-api.yml
```diff
@@ -1,10 +1,12 @@
...
 on:
   workflow_dispatch:
     inputs:
       version:
         description: "Semantic version to release (e.g. 1.2.3 or v1.2.3)"
         required: true
+permissions:
+  contents: write
```
medmes, line 9: Also double this for security compliant. Giving write permission to scoped jobs compared to set globally! 👍

medmes (approved):
LGTM 👍

Issue #2707, opened 2025-09-03
Update to `controller-runtime v0.XX.0` and adapt to native client.Apply


c-pius:
### Description

Bump `controller-runtime` to `v.0.XX.0` and migrate to the native `client.Apply` accordingly. For template-operator, this was done here: https://github.com/kyma-project/template-operator/pull/398 ~~Also note that for SubResourceWriter, it is not supported yet: #2706~~  (supported by now)

The challenge we have in lifecycle-manager is that we have our own [ProxyClient](https://github.com/kyma-project/lifecycle-manager/blob/2059ab91129d2def264ffb9f66479986671fd71a/internal/service/skrclient/client_proxy.go#L24) which must now implement the Apply method accordingly. To match the signature, we need to pass the object as `machineryruntime.ApplyConfiguration`, but it is not obvious what this `machineryruntime.ApplyConfiguration` really is, or how to get the GVK from that object as currently required by ProxyClient to get the resource mapping. This needs to be thought through thoroughly.

### Reasons

`v0.22.0` of `controller-runtime` deprecated `client.Apply` in favor of "Native SSA Support" via `client.Client.Apply`. We need to adapt accordingly.

### Acceptance Criteria

- [ ] in KLM
  - [ ] updated to `controller-runtime v0.XX.0`
  - [ ] fixed the `ClientProxy`
  - [ ] replaced deprecated `client.Apply` usage with `client.Client.Apply`

c-pius:
We agreed to skip this version as proposed by @lindnerby. Let's wait for native SSA support to further mature in the next version of controller-runtime before updating.

In case a CVE should pop up for cr 0.21.0, we can update to 0.22.0 but without adapting to the new client.Client.Apply.

c-pius:
FYI: deprecation was reverted: https://github.com/kubernetes-sigs/controller-runtime/pull/3307

c-pius:
Closing this in favor of https://github.com/kyma-project/lifecycle-manager/issues/3155. See the linked sub-issues for concrete actions.

PR #2800, opened 2025-10-07
chore: Bump controller-runtime to 0.22.2

c-pius:
**Description**

- bumps controller-runtime to `0.22.2`
  - I made a mistake and earlier ignored the whole `0.22` minor https://github.com/kyma-project/lifecycle-manager/pull/2705#issuecomment-3247951247
- with `0.22.0` the new `client.Client.Apply` was introduced and `client.Apply` was deprecated. The deprecation of `client.Apply` was reverted in `0.22.1` as the support for `client.Client.Apply` was incomplete (missing sub-resource writer). Once the full support for `client.Client.Apply` is there and the actual deprecation of `client.Apply` happens, we must make code changes as described in this ticket https://github.com/kyma-project/lifecycle-manager/issues/2707
- we still have the problem with our `ClientProxy`
  - as of now, we expect it to implement the new `Apply` method. However, the signature does not include a `client.Object` which we need to call the `getResourceMapping`. I don't have a good understanding of `getResourceMapping` but my understanding is that we call this to ensure that we can call the object's API which may require a refresh (reset), e.g., when a new CRD has been introduced. I don't know yet how we can solve this and we should investigate that more thoroughly as part of the ticket linked above. To make the code compile, I created a dummy implementation returning a NotImplementedError.

**Related issue(s)**

### DIFF
--- a/internal/remote/modulereleasemeta_syncer_test.go
+++ b/internal/remote/modulereleasemeta_syncer_test.go
@@ -35,6 +35,7 @@ func TestModuleReleaseMetaSyncer_SyncToSKR_happypath(t *testing.T) { //nolint:du
 	skrClient := fake.NewClientBuilder().
 		WithObjects(&mrmSKR2, &mrmSKR3, &mrmSKR4).
 		WithScheme(scheme).
+		WithReturnManagedFields().
 		Build()
 
 	onSyncConcurrentlyFn := func(_ context.Context, kcpModules []v1beta2.ModuleReleaseMeta) {
@@ -95,6 +96,7 @@ func TestModuleReleaseMetaSyncer_SyncToSKR_nilList(t *testing.T) {
 	skrClient := fake.NewClientBuilder().
 		WithObjects(&mtSKR2, &mtSKR3, &mtSKR4).
 		WithScheme(scheme).
+		WithReturnManagedFields().
 		Build()
 
 	// onSyncConcurrentlyFn "pretends" to be the moduleReleaseMetaConcurrentWorker.SyncConcurrently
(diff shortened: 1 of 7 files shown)


internal/service/skrclient/client_proxy.go
```diff
@@ -156,6 +158,12 @@ func (p *ProxyClient) List(ctx context.Context, obj client.ObjectList, opts ...c
 	return nil
 }
 
+// Apply implements client.Apply
+// Will be implemented in https://github.com/kyma-project/lifecycle-manager/issues/2707
+func (p *ProxyClient) Apply(ctx context.Context, applyConfig machineryruntime.ApplyConfiguration, opts ...client.ApplyOption) error {
+	return ErrNotImplemented
```
lindnerby, line 164: It's probably fine to do a dummy implementation here that returns an error, because until now the `Apply` method did not exists yet and is not used. Just make it clear here to `_` parameter variables. And let first sentence of the method's comment reflect that this will only return and error and is a dummy implementation.

lindnerby (changes requested):

internal/remote/modulereleasemeta_syncer_test.go
```diff
@@ -35,6 +35,7 @@ func TestModuleReleaseMetaSyncer_SyncToSKR_happypath(t *testing.T) { //nolint:du
 	skrClient := fake.NewClientBuilder().
 		WithObjects(&mrmSKR2, &mrmSKR3, &mrmSKR4).
 		WithScheme(scheme).
+		WithReturnManagedFields().
```
c-pius, line 38: It seems there were changes to fakeClient. Without providing this flag, the List operation will not include Managed Fields which is required by this and the other tests.

internal/remote/modulereleasemeta_syncer_test.go
```diff
@@ -149,7 +151,10 @@ func moduleReleaseMeta(name, namespace string) v1beta2.ModuleReleaseMeta {
 			Namespace: namespace,
 			ManagedFields: []apimetav1.ManagedFieldsEntry{
 				{
-					Manager: moduleCatalogSyncFieldManager,
+					Manager:    moduleCatalogSyncFieldManager,
+					Operation:  apimetav1.ManagedFieldsOperationApply,
+					APIVersion: v1beta2.GroupVersion.String(),
+					FieldsType: "FieldsV1",
```
c-pius, line 157: `fake.NewClientBuilder` fails if these fields are not provided

PR #2804, opened 2025-10-08
fix: Allow Module Image Names With Digest Only

### LINKED ISSUE #2763: KLM should be able to read valid image address
### Description

At the moment, KLM is possible to handle `tag`, or `tag`:`digest`, however, image only contains `digest` is also a valid case.

https://github.com/kyma-project/lifecycle-manager/blob/fdb9d8116442171fc6db60e45da47b8abb9aa19c/internal/imagerewrite/rewrite_test.go#L18-L25

### Reasons

_No response_

### Acceptance Criteria

- [x] Fix docker image reference to be able to handle valid image without a tag but with a digest https://github.com/kyma-project/lifecycle-manager/blob/fdb9d8116442171fc6db60e45da47b8abb9aa19c/internal/imagerewrite/rewrite.go#L45
  - e.g. something like `example.com/myapp/myimage@sha256:....`
- [x] use a regex to verify/find the images instead of splitting, e.g., use the one by Gardener
  - https://github.com/gardener/gardener-extension-image-rewriter/blob/master/pkg/webhook/operatingsystemconfig/image/mutator.go#L33
- [x] add a test case to modulectl to include a check for `example.com/myapp/myimage:<tag>@sha256:....`: https://github.com/kyma-project/modulectl/blob/main/internal/service/image/image_test.go#L12-L43

### Feature Testing

Unit tests

### Testing approach

_No response_

### Attachments

_No response_

LeelaChacha:
**Description**

- Use regex instead of manually splitting the image name for rewriting images for localisation.

**Related issue(s)**

Closes #2763

### DIFF
--- a/internal/imagerewrite/rewrite_test.go
+++ b/internal/imagerewrite/rewrite_test.go
@@ -18,10 +18,13 @@ func TestDockerImageReference(t *testing.T) {
 		imageRefs, err := imagerewrite.AsImageReferences([]string{
 			"example.com/myapp/myimage:1.2.3",
 			"example.com/myapp/myimage:1.2.3@sha256:837eb50a66bc0915d1986d376920c400d5db18075204339c0b047f5ba2091aa5",
+			"example.com/myapp/myimage@sha256:837eb50a66bc0915d1986d376920c400d5db18075204339c0b047f5ba2091aa5",
 			"example.com:5000/myapp/myimage:3.2.1",
-			"example.com:5000/myapp/myimage:3.2.1@sha256:c140c4dcdfe38aa7b462d9173ff4bad8fbfbb4819c5d9398c53c50abec7",
+			"example.com:5000/myapp/myimage:3.2.1@sha256:9a1de2363c531f585a01e185095498d700fcd10fc8801577e5e4e262832dd3cd",
+			"example.com:5000/myapp/myimage@sha256:9a1de2363c531f585a01e185095498d700fcd10fc8801577e5e4e262832dd3cd",
 			"localhost:5111/myimage:5.4.3",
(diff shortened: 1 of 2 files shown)


internal/imagerewrite/rewrite_test.go
```diff
@@ -18,10 +18,13 @@ func TestDockerImageReference(t *testing.T) {
 		imageRefs, err := imagerewrite.AsImageReferences([]string{
 			"example.com/myapp/myimage:1.2.3",
 			"example.com/myapp/myimage:1.2.3@sha256:837eb50a66bc0915d1986d376920c400d5db18075204339c0b047f5ba2091aa5",
+			"example.com/myapp/myimage@sha256:837eb50a66bc0915d1986d376920c400d5db18075204339c0b047f5ba2091aa5",
 			"example.com:5000/myapp/myimage:3.2.1",
-			"example.com:5000/myapp/myimage:3.2.1@sha256:c140c4dcdfe38aa7b462d9173ff4bad8fbfbb4819c5d9398c53c50abec7",
```
LeelaChacha, line 22: To Reviewer: This was modified because previous digest was not 64 chars and was not matching

internal/imagerewrite/rewrite_test.go
```diff
@@ -125,8 +147,9 @@ func TestDockerImageReference(t *testing.T) {
 				strValue: "example.com/myapp/myimage:1.2.3",
 			},
 			{
-				name:     "Image with digest",
```
LeelaChacha, line 128: To Reviewer: This was modified because previous digest was not 64 chars and was not matching

PR #2819, opened 2025-10-16
refactor: Split kyma controller

lindnerby:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/sec-scanners-config.yaml
+++ b/sec-scanners-config.yaml
@@ -8,7 +8,7 @@ checkmarx-one:
     - "**/testutils/**"
     - "tests/**"
 bdba:
-  - europe-docker.pkg.dev/kyma-project/prod/lifecycle-manager:1.7.3
+  - europe-docker.pkg.dev/kyma-project/prod/lifecycle-manager:1.9.1
   - europe-docker.pkg.dev/kyma-project/prod/lifecycle-manager:latest
 mend:
   language: golang-mod


lindnerby:
>     * => needs to be found out why
>     * log `Kyma deletion reconciliation started` not found (we see other DEBUG level logs)
>       
>       * => looks like the deletion reconciler is not even triggered
>         
>         * `func (r *DeletionReconciler) SetupWithManager(mgr ctrl.Manager, opts ctrlruntime.Options) error {` is unused
>           
>           * probably got removed when merging the conflicts earlier
>           * => re-add Setup DeletionReconciler [59f15bb](https://github.com/kyma-project/lifecycle-manager/commit/59f15bbf2c33c21bb3d865055fea184f8dc5ef0f)

Was added to unreachable code location. Moved it out of the err branch.

c-pius:
## watcher-enqueue

- can observe that the kubeconfig secret is deleted
- test is then stuck in expecting that the SKR webhook TLS cert and secret are removed which still exist
- to remove them, we need to fail initializing the client with a ErrAccessSecret not found: https://github.com/lindnerby/lifecycle-manager/blob/8741eb52ccee9e358cf8ecde1241ed14974e9730/internal/controller/kyma/deletion_controller.go#L96
- however, I assume that we continue reading the client from the cache, therefore we don't try to read the access secret again and don't receive the error
  - invalidation would happen here: https://github.com/lindnerby/lifecycle-manager/blob/8741eb52ccee9e358cf8ecde1241ed14974e9730/internal/controller/kyma/deletion_controller.go#L176-L180
  - but we don't reach that code because looping in error here: https://github.com/lindnerby/lifecycle-manager/blob/8741eb52ccee9e358cf8ecde1241ed14974e9730/internal/controller/kyma/deletion_controller.go#L108-L111
    - (that code I refactored to fix the other failing tests)

c-pius:
## Proposal: Deletion Controller with Use Cases

- we introduce a list of ordered use cases similar to the mandatory module deletion controller: https://github.com/kyma-project/lifecycle-manager/blob/b9e8b79a4314356a8f0e21f825f050b5c5c48c12/internal/service/mandatorymodule/deletion/deletion_service.go#L33-L48
- per reconciliation, exactly one use case will be performed
  - for kyma deletion controller, we expect we can cope with the performance penalty
  - determination of what use case to execute works on cached objects in KCP
  - determination of what use case to execute for SKR items requires API calls to the SKR, but that should be handlable by the individual API servers of the SKR
- use cases return a structured result (useCase, error)
  - depending on the returned result, the controller determines
    - requeue behavior
    - metric written
    - event raised
      - we raise a "Normal" event for every successfully executed use case
      - we raise a "Warning" event for every use case where an error was returned
      - => the KymaCR can stay in Deleting state once deletion was triggered, and the event history for the object reveals the deletion progress and potential errors we are stuck with

The overview of:

deletion-new

- note:
  - red use cases are probably not necessary (as opposed to in mandatory module deletion service), as here we can make it directly part of the reconciler setup
  - the kubeconfig secret exists predicate probably needs to be encoded in each SKR use cases predicate
  - we need to see if use cases like `Delete SKR Kyma` loop until the SKR Kyma is actually gone, or if we just consider it as "trigger deletion" where an eventual last step is verifying if everything is gone before removing finalizers

<details>

<summary>Use case brainstorming</summary>

- Ensure Finalizer
  - condition: "deletion-finalizers" not in Kyma.metadata.finalizers
  - task: SSA finalizer onto Kyma
  - requeue: immediate?
- Skip Non-Deleting
  - condition: deletionTimestampe != nil
  - task: no-op
  - requeue: no
- Skip Reconciliation
  - condition: "skip-reconciliation" in Kyma.metadata.finalizers
  - task: no-op
  - requeue: no
- Set Kyma Status Deleting
  - condition: Kyma.Status.State != Deleting"
  - task: SSA Kyma.Status.State = Deleting
  - requeue: immediate?
- [Delete SKR Resources]
  - condition needs to be added to all the following tasks: Kubeconfig secret exists
  - tasks:
    - Set SKR Kyma Status Deleting
      - condition: if Kyma exists && Kyma.Status.State != Deleting
      - task: SSA Kyma.Status.State = Deleting
      - requeue: immediate
    - Delete SKR Kyma:
      - condition: if Kyma exists && Kyma.deletionTimestamp == nil
      - task: delete Kyma
      - requeue: immediate
    - Delete SKR Webhook
      - condition: if any of the webhook resources exists
      - task: delete ALL SKR webhook resources
      - requeue: immediate
    - Delete SKR MT and MRM
      - condition: if any synced MT or any synced MRM exists
      - task: delete all synced MTs and MRMs
      - requeue: immediate
    - Delete SKR CRDs
      - condition: if any of the sycned CRDs exists
      - task: delete synced CRDs
      - reqeue: immediate
    - Remove SKR Kyma finalizers
      - condition: if Kyma exists and "deletion-finalizers" not in Kyma.metadata.finalizers
      - task: remove "deletion-finalizer" from Kyma.metadata.finalizers
      - requeue: immediate
- Delete SKR Webhook Cert and Secret on KCP
  - condition: if cert exists || secret exists
  - task: delete cert and secret
  - requeue: immediate
- Delete Manifests
  - condition: if manifests for this Kyma exist
  - task: delete manifests
  - requeue: immediate
- Cleanup Metrics
  - ???
- Remove Kyma finalizers
  - condition: if Kyma exists && "deletion-finalizer" in Kyma.finalizers
  - task: remove Kyma.finalizers
  - requeue: no


</details>

c-pius:
Update: "Delete SKR MT and MRM Metadata" + "Delete SKR CRDs" are three individual use cases now: https://github.com/kyma-project/lifecycle-manager/pull/2896

The "Remove SKR Kyma finalizers" one is obsolete. We don't have any finalizers on the SKR and also the ordering above doesn't make sense. Let's also re-iterate once the current behavior is established how to improve it in general. E.g., I think we should introduce a finalizer on the Kyma and first delete all modules.

Issue #2122, opened 2024-12-13
Running integration tests leaves orphaned processes


c-pius:
**time-boxed to 1d**

### Description

Running the integration tests seems to keep K8s-related processes running after they are finished (processes like `kube-apiserver`, `etcd`, ). When consecutively executing them for a larger amount of times, it eats up the machine resources and a restart is necessary.

### Steps to reproduce

run `make test` for a couple of times

### Environment Type

Managed

### Environment Info

Kubernetes Version: 1.30.3
Docker Desktop Version: 4.29.0
macOS: 15.1.1
Command Line Tools for Xcode: 16.1

### Attachments

![image](https://github.com/user-attachments/assets/b3aca2d7-038e-45fc-ae4e-b506bd39f2f2)

PR #2825, opened 2025-10-17
fix: Integration testing leaked processes on DualCluster

medmes:
**Description**

- Ensures subprocesses started by the DualCluster integration tests are launched with process-group semantics so they can be killed reliably.
- Adds explicit cleanup paths and deferred teardown in the test harness to stop background processes and wait for goroutines to exit.
- Adds a test/validation step that asserts there are no leftover processes matching the test helper binaries after test completion (prevents regressions).

**Related issue(s)**
#2122

### DIFF
--- a/tests/integration/commontestutils/skrcontextimpl/dual_cluster.go
+++ b/tests/integration/commontestutils/skrcontextimpl/dual_cluster.go
@@ -28,8 +28,7 @@ type DualClusterFactory struct {
 	clients sync.Map
 	scheme  *machineryruntime.Scheme
 	event   event.Event
-	skrEnv  *envtest.Environment
-	skrEnvs sync.Map
+	SkrEnvs sync.Map
 }
 
 func NewDualClusterFactory(scheme *machineryruntime.Scheme, event event.Event) *DualClusterFactory {
@@ -37,7 +36,7 @@ func NewDualClusterFactory(scheme *machineryruntime.Scheme, event event.Event) *
 		clients: sync.Map{},
 		scheme:  scheme,
 		event:   event,
-		skrEnvs: sync.Map{},
+		SkrEnvs: sync.Map{},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


tests/integration/commontestutils/skrcontextimpl/dual_cluster.go
```diff
@@ -10,27 +10,34 @@ import (
...
 
+type Stopper interface {
+	Stop() error
+}
+
 type DualClusterFactory struct {
 	clients sync.Map
 	scheme  *machineryruntime.Scheme
 	event   event.Event
 	skrEnv  *envtest.Environment
```
LeelaChacha: This `skrEnv` variable is unnecessary. It is never accessed outside of tests. Also it does not make sense to assign it if the skrEnvs map is used.

tests/integration/commontestutils/skrcontextimpl/dual_cluster_test.go
```diff
@@ -0,0 +1,190 @@
...
+
+func Test_LeakPrevention_VerifyStopperInterfaceHandling(t *testing.T) {
+	dualFactory := newFactory()
+	normalStopper := &fakeEnv{name: "normal"}
+	errorStopper := &fakeEnv{name: "error", stopErr: errors.New("stop error")}
+	leakyStopper := &leakyFakeEnv{name: "leaky", shouldSpawnGoroutine: true}
+	dualFactory.StoreEnv("normal", normalStopper)
+	dualFactory.StoreEnv("error", errorStopper)
+	dualFactory.StoreEnv("leaky", leakyStopper)
+	dualFactory.StoreEnv("non-stopper", "this is just a string")
```
LeelaChacha: This test also is not a standard usecase of the dual cluster factory. Instead of having this test, you should focus on ensuring that the value being stored in the map is of type envtest.Environment by having a strongly typed function parameter:

func (f *DualClusterFactory) StoreEnv(name string, env envtest.Environment) {

LeelaChacha (changes requested):
Needs changes

LeelaChacha (approved):
LGTM

PR #2837, opened 2025-10-28
feat: Patch Raw Manifest Deployments with Image Pull Secret

### LINKED ISSUE #2815: KLM configuring the Module deployment to use the pull secret
**Description**

In some of the restricted landscapes, public Docker images are strictly forbidden. In order for Module Manager to start, it needs to be properly configured with a pull secret.

**Acceptance Criteria**
- [ ] If configured, KLM should adjust the manifest of the module to include a pull secret for the module manager image

### LINKED ISSUE #2816: [Private Registries] Add image pull secret env var to module manager containers
### Description

Once an image pull secret is getting [synced and is available on the SKR](https://github.com/kyma-project/lifecycle-manager/issues/2814), all module's workloads (Deployments) that depend on image pulls need this reference to the secret.

Whenever the flag in KLM `--skr-image-pull-secret` introduced [here](https://github.com/kyma-project/lifecycle-manager/issues/2814) is set, KLM should add this reference to the module manager Deployment during module reconciliation, by adding a environment variable `SKR_IMG_PULL_SECRET` to containers.
```
[...]
spec:
  template:
    spec:
      containers:
        - name: name-of-container
          image: manager-image
          env:
            - name: SKR_IMG_PULL_SECRET
              value: name-value-from-flag
[...]
```

### Reasons

Kyma modules (and other Kyma-managed workloads) running on SKRs in private cloud landscapes need to support pulling images from landscape specific private OCI registries. For this scenario we add optional support for referencing image pull secrets for all our Kyma-managed workloads (KLM & modules).

### Notes

[Image pull secrets](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-pod-that-uses-your-secret) are Kubernetes secrets used to authenticate against private container registries when pulling images. This ensures access to restricted images during workload deployment. 
The Secret lookup is done within the namespace where the Deployment or ServiceAccount that defines the `imagePullSecrets` field is present.

### Acceptance Criteria

- [ ] If an image pull secret reference is passed as a flag to KLM, every module Deployment/StatefulState during manifest reconciliation is patched by adding an environment variable for the image pull secret in **every** `spec.template.spec.containers` element
- [ ] If `SKR_IMG_PULL_SECRET` env var entry is already found when parsing containers, Manifest reconciliation should fail

### Testing approach

- Unit
- E2E

LeelaChacha:
**Description**

- Rename ObjectTransform in Manifest Reconciler to ResourceTransform
- Add `Manager` to `ManifestSpec` which is a copy from `ModuleTemplateSpec`
- Have defaults for ResourceTransforms but add SkrImagePullSecretTransform to the list if `--skr-image-pull-secret` flag is set
  - This new ResourceTransform adds the value of `--skr-image-pull-secret` in the `imagePullSecrets` field of pod spec of all Deployments and Statefulsets.
  - And it adds the `SKR_IMG_PULL_SECRET` env var to all containers of the manager resource specified in `ManifestSpec`.
    - Throws `ErrSkrImagePullSecretEnvAlreadyExists` if the env var is set in raw manifest

**Related issue(s)**

Closes #2816
Closes #2815

### DIFF
--- a/tests/e2e/transferred_module_test.go
+++ b/tests/e2e/transferred_module_test.go
@@ -31,8 +31,9 @@ var _ = Describe("Module Transferred to Another OCI Registry", Ordered, func() {
 		// Their values should be visible in the Manifest's Spec.LocalizedImages field
 		// and in the actual values in the target "template-operator-controller-manager" Deployment in the SKR.
 		rewrittenTemplateOperatorImage = "europe-west3-docker.pkg.dev/sap-kyma-jellyfish-dev/restricted-market/" +
-			"kyma-project/prod/template-operator:1.0.3" +
-			"@sha256:e48f087538b17336d6e84c5e276d010158a482ef823f04f296df11c9ca2a047a"
+			"kyma-project/prod/template-operator:1.0.4" +
+			"@sha256:b561d0bad183d72bb4230e83a325587383dc967165bee336756672c82320bee9"
+
 		rewrittenKedaManagerImage = "europe-west3-docker.pkg.dev/sap-kyma-jellyfish-dev/restricted-market/" +
 			"kyma-project/prod/keda-manager:1.7.0" +
 			"@sha256:43b4efb42a08ffb4d463db4a2d859d22407316a813ff2e45e34437cb218bd143"


vvxxvvxx:
Hi @LeelaChacha , the env `SKR_IMG_PULL_SECRET` will be added through the deployment container env, and it can be overridden in the helm values, right?

LeelaChacha:
@vvxxvvxx

Correct. This PR patches the value of the lifecycle-manager flag `--skr-image-pull-secret` (can be overwritten in helm values) in the deployments (and stateful sets) of all modules.
The following fields are patched:
1. `.spec.template.spec.imagePullSecrets`: The secret is appended here.
2. `.spec.template.spec.containers[].env`: The secret is added as an env with name `SKR_IMG_PULL_SECRET`

internal/declarative/v2/image_pull_secret_transform.go
```diff
@@ -0,0 +1,180 @@
...
+	" environment variable already exits in raw manifest")
+
+func CreateSkrImagePullSecretTransform(secretName string) ResourceTransform {
+	return func(_ context.Context, obj Object, resources []*unstructured.Unstructured) error {
+		manifest, ok := obj.(*v1beta2.Manifest)
+		if !ok {
+			return fmt.Errorf("%w, got %T", ErrResourceTransformExpectedManifestType, obj)
+		}
+		for _, resource := range resources {
+			if resource.GetKind() == "Deployment" || resource.GetKind() == "StatefulSet" {
```
Tomasz-Smelcerz-SAP: Something to consider:
Here we're verifying if the resource is a Deployment or a StatefulSet.
Then we're passing the entire resource to the helper functions, where we again check if it's Deployment or StatefulSet - by making implicit assumption that these resources have certain structure: `spec.template.spec` (PodSpec).
This assumption is not, in any way, visible in the function's signatures...

Wouldn't it be more readable if we extract PodSpec right here, where we are explicitly selecting only the resources that have PodSpecs - so there's no doubt about our "intentions" - and just pass these extracted PodSpec to the helper functions?

In this way the knowledgle about resources that have PodSpec is located only here, and the other functions just handle the given PodSpecs.
This make those other functions a little bit more simple/readable - for example they don't need to know how to extract the PodSpec and you can give a meaningful name to the input parameter ('podSpec') thus making the function signature more explicit.
This improves readability because, for example, you don't have to carefully inspect the implementation to know what part of the resource can the function modify.

Of course, there is 'resourceIsManager` function that needs access to entire resource, but you can also evaluate it here or change the structure of the code a little bit - it should be relatively easy.

What do you think?

Tomasz-Smelcerz-SAP (dismissed):

Issue #2827, opened 2025-10-21
Image Rewrite should support `.` in image names


LeelaChacha:
### Description

The image rewrite regex pattern does not support `.` character in image names:
https://github.com/kyma-project/lifecycle-manager/blob/864632ec4f7e690ea5a2d8518d7585fe9246d7be/internal/imagerewrite/rewrite.go#L39C1-L47C2

Therefore, a valid image `europe-docker.pkg.dev/kyma-project/prod/external/gcr.io/kaniko-project/executor:v1.24.0` is causing errors.

We need to fix this bug and any other that might exist in the pattern, which we are not aware of.

### Acceptance Criteria
- [x] All valid image paths are allowed for now until we clarify what format is allowed in Kyma scope (see issue here...)

### Testing
- Unit

PR #2840, opened 2025-10-28
feat: Image rewrite using Docker library (distribution/reference) instead of regex patterns

medmes:
**Description**
As a follow up of this fix PR: https://github.com/kyma-project/lifecycle-manager/pull/2836

Replace fragile regex-based image parsing and rewriting with the official OCI-aware parsing library (github.com/docker/distribution/reference). This makes image normalization and rewrite logic safer, more correct, and easier to maintain.

As the one we use for the `modulectl`: https://github.com/kyma-project/modulectl/blob/main/internal/service/image/image.go

- Replaced the regex-based image parsing and rewrite codepaths with calls to the distribution/reference parsing library.
- Added unit tests covering

**Related issue(s)**
https://github.com/kyma-project/lifecycle-manager/issues/2827

### DIFF
--- a/internal/imagerewrite/rewrite.go
+++ b/internal/imagerewrite/rewrite.go
@@ -44,46 +44,13 @@ func NewDockerImageReference(val string) (*DockerImageReference, error) {
 		return nil, fmt.Errorf("%w: %s", ErrInvalidImageReference, err.Error())
 	}
 
-	res := &DockerImageReference{}
+	imageName := extractImageName(namedRef)
+	repoPath := extractRepositoryPath(namedRef)
 
-	domain := reference.Domain(namedRef)
-	path := reference.Path(namedRef)
-
-	familiarName := reference.FamiliarName(namedRef)
-	imageName := familiarName
-	if idx := strings.LastIndex(familiarName, "/"); idx != -1 {
-		imageName = familiarName[idx+1:]
-	}
-
-	// Derive repository path (everything before last segment).
-	repoPath := ""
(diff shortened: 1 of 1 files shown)


internal/imagerewrite/rewrite.go
```diff
@@ -35,45 +35,63 @@ type DockerImageReference struct {
...
-	matches := ociImagePattern.FindStringSubmatch(val)
-	if matches == nil {
-		return nil, ErrInvalidImageReference
+	// Get the standard components
+	domain := reference.Domain(ref)
+	path := reference.Path(ref)
+
+	// Re-create your non-standard HostAndPath/Name split
+	// This logic finds the last '/' to split the "repository path" from the "image name".
+	lastSlashIdx := strings.LastIndex(path, "/")
```
ruanxin: You can use the native function `reference.FamiliarName()` to get the name

ruanxin (changes requested):
We can use more code from library, and please also align the final implementation with modulectl, create a follow-up issue or a PR.

ruanxin (dismissed):

internal/imagerewrite/rewrite.go
```diff
@@ -35,45 +35,55 @@ type DockerImageReference struct {
...
+		if repoPath != "" {
+			res.HostAndPath = domain + "/" + repoPath
+		} else {
+			res.HostAndPath = domain
+		}
+	} else {
+		res.HostAndPath = repoPath
+	}
 
+	// Populate NameAndTag and Digest
```
ruanxin: If you defined some function with meaningful name, you don't have to leave such comment

PR #2845, opened 2025-10-30
fix: Error status on Kyma is not synced to SKR

Tomasz-Smelcerz-SAP:
**Description**

In some cases Kyma `error` status is not propagated back to SKR.
This happens when SKR Kyma is in `Ready` state and user introduces a change that transitions the KCP Kyma into the `Error` state.
An example of such change is configuring a module to use a non-existing channel.
This puts a KCP Kyma into 'Error' state, and this state should be propagated to the SKR.
However, a recently introduced logic makes an error value to be returned (`ErrKymaInErrorState`). This error value, not being `nil`, changes the code behavior. The method call to update SKR Kyma status never happens as the logic "errors out" earlier in the `reconcile()` method. 

-  Ensure error value returned because of Kyma being in 'Error` state doesn't affect the logic that syncs the status to the SKR.

**Related issue(s)**


internal/controller/kyma/controller.go
```diff
@@ -294,6 +294,11 @@ func (r *Reconciler) reconcile(ctx context.Context, kyma *v1beta2.Kyma) (ctrl.Re
 		return r.requeueWithError(ctx, kyma, fmt.Errorf("could not synchronize remote kyma status: %w", err))
 	}
 
+	if kyma.Status.State == shared.StateError {
+		// Requeue with a new Error in case of Kyma error state, to enable rate limiting for that error.
+		return ctrl.Result{}, ErrKymaInErrorState
+	}
+
 	return res, err
```
c-pius:
```suggestion
	return res, nil
```

Please double check but if I am not mistaken, `err` will always be nil here. Let's make it explicit

Issue #1957, opened 2024-10-14
[Module Catalog] Remove the old way of fetching all ModuleTemplates


nesmabadr:
### Description

We still support the old way of listing all the ModuleTemplates and then filtering them by the channel if there is no ModuleReleaseMeta existing for this module. After all modules teams have their ModuleReleaseMeta CRs deployed to the KCP, we need to remove this old code.

### Acceptance Criteria

- [x] Remove the code which lists all ModuleTemplates and only support the usage of ModuleReleaseMeta
- [x] ~Remove the old E2E tests which run without the ModuleReleaseMeta~
- [x] Refector [by_module_release_meta_strategy.go](https://github.com/kyma-project/lifecycle-manager/blob/d4ab0597b0761f6229feaee88446af5424eed5a1/pkg/templatelookup/moduletemplateinfolookup/by_module_release_meta_strategy.go) for all module, including mandatory module.
- [x] Remove mandatory moduletemplate lookup: https://github.com/kyma-project/lifecycle-manager/blob/6ccede3f8a9b1bac517bcfb433546a2364a9cc19/pkg/templatelookup/mandatory.go

### DevNotes:

The github action relies on old ModuleTemplates can be dropped complete:
https://github.com/kyma-project/lifecycle-manager/tree/df391286f3d549c1625f39bafb5276766e4fe602/.github/actions/deploy-template-operator

All old ModuleTemplates for e2e tests can be dropped: https://github.com/kyma-project/lifecycle-manager/tree/8da47c078172a56a632c3bd6ebb2063792fbfa54/tests/e2e/moduletemplate

After this change, only [by_module_release_meta_strategy.go](https://github.com/kyma-project/lifecycle-manager/blob/d4ab0597b0761f6229feaee88446af5424eed5a1/pkg/templatelookup/moduletemplateinfolookup/by_module_release_meta_strategy.go) will be used. There is no need to keep this strategy pattern.

janmedrek:
Prerequisite: https://github.com/kyma-project/lifecycle-manager/issues/2060

PR #2847, opened 2025-10-31
feat: Adapt templatelookup

lindnerby:
**Description**

- Delete both MT lookup strategies by version and by channel. Only keep MRM based lookup.
- Delete tests that don't have MRM as prerequisite.

**Related issue(s)**
* Relates to https://github.com/kyma-project/lifecycle-manager/issues/1957


pkg/templatelookup/moduletemplateinfolookup/lookup_test.go
```diff
@@ -0,0 +1,373 @@
...
+			Module: v1beta2.Module{Name: "test-module"},
+		},
+		&v1beta2.Kyma{},
+		moduleReleaseMeta)
+
+	require.Error(t, result.Err)
+	assert.Nil(t, result.ModuleTemplate)
+}
+
+func TestLookup_ReturnsError_WhenGetChannelVersionFails(t *testing.T) {
```
c-pius:
```suggestion
func TestLookup_ReturnsError_WhenNoVersionMappedToChannel(t *testing.T) {
```

See above

pkg/templatelookup/moduletemplateinfolookup/lookup_test.go
```diff
@@ -0,0 +1,373 @@
...
+		},
+		moduleReleaseMeta)
+
+	require.NoError(t, result.Err)
+	assert.NotNil(t, result.ModuleTemplate)
+	assert.Equal(t, "test-module", result.Spec.ModuleName)
+	assert.Equal(t, "1.0.0", result.Spec.Version)
+	assert.Equal(t, v1beta2.DefaultChannel, result.DesiredChannel)
+	assert.NotNil(t, result.ComponentId)
+}
```
c-pius: Is this a expected case? AFAIK, `KymaCR.spec.channel` is a required attribute that has a min length.

c-pius (commented):
Code-wise looks good, just some minor suggestions. Up to you if you want to apply them.

Regarding the ticket AC (I may not have been there when you discussed this in the team):

- Remove the old E2E tests which run without the ModuleReleaseMeta
  - this was done already earlier IIRC?
- Refector [by_module_release_meta_strategy.go](https://github.com/kyma-project/lifecycle-manager/blob/d4ab0597b0761f6229feaee88446af5424eed5a1/pkg/templatelookup/moduletemplateinfolookup/by_module_release_meta_strategy.go) for all module, including mandatory module
  - this was already done by Tomek earlier?

Regarding Dev Notes

- [/.github/actions/deploy-template-operator](https://github.com/kyma-project/lifecycle-manager/tree/df391286f3d549c1625f39bafb5276766e4fe602/.github/actions/deploy-template-operator)
  - was also already dropped earlier?
- All old ModuleTemplates for e2e tests can be dropped
  - there is still one left, but that is intentional right as we need it for the `module-transferred-to-another-oci-registry` test right?

PR #2867, opened 2025-11-11
chore: Hotfix setting label `mandatory: true` on Manifest

c-pius:
**Description**

- hotfix to fix a bug where Manifests for mandatory modules are not labeled with `operator.kyma-project.io/mandatory-module: "true"`
  - `spec.mandatory` has been removed from CRs and we consecutively failed to set this label on the Manifest which in turn led to these manifests beeing deleted as orphaned

**Related issue(s)**


LeelaChacha (approved):
LGTm

Issue #2856, opened 2025-11-06
Enhance SKR client caching


Tomasz-Smelcerz-SAP:
### Description

Recently observed problems with expired SKR clients on DEV can be adressed by introducing few improvements:

- Explicitly handle the "tls: expired certificate" error. While currently observed only on DEV, this is a legitimate error that is returned by standard library `tls` package, so it may occur in any environment. Adding support for doesn't require much code changes and the performance cost is acceptable (parsing error message).
- [Make the configured TTL for SKR clients to be respected](https://github.com/kyma-project/lifecycle-manager/issues/2853)
- Make the TTL for SKR clients configurable per-environment: Dev: 30 minutes, Stage: 4 hours, Prod: 8 hours. This provides additional "line of defence" against conditions we don't yet recognize in code. If a new error type appears that we don't yet explicitly recognize as a reason to evict the client from the cache, the TTL eviction will eventually restore a working connection to an SKR cluster.
- Reduce "invalidation bursts" - If there's a cluster of networking errors (like 5 in a row), which happens quite often on kubeconfig rotations, we end up in invalidating client cache 5 times. It means we'll re-create a client 5 times, which is waste of resources. We should extend the cache with a "rate limiter" that allows for just one invalidation in a second (an example), regardless of how many "invalidation requests" were issued.
- Eviction logging: We may want to log all evictions - just to closely monitor the cache behavior in case of problems - like the ones we have now. The logging should be off by default, but there should be a flag to enable it, if necessary.

### Acceptance Criteria

- [ ] Acceptance criterion 1
- [ ] Acceptance criterion 2

Issue #2853, opened 2025-11-06
TTL in remote client cache does not work as expected


Tomasz-Smelcerz-SAP:
### Description

TLDR: Instantiate the cache with [disableTouchOnHit option](https://github.com/jellydator/ttlcache/blob/db85e4f64251c73b33ba055e3fe07d70870992ce/options.go#L85)

#### Long version:

We're having two remote client caches:
- `internal/remote/client_cache.go` for Kyma reconciler
- `internal/service/skrclient/cache/client_cache.go` for Manifest reconciler

In both cases we're instantiating new cache with [default options](https://github.com/kyma-project/lifecycle-manager/blob/60474dcbc24f18f67c313fac42fe2ca2543d46a5/internal/remote/client_cache.go#L22) (empty list) and we're storing items (clients) with a TTL value of about 24 hours.
The idea is to automatically evict the clients from the cache after that time.
The regular eviction is important because we know that we **must stop using** the clients after some time, because the underlying kubeconfigs have expiration times on them.
Eviction is not a problem - if the client is not in the cache we generate a new one, based on the current kubeconfig value fetched from the secret.
This mechanism (if working) guarantees that we will not keep the outdated (expired) clients forever.

Unfortunately we made wrong assumptions with respect to how the `jellydator/ttlcache` works: The default behavior is to [**extend an item's expiration timestamp on successful retrieval**](https://github.com/jellydator/ttlcache/blob/0e73dcd73dc163074cf12e171c95b6d720e3d345/cache.go#L366-L368)
It means that if we *Get()* a client from these caches frequently enough, it will **never expire**, as each successful *Get()* extends cache's item (client) expiration time back to the initially configured value: *now() + 24 hours*.

Overall, this results in the TTL mechanism to be effectively disabled in the current codebase.

### Steps to reproduce

n/a

### Environment Type

Managed

### Environment Info
n/a

### Related Issues
#2856

c-pius:
State on DEV 2026-06-17T09:35

<img width="1110" height="645" alt="Image" src="https://github.com/user-attachments/assets/bb5ae568-8872-4468-a634-d1079e4cb4d9" />

c-pius:
State on DEV 2026-06-22T14:11

<img width="1165" height="445" alt="Image" src="https://github.com/user-attachments/assets/85618878-56cf-4862-819b-daacbbb15f76" />

State on DEV 2026-06-23T09:06

[...]

PR #2870, opened 2025-11-12
fix: Evict remote SKR clients on certificate expiration errors

Tomasz-Smelcerz-SAP:
**Description**

We're observing a service degradation on DEV: After a few hours from deployment all the Manifests and Kymas transition into `error` state.
The log analysis indicate expired TLS certificates: `remote error: tls: expired certificate`. This is a "kubeconfig" certificate used by go-client to interact with remote (SKR) API-Server.
This PR extends error handling in KLM to handle this specific issue.

**Technical Details**

Currently our remote clients cache in the Manifest Reconciler is evicted only in specific cases - there is a hard-coded list of error types and only these errors cause client invalidation.
The list is fixed for a reason: Consider temporary network outage. This should **not** cause eviction of the client - that particular client instance is not "wrong" in any way, we don't have to replace it with a "fixed" one.

The existing list lacks a "tls: expired certificate" error which means we're not evicting clients upon this error.
This is a problem as such client is definitely invalid and will **always** return error on every call until finally evicted by the cache's TTL mechanism. Which - to make things worse - [is currently not working.](https://github.com/kyma-project/lifecycle-manager/issues/2853)
Because of the mentioned problems we should invalidate such clients immediately.

- Detect the`tls: expired certificate` error from the client and use a custom error value to represent that condition:`ErrClientTLSCertExpired`. By doing so, we don't have to log the original error message which usually contains the entire serialized resource for which the "patch" operation failed. This makes our logs smaller and more readable. 
- Extend the list of errors that should evict the client cache with `ErrClientTLSCertExpired` error.
- "Collapse" the error from the client: The errors from the client are generated once for every resource in a module (paralles SSA). Normally all the errors are reported. To avoid a big and redundant message, we replace all the identical errors with a single one).

**Related issue(s)**
#2856

### DIFF
--- a/.github/actions/deploy-template-operator-with-modulereleasemeta/action.yml
+++ b/.github/actions/deploy-template-operator-with-modulereleasemeta/action.yml
@@ -4,7 +4,7 @@ inputs:
   module_version:
     description: "Version of the template operator to be deployed, should be aligned with the latest version released in github."
     required: true
-    default: "1.0.3"
+    default: "1.0.4"
 runs:
   using: composite
   steps:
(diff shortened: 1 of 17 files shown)


pkg/util/error.go
```diff
@@ -12,7 +12,14 @@ import (
 	"k8s.io/client-go/discovery"
 )
 
-var ErrClientUnauthorized = errors.New("ServerSideApply is unauthorized")
+const (
+	MsgTLSCertificateExpired = "expired certificate" // stdlib: crypto/tls/alert.go
```
c-pius: Just wondering if this could also be a private member of the ssa.go? Do we use it somewhere outside?

PR #2875, opened 2025-11-18
feat: Add kcp status usecase

lindnerby:
* Add usecase interface
* implement first usecase

### DIFF
--- a/internal/service/kyma/deletion/usecases/set_status_deleting_test.go
+++ b/internal/service/kyma/deletion/usecases/set_status_deleting_test.go
@@ -1,4 +1,4 @@
-package usecases
+package usecases_test
 
 import (
 	"context"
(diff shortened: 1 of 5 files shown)


internal/repository/kyma/status/kyma_status_repository.go
```diff
@@ -0,0 +1,56 @@
+package status
```
c-pius: We usually use a short for `repo` in the file names: `internal/repository/kyma/status/kyma_status_repo.go`

internal/service/kyma/deletion/usecases/set_status_deleting_test.go
```diff
@@ -0,0 +1,204 @@
+package usecases
```
c-pius: Tests not yet updated to the new repo func

PR #2876, opened 2025-11-19
feat: UseCase SetSkrKymaStateDeleting

c-pius:
**Description**

- implements UseCase SetSkrKymaStateDeleting

**Related issue(s)**

### DIFF
--- a/internal/remote/client_cache.go
+++ b/internal/remote/client_cache.go
@@ -14,6 +14,9 @@ const (
 	ttlInSecondsLower, ttlInSecondsUpper = 23 * 60 * 60, 25 * 60 * 60
 )
 
+// TODO: consider to introduce a type/alias for "kymaId" which is used uniformly to identify
+// a Kyma instance and is used here in the client as well as when calling services/repo that only
+// need info about what Kyma to process
 type ClientCache struct {
 	internal *ttlcache.Cache[client.ObjectKey, client.Client]
 }


internal/service/kyma/deletion/usecases/set_skr_kyma_state_deleting_execute_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	apimetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+	"github.com/kyma-project/lifecycle-manager/internal/result/kyma/usecase"
+	"github.com/kyma-project/lifecycle-manager/internal/service/kyma/deletion/usecases"
+	"github.com/kyma-project/lifecycle-manager/pkg/testutils/random"
+)
+
```
lindnerby: Let's make sure `IsApplicable()` is also behaving as expected.

internal/service/kyma/deletion/usecases/set_skr_kyma_state_deleting.go
```diff
@@ -0,0 +1,63 @@
...
+func (u *SetSkrKymaStateDeleting) IsApplicable(ctx context.Context, kyma *v1beta2.Kyma) (bool, error) {
+	if kyma.DeletionTimestamp.IsZero() {
+		return false, nil
+	}
+
+	if exists, err := u.skrAccessSecretRepo.Exists(ctx, kyma.GetName()); !exists || err != nil {
+		return false, err
+	}
+
+	status, err := u.kymaStatusRepo.Get(ctx, kyma.GetNamespacedName())
```
lindnerby: Not sure if this will work? the KCP Kyma will have a different namespace than the SKR Kyma obviously, but we are interested in the SKR Kyma status here, right?

internal/repository/skr/kyma/status/repo.go
```diff
@@ -0,0 +1,96 @@
...
+			APIVersion: v1beta2.GroupVersion.String(),
+		},
+		ObjectMeta: apimetav1.ObjectMeta{
+			Name:      shared.DefaultRemoteKymaName,
+			Namespace: shared.DefaultRemoteNamespace,
+		},
+		Status: v1beta2.KymaStatus{
+			State: shared.StateDeleting,
+			LastOperation: shared.LastOperation{
+				Operation:      ".status.State set to Deleting",
```
lindnerby: Let's `const` this message out

internal/repository/skr/kyma/status/repo.go
```diff
@@ -0,0 +1,96 @@
...
+		kyma,
+		client.Apply,
+		client.ForceOwnership,
+		fieldowners.LifecycleManager,
+	)
+}
+
+// TODO: this should work as long as we use the same client cache that we passed to KymaSkrContextProvider
+// it however depends on KymaSkrContextProvider.Init being called. As of now, this is the case, but we
+// should re-think how we use the client cache.
```
lindnerby: Good point! We should for sure enforce this as a Singleton and make it apparent that it actually is the same reference used throughout kyma reconciliation. ☝️

internal/service/kyma/deletion/usecases/set_skr_kyma_state_deleting.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (u *SetSkrKymaStateDeleting) Execute(ctx context.Context, kyma *v1beta2.Kyma) result.Result {
+	return result.Result{
+		UseCase: u.Name(),
+		Err:     u.kymaStatusRepo.SetStateDeleting(ctx, kyma.GetNamespacedName()),
+	}
+}
+
+func (u *SetSkrKymaStateDeleting) Name() result.UseCase {
+	return usecase.SetSkrKymaStateDeleting
```
lindnerby: May sound a bit redundant, but can we establish in these usecases that the usecase.* that is returned here is taken from a private const on top, similar to the other state setting usecase? So the mapping is immediately visible.
also no strong feelings on it, just for consistency and clarity.
So all functions point to the same constant

internal/service/kyma/deletion/usecases/set_skr_kyma_state_deleting.go
c-pius: It works, but agree that it may be confusing. The thing is that the passed `kcpKyma.GetNamespacedName()` is required to get the related skr client. Once we have the client, the Get actually works on the default names.

Agree that this is not ideal, but also don't have a good idea right now how this could be made more obvious.

```go
func (r *Repository) Get(ctx context.Context, kymaName types.NamespacedName) (*v1beta2.KymaStatus, error) {
	skrClient, err := r.getSkrClient(kymaName)
	if err != nil {
		return nil, err
	}

	kyma := &v1beta2.Kyma{}
	if err := skrClient.Get(ctx,
		types.NamespacedName{
			Name:      shared.DefaultRemoteKymaName,
			Namespace: shared.DefaultRemoteNamespace,
		},
		kyma,
	); err != nil {
		return nil, err
	}

	return &kyma.Status, nil
}
```

internal/repository/skr/kyma/status/repo.go
c-pius: Yes. But I am also still contemplating with refactoring the client cache overall. Let's also quickly discuss where we put in our effort first 👍🏻

internal/service/kyma/deletion/usecases/set_skr_kyma_state_deleting.go
c-pius: Let's also quickly talk about this one. Not sure if I understand you right. If I want to do a `const useCase = usecase.SetSkrKymaStateDeleting` at the top of the file, I get a conflict since we are sharing the same package between usecases and `useCase` const is already defined. I could go for `const useCaseSetSkrKymaStateDeleting = usecase.SetSkrKymaStateDeleting`, but then I am not sure if I address your concern.

PR #2877, opened 2025-11-19
chore: Drop `remote.Client` + `remote.ConfigAndClient` and configure SKR QPS + Burst explicitly

c-pius:
**Description**

- drops the `remote.Client` interface and `remote.ClientWithConfig` struct
- the only usage of this was in the SkrContextProvider to fetch the QPS/Burst settings from the KCP client and apply it to created SKR clients
  - I doubt that this even makes sense as the values should vary a lot between KCP and SKR
- for now, I made those settings explicit in the SkrContextProvider, and I made them configurable via flag with the same defaults as coming from client-go. We can use the flags to reduce the values and test how stage behaves. Once we found reasonable values, we can remove the flag again and just configure it in code

**Related issue(s)**

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -203,7 +203,7 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
 		remoteClientCache,
 		eventRecorder,
 		accessManagerService,
-	flagVar.SkrClientQPS,
+		flagVar.SkrClientQPS,
 		flagVar.SkrClientBurst)
 	var skrWebhookManager *watcher.SkrWebhookManifestManager
 	var options ctrlruntime.Options


internal/remote/skr_context_provider.go
```diff
@@ -63,7 +69,7 @@ func (k *KymaSkrContextProvider) Init(ctx context.Context, kyma types.Namespaced
 		return fmt.Errorf("failed to create lookup client: %w", err)
 	}
 
-	skrClient := NewClientWithConfig(clnt, restConfig)
+	skrClient := clnt
```
lindnerby: `skrClient` should be the var of the already above created client, and this statement removed.

c-pius: [a9e9233](https://github.com/kyma-project/lifecycle-manager/pull/2877/commits/a9e9233df6d1da423272b44d3d15aaafacd5c9df)

internal/pkg/flags/flags.go
```diff
@@ -180,13 +182,18 @@ func DefineFlagVar() *FlagVar {
...
 		DefaultWatcherRequeueSuccessInterval,
 		"Duration after which a Watcher in Ready state is enqueued for reconciliation.")
-
 	flag.Float64Var(&flagVar.ClientQPS, "k8s-client-qps", DefaultClientQPS,
 		"Maximum queries per second (QPS) limit for the Kubernetes client. Controls how many requests"+
 			" can be made to the Kubernetes API server per second in steady state.")
 	flag.IntVar(&flagVar.ClientBurst, "k8s-client-burst", DefaultClientBurst,
 		"Maximum burst size for throttling Kubernetes API requests. Allows temporarily exceeding the QPS"+
 			" limit when there are sudden spikes in request volume.")
+	flag.Float64Var(&flagVar.SkrClientQPS, "k8s-skr-client-qps", DefaultSkrClientQPS,
```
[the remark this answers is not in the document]
lindnerby: Doubting we will need to configure a floating point QPS below zero, and everything above is fine to be int imo.

internal/pkg/flags/flags_test.go
```diff
@@ -93,6 +93,16 @@ func Test_ConstantFlags(t *testing.T) {
...
 		},
+		{
+			constName:     "DefaultSkrClientQPS",
+			constValue:    strconv.Itoa(DefaultSkrClientQPS),
+			expectedValue: "5",
+		},
+		{
+			constName:     "DefaultSkrClientBurst",
+			constValue:    strconv.Itoa(DefaultSkrClientBurst),
+			expectedValue: "10",
```
lindnerby:
```suggestion
			expectedValue: "100",
```

mmitoraj (dismissed):

internal/remote/skr_context_provider.go
```diff
@@ -21,21 +21,27 @@ type SkrContextProvider interface {
...
+	skrQps               float32
+	skrBurst             int
 }
 
-func NewKymaSkrContextProvider(kcpClient Client,
+func NewKymaSkrContextProvider(kcpClient client.Client,
 	clientCache *ClientCache,
 	event event.Event,
 	accessManagerService *accessmanager.Service,
+	skrQps float32,
```
lindnerby:
```suggestion
	skrQps int,
```

c-pius:
This is part of the earlier 1.9.2 release which is looking good and STAGE and PROD.

PR #2884, opened 2025-11-24
chore: Test SelfEvictingClient

c-pius:
**Description**

- test the idea of a self-evicting client

**Related issue(s)**


c-pius:
I think this will work, but I am struggling with some tests and especially the integration test. Parking this for now until we have the deletion reconciler working.

PR #2896, opened 2025-11-27
feat: UseCase DeleteSkrCrds with instances for MT, MRM, Kyma

c-pius:
**Description**

- Implements UseCase for Deleting SKR CRDs
- provides instances for ModuleTemplate, ModuleReleaseMeta and Kyma CRDs to the deletion service

**Related issue(s)**


lindnerby:
> The only CRD we managed left at the SKR after this use case is the Kyma one. I was first planning on having a separate use case for this one but now thinking about it, actually I would merge them together. So `DeleteSkrModuleMetadata` would become a `DeleteSkrCrds`. WDYT?

How about we try with even further granularity and try with one usecase per CRD, with Kyma last so in the events we would also have this distinction. If it figures this is taking too many reconciliations until final deletion we can still combine usecases into one very easily.

PR #2897, opened 2025-11-27
chore: Refine and test deletion service

c-pius:
**Description**

- Refines the DeletionService to:
  - include the UseCase name if IsApplicable fails
  - include a default UseCase name on fall through (otherwise the event name would be empty)
- Adds unit tests for the DeletionService

**Related issue(s)**

### DIFF
--- a/internal/result/event/event_recorder.go
+++ b/internal/result/event/event_recorder.go
@@ -11,7 +11,11 @@ import (
 	"github.com/kyma-project/lifecycle-manager/internal/result"
 )
 
-const success = "Success"
+const (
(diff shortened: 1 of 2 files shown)


c-pius:
> If there is no usecase applicable anymore, we need to assume the deletion is done. I would not even introduce this Warning event. After all usecases have been handled, we even should not even requeue anymore, right?

Not entirely sure. As long as the KCP Kyma CR still exists, one use case must be applicable I think. If we continue to reconcile a Kyma but no use case applies, than it should be an error situation that we need to follow-up with?

PR #2906, opened 2025-12-02
feat: SKR Webhook & Watcher Cert Setup Deletion

lindnerby:
**Description**

- Provide usecase for the deletion of skr-webhook resources
- Provide usecase for the deletion of the watcher cert setup (certificate and secret)

### DIFF
--- a/internal/service/kyma/deletion/deletion_service_test.go
+++ b/internal/service/kyma/deletion/deletion_service_test.go
@@ -121,6 +121,8 @@ func Test_Delete_Fallthrough_WhenNoUseCaseIsApplicable(t *testing.T) {
 	uc7 := &useCaseStub{isApplicable: false, err: nil}
 	uc8 := &useCaseStub{isApplicable: false, err: nil}
 	uc9 := &useCaseStub{isApplicable: false, err: nil}
+	uc10 := &useCaseStub{isApplicable: false, err: nil}
+	uc11 := &useCaseStub{isApplicable: false, err: nil}
 
 	svc := kymadeletionsvc.NewService(
 		uc1,
@@ -131,9 +133,9 @@ func Test_Delete_Fallthrough_WhenNoUseCaseIsApplicable(t *testing.T) {
 		uc6,
 		uc7,
 		uc8,
-		nil,
-		nil,
 		uc9,
+		uc10,
+		uc11,
 	)
 
 	rslt := svc.Delete(t.Context(), kyma)
@@ -158,6 +160,10 @@ func Test_Delete_Fallthrough_WhenNoUseCaseIsApplicable(t *testing.T) {
 	assert.False(t, uc8.executeCalled)
 	assert.True(t, uc9.isApplicableCalled)
 	assert.False(t, uc9.executeCalled)
+	assert.True(t, uc10.isApplicableCalled)
+	assert.False(t, uc10.executeCalled)
+	assert.True(t, uc11.isApplicableCalled)
+	assert.False(t, uc11.executeCalled)
 	assert.Equal(t, kyma, uc1.receivedKyma)
 	assert.Equal(t, kyma, uc2.receivedKyma)
 	assert.Equal(t, kyma, uc3.receivedKyma)


internal/service/kyma/deletion/usecases/watcher_certificate_cleanup.go
```diff
@@ -0,0 +1,85 @@
+package usecases
```
c-pius: See the other comment for errors.Join for this file

internal/repository/skr/webhook/skrwebhook_repo.go
```diff
@@ -0,0 +1,166 @@
...
+// ResourcesExist checks if any of the skr-webhook resources exist in the SKR cluster for the given Kyma.
+func (r *ResourceRepository) ResourcesExist(kymaName string) (bool, error) {
+	kymaNamespacedName := types.NamespacedName{
+		Name:      kymaName,
+		Namespace: r.remoteSyncNamespace,
+	}
+
+	skrClient := r.skrClientCache.Get(kymaNamespacedName)
+
+	ctx, cancel := context.WithCancel(context.Background())
```
c-pius: Add the existing context to the `ResourcesExist` func signature instaed of creating a new background context?

internal/repository/skr/webhook/skrwebhook_repo.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// DeleteWebhookResources deletes all skr-webhook resources from the SKR cluster for the given Kyma.
+func (r *ResourceRepository) DeleteWebhookResources(ctx context.Context, kymaName string) error {
+	kymaNamespacedName := types.NamespacedName{
+		Name:      kymaName,
+		Namespace: r.remoteSyncNamespace,
+	}
+
+	skrClient := r.skrClientCache.Get(kymaNamespacedName)
```
c-pius: see comment above

internal/repository/skr/webhook/skrwebhook_repo.go
```diff
@@ -0,0 +1,166 @@
...
+			ref := r.resources[resIdx]
+			resource := &unstructured.Unstructured{}
+			resource.SetGroupVersionKind(schema.FromAPIVersionAndKind(ref.APIVersion, ref.Kind))
+			resource.SetName(ref.Name)
+			resource.SetNamespace(r.remoteSyncNamespace)
+
+			err := skrClient.Delete(grpCtx, resource)
+			if err != nil {
+				return fmt.Errorf("failed to delete resource %s: %w", ref.Name, err)
+			}
```
c-pius: I think we already need to ignore not found here. If I understant the errGrp.Wait correclty, it returns the first non nill error. If we have something like:

- err1: not found
- err2: some real error

Then the `err` below is `err1` and since we ignore not found there, we would miss the real error in err2.

internal/repository/skr/webhook/skrwebhook_repo_test.go
```diff
@@ -0,0 +1,405 @@
...
+		repo := webhook.NewResourceRepository(clientCache, remoteSyncNamespace, baseResources)
+
+		exists, err := repo.ResourcesExist(kymaName)
+
+		require.Error(t, err)
+		assert.False(t, exists)
+		assert.Contains(t, err.Error(), "failed to check resource")
+	})
+
+	t.Run("short-circuits on first found resource", func(t *testing.T) {
```
c-pius: What is this test adding on to pof the `returns true when at least one resource exists`?

internal/repository/skr/webhook/skrwebhook_repo.go
```diff
@@ -0,0 +1,166 @@
...
+	skrClient := r.skrClientCache.Get(kymaNamespacedName)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	errGrp, grpCtx := errgroup.WithContext(ctx)
+	resourceExists := make(chan bool, 1)
+
+	for idx := range r.resources {
+		resIdx := idx
```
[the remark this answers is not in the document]
lindnerby: No reason 😆 Removed

internal/repository/skr/webhook/skrwebhook_repo.go
c-pius: I think we don't have a cancel from outside anyway. So would only be the ctx we receive from the Reconcile func, and we add the cancel within the func

PR #2914, opened 2025-12-03
chore: Add `instance-id` label to `deploy_kyma.sh` test script

c-pius:
**Description**

- updating state metrics requires instance id label: https://github.com/kyma-project/lifecycle-manager/blob/main/internal/pkg/metrics/kyma.go#L82-L85
- in local testing we were seeing failures for writing metrics, therefore providing an arbitrary uuid to the test script

**Related issue(s)**


scripts/tests/deploy_kyma.sh
```diff
@@ -41,6 +41,8 @@ metadata:
     skr-domain: "example.domain.com"
   name: kyma-sample
   namespace: kcp-system
+  labels:
+    kyma-project.io/instance-id: "b34bed57-ecd0-4d68-81a7-966e114954b8"
```
lindnerby:
```suggestion
    kyma-project.io/instance-id: "b34bed57-ecd0-4d68-81a7-966e114954b8"
    kyma-project.io/runtime-id: "kyma-sample"
```

PR #2918, opened 2025-12-04
feat: Use Runtime ID in SkrEventHandler Instead of Owner Field

### LINKED ISSUE #2439: [SKR Watch] Refactor skrEventHandler to map GenericEvent to the namespaced-name of the affected Kyma
### Description

Refactor the [`skrEventHandler`](https://github.com/kyma-project/lifecycle-manager/blob/main/internal/controller/kyma/setup.go#L79) in KLM to work with the updated GenericEvent, using the new skr field for SKR metadata, instead of relying on the deprecated owner field.

### Reasons

Simplifying and centralizing the SKR identity handling avoids redundant ownership lookups and SAN pinning checks, which reduces complexity and improves reliability in KLM.

### Acceptance Criteria
- [x] Update `skrEventHandler` to extract necessary info from watched, watched-gvk, and skr fields only
- [x] Map the event data to the correct namespaced-name of the Kyma resource to reconcile

### Testing Approach
- Unit Testing of skrEventHandler
- E2E Testing of SKR event processing still works, but now works on the updated event format

LeelaChacha:
**Description**

- Changed the watcher event handling logic: now it's based on `runtimeID` value instead of the `owner` field
- Refactored the relevant code for the incoming watcher event handling
- Removed unused watch events listener from the Manifest reconciler: `internal/controller/manifest/setup.go`
- Added Kyma Lookup Service allowing to lookup Kymas
 by labels

**Related issue(s)**

Closes #2439


internal/controller/manifest/setup.go
```diff
@@ -53,48 +41,12 @@ func SetupWithManager(mgr manager.Manager,
...
-	if settings.EnableDomainNameVerification {
-		// Verifier used to verify incoming listener requests
-		verifyFunc = security.NewRequestVerifier(mgr.GetClient()).Verify
-	} else {
-		verifyFunc = func(r *http.Request, watcherEvtObject *types.WatchEvent) error {
-			return nil
-		}
-	}
-
-	runnableListener := watcherevent.NewSKREventListener(
```
LeelaChacha: For Review: This was not being used and is conceptually wrong to place the listener in the manifest reconciler. Hence, removed after discussing with the team

Issue #2484, opened 2025-05-30
[SKR Watch] Introduce Explicit manager Field in Watcher CR to Replace Overloaded managed-by Label


ruanxin:
### Description

Currently, the operator.kyma-project.io/managed-by label in the Watcher custom resource (CR) is used to indicate which component is responsible for handling the webhook requests. For example, when the request is directed to KLM (Kyma Lifecycle Manager), the value is set to lifecycle-manager.

However, the same label is also reused for a different purpose: to indicate which module resources in SKR are managed by Kyma. Within the KLM codebase, the label is primarily interpreted in this second context: https://github.com/kyma-project/lifecycle-manager/blob/80de252cc8320f4c9f96723a59516f63afa525b6/api/shared/operator_labels.go#L12-L14

This dual usage creates confusion and makes the business logic ambiguous.

### Reasons

Introduce a dedicated field in the Watcher CR to explicitly define the manager component responsible for the webhook, thereby separating concerns and clarifying intent.

### Acceptance Criteria

- [x] Add a new optional `spec.manager` field in the Watcher CR to designate the component managing the webhook.
  - [x] Update in the kustomization watcher file: https://github.com/kyma-project/lifecycle-manager/blob/2b4b9a82f9501b4d47fba0ef1f4a32ee3e1fc8dd/config/watcher/kyma_watcher.yaml#L7 
  - [x] Refactor the method to read manager from new field instead of label https://github.com/ruanxin/kyma-operator/blob/39813abda62c497fa8d0e8f7b2c63dc218006d36/api/v1beta2/watcher_types.go#L114
- [x] Update watcher document: https://github.com/kyma-project/runtime-watcher/blob/main/docs/api.md#labels
- [x] Have a integration test to make sure the deployed `kyma_watcher.yaml` `spec.manager` equals with `shared.OperatorName`
- [x] Remove `strings.ToLower` https://github.com/kyma-project/lifecycle-manager/blob/d516291831aaf94f4cb12a40c8bd9482c0c26aa6/internal/controller/manifest/setup.go#L55 
- [x] Check in the management-plane-charts PR that the watcher spec filed is configured properly and `managed-by` label in Watcher CR is removed (Canary Verification).
- [x] Create new issue about Watcher API upgrade to make this field as mandatory 

File in the `management-plane-charts` repository: `management-plane-charts/blob/chart/lifecycle-manager/lifecycle-manager/templates/watcher.yaml`
### Feature Testing

[...]

PR #2921, opened 2025-12-04
feat: Introduce Explicit manager Field in Watcher CR to Replace Overloaded managed-by Label

medmes:
**Description**

Adds explicit spec.manager to Watcher and deprecates the managed-by label; updates code, CRD, configs and tests.

**Related issue(s)**
#2484

### DIFF
--- a/api/v1beta2/watcher_types.go
+++ b/api/v1beta2/watcher_types.go
@@ -114,7 +114,18 @@ type Watcher struct {
 	Status WatcherStatus `json:"status,omitempty"`
 }
 
-// GetManagerName returns the manager component name responsible for handling webhook requests.
+// GetManagerName returns the name of the manager component responsible for handling webhook requests.
+//
+// The manager name is determined from spec.Manager field (preferred) or falls to the
+// 'operator.kyma-project.io/managed-by' label for backward compatibility.
+//
+// This value is used in two places:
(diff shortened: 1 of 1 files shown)


.golangci.yaml
```diff
@@ -242,6 +242,9 @@ linters:
       - linters:
           - lll
         source: "^// http "
+      - linters:
+          - staticcheck
+        text: "SA1019.*shared.ManagedBy is deprecated"
```
medmes: For the reviewer as we introduced the deprecation of the managedBy label. The linter claims of still usage of the managedBy which are not related to the watcher CR. For now , supressing the deprecation message only temporary and will be removed on the follow up ticket.

config/crd/bases/operator.kyma-project.io_watchers.yaml
```diff
@@ -340,6 +340,11 @@ spec:
                   type: string
                 description: LabelsToWatch describes the labels that should be watched
                 type: object
+              manager:
+                description: |-
+                  Manager specifies the component responsible for handling webhook requests.
+                  If not set, falls back to the 'operator.kyma-project.io/managed-by' label.
```
lindnerby: The description here is wrong. The `manager` field does not fall back to anything.

Issue #2928, opened 2025-12-09
Cache invalidation logic incomplete in Manifest Reconciler


Tomasz-Smelcerz-SAP:
### Description

The error:
```
{"level":"ERROR","date":"[...]","caller":"controller/controller.go:474","msg":"Reconciler error",
"context":{"controller":"manifest","controllerGroup":"operator.kyma-project.io","controllerKind":"Manifest",
    "Manifest":{"name":"<some-name>","namespace":"kcp-system"},"namespace":"kcp-system",
        "name":"<some-name>1","reconcileID":"<some-reconcile-id>",
        "error":"failed rest mapping [PriorityClass.scheduling.k8s.io, v1]:
             Get \"https://api.<some-url>.dev.kyma.ondemand.com/api?timeout=32s\": remote error: tls: expired certificate
```

The error originated from here: 

### Steps to reproduce

Unknown. The `getResourceMapping()` function is executed in *many* places, but so far we did not recognize this problem. Probably because this resource mapping is cached internally by the code in `apimachinery` module and it returns cached contents. This mapping must be available before any other important call: List, Create, Apply, Delete is made.
What we know is that it happens very rarely

### Environment Type

Managed

### Environment Info

Kubernetes Version: x.y.z
Docker Version: x.y.z
modulectl Version: x.y.z
If used - k3d/k3s/minikube versions

PR #2931, opened 2025-12-10
fix: Ensure client eviction on SKR client certificate expiration

Tomasz-Smelcerz-SAP:
**Description**

- detect "raw" certificate expiration-related errors to the `IsConnectionRelatedError()` function
- updated go docs with "why" the additional checks were added
- moving error-related constants/functions to the common `errors` package

**Related issue(s)**
Fixes: #2928


pkg/util/error.go
```diff
@@ -84,3 +90,23 @@ func NestedErrorMessage(err error) string {
...
+// we need to handle these as well.
+func isRawCertRelatedError(err error) bool {
+	return IsUnauthorizedError(err) || IsTLSCertExpiredError(err)
+}
+
+func IsUnauthorizedError(src error) bool {
+	return strings.HasSuffix(strings.TrimRight(src.Error(), " \n"), msgUnauthorized)
+}
+
+func IsTLSCertExpiredError(src error) bool {
```
lindnerby: Same here: Would be interesting to see some wrapped errors that will result in `true` here

PR #2932, opened 2025-12-10
chore: Compose KymaDeletionService in integration tests

c-pius:
**Description**

- fixes integration tests by providing a deletion service to the test suites
- note that `withwatcher` already had the SkrWebhookManager setup which is now also required by `kcp` and `kyma` testsuites
  - moved out parts of the setup logic from `withwatcher` to a `tests/integration/controller/composition` pkg for re-use from `kcp` and `kyma` suites

**Related issue(s)**

### DIFF
--- a/tests/integration/commontestutils/skrcontextimpl/dual_cluster.go
+++ b/tests/integration/commontestutils/skrcontextimpl/dual_cluster.go
@@ -91,7 +91,7 @@ func (f *DualClusterFactory) Init(_ context.Context, kyma types.NamespacedName) 
 	f.clientCache.Add(kyma, skrClient)
 
 	// track this envtest so Stop() can stop all started envs
-	f.SkrEnvs.Store(kyma, skrEnv)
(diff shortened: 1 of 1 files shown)


tests/integration/commontestutils/skrcontextimpl/dual_cluster.go
```diff
@@ -137,7 +142,6 @@ func (f *DualClusterFactory) Stop() error {
 			}
 		}
 		f.SkrEnvs.Delete(key)
-		f.clients.Delete(name)
```
c-pius: This doesn't work anymore since we don't have the NamespacedName available. Providing it to here would also be a not entirely trivial thing.

I opted to remove and not delete it as it is anyway in memory and will be cleaned up by the garbage collection. The important thin gis that we call `stopper.Stop()` above to kill etcd/kube-apiserver processes on the machine (which is anyway not working probably currently...)

lindnerby (approved):

Issue #2446, opened 2025-05-05
Implement a dedicated service for handling Kyma deprovisioning


ruanxin:
### Description
Currently, Kyma controller handles all reconciliation tasks. As the complexity of the reconciliation logic increases, mixing both provisioning and deprovisioning together makes the business logic hard to maintain and read.

Typical Symptom: We can find many specific condition handling `if GetDeletionTimestamp().IsZero()` to control when the CR is under deletion, and even worse, the code must follow a specific order.

To improve clarity and simplify the logic, we can extract deletion logic into a separate service.

### AC
- [x] Implement a dedicated service for handling Kyma deprovisioning
  - [x] Invoke this service when the Kyma's DeletionTimestamp is not zero
- [ ] Remove existing Kyma deprovisioning logic from the current Kyma controller

PR #2942, opened 2025-12-12
feat: Dedicated Kyma Deletion Service

lindnerby:
**Description**

- Extract a dedicate Service for deprovisioning when Kyma is deleting
- The usage of the new Service can be rolled back via env var to use legacy deletion in case something goes wrong in the landscapes
- Apply events and requeue behavior after each deletion step (usecase)
- Usecases will be run in sequential order and only continue with the next one if the previous one is completed
- Follow-up for dropping the legacy deletion parts if this is working as expected can be found [here](https://github.com/kyma-project/lifecycle-manager/issues/2943) 

**Related issue(s)**
* https://github.com/kyma-project/lifecycle-manager/issues/2446

### DIFF
--- a/tests/integration/controller/withwatcher/controller_test.go
+++ b/tests/integration/controller/withwatcher/controller_test.go
@@ -11,6 +11,7 @@ import (
 	apiappsv1 "k8s.io/api/apps/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/kyma-project/lifecycle-manager/api/shared"
 	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
 	skrwebhookresources "github.com/kyma-project/lifecycle-manager/internal/service/watcher/resources"
 	. "github.com/kyma-project/lifecycle-manager/pkg/testutils"


internal/service/kyma/deletion/deletion_service.go
```diff
@@ -0,0 +1,71 @@
...
+	Execute(ctx context.Context, kyma *v1beta2.Kyma) result.Result
+	Name() result.UseCase
+}
+
+type Service struct {
+	deletionSteps []UseCase
+}
+
+func NewService(
+	setKcpKymaStateDeleting UseCase,
```
Tomasz-Smelcerz-SAP: This is very brittle... There's no compile-time check of these types...
There are 11 arguments here, one can provide the same object 11 times and the compiler will be happy. Any order of these arguments also makes compiler happy.
The only thing that makes this work is the developer being very careful in providing the right objects at the correct positions...
I think we should find an alternative that makes the mentioned mistakes compile-time errors.

internal/service/kyma/deletion/usecases/delete_skr_crd.go
```diff
@@ -0,0 +1,53 @@
...
+		return false, err
+	}
+
+	return u.skrCrdRepo.Exists(ctx,
+		kcpKyma.GetNamespacedName(),
+	)
+}
+
+func (u *DeleteSkrCrd) Execute(ctx context.Context, kcpKyma *v1beta2.Kyma) result.Result {
+	// deleting the CRDs is sufficient as this also deletes related CRs
```
Tomasz-Smelcerz-SAP: to be honest I don't get this comment.

internal/controller/kyma/controller.go
```diff
@@ -144,6 +168,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return r.requeueWithError(ctx, kyma, err)
 	}
 
+	if !kyma.DeletionTimestamp.IsZero() {
+		useLegacyKymaDeletion, ok := os.LookupEnv("ENABLE_LEGACY_KYMA_DELETION")
+		if !ok || useLegacyKymaDeletion != "true" {
```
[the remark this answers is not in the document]
c-pius: https://github.com/kyma-project/lifecycle-manager/pull/2942/changes/8afdaf7debff8d5d735186449baf56a35fe0b83e

paranoid me also tested that it works locally 👍🏻

Tomasz-Smelcerz-SAP (approved):

c-pius (approved):

PR #2946, opened 2025-12-15
chore: Bump sec scanners to 1.10.0; RW to 2.0.9

c-pius:
**Description**

- ...
- ...
- ...

**Related issue(s)**


c-pius:
> ❌ Manifests created with 'make dry-run-control-plane' changed! Please make sure to check if changes are needed in related repositories like management-plane-charts, runtime-watcher, etc

probably the RW bump, need to investigate

Issue #2440, opened 2025-04-29
[SKR Watch] Remove Verify func


lindnerby:
### Description

Update the Kyma Lifecycle Manager (KLM) to clean up deprecated verification logic for listener event parsing and move to the new Runtime Watcher version.

### Reasons

Since certificate parsing now happens on the Runtime Watcher side, KLM no longer needs to verify SKR domains manually. This simplifies KLM’s event handling code significantly.

### Acceptance Criteria
- [x] Update KLM to consume the latest RW version, after https://github.com/kyma-project/runtime-watcher/issues/549 implemented and released
- [x] Remove the [verification logic](https://github.com/kyma-project/lifecycle-manager/blob/493276d28f7af81158bbad5c7f3f7a3e44c42344/pkg/security/san_pinning.go#L56) based on san_pinning.go
- [x] Cleanup and remove related tests around verification
- ~~[ ] Ensure skrEventHandler depends only on event payload fields (watched, watched-gvk, skr)~~

### Testing Approach
- Unit Testing
- E2E Testing for event-driven reconciliation

PR #2965, opened 2025-12-31
chore: Deprecate san-pinning flag

Tomasz-Smelcerz-SAP:
**Description**
See: #2966 
This PR cleans up the related code, documentation and marks the `enable-domain-name-pinning` flag as deprecated.

- Cleanup unused code
- Cleanup the docs
- Mark the flag as deprecated

**Related issue(s)**
#2966

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,8 +40,8 @@ require (
 )
 
 require (
-	istio.io/api v1.28.1
-	istio.io/client-go v1.28.1
+	istio.io/api v1.28.2
+	istio.io/client-go v1.28.2
 )
 
 require (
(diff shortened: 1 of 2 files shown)


c-pius (approved):
I am a bit surprised why parts of the san pinning functionality have been removed here already, but not entirely? https://github.com/kyma-project/lifecycle-manager/pull/2961

do you know why we did that?

Tomasz-Smelcerz-SAP:
> I am a bit surprised why parts of the san pinning functionality have been removed here already, but not entirely? #2961
> 
> do you know why we did that?

I think it was simply overlooked/forgotten by the author. @medmes any comment on that?
I noticed these parts should be removed while reviewing the original PR. But then it became obvious that the flag is no longer necessary.
To avoid making the PR even bigger and adding additional concerns to it, I decided It was simpler and faster to approve the PR as it was, and collect these left-overs here.
In addition we have a separate issue for the cleanup: https://github.com/kyma-project/lifecycle-manager/issues/2440

PR #3012, opened 2026-01-23
fix: Check all Module CRs across namespaces before deletion

### LINKED ISSUE #3006: KLM doesn't seem to consider Module CRs outside of the Default Module CRs namespace for safe deletion
### Description

In this ADR, it was aligned how KLM shall handle Module CRs as a gate before module deletion: https://github.com/kyma-project/community/issues/972. Specifically it reads in the consequences:

> For both CustomResourcePolicy: CreateAndDelete and CustomResourcePolicy: Ignore, lifecycle-manager uses ALL CRs of the Module CRD as the "gate" before proceeding with the deletion of the module. Lifecycle-manager does NOT report any configuration status of the module in the Kyma status.

However, it looks like the check implemented only checks for Module CRs part of namespace of the Default Module CR: https://github.com/kyma-project/lifecycle-manager/blob/a701253e856c4973eea0175cdc07f97a6fb7f4a9/internal/manifest/modulecr/client.go#L149-L156

This seems to be wrong and should be fixed.

### Steps to reproduce

- create test cluster on DEV
- enable telemetry module
- deploy a second Telemetry CR in default namespace
- remove the module from Kyma CR and observe if deletion is blocked or not
- if it is not blocked, then we have a problem

### Environment Type

Managed

### Environment Info

Kubernetes Version: x.y.z
Docker Version: x.y.z
modulectl Version: x.y.z

### Attachments

_No response_

medmes:

 **Core fix:**
- Modified `GetAllModuleCRsExcludingDefaultCR` to search Module CRs across ALL namespaces using empty `ListOptions{}`
- Implemented `listResourcesByGroupKindInAllNamespaces` with `REST mapper` for API version compatibility
- Removed redundant namespace-specific function

**Testing:**
- Consolidated 4 similar e2e tests into single file with shared helper function `testModuleDeletionBlocking`
- Added unit test `TestClient_GetAllModuleCRsExcludingDefaultCR_WithCRsInDifferentNamespaces`])
- Test coverage: multiple CRs (same/cross-namespace) × policies (Ignore/CreateAndDelete)

**Code quality improvements:**
- Fixed test to create default CR in correct client for proper filtering validation
- Changed parameter from `GroupVersionKind` → `GroupKind` for semantic accuracy
- Reduced test code by ~60% (800+ → 285 lines) through consolidation

**Related issue:**
Fixes #3006

### DIFF
--- a/.github/actions/deploy-template-operator-with-modulereleasemeta/action.yml
+++ b/.github/actions/deploy-template-operator-with-modulereleasemeta/action.yml
@@ -130,6 +130,7 @@ runs:
         matrix.e2e-test == 'modulereleasemeta-watch-trigger' ||
         matrix.e2e-test == 'modulereleasemeta-with-obsolete-moduletemplate' ||
         matrix.e2e-test == 'labelling' ||
+        matrix.e2e-test == 'module-deletion-with-only-default-cr' ||
         matrix.e2e-test == 'module-deletion-blocking-with-multiple-crs-ignore-policy' ||
         matrix.e2e-test == 'module-deletion-blocking-with-multiple-crs-create-and-delete-policy' ||
         matrix.e2e-test == 'module-deletion-blocking-with-default-cr-with-finalizer' ||
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -54,6 +54,7 @@ jobs:
           - module-without-default-cr
           - module-consistency
           - non-blocking-deletion
+          - module-deletion-with-only-default-cr
           - modulereleasemeta-upgrade-under-deletion
           - module-api-upgrade-under-blocked-deletion
           - purge-controller
--- a/tests/e2e/Makefile
+++ b/tests/e2e/Makefile
@@ -106,6 +106,9 @@ mandatory-module:
 non-blocking-deletion:
 	$(GO_TEST) "Non Blocking Kyma Module Deletion"
 
+module-deletion-with-only-default-cr:
+	$(GO_TEST) "Non Blocking Module Deletion With Only Default CR"
+
 skip-manifest-reconciliation:
 	$(GO_TEST) "Manifest Skip Reconciliation Label"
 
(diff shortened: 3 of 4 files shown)


internal/manifest/modulecr/client.go
```diff
@@ -211,6 +210,38 @@ func (c *Client) listResourcesByGroupKindInNamespace(ctx context.Context,
 	return allItems, nil
 }
 
+// listResourcesByGroupKindInAllNamespaces lists all resources matching the given GroupVersionKind
+// across ALL namespaces. This is required by ADR #972 to check all Module CRs before deletion.
+func (c *Client) listResourcesByGroupKindInAllNamespaces(ctx context.Context,
+	gvk schema.GroupVersionKind,
```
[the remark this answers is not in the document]
c-pius: The GroupKind can be retrieved from the moduleCR by `defaultModuleCR.GroupVersionKind().GroupKind()`

tests/e2e/module_deletion_blocking_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+		It("When remaining user-managed CRs are deleted", func() {
+			for _, cr := range userCRs[1:] {
+				Eventually(DeleteModuleCR).
+					WithContext(ctx).
+					WithArguments(cr.name, cr.namespace, skrClient).
+					Should(Succeed())
+			}
+		})
+	} else {
```
c-pius: Better use a else if len == 1. Otherwise we may get hard to find problems when we misconfigure the test

tests/e2e/module_deletion_blocking_test.go
```diff
@@ -0,0 +1,332 @@
...
+			moduleCR := &unstructured.Unstructured{}
+			moduleCR.SetGroupVersionKind(schema.GroupVersionKind{
+				Group:   templatev1alpha1.GroupVersion.Group,
+				Version: templatev1alpha1.GroupVersion.Version,
+				Kind:    string(templatev1alpha1.SampleKind),
+			})
+			Expect(skrClient.Get(ctx, defaultCRNamespacedName, moduleCR)).To(Succeed())
+			Expect(AddFinalizerToModuleCR(ctx, skrClient, moduleCR, "test-finalizer")).To(Succeed())
+		})
+	} else if len(userCreatedCRs) > 0 {
```
c-pius: I don't understand why this is an `else if` rather then a separate `if`. Can you elaborate?

Issue #2981, opened 2026-01-19
Investigate `Issuing certificate as Secret does not exist` error upon `istio-system/klm-watcher-serving` Cert rotation


c-pius:
### Description

In December and January we observed `Issuing certificate as Secret does not exist` errors when the `istio-system/klm-watcher-serving` Cert (CA cert) was rotated. December was an automated rotation, January was a manual one. In Jan, we observed the error both on STAGE and PROD.

In KLM, the rotation is implemented by observing the CA cert. Once it was rotated, KLM starts deleting all Secrets belonging to a SKR client cert to force cert-manager to re-issue the certificate based on the new CA cert. It seems like KLM is deleting the secrets faster than cert-manager re-issuing them. Therefore, many SKR certs transition to unready state with the message `Issuing certificate as Secret does not exist`. Further, this error even propagates to Kyma status setting them to error state. All of them self-heal after some time, but still this observable behavior is problematic.

We have only started to observe this behavior from December on. Shortly before (November), the KCP version of cert-manager was updated from 1.18.X to 1.19.X. It could be that the behavior from cert-manager changed with this update. It could also as well be that the behavior was always like described above, we just missed to observe it.

We need to investigate if this behavior was introduced with the cert-manager minor bump. In any case, we need to find a way to prevent Kymas transitioning to Error state during CA cert rotation. First ideas:

- find a way to get the SKR client certs rotated without deleting the secret
- throttling KLM deleting secrets so that cert-manager can keep up with re-issuing them

### Steps to reproduce

Try the following with cert-manager 1.18 and 1.19

- create a local test step
- run a kubectl script creating 1000 certs
- wait for cert-manager to issue all certs
- run a kubectl script deleting all 1000 secrets belonging to the cert
- observe whether the certs transition to unready state with `Issuing certificate as Secret does not exist`

### Environment Type

Managed

### Environment Info

Kubernetes Version: x.y.z
Docker Version: x.y.z
modulectl Version: x.y.z
If used - k3d/k3s/minikube versions

c-pius:
Further side quest:

[...]

PR #3019, opened 2026-01-28
feat: Re-Issue SKR Certificate by setting 'Issuing' condition 

c-pius:
**Description**

- apologies for the big PR, but it appears to be bigger than it actually is because code (especially test code) was moved
- the actual change this PR introduces is in the renewal of the SKR certificate. This includes two things
  - 1) whether the renewal is due is checked based on the "Not Before" time of the certificate, not as previously the creation timestamp of the secret.
  - 2) instead of deleting the related secret to force renewal, the condition as explain in https://github.com/kyma-project/lifecycle-manager/issues/2981#issuecomment-3798245513 is set on the certificate.
- these two changes above allowed for some further code alignment/restructuring
  - previously, there were specific renewal services for GCM and cert-manager, partly also relying on renewal specific repositories. Now, we have one shared renewal service that can work with both, GCM and cert-manager repo implementations.
  - most of the PR diff originates from this restructuring as also the tests had to be moved and adapted.

**Related issue(s)**

- closes https://github.com/kyma-project/lifecycle-manager/issues/2981


internal/repository/watcher/certificate/gcm/certificate/certificate_repo.go
```diff
@@ -121,12 +122,28 @@ func (r *Repository) Delete(ctx context.Context, name string) error {
...
+
+	cert.Spec.Renew = boolPtr(true)
+
+	if err = r.update(ctx, cert); err != nil {
+		return fmt.Errorf("failed to update certificate for renewal: %w", err)
+	}
 
-	err := r.kcpClient.Get(ctx, client.ObjectKeyFromObject(cert), cert)
+	return nil
+}
```
c-pius: This moved from the gcm renewal service into the repo

cmd/composition/service/skrwebhook/skr_webhook_manager.go
```diff
@@ -41,10 +41,19 @@ var (
...
 
+type CertificateRepository interface {
+	Create(ctx context.Context, name, commonName string, dnsNames []string) error
+	Delete(ctx context.Context, name string) error
+	Exists(ctx context.Context, certName string) (bool, error)
+	Renew(ctx context.Context, name string) error
+	GetRenewalTime(ctx context.Context, name string) (time.Time, error)
+	GetValidity(ctx context.Context, name string) (time.Time, time.Time, error)
+}
+
```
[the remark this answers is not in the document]
lindnerby: Sure this is reasonable to merge and centralize. Will check for a better location on the side though

cmd/composition/service/skrwebhook/skr_webhook_manager.go
```diff
@@ -147,16 +153,14 @@ func setupSKRCertService(kcpClient client.Client,
...
-	case gcertv1alpha1.SchemeGroupVersion.String():
-		renewalCertRepo := renewal.NewRepository(kcpClient, flagVar.IstioNamespace)
-		renewalService = gcm.NewService(renewalCertRepo)
-	default:
-		return nil, errCertificateManagementNotSupported
-	}
+	renewalService := renewal.NewService(certificateRepository,
+		secretRepository,
+		shared.GatewaySecretName,
+	)
```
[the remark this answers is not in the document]
lindnerby: Glad this is gone tbh

internal/service/watcher/certificate/certificate_service.go
```diff
@@ -157,23 +157,23 @@ func (c *Service) GetSkrCertificateSecretData(ctx context.Context,
...
 	}
 
-	return gatewayCertSecret, nil
+	return data.NewGatewaySecretData(gatewayCertSecret)
 }
 
-func (c *Service) GetGatewayCertificateSecretData(ctx context.Context) (*data.GatewaySecretData, error) {
-	gatewayCertSecret, err := c.GetGatewayCertificateSecret(ctx)
+// GetGatewayCertificateSecret returns the gateway certificate secret.
+func (c *Service) getGatewayCertificateSecret(ctx context.Context) (*apicorev1.Secret, error) {
```
lindnerby: Guess there is no need for this to be a private and seperate func. The few lines can be included in the single one consuming method.

internal/service/kyma/deletion/usecases/delete_watcher_certificate_setup.go
```diff
@@ -19,18 +18,18 @@ var (
 	errFailedToDeleteWatcherSkrSecret               = errors.New("failed to delete SKR secret")
 )
 
-type SecretRepository interface {
-	Exists(ctx context.Context, name string) (bool, error)
-	Delete(ctx context.Context, name string) error
-}
+type (
+	SecretRepository      = ExistsDeleteByNameRepo
+	CertificateRepository = ExistsDeleteByNameRepo
```
lindnerby: Does the linter complain here that the interfaces have the same signatures defined and one should be reused?
Imo it is totally fine to have two interface deps defined where only the name differs, because one might actually be extended/removed.
WDYT? Can we define SecretRepository and CertificateRepository here seperatly, not as type alias?

internal/service/watcher/certificate/certificate_service.go
c-pius: https://github.com/kyma-project/lifecycle-manager/pull/3019/commits/bbe6f67d11a88f63d6ee6c1aa582b4c9b3a9651e

internal/repository/watcher/certificate/gcm/certificate/certificate_repo.go
```diff
@@ -206,3 +217,27 @@ func parseValidity(input string) (time.Time, time.Time, error) {
...
+	return cert, nil
+}
+
+func (r *Repository) update(ctx context.Context, cert *gcertv1alpha1.Certificate) error {
+	if err := r.kcpClient.Update(ctx, cert); err != nil {
+		return fmt.Errorf("failed to update GCM Certificate %s-%s: %w", cert.Name, cert.Namespace, err)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
c-pius: https://github.com/kyma-project/lifecycle-manager/pull/3019/commits/d8011509981af00a30b77908a5c0b53814cc6c67

PR #3021, opened 2026-01-28
test: Add E2E test for module api upgrade under blocked deletion

### LINKED ISSUE #3005: ModuleCR version upgrade of module under deletion makes KLM ignore safe deletion
### Description

See:

- `/kyma/backlog/issues/8507`
- `/kyma/backlog/issues/8508`

Huskies released a new 1.56.1 version of their module. This new version contains a version update of the Telemetry CR from v1alpha1 to v1beta1. There were two instances where the telemetry module was under deletion waiting for the customer-managed resourced to be cleaned up, so Module CR was blocking the deletion as expected through a finalizer from telemetry manager.

However, as the new version of the module arrived, KLM processed with deleting the module resources (manager, roles, crds, ...) and ended up being blocked since Telemetry CRD still exsists and is blocked by an existing Telemetry CR.

### Potential Cause

KLM updated the `.spec.resource` in the related Manifest with the new version. The check for existing Module CRs is version specific:

https://github.com/kyma-project/lifecycle-manager/blob/a701253e856c4973eea0175cdc07f97a6fb7f4a9/internal/manifest/modulecr/client.go#L151

The API server probably returns something like this version does not exist. In the processing of a received error, we call a custom is not found function: https://github.com/kyma-project/lifecycle-manager/blob/a701253e856c4973eea0175cdc07f97a6fb7f4a9/internal/manifest/modulecr/client.go#L83

This is not found function includes a check for "IsNotRegisteredError": https://github.com/kyma-project/lifecycle-manager/blob/a701253e856c4973eea0175cdc07f97a6fb7f4a9/pkg/util/error.go#L36, Also it contains a check for "failed to get restmapping" https://github.com/kyma-project/lifecycle-manager/blob/a701253e856c4973eea0175cdc07f97a6fb7f4a9/pkg/util/error.go#L53-L61

This probably matched the error returned from the API server and therefore KLM assumed that we have a not found situation and continued deleting the module manager.

### Steps to reproduce

- run the steps from `Kyma Module with ModuleReleaseMeta Upgrade Under Deletion` but modify the new version introduced during deletion to contain a new version of the CR

### Environment Type

Managed

### Environment Info

Kubernetes Version: x.y.z
Docker Version: x.y.z
modulectl Version: x.y.z

### Attachments

_No response_

LeelaChacha:
**Description**

- Bugfix: for Unwanted Module Cleanup when it is stuck in deletion, and a version upgrade with API upgrade happens.
- Added E2E Test `Module API Upgrade Under Blocked Deletion`

**Related issue(s)**

Closes #3005

### DIFF
--- a/tests/e2e/Makefile
+++ b/tests/e2e/Makefile
@@ -71,7 +71,7 @@ e2e-coverage: ginkgo ## Generate the effective Acceptance Criteria for all the t
     done
 
 ##@ E2E Tests
-GO_TEST := GOFIPS140=$(FIPS140_MODULE_VERSION) go test -timeout 10m -ginkgo.v -ginkgo.focus
+GO_TEST := GOFIPS140=$(FIPS140_MODULE_VERSION) go test -timeout 20m -ginkgo.v -ginkgo.focus
 
 kyma-deprovision-with-foreground-propagation:
 	$(GO_TEST) "KCP Kyma CR Deprovision With Foreground Propagation After SKR Cluster Removal"


internal/manifest/modulecr/client.go
```diff
@@ -139,30 +141,55 @@ func (c *Client) SyncDefaultModuleCR(ctx context.Context, manifest *v1beta2.Mani
...
+	for _, mapping := range mappings {
+		list := &unstructured.UnstructuredList{}
+		list.SetGroupVersionKind(schema.GroupVersionKind{
+			Group:   mapping.GroupVersionKind.Group,
+			Version: mapping.GroupVersionKind.Version,
+			Kind:    mapping.GroupVersionKind.Kind,
+		})
+
+		if err := c.List(ctx, list, &client.ListOptions{
+			Namespace: resource.GetNamespace(),
```
c-pius: Just for future reference, I think this will need to change based on https://github.com/kyma-project/lifecycle-manager/issues/3006

internal/manifest/modulecr/client_test.go
```diff
@@ -218,3 +223,13 @@ func TestClient_GetAllModuleCRsExcludingDefaultCR_WithIgnorePolicy(t *testing.T)
 	assert.Equal(t, moduleCRs[1].GetName(), moduleCR2.GetName())
 	assert.Equal(t, moduleCRs[1].GetNamespace(), moduleCR2.GetNamespace())
 }
+
+func addSampleToScheme(sc *machineryruntime.Scheme) error {
```
c-pius: Just for my understanding... We now have to add Sample to the client scheme because the client explicitly requires the client scheme right. So this fixed existing tests.

internal/manifest/modulecr/client.go
```diff
@@ -59,22 +60,35 @@ func (c *Client) GetDefaultCR(ctx context.Context, manifest *v1beta2.Manifest) (
...
-	if err != nil {
-		if util.IsNotFound(err) {
-			return true, nil
-		}
-		return false, fmt.Errorf("%w: failed to fetch default resource CR", err)
+	defaultModuleCR := manifestCR.Spec.Resource
+	moduleCRGvk := defaultModuleCR.GroupVersionKind()
+	allModuleCRs, err := c.listResourcesByGroupKindInNamespace(ctx, moduleCRGvk, defaultModuleCR.GetNamespace())
+	if util.IsNotFound(err) {
+		return true, nil
```
c-pius: I am afraid this also won't do the trick long term. If it fixes the current problem, we can move forward with it. The thing is that we could query with a cached scheme that the SKR apiserver already unserved, then we would return true here even though there may still be an object. I assume this is unlikely and shouldn't block the fix for huskies, but for the target solution we need to consider this.

PR #3022, opened 2026-01-29
docs: ADR007 - Change step 3 from deleting the secret to using the Issuing condition

### LINKED ISSUE #2981: Investigate `Issuing certificate as Secret does not exist` error upon `istio-system/klm-watcher-serving` Cert rotation

c-pius:
**Description**

- Update the ADR docs from deleting the client certificate secret to using the Issuing condition

**Related issue(s)**

- closes #2981

### DIFF
--- a/docs/contributor/adr/007-pki-certs-and-rotation.md
+++ b/docs/contributor/adr/007-pki-certs-and-rotation.md
@@ -38,19 +38,19 @@ KLM watches the *klm-watcher* secret (2a). When it changes, KLM adds the new CA 
 
 In the [figure for step 2](../assets/adr-007/watcher-certificates-2.png), the new blue CA certificate is pushed into the CA bundle.
 
-### 3. Trigger Re-issuance of Client Certificates
+### 3. Trigger Reissuance of Client Certificates
 
-As part of the Kyma CR reconciliation, KLM gets the *klm-istio-gateway* to determine the last time the CA bundle was updated, indicated by the `lastModifiedAt` annotation (3a). If the `lastModifiedAt` of the CA bundle is more recent than the issued client certificate, KLM triggers the client certificate to be re-issued (3b).
+As part of the Kyma CR reconciliation, KLM gets the *klm-istio-gateway* to determine the last time the CA bundle was updated, indicated by the `lastModifiedAt` annotation (3a). If the `lastModifiedAt` of the CA bundle is more recent than the issued client certificate, KLM triggers reissuance of the client certificate. (3b).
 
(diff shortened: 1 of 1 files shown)


docs/contributor/adr/007-pki-certs-and-rotation.md
```diff
@@ -38,17 +38,17 @@ KLM watches the *klm-watcher* secret (2a). When it changes, KLM adds the new CA
 
 In the [figure for step 2](../assets/adr-007/watcher-certificates-2.png), the new blue CA certificate is pushed into the CA bundle.
 
-### 3. Delete Client Certificate Secrets
+### 3. Trigger Re-issuance of Client Certificates
```
mmitoraj:
```suggestion
### 3. Trigger Reissuance of Client Certificates
```

mmitoraj (approved):

Issue #2276, opened 2025-02-26
[E2E from Makefile] Port First E2E Test to the New Approach


LeelaChacha:
### Description

A new E2E workflow 

### Acceptance Criteria

- [ ] Decide on a test case to port
- [ ] Create a separate Makefile for new approach
  - [ ] Two targets (for setup and execution) for the test
  - [ ] Setup target should setup clusters, perform test-specific Kustomizations, deploy KLM, and create test-specific module metadata
    - (KLM deployment should used registry image for testing in the pipeline, otherwise build the image from sources)
- [ ] Create a separate workflow for GitHub Actions
  - (It should only setup tools and execute the Makefile target=
- [ ] Scripts are adjusted for both darwin and linux (see PoC)

### Attachments

PoC [PR #2274](https://github.com/kyma-project/lifecycle-manager/pull/2274)

PR #3030, opened 2026-02-03
test: Separate makefile for mandatory-module e2e test

Tomasz-Smelcerz-SAP:
**Description**
Change how `mandatory-module` e2e test is executed from logic embedded in Github actions to a makefile-based approach.

- introduced a new makefile. The name of the makefile corresponds to the filename of the related e2e golang test file.
- add scripts to install necessary tooling in the same way on local machine and in Github workflow
- switch e2e test execution to the new makefile-based mechanism

**Related issue(s)**
Fix: #2276

### DIFF
--- a/tests/e2e/mandatory_module_test.mk
+++ b/tests/e2e/mandatory_module_test.mk
@@ -155,9 +155,21 @@ deploy-klm: ## Deploy KLM in the KCP test cluster.
 	@echo "::endgroup::"
 
 
+clean-test-artifacts: ## Clean up test-specific resources.
+	@echo "::group::Cleaning up test-specific resources"
+	@export PATH=$(LOCALBIN):$$PATH # Add LOCALBIN to PATH
+	rm -f $(LIFECYCLE_MANAGER_DIR)/tests/e2e/mandatory_template_v2.yaml
+	@echo "::endgroup::"
+
+
 test-setup: ## Set up test-specific resources.
 	@echo "::group::Setting up test-specific resources"
 	@export PATH=$(LOCALBIN):$$PATH # Add LOCALBIN to PATH
+	# Exit early with an error if the file $(LIFECYCLE_MANAGER_DIR)/tests/e2e/mandatory_template_v2.yaml already exists
+	if [ -f $(LIFECYCLE_MANAGER_DIR)/tests/e2e/mandatory_template_v2.yaml ]; then
+		echo "File $(LIFECYCLE_MANAGER_DIR)/tests/e2e/mandatory_template_v2.yaml already exists. Please remove it before running the tests."
+		exit 1
+	fi
 	@pushd $(TEMPLATE_OPERATOR_DIR) > /dev/null
 	#
 	yq eval '.images[0].newTag = "$(OLDER_VERSION_FOR_MANDATORY_MODULE)"' -i config/manager/deployment/kustomization.yaml
@@ -203,11 +215,11 @@ test-run: ginkgo-install ## Run E2E test for mandatory module installation and d
 	exit $${status}
 
 
-test: create-clusters deploy-klm test-setup test-run ## Run all E2E tests for mandatory module installation and deletion.
+test: clean-test-artifacts teardown create-clusters deploy-klm test-setup test-run ## Run all E2E tests for mandatory module installation and deletion.
 
 
 teardown:
-	@echo "::group::Cleaning up..."
+	@echo "::group::Shutting down local clusters..."
 	@export PATH=$(LOCALBIN):$$PATH # Add LOCALBIN to PATH
 	@$(SCRIPTS_DIR)/clusters_cleanup.sh
 	@echo "::endgroup::"
--- a/tests/e2e/mandatory_modules_metrics_test.mk
+++ b/tests/e2e/mandatory_modules_metrics_test.mk
@@ -166,12 +166,6 @@ test-setup: ## Set up test-specific resources.
 	#
 	$(SCRIPTS_DIR)/deploy_mandatory_modulereleasemeta.sh "$(MODULE_NAME)" "$(OLDER_VERSION_FOR_MANDATORY_MODULE)"
 	#
-	#yq eval '.images[0].newTag = "$(NEWER_VERSION_FOR_MANDATORY_MODULE)"' -i config/manager/deployment/kustomization.yaml
-	#make build-manifests
-	#yq eval '(. | select(.kind == "Deployment") | .metadata.name) = "$(MODULE_DEPLOYMENT_NAME_IN_NEWER_VERSION)"' -i template-operator.yaml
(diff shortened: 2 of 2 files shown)


.github/actions/deploy-template-operator-with-modulereleasemeta/action.yml
```diff
@@ -174,17 +169,6 @@ runs:
 
         ./deploy_mandatory_modulereleasemeta.sh ${{ env.ModuleName }} ${{ env.OlderVersionForMandatoryModule }}
 
-    - name: Create ModuleTemplate in a new version for Mandatory module
```
Tomasz-Smelcerz-SAP: [Note for reviewer] This entire block has been migrated to the dedicated makefile.

tests/e2e/mandatory_module_test.mk
```diff
@@ -0,0 +1,210 @@
...
+
+# Ginkgo binary metadata.
+GINKGO_CMD ?= $(LOCALBIN)/ginkgo
+
+
+# Setting SHELL to bash allows bash commands to be executed by recipes.
+# This is a requirement for 'setup-envtest.sh' in the test target.
+# Options are set to exit when a recipe line exits non-zero or a piped command fails.
+.ONESHELL:
+SHELL := bash
```
medmes: `.ONESHELL Incompatibility` with macOS (GNU Make 3.81)
**Problem**:
I think it's on your knowledge that macOS ships with GNU Make 3.81 (from 2006) which does not support `.ONESHELL` (this directive was added in GNU Make 3.82, released in 2010). This is a hard blocker for local development on macOS without upgrades.

**Impact**:
developers. who use macOS will see silent failures or unpredictable behavior
pushd/popd sequences will execute in separate subshells, breaking directory context
Variables set in one command won't be available in the next (even with export).

**Solutions**
**Path 1 :  Remove .ONESHELL and use semicolons + backslashes**

```
deploy-klm: ## Deploy KLM in the KCP test cluster.
    @echo "::group::Deploying KLM"; \
    export PATH=$(LOCALBIN):$$PATH; \
    pushd $(LIFECYCLE_MANAGER_DIR)/config/watcher_local_test > /dev/null && \
    kustomize edit add patch --path patches/oci_registry_host.yaml --kind Deployment && \
    popd > /dev/null; \
    if [ -z $$GITHUB_ACTIONS ]; then \
        $(SCRIPTS_DIR)/deploy_klm_from_sources.sh; \
    else \
        $(SCRIPTS_DIR)/deploy_klm_from_registry.sh --image-registry $${KLM_IMAGE_REPO} --image-tag $${KLM_VERSION_TAG}; \
    fi; \
    echo "::endgroup::"; \
    echo "::group::Patching KCP metrics endpoint"; \
    $(SCRIPTS_DIR)/patch_kcp_metrics_endpoint.sh; \
    echo "::endgroup::"
```

**Or the opther path  is to document requirement for GNU Make ≥3.82**
```
## macOS Prerequisites
macOS ships with GNU Make 3.81. To run these tests locally, upgrade via Homebrew:
```bash
brew install make
# Add to ~/.zshrc or ~/.bashrc:
export PATH="/opt/homebrew/opt/make/libexec/gnubin:$PATH"
```

scripts/tests/setup_cluster_context.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
medmes: Missing Error Handling in Critical Scripts here and it apply to other scripts too:
-setup_cluster_context.sh
- patch_kcp_metrics_endpoint.sh
- clusters_cleanup.sh


Problem:
These scripts use #!/bin/bash without set -e, set -o pipefail, or set -o nounset. They can silently fail.

scripts/tests/setup_cluster_context.sh
```diff
@@ -0,0 +1,6 @@
...
+
+k3d kubeconfig merge -a -d
+export KCP_KUBECONFIG=$(k3d kubeconfig write skr)
```
medmes: There's a copy-paste bug here - KCP_KUBECONFIG is being assigned the output of k3d kubeconfig write skr instead of k3d kubeconfig write kcp.

**Solution** with the issues mentioned comment above:

```
#!/usr/bin/env bash
set -o nounset
set -o errexit
set -E
set -o pipefail

k3d kubeconfig merge -a -d
export KCP_KUBECONFIG=$(k3d kubeconfig write kcp)  
export SKR_KUBECONFIG=$(k3d kubeconfig write skr)
kubectl config use-context k3d-kcp
```

tests/e2e/mandatory_module_test.mk
```diff
@@ -0,0 +1,210 @@
...
+	make build-manifests
+	yq eval '(. | select(.kind == "Deployment") | .metadata.name) = "$(MODULE_DEPLOYMENT_NAME_IN_NEWER_VERSION)"' -i template-operator.yaml
+	$(SCRIPTS_DIR)/deploy_moduletemplate.sh "$(MODULE_NAME)" "$(NEWER_VERSION_FOR_MANDATORY_MODULE)" "$(DEPLOYABLE_MODULE_VERSION)" true true false false
+	cp template.yaml $(LIFECYCLE_MANAGER_DIR)/tests/e2e/mandatory_template_v2.yaml
+	#
+	@popd > /dev/null
+	@echo "::endgroup::"
+
+
+test-run: ginkgo-install ## Run E2E test for mandatory module installation and deletion.
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: The decision is to use `.ONESHELL`
If it can't be guaranteed, I'd say we should stop using make entirely. I will probably suggest that anyway in the near future. `make` is an old tool with terrible ergonomics. It's designed to for building C-language projects so it is compatible with how C compiler works. None of this is relevant for Go projects. IMO we'd have much better solution if we rewrite all of this `bash` + `make` combo in `Python` or `Lua`.
These days one doesn't really need any dedicated "build tool" at all, mentioned `Python` or `Lua` are powerful enough to handle any build-related task and they are much safer and more readable..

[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: The decision is to use `.ONESHELL`.
If it can't be used, I'd say we should drop make entirely. I will probably suggest that anyway in the near future.
`make` is an almost 50-years old tool with terrible ergonomics.
It was designed for building C-language projects so it is compatible with how C compiler used to work. None of this is relevant for Go projects.
IMO we'd have much better solution if we rewrite all of this `bash` + `make` combo in `Python` or `Lua`.
These days one doesn't really need any dedicated "build tool" at all, mentioned `Python` or `Lua` are powerful enough to handle any build-related task, in addition being safer (error handling) and more readable.

scripts/tests/setup_cluster_context.sh
Tomasz-Smelcerz-SAP: Done.

scripts/tests/patch_kcp_metrics_endpoint.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Done.

Issue #2983, opened 2026-01-19
Keep all unexpired CA certs in the CA bundle


c-pius:
### Description

The CA bundle in `istio-system/klm-istio-gateway` Secret serves as the bundle of CA certs both the server and the clients trust. Right now, we only keep the latest two which makes rotation sometimes a bit brittle if configured cert switches / rotation don't exactly match schedule (e.g., due to a manual rotation).

We should rather keep all unexpired CA certs in the bundle. Only once they have expired, we remove them from the bundle.

### Reasons

Improved robustness. Both server and clients continue to trust old CA certificates until their expiry. This eliminates the problem the server cert is CA-2, whereas the CA bundle used by server and clients only includes CA-0 and CA-1.

> If we have an emergency like a leak of the CA certificate, we can still force an update by renewing the CA certificate and then deleting the `klm-istio-gateway` secret. This will cause some downtime of the watch mechanism but the re-created CA bundle will only include the new CA certificate and will be propagated to both server and clients.

### Acceptance Criteria

- [x] CA bundle in Secret `istio-system/klm-istio-gateway#data."ca.crt"` contains all CA certs until their expiry
- [x] when the CA cert is renewed, the new CA cert is inserted into the bundle
  - [x] only when this happens, the `lastModifiedAt` annotation is updated to the current time
- [x] when a CA cert has expired, it is removed from the bundle
- [ ] secret type is updated to `kubernetes.io/tls` (currently Opaque)
  - (not entirely related, but also a small improvement if we are anyway touching the secret)
- [ ] [ADDED] see https://github.com/kyma-project/lifecycle-manager/pull/3033#discussion_r2773995628. Once refactored handle.go, also update the unit tests and use a mocked instance instead of the real bundler

### Dev Notes

- for parsing the certificate, we have existing code in runtime-watcher here that could serve as an example: https://github.com/kyma-project/runtime-watcher/blob/main/listener/pkg/v2/certificate/parse_certificate.go#L27-L65
  - https://github.com/kyma-project/lifecycle-manager/issues/3001 will likely also need to parse the certificate. Ideally implement it in KLM for easy re-use.
- it should be possible to regularly loop over all certificates in the bundle and check their Not After timestamp
  - since "regularly" means something like every 5 mins, it should not be problematic from compute perspective
  - we probably cannot loop from the back and stop at the first not expired ones since there could be changes to validity in the meantime
[...]

PR #3033, opened 2026-02-05
feat: Keep all unexpired certs in istio-gatway-secret ca.crt

c-pius:
**Description**

- introduces the `bundler.go` with the functionality to
  - add new certs to a bundle
  - drop expired certs from a bundler
- updated the `handler.go` to use bundler adding for managing the `istio-gateway-secret` CA bundle (`ca.crt`)
- this PR just introduces the plain changes needed to fulfill the story so we can included in this weeks release. There will be a follow-up PR with some refactorings in the handler.go to make things nicer.

**Related issue(s)**

- https://github.com/kyma-project/lifecycle-manager/issues/2983

### DIFF
--- a/internal/gatewaysecret/cabundle/handler_test.go
+++ b/internal/gatewaysecret/cabundle/handler_test.go
@@ -20,12 +20,16 @@ import (
 	"github.com/kyma-project/lifecycle-manager/tests/fixtures/certificates"
 )
 
-const gatewaySwitchCertBeforeExpirationTime = 1 * time.Hour
+const gatewayServerCertSwitchGracePeriod = 90 * time.Minute
 
 var (
 	notBefore = time.Now().Add(-1 * time.Hour)
 	notAfter  = time.Now().Add(2 * time.Hour)
 )
+
+// With the above notBefore and notAfter, the "new" CA cert is valid from 1 hour ago till the next two hours.
+// In the good path, we bundle the CA cert immediately, since the "new" CA cert is already valid since 1 hour.
+// And we will switch the TLS cert and key 30 minutes from now, since the grace period is 90 minutes from the notBefore.
 
 func TestManageGatewaySecret_WhenGetWatcherServingCertValidityReturnsError_ReturnsError(t *testing.T) {
 	// ARRANGE
@@ -35,7 +39,7 @@ func TestManageGatewaySecret_WhenGetWatcherServingCertValidityReturnsError_Retur
 
 	handler := cabundle.NewGatewaySecretHandler(mockClient,
 		nil,
-		gatewaySwitchCertBeforeExpirationTime,
+		gatewayServerCertSwitchGracePeriod,
 		certificate.NewBundler(),
 	)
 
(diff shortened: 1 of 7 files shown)


internal/service/watcher/certificate/bundler/bundler.go
```diff
@@ -0,0 +1,175 @@
...
+	ErrInvalidPEM         = errors.New("invalid PEM")
+	ErrFailedToParseX509  = errors.New("failed to parse x509 certificate")
+	ErrX509NotAfterIsZero = errors.New("x509 certificate NotAfter is zero")
+)
+
+type Bundler struct {
+	parseX509Func func([]byte) (*x509.Certificate, error)
+}
+
+func NewBundler(opts ...func(*Bundler) *Bundler) *Bundler {
```
lindnerby: Rename to `CertBundler`?
The composition for `bundler.NewBundler()` seems too non-descriptive. WDYT?

internal/service/watcher/certificate/bundler/bundler_test.go
```diff
@@ -0,0 +1,257 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/kyma-project/lifecycle-manager/internal/service/watcher/certificate/bundler"
+	"github.com/kyma-project/lifecycle-manager/tests/fixtures/certificates"
+
+	_ "embed"
```
lindnerby: Interesting, what does this do?
Also, `appendCerts()` is not defined in this file, related?

c-pius: Embed is a leftover, will remove. 

`appendCerts()` is defined in this file at the very bottom.

lindnerby: It seems blindness has kicked in already 🙈 
👍

internal/gatewaysecret/cabundle/handler_test.go
```diff
@@ -164,32 +192,32 @@ func TestManageGatewaySecret_WhenLegacySecret_BootstrapsLegacyGatewaySecret(t *t
 	// ASSERT
 	require.NoError(t, err)
 	mockClient.AssertNumberOfCalls(t, "UpdateGatewaySecret", 1)
-	mockClient.AssertCalled(t, "UpdateGatewaySecret", mock.Anything, mock.MatchedBy(
-		func(secret *apicorev1.Secret) bool {
-			return string(secret.Data["temp.ca.crt"]) == "value3"
-		}))
```
c-pius: Regarding this test I am not entirely sure anymore. My updates removed the need for asserting this temp.ca.crt. Chachas PR removed the need for checking `secret.Annotations[cabundle.CurrentCAExpirationAnnotation] != ""`. Maybe this test can be dropped entirely now as it is also related to the legacy gatway secret. But I would propose that we also do that in the follow-up to not block today's release

PR #3049, opened 2026-02-13
refactor: Move all Field Owners to `fieldowners` package

### LINKED ISSUE #2501: Align field-owners
### Description

Align the values for field-owners we use. Define them at a [consistent location](https://github.com/kyma-project/lifecycle-manager/blob/main/internal/common/fieldowners/const.go) in the codebase and re-use the values.

### Reasons

Consistency. We currently use different values, patterns (plain strings, FQDNs, ...) and re-define it multiple times in the codebase. In the future we need a consistent pattern for field-owners for eventual reversal of changes of fields, as well as audit logs that contains changes with references to field-owners.

### Acceptance Criteria

- [ ] all field owners follow a clear structure
- [ ] all field owner values defined at: https://github.com/kyma-project/lifecycle-manager/blob/main/internal/common/fieldowners/const.go

LeelaChacha:
**Description**

- Move all different field owners to `fieldowners` package:
  - `lifecycle-manager`
  - `declarative.kyma-project.io/applier`
  - `kyma-context-sync`
  - `catalog-sync`
  - `resource.kyma-project.io/finalizer` (yes, we were using the finalizer as a field owner when setting the finalizer)

**Related issue(s)**

Closes #2501


c-pius (commented):
The rollout of this change may indeed be problematic. 

https://kubernetes.io/docs/reference/using-api/server-side-apply/
> When trying to apply an object, fields that have a different value and are owned by another [manager](https://kubernetes.io/docs/reference/using-api/server-side-apply/#managers) will result in a [conflict](https://kubernetes.io/docs/reference/using-api/server-side-apply/#conflicts). This is done in order to signal that the operation might undo another collaborator's changes. Writes to objects with managed fields can be forced, in which case the value of any conflicted field will be overridden, and the ownership will be transferred.
> Whenever a field's value does change, ownership moves from its current manager to the manager making the change.
> A Server-Side Apply patch request requires the client to provide its identity as a [field manager](https://kubernetes.io/docs/reference/using-api/server-side-apply/#managers). When using Server-Side Apply, trying to change a field that is controlled by a different manager results in a rejected request unless the client forces an override. For details of overrides, see [Conflicts](https://kubernetes.io/docs/reference/using-api/server-side-apply/#conflicts).

There are some places where we do a SSA without forcing ownership. Probably things would work given that we don't change values, but if we do that in the future, we may see rejected requests. In the e2e tests, we currently don't observe that because we don't have historic objects. Would you mind testing this locally by spinning up a previous version of KLM with Kyma and Module synced. Then apply this version of KLM and see if we already observe errors?

Issue #3002, opened 2026-01-21
Rename `lastModifiedAt` annotation of `istio-system/klm-istio-gateway` to `caAddedToBundledAt`


c-pius:
### Description

Rename `lastModifiedAt` annotation of `istio-system/klm-istio-gateway` to a more expressive name.

### Reasons

`lastModifiedAt` was used for legacy reasons, but it is not expressing what it is for. We actually only write it once we add a new CA cert into the CA bundle.

### Acceptance Criteria

- [ ] annotation `caAddedToBundledAt` written to `istio-system/klm-istio-gateway`
- [ ] code migrated to to use the new `caAddedToBundledAt` annotation instead of `lastModifiedAt`
- [ ] `lastModifiedAt` annotation dropped by code (to roll it out automatically to all landscapes)

### Dev Notes

- we may need to do a three step approach where we
  - first add a new annotation to the secret
  - then update the other parts in the code to use this new annotation
  - then drop the old one
  - drop the code dropping the old annotation

Issue #3003, opened 2026-01-21
Drop legacy gateway secret support


c-pius:
### Description

We still have some old code from where legacy gateway secret with downtime rotation, e.g.: https://github.com/kyma-project/lifecycle-manager/blob/a701253e856c4973eea0175cdc07f97a6fb7f4a9/internal/gatewaysecret/cabundle/handler.go#L56-L57

We need to clean that up.

### Reasons

Clean codebase.

### Acceptance Criteria

- [ ] Acceptance criterion 1
- [ ] Acceptance criterion 2

PR #3104, opened 2026-03-06
fix: Ensure proper annotation migration

Tomasz-Smelcerz-SAP:
**Description**

A bug has been revealed in GW Secret annotation handling: We blindly changed the annotation name from `lastModifiedAt` to `caAddedToBundleAt`. New code wasn't looking at all at  `lastModifiedAt` annotation anymore.

The problem:
If we deploy the new version to a system where existing GW Secret only has the "old"  `lastModifiedAt` annotation (as is the case), the new code cannot find the value it requires to properly execute the cert-related logic.
The  `lastModifiedAt` will be eventually replaced by `caAddedToBundleAt` annotation - this logic exists in code - but this only happens at "bundling time", i.e: after the root certificate is rotated, which happens once every week or once every month.

The fix: Introduce fallback to "old"  `lastModifiedAt` annotation value if the `caAddedToBundleAt` annotation does not exist.

- ensure we read the value of `lastModifiedAt` annotation if `caAddedToBundleAt` does not exist - as a fallback

**Related issue(s)**
#3002

### DIFF
--- a/unit-test-coverage-lifecycle-manager.yaml
+++ b/unit-test-coverage-lifecycle-manager.yaml
@@ -1,6 +1,6 @@
 packages:
   internal/controller: 100
-  internal/controller/istiogatewaysecret: 28
+  internal/controller/istiogatewaysecret: 26
   internal/crd: 92
   internal/descriptor/cache: 90
   internal/descriptor/provider: 100
--- a/internal/controller/istiogatewaysecret/setup.go
+++ b/internal/controller/istiogatewaysecret/setup.go
@@ -42,17 +42,7 @@ func SetupReconciler(mgr ctrl.Manager,
 
 	var handler gatewaysecret.Handler
 	if flagVar.UseLegacyStrategyForIstioGatewaySecret {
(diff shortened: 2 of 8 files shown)


unit-test-coverage-lifecycle-manager.yaml
```diff
@@ -1,6 +1,6 @@
 packages:
   internal/controller: 100
-  internal/controller/istiogatewaysecret: 28
+  internal/controller/istiogatewaysecret: 26
```
Tomasz-Smelcerz-SAP: This is because we are not testing here the legacy handler at all, and few more lines was added to the legacy handler. It is going to be removed soon anyway along with #3003, and then we'll observe coverage increase here, for the same reason: there will be less LoC in this package.

medmes (approved):
The fix is correct, minimal, and appropriately scoped.

medmes (approved):

c-pius:
Moving this to accept. @Tomasz-Smelcerz-SAP verified that the annotation was migrated properly, so no need to verify the fallback anymore. See message from @Tomasz-Smelcerz-SAP in `#kyma-sre Mar 13th at 09:47`

Issue #3113, opened 2026-03-09
Support configuration of explicit OCI registry sub-path


c-pius:
### Description

KLM needs to be configurable with a OCI repository sub-path. If the configured `oci-registry-cred-secret` is only containing the `<registry-host>`, KLM needs to be configurable with a `<repository-sub-path>` for the OCM lookup of modules.

### Reasons

There were problems with KLM on a US pre-prod landscape. The operators transferred the kyma product into a registry with a repository sub-path, e.g. `<registry-host>/<repository-sub-path>`. KLM was configured with a `oci-registry-cred-secret` where only `<registry-host>` was given. This made KLM fail the OCM descriptor lookup because it was searching at `<registry-host>/component-descriptors`, not at `<registry-host>/<repository-sub-path>/component-descriptors`. According to SRE, the secret passed to `oci-registry-cred-secret` may be a more general secret to an OCI registry. It is a valid use case that this secret points to the `<registry-host>` only where it is necessary that KLM is configured additionally with the sub-path in this registry where the module OCMs are stored.

### Acceptance Criteria

- [x] OCI registry host and sub path configurable independent of each other
- [x] No flag name change

c-pius:
Verified on STAGE with `kyma/management-plane-config/pull/7558`

PR #3141, opened 2026-03-19
chore: Refactor resolving of OCI registry from flags

c-pius:
**Description**

- refactors the OCIRegistry resolving with the aim for easier usage and to get more clarity on the naming
  - see also https://learn.microsoft.com/en-us/azure/container-registry/container-registry-concepts
- refactors main extracting composition functionality into `cmd/composition` package
- another candidate to consider to include in the OCIRegistry is the related key chain lookup: https://github.com/kyma-project/lifecycle-manager/blob/dc6c132126447868822ba4225d34c7518d926b9c/cmd/main.go#L262
  - but this would require more investigation and for the sake of time, I would opt to this at a later point in time

**Related issue(s)**

- resolves https://github.com/kyma-project/lifecycle-manager/issues/3113


unit-test-coverage-lifecycle-manager.yaml
```diff
@@ -48,6 +48,7 @@ packages:
   internal/service/mandatorymodule/deletion/usecases: 100
   internal/service/skrsync: 100
 
+  internal/setup: 71
```
c-pius: There are other structs in this package. OCIRegistry has 100% coverage.

internal/setup/ociregistry.go
c-pius: This used to be `internal/setup/ociregistryhostprovider.go`

PR #3161, opened 2026-03-31
fix: Skip mandatory module reconciliation if Kyma is deleting

lindnerby:
**Description**

- Skip mandatory module installation when Kyma is deleting

**Related issue(s)**


c-pius (approved):

c-pius (approved):

c-pius:
This is hard to verify. I have tried to block Kyma deletion by finalizers on KCP Manifests and SKR Kyma / Module CR. But KEB is almost instantly removing the SKR secret and therefore the entire KCP Kyma is deleted and finalizers fremoved from KCP Manifests. There is no time to really observe if mandatory manifests are re-created.

c-pius:
Verified on DEV instead of STAGE. Used this PR to increase requeue intervals for Kyma and set my test cluster to deleting. `kyma/management-plane-charts/pull/5018`

This lead to all Manifests being deleted within a couple of seconds. Contrary to before, we are not seeing Manifests for mandatory modules re-appearing while the Kyma is still in Deletion.

PR #3185, opened 2026-04-14
feat: Spread Kyma requeues on New Module Rollout

### LINKED ISSUE #2767: Fine tune KLM to reduce the speed of rolling out module updates
**2-day technical spike**

### Description

Fine tune KLM to reduce the speed of rolling out module updates. Two points for tuning may be

- reducing the number of concurrent workers
  - https://github.com/kyma-project/lifecycle-manager/blob/main/internal/pkg/flags/flags.go#L112-L115
  - see also `/kyma/management-plane-charts/pull/1749`
- adjusting QPS and burst
  - https://github.com/kyma-project/lifecycle-manager/blob/main/internal/pkg/flags/flags.go#L184-L189

### Reasons

SRE were observing rate limiting issues while rolling out istio 1.21.0 in regular channel on 2025-09-22 10AM:

```
Failed to pull image "europe-docker.pkg.dev/kyma-project/prod/external/istio/install-cni:1.27.1-distroless": failed to pull and unpack image "europe-docker.pkg.dev/kyma-project/prod/external/istio/install-cni:1.27.1-distroless": failed to copy: httpReadSeeker: failed open: unexpected status code https://europe-docker.pkg.dev/v2/kyma-project/prod/external/istio/install-cni/blobs/sha256:6f6914a0b8c5d36408b7155aa9bd41a6d6a12fa974314e11a802161c2e5b28bd: 429 Too Many Requests - Server message: toomanyrequests: Quota exceeded for quota metric 'Requests per project in the Europe multi-region' and limit 'Requests per project in the Europe multi-region per minute' of service 'artifactregistry.googleapis.com' for consumer 'project_number:905162714753'.
```

@kyma-project/neighbors already increased quota (see `kyma/test-infra/issues/920`), but we should also try to mitigate the risk of this happening again from KLM perspective.

This is also related to `/kyma/backlog/issues/7507`. Here we are looking for a short term solution just increasing the rollout time. In 7507, it looks for a long term solution also considering requirements like rolling out region by region for example.

### Acceptance Criteria

- [x] Module updates rolled out less aggressively
- [ ] dependencies like API server, artifact registry, signify service ... are not overwhelmed with requests
- [x] rollout is still executed siwftly
- [ ] found kind of a balance between KLM working sufficiently fast but also not overwhelming downstream dependencies

### Dev Note

Get in contact with otters and neighbors to see if we can observe that the rollout is indeed slowed down. E.g, when otters run a new version of Warden, the notary service should not be overwhelmed (`/kyma/warden/issues/36`). With neighbors, check if the GCP chart of the image registry shows lowered (but longer lasting) spikes (`/kyma/test-infra/issues/920#issuecomment-15713122`)

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

Workers have been maxed out when doing the istio update:

Image

Processing latency shortly spiked:

Image

LeelaChacha:
When a new module version is rolled out the MRM update event previously enqueued all affected Kymas at once, causing a reconciliation burst that could exhaust API-server rate limits.

The ModuleReleaseMetaEventHandler now accepts an `updateRequeueMaxDelay` and, for Update events only, picks a uniform random delay in [0, updateRequeueMaxDelay) per Kyma via `rli.AddAfter`.  Create / Delete / Generic events are unchanged and still enqueue immediately.

The delay window is wired up to `r.Success` (the normal success-requeue interval) so the spread is proportional to the configured reconcile cadence.

Closes #2767


internal/controller/kyma/setup.go
```diff
@@ -42,7 +42,7 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, opts ctrlruntime.Options
 		WithEventFilter(predicate.Or(predicate.GenerationChangedPredicate{}, predicate.LabelChangedPredicate{})).
 		Watches(&v1beta2.ModuleTemplate{},
 			handler.EnqueueRequestsFromMapFunc(watch.NewTemplateChangeHandler(r).Watch())).
-		Watches(&v1beta2.ModuleReleaseMeta{}, watch.NewModuleReleaseMetaEventHandler(r)).
+		Watches(&v1beta2.ModuleReleaseMeta{}, watch.NewModuleReleaseMetaEventHandler(r, r.Success)).
```
c-pius: Even with the behavior that a enqueue n overwrites the enqueue n-1, I think it makes sense to re-use the success interval here. But would you mind adding a comment just to indicate that this is done by intent. Maybe something like `// re-using the success interval as the upper limit for jittered requeues`

internal/watch/modulereleasemeta_change.go
c-pius, line 1: Would you mind refactoring this and the related mandatory module handlers a bit? The functionality of `getKymaList()` is already avaialble from the repo: https://github.com/kyma-project/lifecycle-manager/blob/171ecba82f3dbbb093f45ce8f94e98c97fbac1cd/internal/repository/kyma/kyma_repo.go#L37-L46. Can we rather inject and re-use the repo instead of the entire kcp client as Reader?

I am also not entirely sure about the event types other than Update. The Create event should be a no-op. MRM creation equals the introduction of a new module. Since it is new, no Kyma is using this module yet and "GetAffectedKymas" should be an empty list. I think we could also just make this explicit bay making `Create` a no-op function with related comment. The Delete event, I think we haven't defined yet what the behavior actually is. We don't even have a finalizer on MRM that prevents its deletion until no Kyma is using that module anymore. Here, I would suggest that we create a ticket to clarify this story and here also make the Delete a no-op with reference to this ticket that this still needs to be clarified. And since we don't expect Generic events, I would also make this one a no-op. With this, we can even drop the `handleEvent` and `requeueKymas` funcs, maybe even more.

The mandatory module one could also use the Kyma Repo and MRM Repo:
https://github.com/kyma-project/lifecycle-manager/blob/171ecba82f3dbbb093f45ce8f94e98c97fbac1cd/internal/repository/modulereleasemeta/mrm_repo.go#L52-L59

internal/watch/modulereleasemeta_change.go
```diff
@@ -27,8 +34,13 @@ type TypedModuleReleaseMetaEventHandler[object any, request comparable] struct {
...
 }
 
-func NewModuleReleaseMetaEventHandler(handlerClient ChangeHandlerClient) *ModuleReleaseMetaEventHandler {
-	return &ModuleReleaseMetaEventHandler{Reader: handlerClient}
+// NewModuleReleaseMetaEventHandler creates a handler that, on update events, spreads Kyma reconciliations
+// over a random delay of up to updateRequeueMaxDelay to avoid rate-limiting bursts.
+func NewModuleReleaseMetaEventHandler(handlerClient ChangeHandlerClient, updateRequeueMaxDelay time.Duration) *ModuleReleaseMetaEventHandler {
+	return &ModuleReleaseMetaEventHandler{
+		Reader:                handlerClient,
+		updateRequeueMaxDelay: updateRequeueMaxDelay,
```
c-pius: I actually wouldn't mind if we re-use the Jitter from `pkg/queue`, just with different parameters. We could inject a jitter with 100% probability and 100% percentage. This would mean that it always jitters. If we supply it with `r.Success / 2`, we would get the jittering across the entire interval.

May sound a bit complicated but if we don't use the Jitter, I am afraid we will ask ourselves in the future why we use it at one place but not the other.

LeelaChacha: Would argue that the keyword jitter should not be used in this context and we don't reuse the jitter package as it would hurt the readability considerably

internal/controller/kyma/setup.go
```diff
@@ -36,13 +36,16 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, opts ctrlruntime.Options
 		return fmt.Errorf("KymaReconciler %w", err)
 	}
 
+	mtEventHandlerMapFunc := watch.ComposeTemplateChangeHandlerMapFunc(r.Client)
+	mrmEventHandler := watch.ComposeMrmEventHandler(r.Client,
+		r.Success) // re-using the success interval as the max delay for jittered requeues
+
```
c-pius: Please construct these in the `main.go` and hand them to the reconciler from there.See also ADR 002:

> Composition functions are called from the `main` function, and the dependency tree is resolved

In the main function, you should also find an existing instance of the kyma repo that you can re-use instead of initializing a new one in the composition function. You can also find that MRM and MT Repos are initialized within multiple composition functions already. Please try to extract them from there and initialize a single repo in main and pass it to the composition functions.

internal/repository/kyma/kyma_repo_get_all_test.go
```diff
@@ -0,0 +1,59 @@
...
+	foundKymas, err := repo.GetAll(t.Context())
+
+	// then
+	require.NoError(t, err)
+	require.Len(t, foundKymas.Items, 2)
+	require.Equal(t, kymaName+"-1", foundKymas.Items[0].GetName())
+	require.Equal(t, kymaNamespace, foundKymas.Items[0].GetNamespace())
+	require.Equal(t, kymaName+"-2", foundKymas.Items[1].GetName())
+	require.Equal(t, kymaNamespace, foundKymas.Items[1].GetNamespace())
+}
```
c-pius, line 56: Other funcs don't contain given, when, then comments. I also don't see the need here, but also okay to keep it.

internal/watch/modulereleasemeta/mandatory_mrm_handler.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+)
+
+type mrmRepository interface {
+	Get(ctx context.Context, name string) (*v1beta2.ModuleReleaseMeta, error)
+}
+
+type mandatoryKymaRepository interface {
+	GetAll(ctx context.Context) (*v1beta2.KymaList, error)
+}
```
c-pius: Please just call it `kymaRepository` instead.

internal/watch/modulereleasemeta/mrm_handler.go
```diff
@@ -0,0 +1,85 @@
...
+	// bursts when a new module version is rolled out to many clusters simultaneously.
+	updateRequeueMaxDelay time.Duration
+
+	CreateFunc func(context.Context, event.TypedCreateEvent[object], workqueue.TypedRateLimitingInterface[request])
+
+	UpdateFunc func(context.Context, event.TypedUpdateEvent[object], workqueue.TypedRateLimitingInterface[request])
+
+	DeleteFunc func(context.Context, event.TypedDeleteEvent[object], workqueue.TypedRateLimitingInterface[request])
+
+	GenericFunc func(context.Context, event.TypedGenericEvent[object], workqueue.TypedRateLimitingInterface[request])
```
c-pius: What are those needed for? Can they be removed?

internal/watch/modulereleasemeta/mrm_handler.go
```diff
@@ -0,0 +1,85 @@
...
+// to spread reconciliations and avoid rate-limiting bursts when a new module version is released.
+func (m TypedEventHandler[object, request]) Update(ctx context.Context, evt event.UpdateEvent,
+	rli workqueue.TypedRateLimitingInterface[reconcile.Request],
+) {
+	kymaList, err := m.kymaRepository.LookupByLabel(ctx, shared.ManagedBy, shared.OperatorName)
+	if err != nil {
+		return
+	}
+	events.HandleUpdate(evt, rli, kymaList, m.updateRequeueMaxDelay)
+}
```
c-pius, line 115: I think I would re-shape the boundaries here a bit. As of now, both the functions here comply with the interface required by controller-runtime, and we push those types also to the `Handle*` funcs (`evt event.UpdateEvent, rli workqueue.TypedRateLimitingInterface[reconcile.Request]`).

I would rather see that the funcs here take care of complying with the controller-runtime interface, and the `Handle` funcs only take care of determining what kymas to enqueue and return a kyma list. So here, it would look something like:

```Go
func (m TypedEventHandler[object, request]) Delete(ctx context.Context, evt event.DeleteEvent,
	rli workqueue.TypedRateLimitingInterface[reconcile.Request],
) {
    moduleReleaseMeta, ok := evt.Object.(*v1beta2.ModuleReleaseMeta)
	if !ok {
		return
	}

	kymaList, err := m.kymaRepository.LookupByLabel(ctx, shared.ManagedBy, shared.OperatorName)
	if err != nil {
		return
	}

    // events.HandleDElete would need to be renamed
    kymasToEnqeue := events.HandleDelete(moduleReleaseMeta, kymaList)

    // these function would also move into the handler here, or you would could also separate them out. What you prefer.
	requeueKymas(kymasToEnqueue)
}
```

internal/watch/template_change.go
c-pius, line 1: Wondering if we should rather drop this entirely. MTs should be considered immutable. Why would we react on change events? WDYT?

c-pius (commented):
Please also update the coverage file and fix the linting.

internal/watch/modulereleasemeta/mandatory_mrm_handler.go
LeelaChacha: There is one in the same package already. The one used by `mrm_handler.go`. Lemme see how to refactor the best way

internal/watch/modulereleasemeta/mrm_handler.go
LeelaChacha: Done. Moved the controller-runtime interface concerns (event extraction, enqueuing) into the handler, and the `Handle*` functions in `events/` are now pure logic: they take the already-extracted MRM object(s) and a `KymaList`, and return `[]*types.NamespacedName`. No `event.*` or `workqueue.*` types leak into `events/` anymore.

Renamed them to `AffectedKymasOnDelete` / `AffectedKymasOnUpdate` to reflect what they actually return. `requeueKymas` and `requeueKymasWithRandomDelay` moved up into `mrm_handler.go` alongside the rest of the enqueue logic.

Also caught and fixed a pre-existing swapped-argument bug in `findDeletedChannels` while I was in there — all tests pass now.

internal/watch/template_change.go
LeelaChacha: I would reason that we may want to leave it like that. We do react on label changes atm. Perhaps we want to reconcile all Kymas using a particular MT. Or if add a field to existing MTs, 🤷‍♂️ 
I don't see much utility but also no harm in keeping it there

internal/watch/modulereleasemeta/mandatory_mrm_handler.go
LeelaChacha: Okay. I have added the GetAll method to the local kymaRepository interface in `mrm_handler.go`. This is not used in the typed mrm event handler but used in `mandatory_mrm_handler.go` in the same package. So it should be fine

internal/watch/modulereleasemeta/events/delete.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+// AffectedKymasOnDelete returns the Kymas that should be requeued when an MRM is deleted.
+// All channels are affected because the entire MRM is going away.
+func AffectedKymasOnDelete(mrm *v1beta2.ModuleReleaseMeta, kymaList *v1beta2.KymaList) []*types.NamespacedName {
+	return GetAffectedKymas(kymaList, mrm.Spec.ModuleName, mrm.GetAllChannels())
+}
+
+// GetAffectedKymas determines which Kymas are affected by the update. It returns a list of Kymas that have modules
+// assigned to the updated channels.
```
c-pius:
```suggestion
// GetAffectedKymas determines which Kymas are affected by the update. It returns a list of Kymas that have the affected
// modules installed from the affected channels.
```

internal/watch/modulereleasemeta/events/delete.go
```diff
@@ -0,0 +1,46 @@
...
+			moduleChannel := getModuleChannel(module.Channel, kyma.Spec.Channel)
+			if slices.Contains(affectedChannels, moduleChannel) {
+				affectedKymas = append(affectedKymas,
+					&types.NamespacedName{Name: kyma.GetName(), Namespace: kyma.GetNamespace()})
+				break
+			}
+		}
+	}
+	return affectedKymas
+}
```
c-pius: Consider to move this one to a `events/common.go` file this is used by Delete and Update. We use that approach at a few other places already. But also okay to keep it here.

I would also prefer to have this func private, and testing it via the actually outside used `AffectedKymasOnDelete` and `AffectedKymasOnUpdate` funcs instead.

internal/watch/modulereleasemeta/events/update.go
```diff
@@ -0,0 +1,55 @@
...
+func AffectedKymasOnUpdate(oldMRM, newMRM *v1beta2.ModuleReleaseMeta,
+	kymaList *v1beta2.KymaList,
+) []*types.NamespacedName {
+	affectedChannels := DiffModuleReleaseMetaChannels(oldMRM, newMRM)
+	return GetAffectedKymas(kymaList, newMRM.Spec.ModuleName, affectedChannels)
+}
+
+// DiffModuleReleaseMetaChannels returns the names of channels whose version changed or that were added/removed
+// between the old and new ModuleReleaseMeta.
+func DiffModuleReleaseMetaChannels(
```
c-pius: Would also prefer this one to be private func and tested via `AffectedKymasOnUpdate`.

internal/watch/modulereleasemeta/mandatory_mrm_handler.go
c-pius, line 1: Nice!

c-pius: Should also be rather easy to unit test now.

internal/watch/modulereleasemeta/mrm_handler.go
```diff
@@ -0,0 +1,119 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/kyma-project/lifecycle-manager/api/shared"
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+	"github.com/kyma-project/lifecycle-manager/internal/watch/modulereleasemeta/events"
+)
+
+type kymaRepository interface {
+	LookupByLabel(ctx context.Context, labelKey, labelValue string) (*v1beta2.KymaList, error)
+}
```
c-pius: May also be moved to a `common.go` file.

internal/watch/modulereleasemeta/mrm_handler.go
c-pius, line 1: Also nice, and should be easily unit testable.

internal/watch/template_change.go
```diff
@@ -3,61 +3,75 @@ package watch
...
 }
 
 func (h *TemplateChangeHandler) Watch() handler.MapFunc {
 	return func(ctx context.Context, o client.Object) []reconcile.Request {
-		emptyRequest := make([]reconcile.Request, 0)
-		template := &v1beta2.ModuleTemplate{}
-
-		if err := h.Get(ctx, client.ObjectKeyFromObject(o), template); err != nil {
-			return emptyRequest
+		template, err := h.templateRepository.Get(ctx, o.GetName())
```
c-pius: Just wondering... Is `o` the entire MT object already, so we actually don't need the Get from the repo?

c-pius (commented):
Looks good, just some minor comments left, thanks for the effort! Consider them best effort. If you like to do them and you can do them until 15:30 please go ahead. Otherwise, let me know once the linting stuff is fixed and I will approve.

c-pius (approved):

c-pius (approved):

PR #3194, opened 2026-04-17
fix: Update RBAC for new eventing API

Tomasz-Smelcerz-SAP:
**Description**

Add RBAC for new eventing API that we "enabled" when bumping `controller-runtime` version in #3099 
Note: I haven't removed access to the "old" (core) events API, as it's hard to tell if it's really never used by the controller-runtime internally. In addition I don't think this is a problem.

- Add missing RBAC for `events.k8s.io`.

### DIFF
--- a/config/rbac/leader_election_role.yaml
+++ b/config/rbac/leader_election_role.yaml
@@ -35,3 +35,10 @@ rules:
     verbs:
       - create
       - patch
+  - apiGroups:
+      - "events.k8s.io"
+    resources:
+      - events
+    verbs:
+      - create
+      - patch
--- a/config/rbac/manager_role.yaml
+++ b/config/rbac/manager_role.yaml
@@ -24,6 +24,16 @@ rules:
       - patch
       - watch
   - apiGroups:
+      - "events.k8s.io"
+    resources:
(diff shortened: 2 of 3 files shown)


c-pius:
> Note: I haven't removed access to the "old" (core) events API, as it's hard to tell if it's really never used by the controller-runtime internally. In addition I don't think this is a problem.

We can check as follow-up. Just one question, are the rules really needed for both the manager and leader election role?

Tomasz-Smelcerz-SAP:
> > Note: I haven't removed access to the "old" (core) events API, as it's hard to tell if it's really never used by the controller-runtime internally. In addition I don't think this is a problem.
> 
> We can check as follow-up. Just one question, are the rules really needed for both the manager and leader election role?

I think so. I mean there's a reason why we have both of these. I am pretty sure I've seen one event emitted by leader election (leader election "result") and then there are "normal" events emitted by "lifecycle-manager" - notifying about reconciliation progress or errors. So IMO yes, we need both. And I think once `controller-runtime` is updated, both of these require permission to use the new API `events.k8s.io`.

c-pius (approved):

PR #3225, opened 2026-04-28
feat: Add RestrictedModuleMatch helper func

### LINKED ISSUE #3219: Implement `RestrictedModuleMatch` helper func
## Description

Implement a shared function that extracts the `kymaSelector` from a ModuleReleaseMeta and checks if it matches a Kyma. If `mrm.Spec.KymaSelector` is nil, the selector matches nothing by default `labels.Nothing()`, if `mrm.Spec.KymaSelector` is empty (`{}` in the MRM yaml), the selector matches nothing by default.

```Go
func RestrictedModuleMatch(mrm *v1beta2.ModuleReleaseMeta, kyma *v1beta2.Kyma) (bool, error) {
	// ...
}
```

## Reasons

There are at least three places in code where it needs to be checked whether a Kyma matches the selector from a `MRM.spec.kymaSelector`. Introduce this centrally for re-use.

## AC

- [x] `RestrictedModuleMatch` parses the selector from the MRM
  - [x] if parsing fails, false and an error is returned
  - [x] if selector is nil, false is returned
  - [x] if selector is empty (`{}` in the MRM yaml), false is returned
- [x] `RestrictedModuleMatch` matches the parsed selector against the `kyma.ObjectMeta.Labels`
  - [x] if match, returns true
  - [x] if no match, returns false
- [x] Unit test coverage 100%
  - [x] includes one test case for `matchLabels`
  - [x] includes one test case for `matchExpressions`
  - [x] includes ones test case for `matchLabels` && `matchExpressions`

## Dev Notes

`mrm.Spec.KymaSelector == nil` should already create a nothing selector, but still cover this explicitly. `mrm.Spec.KymaSelector` is empty (`{}` in the MRM yaml) should create a always selector, but we want to be rather defensive and therefore rather also match nothing in that case.

Tomasz-Smelcerz-SAP:
**Description**

Fixes #3219 

- Add RestrictedModuleMatch helper func
- Add unit tests

**Related issue(s)**
Fix: #3219

### DIFF
--- a/internal/service/restrictedmodule/match.go
+++ b/internal/service/restrictedmodule/match.go
@@ -1,0 +1,37 @@
+package restrictedmodule
+
+import (
+	"errors"
+	"fmt"
+
+	apimetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8slabels "k8s.io/apimachinery/pkg/labels"
+
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+)
+
+var ErrSelectorParse = errors.New("failed to parse kymaSelector on ModuleReleaseMeta")
+
+// RestrictedModuleMatch extracts the kymaSelector from a ModuleReleaseMeta
+// and checks if it matches labels on a specified Kyma instance.
+// A nil or empty selector returns false.
+func RestrictedModuleMatch(mrm *v1beta2.ModuleReleaseMeta, kyma *v1beta2.Kyma) (bool, error) {
+	if mrm == nil || kyma == nil {
+		return false, nil
+	}
+
+	kymaSelector := mrm.Spec.KymaSelector
+	if kymaSelector == nil {
+		return false, nil
+	}
+	if len(kymaSelector.MatchLabels) == 0 && len(kymaSelector.MatchExpressions) == 0 {
+		return false, nil
+	}
+
+	selector, err := apimetav1.LabelSelectorAsSelector(kymaSelector)
+	if err != nil {
+		return false, fmt.Errorf("%w: %w", ErrSelectorParse, err)
+	}
+
+	return selector.Matches(k8slabels.Set(kyma.Labels)), nil
+}
--- a/internal/service/restrictedmodule/match_test.go
+++ b/internal/service/restrictedmodule/match_test.go
@@ -1,0 +1,235 @@
+package restrictedmodule_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	apimetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+	"github.com/kyma-project/lifecycle-manager/internal/service/restrictedmodule"
+)
+
+func TestRestrictedModuleMatch_NilModuleReleaseMeta(t *testing.T) {
+	kyma := &v1beta2.Kyma{}
+
+	matched, err := restrictedmodule.RestrictedModuleMatch(nil, kyma)
+
+	require.NoError(t, err)
+	assert.False(t, matched)
+}
+
+func TestRestrictedModuleMatch_NilKyma(t *testing.T) {
+	mrm := &v1beta2.ModuleReleaseMeta{Spec: v1beta2.ModuleReleaseMetaSpec{}}
+
+	matched, err := restrictedmodule.RestrictedModuleMatch(mrm, nil)
+
+	require.NoError(t, err)
+	assert.False(t, matched)
+}
+
+func TestRestrictedModuleMatch_NilSelector(t *testing.T) {
+	mrm := &v1beta2.ModuleReleaseMeta{Spec: v1beta2.ModuleReleaseMetaSpec{KymaSelector: nil}}
+	kyma := &v1beta2.Kyma{}
+
+	matched, err := restrictedmodule.RestrictedModuleMatch(mrm, kyma)
+
+	require.NoError(t, err)
+	assert.False(t, matched)
+}
+
+func TestRestrictedModuleMatch_EmptySelector(t *testing.T) {
+	mrm := &v1beta2.ModuleReleaseMeta{
+		Spec: v1beta2.ModuleReleaseMetaSpec{KymaSelector: &apimetav1.LabelSelector{}},
+	}
+	kyma := &v1beta2.Kyma{}
+
+	matched, err := restrictedmodule.RestrictedModuleMatch(mrm, kyma)
+
+	require.NoError(t, err)
+	assert.False(t, matched)
+}
+
+func TestRestrictedModuleMatch_NilLabelsOnKyma(t *testing.T) {
+	mrm := &v1beta2.ModuleReleaseMeta{
+		Spec: v1beta2.ModuleReleaseMetaSpec{
+			KymaSelector: &apimetav1.LabelSelector{
+				MatchLabels: map[string]string{"env": "production"},
+			},
+		},
+	}
+	kyma := &v1beta2.Kyma{}
+	kyma.SetLabels(nil)
+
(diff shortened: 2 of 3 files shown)


internal/service/restrictedModule/restrictedModuleMatch.go
```diff
@@ -0,0 +1,33 @@
+package restrictedModule
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Done.

internal/service/restrictedModule/restrictedModuleMatch_test.go
```diff
@@ -0,0 +1,198 @@
+package restrictedModule_test
+
+import (
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Done.

internal/service/restrictedModule/restrictedModuleMatch_test.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	kyma := &v1beta2.Kyma{
+		ObjectMeta: apimetav1.ObjectMeta{Labels: map[string]string{"env": "production", "tier": "backend"}},
+	}
+
+	_, err := restrictedModule.RestrictedModuleMatch(mrm, kyma)
+
+	require.Error(t, err)
+	require.ErrorIs(t, err, restrictedModule.ErrSelectorParse)
+	require.ErrorContains(t, err, "\"InvalidOperator\" is not a valid label selector operator")
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Done.

medmes (commented):
Overall is good , just some minor suggestions!

internal/service/restrictedmodule/match.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+)
+
+var ErrSelectorParse = errors.New("failed to parse kymaSelector on ModuleReleaseMeta")
+
+// RestrictedModuleMatch extracts the kymaSelector from a ModuleReleaseMeta
+// and checks if it matches labels on a specified Kyma instance.
+// A nil or empty selector returns false.
+func RestrictedModuleMatch(mrm *v1beta2.ModuleReleaseMeta, kyma *v1beta2.Kyma) (bool, error) {
```
medmes, line 18: Function name stutters with the package name — call site reads `restrictedmodule.RestrictedModuleMatch(...)`. Rename to `Match` → `restrictedmodule.Match(mrm, kyma)`.

internal/service/restrictedmodule/match.go
```diff
@@ -0,0 +1,33 @@
+package restrictedmodule
```
medmes, line 1: Non-idiomatic filename — Go convention is snake_case or flat lowercase. Rename to `restricted_module_match.go` or just `match.go`.

internal/service/restrictedmodule/match.go
```diff
@@ -0,0 +1,33 @@
...
+	k8slabels "k8s.io/apimachinery/pkg/labels"
+
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+)
+
+var ErrSelectorParse = errors.New("failed to parse kymaSelector on ModuleReleaseMeta")
+
+// RestrictedModuleMatch extracts the kymaSelector from a ModuleReleaseMeta
+// and checks if it matches labels on a specified Kyma instance.
+// A nil or empty selector returns false.
```
medmes, line 17: Add a sentence explaining *why* — this deviates from standard Kubernetes semantics where an empty selector matches everything. Intent here is: no selector means the module is unrestricted, so it matches nothing.

internal/service/restrictedmodule/match.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+	if len(kymaSelector.MatchLabels) == 0 && len(kymaSelector.MatchExpressions) == 0 {
+		return false, nil
+	}
+
+	selector, err := apimetav1.LabelSelectorAsSelector(kymaSelector)
+	if err != nil {
+		return false, fmt.Errorf("%w: %w", ErrSelectorParse, err)
+	}
+
```
medmes, line 35: `kyma` has no nil guard — `kyma.Labels` will panic if nil is passed. Either add a check or document the precondition explicitly.

internal/service/restrictedmodule/match_test.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	matched, err := restrictedmodule.RestrictedModuleMatch(mrm, kyma)
+
+	assert.False(t, matched)
+	require.Error(t, err)
+	require.ErrorIs(t, err, restrictedmodule.ErrSelectorParse)
+	require.ErrorContains(t, err, "InvalidOperator")
+	require.ErrorContains(t, err, "operator")
+}
```
medmes, line 235: Missing cases worth adding: Kyma with a nil labels map, at least one test each for `NotIn`/`Exists`/`DoesNotExist` operators, and a nil `kyma` test if a guard is added.

internal/service/restrictedmodule/match.go
Tomasz-Smelcerz-SAP: I have mixed feelings about it... This function will be primarily used by the service code that will be in the same package (probably). If so, then it's better to have the name `RestrictedModuleMatch()`, as simply `Match()` is too generic. I would leave such renaming to when the service and other code is in place.

lindnerby: Let's take the name as is, now (`RestrictedModuleMatch`). A renaming is quickly done if really needed. 👍

internal/service/restrictedmodule/match_test.go
Tomasz-Smelcerz-SAP: I don't see a point in testing the operators. This logic comes from k8s packages, they are already tested. It's enough to ensure that any of them works. We're not "constructing" these operators here, we're just using the results of parsing (which is also done by k8s packages).

The kyma==nil test is a bit paranoid, considering we want to use this in a reconciler, where we guarantee that kyma is not nil, but OK, I can add this check.
Nil labels are already tested, but I added a specific test case for that, just for clarity.

internal/service/restrictedmodule/match.go
medmes: Alright!

internal/service/restrictedmodule/match.go
Tomasz-Smelcerz-SAP: The kyma==nil test is a bit paranoid, considering we want to use this in a reconciler, where we guarantee that kyma is not nil, but OK, I can add this check.

internal/service/restrictedmodule/match.go
Tomasz-Smelcerz-SAP: Renamed to match.go

internal/service/restrictedmodule/match.go
Tomasz-Smelcerz-SAP: I disagree. The contract description here is clearly specified. This function is just a helper that will be used by the service.
The elaborate description for _why_ certain things are done in this way or the other should be placed on the Service and maybe in the project documentation, where we can refer to additional context of "restricted modules", what they are, what is their lifecycle etc.
What is the point of repeating the same message here?
IMO this is a low level function  and we __shoudn't__ include the information on _why_ our API contract is defined like this  __on this function__.
Otherwise our API contract description is "spread out" across entire codebase and it's extremely hard to find out all the pieces that are relevant to it. The _why_ for API contract should be centralized: On the Service level, in the project docs.
Here we just should provide a minimal, precise description about what this particular function is doing. This description is already provided.

internal/service/restrictedmodule/match.go
Tomasz-Smelcerz-SAP: Added.

medmes (approved):
After the new changes, looks good for me!
Just the e2e needed to be fixed!

c-pius (approved):

PR #3229, opened 2026-04-28
feat: Enable Restricted Modules when defaulted

### LINKED ISSUE #3214: Enable restricted module by default for matching GA created Kyma
## Description

Extend the Kyma controller to configure restricted default modules. If `flagVar.GetRestrictedDefaultModules()` (#3211) contains module names, fetch the related MRMs. Only if the `RestrictedModuleMatch(mrm, kyma)` func returns a match, add the module to the KCP Kyma's `.spec.modules[]` if not present already. Add it with:

```yaml
spec:
  modules:
  - name: <restricted-default-module-name>
    managed: true
```

This needs to happen directly after we get the KCP Kyma in the reconcliation loop.

## Reason

KLM needs to extend the default module list with the modules (only one expected for now) listed in the `RestrictedDefaultModules` flag (#3211). It was explicitly decided that the module shall not be a mandatory module, but KEB at the same time can't create the Kyma CR with this module listed in the KCP `Kyma.spec.modules`. Therefore, KLM will do the hack and add it to the `Kyma.spec.modules` list on KCP if the Kyma matches the `kymaSelector` of the MRM corresponding to the RestrictedDefaultModules entry.

>[!Caution]
>This hack does not install this module to existing Kymas matching the `kymaSelector`

>[!Caution]
>When a global account is removed from the `kymaSelector`, this hack does not remove the module from the `.spec.modules` list of the Kyma. The module is uninstalled from the cluster and removed from the module catalog as it is now allowed anymore, but the Kyma will have an entry in the modules list in error state saying that the module is not allowed.

## AC

- [x] ‼️ Verify with KEB that they are NOT reconciling the Kyma spec modules and thereby revert our addition to the modules list.
  - KEB does NOT edit the Kyma spec after creation
- [x] When `flagVar.GetRestrictedDefaultModules()` is empty
  - [x]  do nothing (continue reconciliation as normal)
- [x] When `flagVar.GetRestrictedDefaultModules()` contains entries
  - [x] do NOT enable the module by default
  - [x] only if the `RestrictedModuleMatch(mrm, kyma)` returns `true`, and if the module is not already in the spec, add the module to the spec of the KCP Kyma
- [x] When the module is already installed and the matching criteria changes (either on Kyma or MRM side)
  - [x] do nothing
- [x] Error Handling
  - [x] When the MRM is not found => make the problem visible but don't fail the reconciliation
    - [x] do NOT enable the module by default
  - [x] When the selector cannot be parsed => make the problem visible but don't fail the reconciliation
    - [x] do NOT enable the module by default
  - [x] When the Update of the KCP Kyma fails => return the error and fail the reconciliation
- [x] 100% unit test coverage for the new service and changes in the repos
- [x] New E2E test suite introduced
  - [x] When configured default modules
    - [x] When MRM cannot be fetched => Kyma spec not updated, problem logged
    - [x] When selector from MRM cannot be parsed => Kyma spec not updated, problem logged
    - [x] When selector doesn't match the Kyma => Kyma spec not updated
    - [x] When selector matches the Kyma
      - [x] And spec update fails => Kyma in error state
      - [x] And spec update succeeds => Kyma spec contains module

## Dev Notes

- should be implemented as a self-contained service that is injected to the Kyma controller
  - args should only need Kyma
  - service encapsulates MRM repo, Kyma repo (for the Update) call, `RestrictedModuleMatch` func, `flagVar.GetRestrictedDefaultModules()` list
- there should be no-need to requeue or re-fetch the KCP Kyma since the potential Update from controller-runtime populates the passed object with the data from the object living in the cluster. The continued reconciliation uses the update spec module list already
- first step in the service should be to check if the `Kyma.spec.modules` already contains the `flagVar.GetRestrictedDefaultModules()`. If so, bail out early so that we don't have potentially failing calls getting the MRM or parsing the selector

c-pius:
Implements a defaulter adding restriced modules configured in `--restricted-default-modules` to the KCP Kyma spec if the KCP Kyma matches the kymaSelector from the related ModuleReleaseMeta.

Notes:

- in a general context, the logging may be a bit verbose but in this context I would prefer to have a good amount of logs so we can reason about (miss-)behavior.

closes #3214

### DIFF
--- a/internal/controller/kyma/controller.go
+++ b/internal/controller/kyma/controller.go
@@ -139,6 +139,8 @@ type Reconciler struct {
 // Reconcile reconciles Kyma resources.
 //
 // See https://github.com/kyma-project/lifecycle-manager/issues/2943.
+//
+//nolint:funlen // disable for kyma controller until further refactoring
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logger := logf.FromContext(ctx)
 	logger.V(log.DebugLevel).Info("Kyma reconciliation started")
--- a/internal/controller/kyma/controller.go
+++ b/internal/controller/kyma/controller.go
@@ -99,6 +99,10 @@ type SkrSyncService interface {
 	SyncImagePullSecret(ctx context.Context, kyma types.NamespacedName) error
 }
 
+type RestrictedModules interface {
+	Default(ctx context.Context, kyma *v1beta2.Kyma) error
+}
+
 // ReconcilerConfig holds configuration values for the Kyma Reconciler.
 // Usually read from flags or environment variables.
 type ReconcilerConfig struct {
@@ -125,10 +129,11 @@ type Reconciler struct {
 	RemoteCatalog  *remote.RemoteCatalog
 	TemplateLookup *templatelookup.TemplateLookup
 
-	DeletionMetrics DeletionMetricWriter
-	DeletionEvents  DeletionEventRecorder
-	DeletionService DeletionService
-	LookupService   LookupService
+	DeletionMetrics   DeletionMetricWriter
+	DeletionEvents    DeletionEventRecorder
+	DeletionService   DeletionService
+	LookupService     LookupService
+	RestrictedModules RestrictedModules
 }
 
 // Reconcile reconciles Kyma resources.
@@ -152,6 +157,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, fmt.Errorf("KymaController: %w", err)
 	}
 
+	err := r.RestrictedModules.Default(ctx, kyma)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("KymaController: failed to default restricted modules: %w", err)
+	}
+
 	status.InitConditions(kyma, r.WatcherEnabled(), r.SkrImagePullSecretSyncEnabled())
 
 	if kyma.SkipReconciliation() && kyma.DeletionTimestamp.IsZero() {
@@ -159,7 +169,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{RequeueAfter: r.Success}, nil
 	}
 
-	err := r.SkrContextFactory.Init(ctx, kyma.GetNamespacedName())
+	err = r.SkrContextFactory.Init(ctx, kyma.GetNamespacedName())
 	if !kyma.DeletionTimestamp.IsZero() && errors.Is(err, accessmanager.ErrAccessSecretNotFound) {
 		return r.handleDeletedSkr(ctx, req, kyma)
 	}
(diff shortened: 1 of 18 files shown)


internal/service/restrictedmodule/defaulter.go
```diff
@@ -0,0 +1,128 @@
...
+			WithValues("kyma", kyma.Name,
+				"modules", d.restrictedDefaultModules,
+				"service", "restricted module defaulter",
+			).
+			Error(err, "Failed to update Kyma")
+		return fmt.Errorf("failed to update Kyma %s with restricted default modules %v: %w",
+			kyma.Name,
+			d.restrictedDefaultModules,
+			err,
+		)
```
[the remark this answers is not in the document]
c-pius: Removed the list. We can see from the previous logs what was added.

internal/service/restrictedmodule/defaulter.go
```diff
@@ -0,0 +1,126 @@
...
+	)
+
+	if !kyma.GetDeletionTimestamp().IsZero() {
+		log.Info("Skipping as Kyma is Deleting")
+		return nil
+	}
+
+	// nothing to do
+	if len(d.restrictedDefaultModules) == 0 {
+		log.Info("Skipping as there are no restriced default modules configured")
```
Tomasz-Smelcerz-SAP: This will be logged for every reconciliation? If so, I think it will just spam the log stream with redundant messages. I think it's better to produce a log entry only when we're adding a module to the list, as this is a "change in the system". No log entries ==> no changes in the system.

internal/controller/kyma/controller.go
```diff
@@ -152,14 +159,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, fmt.Errorf("KymaController: %w", err)
 	}
 
+	err := r.RestrictedModules.Default(ctx, kyma)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("KymaController: failed to default restricted modules: %w", err)
+	}
```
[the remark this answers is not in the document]
c-pius: Introduced with e2e tests now.

internal/service/restrictedmodule/defaulter.go
```diff
@@ -0,0 +1,126 @@
...
+		return fmt.Errorf("failed to update Kyma %s with restricted default modules: %w",
+			kyma.Name,
+			err,
+		)
+	}
+
+	return nil
+}
+
+func skipAlreadyEnabled(kyma *v1beta2.Kyma, moduleName string) bool {
```
Tomasz-Smelcerz-SAP: I know it's nitpicking, but `isAlreadyEnabled` would be a better name, as this function is not "skipping" anything.

internal/service/restrictedmodule/defaulter.go
c-pius: Removed the no-op info logs

internal/service/restrictedmodule/defaulter.go
c-pius: Renamed

Tomasz-Smelcerz-SAP (approved):

Issue #3209, opened 2026-04-24
Add Global Account ID restricted installation list to module


c-pius:
### Description

The ModuleReleaseMeta type has been extended with the `KymaSelector *apimetav1.LabelSelector json:"kymaSelector,omitempty"`. This selector needs to be populated with data for certain modules. As of now, we want to restrict the module teams to be allowed to only match Global Accounts.

If possible restrict this feature only to the module called `deployer-service`, meaning `kyma/module-manifests/modules/deployer-servicemodule-releases.yaml`, if not, then allow the `kyma/module-manifests/modules/<module-name>/module-releases.yaml` file to be extended with the following:

```yaml
restrictToGlobalAccounts: # list[string]
- <uuid-1>
- <uuid-2>
- <uuid-3>
```

If `restrictToGlobalAccounts` is set, the Submission Pipeline shall add the following to the ModuleReleaseMeta file of EVERY landscape (`kyma/kyma-modules/<module-name>/<dev|stage|prod>/module-release-meta.yaml`):

```yaml
spec:
  kymaSelector:
    matchExpressions:
    - key: kyma-project.io/global-account-id
      operator: In
      values:
      - <uuid-1>
      - <uuid-2>
      - <uuid-3>
```

If `restrictToGlobalAccounts` is missing, then do NOT provide a `kymaSelector` to the MRM. If the `restrictToGlobalAccounts` list is empty, fail the pipeline. If `.restrictToGlobalAccounts` and `.mandatory` are present at the same time, fail the pipeline.

>[!Note]
> This feature may be reverted in the future.

### Reasons

The module needs to be marked as being selectively installed. If this marker is present, KLM looks up the selection criteria from the label selector. 

### Acceptance Criteria

- [ ] When valid `restrictToGlobalAccounts` is present
  - [ ] Then create the `kymaSelector` in the generated ModuleReleaseMeta for ALL landscapes
  - [ ] ‼️ ensure that this update propagates to KCP even though no version may be bumped alongside
- [ ] When `restrictToGlobalAccounts == nil
  - [ ]  Then do NOT create a `kymaSelector`
- [ ] When `restrictToGlobalAccounts != nil && isEmpty(restrictToGlobalAccounts)`
  - [ ] Then the pipeline fails
- [ ] When `restrictToGlobalAccounts != nil && .mandatory != nil`
  - [ ] Then fail the pipeline

c-pius:
[...]

PR #3230, opened 2026-04-29
feat: Support restricted module installation

### LINKED ISSUE #3213: Support restricted module installation in module reconciler
## Description

Update the template mode validation for restricted modules. If the `.spec.moduleName` of the ModuleReleaseMeta is in the list of restrictedDefaultModules (https://github.com/kyma-project/lifecycle-manager/issues/3211), the module MUST NOT be installable unless we get an explicit match from the matcher func (https://github.com/kyma-project/lifecycle-manager/issues/3219). If the module name is NOT included in the restricted default modules, the module should install as usual.

If the matching criteria changes, and a previously installed module is not allowed for this Kyma anymore, the module shall be uninstalled.

## Reason

Restricted modules MUST NOT be installable to Kymas not matching the selection criteria. For the restricted default modules, also have no install by default.

## AC

- [x] When `flagVar.GetRestrictedDefaultModules()` includes MRM `.spec.moduleName`, then
  - [x] do NOT allow module installation by default
  - [x] only allow module installation if `RestrictedModuleMatch(mrm, kyma)` returns `true`
- [x] When a Kyma matched the criteria and had the module installed, and the matching criteria gets modified so that the Kyma doesn't match anymore, uninstall the module from the cluster
- [x] 100% unit test coverage for the related code parts in validate template mode
- [x] E2E tests `ModuleReleaseMeta Not Allowed Installation` cover
  - [x] Kyma and nil selector in MRM => module NOT installed
  - [x] Kyma and empty selector in MRM => module NOT installed
  - [x] Kyma matching the selector => module installed
  - [ ] Selector in MRM updated, Kyma still matches => module installed
  - [ ] Selector in MRM updated, Kyma doesn't match anymore => module NOT installed
  - [ ] Kyma labels updated, Kyma matches again => module installed
  - [ ] Selector in MRM updated, Kyma doesn't match anymore => module NOT installed
  - [ ] Selector in MRM updated, Kyma matches again => module installed

## Dev Notes

- this where validate template mode checks that the module is allowed for the Kyma
  - https://github.com/kyma-project/lifecycle-manager/blob/14420ec01eb661665bd3f226a23bd0e82b2a0055/pkg/templatelookup/regular.go#L141-L156
- the MRM is already available and can be passed as argument
  - https://github.com/kyma-project/lifecycle-manager/blob/14420ec01eb661665bd3f226a23bd0e82b2a0055/pkg/templatelookup/regular.go#L84-L87
-  if we set the `ErrTemplateNotAllowed`, the `ReconcileManifests` function already takes care of uninstalling the module again
  - https://github.com/kyma-project/lifecycle-manager/blob/14420ec01eb661665bd3f226a23bd0e82b2a0055/pkg/module/sync/runner.go#L54-L58

Notes:
* Module installation should detect if a module that is marked as being selectively installed
* checks the name of every module to be installed against the provided KLM flag containing of restricted modules
  * if the MRM does not contain a global account label selector AND module name does not match a configured restricted module name -> continue installation of module
  * if the MRM does not contain a global account label selector AND module name matches a configured restricted module name -> skip installation
  * if the MRM does contain a global account label selector, but does not match a configured name -> skip installation
  *  if the MRM does contain a global account label selector, and matches the configured name -> continue installation
* check the list of GA IDs provided by selector against the kyma's global account id:
  * only matches will lead to module being installed.

lindnerby:
## Summary

- Extends `ValidateTemplateMode` to block installation of restricted modules unless the MRM `kymaSelector` explicitly matches the Kyma's labels
- Modules in the `--restricted-default-modules` flag list are denied by default (`ErrTemplateNotAllowed`), which triggers uninstallation via the existing manifest sync runner when matching criteria change
- Adds `WithKymaSelector` and `WithGlobalAccountKymaSelector` builder helpers for tests

Closes #3213

## Test plan

- [x] Units tests
- [x] E2E tests


pkg/templatelookup/regular.go
```diff
@@ -152,6 +161,33 @@ func validateTemplateMode(template ModuleTemplateInfo, kyma *v1beta2.Kyma) Modul
...
+		case !inRestrictedList && !hasSelector:
+			// normal module, allow
+		case !inRestrictedList && hasSelector:
+			template.Err = fmt.Errorf("%w: module has kymaSelector but is not in restricted modules list",
+				ErrTemplateNotAllowed)
+			return template
+		case inRestrictedList && !hasSelector:
+			template.Err = fmt.Errorf("%w: restricted module has no kymaSelector configured",
+				ErrTemplateNotAllowed)
+			return template
```
c-pius, line 177: One thought that came to mind is the error message. These will surface at the SKR. We could think about just setting the generic ErrTemplateNotAllowed, and complement it with an Error Log giving more details but is only available on KCP.

The same is true with the error from kymaRepo.Update() in my PR actually. Maybe a bit less prominent as this logic should only execute before the SKR Kyma is created and therefore no status is visible to the customer. But it could surface if a new restricted module is added.

pkg/templatelookup/regular.go
```diff
@@ -152,6 +161,33 @@ func validateTemplateMode(template ModuleTemplateInfo, kyma *v1beta2.Kyma) Modul
...
 	}
+	if mrm != nil {
+		inRestrictedList := slices.Contains(t.restrictedModules, mrm.Spec.ModuleName)
+		hasSelector := mrm.Spec.KymaSelector != nil
+
+		switch {
+		case !inRestrictedList && !hasSelector:
+			// normal module, allow
+		case !inRestrictedList && hasSelector:
+			template.Err = fmt.Errorf("%w: module has kymaSelector but is not in restricted modules list",
```
c-pius, line 172: Maybe also something that we could ask neighbors to do. Only allow specifying the GAs if module name is `deployer-service`.

config/watcher_local_test/kustomization.yaml
```diff
@@ -81,6 +81,9 @@ patches:
       - op: replace
         path: /spec/template/spec/containers/0/imagePullPolicy
         value: Always
+      - op: add
+        path: /spec/template/spec/containers/0/args/-
+        value: --restricted-default-modules=not-template-operator
```
medmes: This static patch has two problems.

First, the value `not-template-operator` means `template-operator` (the standard test module) is never restricted. The AC scenarios in #3213 require the actual test module to be in the restricted list so the selector gating logic is exercised. This value can't drive those scenarios.

Second, `restricted_default_module_test.mk` already injects `--restricted-default-modules` via `kustomize edit add patch` at runtime. With this static patch also present, the container ends up with two `--restricted-default-modules` args (last-value wins in Go's flag package, so it functions but is fragile).

The established pattern from `restricted_default_module_test.mk` is to apply the flag in the test-specific `klm-patch` target, not in the shared kustomization. This should be removed and handled in the `.mk` file for the new E2E scenarios.

pkg/templatelookup/regular.go
```diff
@@ -152,6 +161,33 @@ func validateTemplateMode(template ModuleTemplateInfo, kyma *v1beta2.Kyma) Modul
 			common.ErrNoTemplatesInListResult, template.Name, template.DesiredChannel)
 		return template
 	}
+	if mrm != nil {
+		inRestrictedList := slices.Contains(t.restrictedModules, mrm.Spec.ModuleName)
+		hasSelector := mrm.Spec.KymaSelector != nil
```
medmes, line 166: `RestrictedModuleMatch` explicitly returns `false` for both `nil` and empty `{}` selectors (guarded by the `len(MatchLabels)==0 && len(MatchExpressions)==0` check). But `hasSelector := mrm.Spec.KymaSelector != nil` treats an empty `{}` as `true`, so a restricted module with `kymaSelector: {}` routes into `inRestrictedList && hasSelector`, calls the matcher, gets `false`, and surfaces `"restricted module not allowed for this Kyma"`. The correct message is `"restricted module has no kymaSelector configured"`. The unit test for the empty-selector case passes for the wrong reason.

Fix to align with `RestrictedModuleMatch` semantics:
```go
hasSelector := mrm.Spec.KymaSelector != nil &&
    (len(mrm.Spec.KymaSelector.MatchLabels) > 0 || len(mrm.Spec.KymaSelector.MatchExpressions) > 0)
```

medmes:
The AC requires: *"When a Kyma matched the criteria and had the module installed, and the matching criteria gets modified so that the Kyma doesn't match anymore, uninstall the module."*

The uninstall relies on the existing `ErrTemplateNotAllowed → deleteManifest` path in `runner.go`. The unit tests verify the error is set, but nothing covers the full state transition: module installed → `kymaSelector` updated → module removed. This is the most operationally critical AC item and it has no test coverage at any level (unit, integration, or E2E). A targeted test for this transition would be the minimum.

config/watcher_local_test/kustomization.yaml
lindnerby: Yes, I was only testing the current E2E if they work with a `--restricted-default-modules` defined 👍

pkg/templatelookup/regular.go
lindnerby: Good idea, will let them know.

lindnerby: Communicated and issue updated: https://github.com/kyma-project/lifecycle-manager/issues/3209

pkg/templatelookup/regular_test.go
```diff
@@ -130,14 +132,222 @@ func TestValidateTemplateMode_ForOldModuleTemplates(t *testing.T) {
...
+			kyma: builder.NewKymaBuilder().
+				WithLabel(shared.GlobalAccountIDLabel, "account-1").
+				Build(),
+			template: templatelookup.ModuleTemplateInfo{
+				ModuleTemplate: builder.NewModuleTemplateBuilder().Build(),
+			},
+			wantErr: templatelookup.ErrTemplateNotAllowed,
+		},
+		{
+			name:              "When module in restricted list but no selector configured, Then not allowed",
```
Tomasz-Smelcerz-SAP: This test case is identical to the very first one.

pkg/templatelookup/regular_test.go
```diff
@@ -130,14 +132,222 @@ func TestValidateTemplateMode_ForOldModuleTemplates(t *testing.T) {
...
+				WithLabel(shared.GlobalAccountIDLabel, "account-1").
+				Build(),
+			template: templatelookup.ModuleTemplateInfo{
+				ModuleTemplate: builder.NewModuleTemplateBuilder().Build(),
+			},
+			wantErr:     restrictedmodulesvc.ErrSelectorParse,
+			wantErrAlso: templatelookup.ErrTemplateNotAllowed,
+		},
+		{
+			name:              "When module in restricted list and matching matchLabels selector, Then allowed",
```
Tomasz-Smelcerz-SAP, line 311: This test case is almost identical to the one defined in line 187. The difference is only in the selector configuration and so it looks like we're focusing on testing differences in the "selector matching" logic, which is already tested: primarily by the k8s code and also by our "MatchingFunc" unit tests.
So we don't need to test "selector matching" logic here and only one test case is enough.

pkg/templatelookup/regular_test.go
```diff
@@ -130,14 +132,222 @@ func TestValidateTemplateMode_ForOldModuleTemplates(t *testing.T) {
...
+			kyma: builder.NewKymaBuilder().
+				WithLabel(shared.GlobalAccountIDLabel, "account-1").
+				Build(),
+			template: templatelookup.ModuleTemplateInfo{
+				ModuleTemplate: builder.NewModuleTemplateBuilder().Build(),
+			},
+			wantErr: templatelookup.ErrTemplateNotAllowed,
+		},
+		{
+			name:              "When restricted modules list is empty, Then allowed",
```
Tomasz-Smelcerz-SAP: For completeness, we could add "When restricted modules list is empty and has selector, Then not  allowed"
This is a corner case, but it could be that the implementation has a specific branch for handling "an empty" list and in that branch this condition is not handled properly

pkg/templatelookup/regular_test.go
lindnerby: True. Are you fine with keeping it?

Tomasz-Smelcerz-SAP: yes, it's not 100% identical, it was just suggestion from my side that we can have one less test case here. But If you think it verifies a specific case in the API contract (just a single value in the selector), I am fine with it.

Tomasz-Smelcerz-SAP (approved):

PR #3231, opened 2026-04-29
feat: Restrict Modules in Remote Catalog Sync

### LINKED ISSUE #3212: Remove restricted modules from remote catalogue
## Description

Update the remote catalog sync for restricted modules. If the `.spec.moduleName` of the ModuleReleaseMeta is in the list of restrictedDefaultModules (https://github.com/kyma-project/lifecycle-manager/issues/3211), the ModuleReleaseMeta and related ModuleTemplates MUST NOT be synced unless we can an explicit match from the matcher func (https://github.com/kyma-project/lifecycle-manager/issues/3219).  If the module name is NOT included in the restricted default modules, the MRMs and MTs should sync as usual.

If the matching criteria changes, and the module is not allowed for this Kyma anymore, the module catalog shall be removed.

## Reason

Restricted modules MUST NOT be visible to Kymas not matching the selection criteria. For the restricted default modules, also have a no sync by default.

## AC

- [ ] When `flagVar.GetRestrictedDefaultModules()` includes MRM `.spec.moduleName`, then
  - [ ] do NOT sync MRM and MT to Kyma by default
  - [ ] only sync MRM and MT if `RestrictedModuleMatch(mrm, kyma)` returns `true`
- [ ] When a Kyma matched the criteria and had MRMs and MTs synced, and the matching criteria gets modified so that the Kyma doesn't match anymore, the MRMs and MTs get removed from the cluster
- [ ] 100% unit test coverage for the related code parts in the catalog sync
- [ ] E2E tests `ModuleReleaseMeta Sync` cover
  - [ ] Kyma and nil selector in MRM => MRMs + MTs NOT synced
  - [ ] Kyma and empty selector in MRM => MRMs + MTs NOT synced
  - [ ] Kyma matching the selector => MRMs + MTs synced
  - [ ] Selector in MRM updated, Kyma still matches => MRMs + MTs synced
  - [ ] Selector in MRM updated, Kyma doesn't match anymore => MRMs + MTs removed
  - [ ] Kyma labels updated, Kyma matches again => MRMs + MTs synced
  - [ ] Selector in MRM updated, Kyma doesn't match anymore => MRMs + MTs removed
  - [ ] Selector in MRM updated, Kyma matches again => MRMs + MTs synced

## Dev Notes

- this is where we determine the MRMs to sync
  - https://github.com/kyma-project/lifecycle-manager/blob/14420ec01eb661665bd3f226a23bd0e82b2a0055/internal/remote/remote_catalog.go#L131-L150
- MTs only sync if they are referenced from a MRM
  - https://github.com/kyma-project/lifecycle-manager/blob/14420ec01eb661665bd3f226a23bd0e82b2a0055/internal/remote/remote_catalog.go#L169-L178
- sync funcs remove MRMs and MTs from the Runtimes when not provided anymore
  - https://github.com/kyma-project/lifecycle-manager/blob/14420ec01eb661665bd3f226a23bd0e82b2a0055/internal/remote/moduletemplate_syncer.go#L49-L75
  - https://github.com/kyma-project/lifecycle-manager/blob/14420ec01eb661665bd3f226a23bd0e82b2a0055/internal/remote/modulereleasemeta_syncer.go#L49-L78

LeelaChacha:
**Description**

- ...
- ...
- ...

**Related issue(s)**

Closes #3212

### DIFF
--- a/internal/remote/remote_catalog.go
+++ b/internal/remote/remote_catalog.go
@@ -8,6 +8,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
 	"github.com/kyma-project/lifecycle-manager/internal/common/fieldowners"
@@ -124,7 +125,7 @@ func (c *RemoteCatalog) Delete(
 // A ModuleReleaseMeta is synced if it has at least
 // one channel-version pair whose ModuleTemplate is allowed to be synced.
 // Restricted modules (in the restrictedModules list) are only synced if the MRM's kymaSelector matches the Kyma.
-// Non-restricted modules with a kymaSelector set are also checked against the Kyma labels.
+// Non-restricted modules with a kymaSelector are skipped.
 func (c *RemoteCatalog) GetModuleReleaseMetasToSync(
 	ctx context.Context,
 	kyma *v1beta2.Kyma,
@@ -152,7 +153,8 @@ func (c *RemoteCatalog) GetModuleReleaseMetasToSync(
 				continue
 			}
 		} else if moduleReleaseMeta.Spec.KymaSelector != nil {
-			// this shouldn't happen. MRM should not have a label selector if it is not a restricted module.
+			logf.FromContext(ctx).Info("WARNING: skipping catalog sync: MRM has kymaSelector but is not a restricted module",
+				"moduleName", moduleReleaseMeta.Spec.ModuleName)
 			continue
 		}
 
--- a/internal/remote/remote_catalog.go
+++ b/internal/remote/remote_catalog.go
@@ -124,6 +124,7 @@ func (c *RemoteCatalog) Delete(
 // A ModuleReleaseMeta is synced if it has at least
 // one channel-version pair whose ModuleTemplate is allowed to be synced.
 // Restricted modules (in the restrictedModules list) are only synced if the MRM's kymaSelector matches the Kyma.
+// Non-restricted modules with a kymaSelector set are also checked against the Kyma labels.
 func (c *RemoteCatalog) GetModuleReleaseMetasToSync(
 	ctx context.Context,
 	kyma *v1beta2.Kyma,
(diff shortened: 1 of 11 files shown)


internal/remote/remote_catalog.go
```diff
@@ -130,12 +139,24 @@ func (c *RemoteCatalog) GetModuleReleaseMetasToSync(
...
 	for _, moduleReleaseMeta := range moduleReleaseMetaList.Items {
 		if moduleReleaseMeta.Spec.Mandatory != nil {
-			// Skip mandatory ModuleReleaseMetas as they are not allowed to be synced
+			continue
+		}
+
+		if c.isRestrictedModule(moduleReleaseMeta.Spec.ModuleName) {
+			matched, err := restrictedmodule.RestrictedModuleMatch(&moduleReleaseMeta, kyma)
+			if err != nil {
+				return nil, fmt.Errorf("failed to evaluate restricted module match for %s: %w",
```
LeelaChacha: Still deciding if we throw an error here or just skip

lindnerby: We can treat this the same as not matched right?

internal/remote/remote_catalog_test.go
```diff
@@ -305,6 +305,403 @@ func Test_IsAllowedModuleVersion_ForInternalKyma_InternalModule(t *testing.T) {
...
+	kyma := newKymaBuilder().withLabel("kyma-project.io/global-account-id", "account-999").build()
+	mts := &v1beta2.ModuleTemplateList{Items: []v1beta2.ModuleTemplate{*restrictedMT}}
+
+	mrms, err := remoteCatalog.GetModuleReleaseMetasToSync(t.Context(), kyma, mts)
+
+	require.NoError(t, err)
+	assert.Empty(t, mrms)
+}
+
+func Test_GetModuleReleaseMetasToSync_NonRestrictedModule_NotSynced_WhenSelectorDoesNotMatch(t *testing.T) {
```
Tomasz-Smelcerz-SAP: I have a problem with this test case. It is almost identical to the next one `[...]WhenSelectorSetButNotInList` - the module name is also *not in the list*. It looks like the test is doing the same check as the next one. Am I missing anything?

LeelaChacha: That's the case that we discussed on Slack. In the hightlighted test, the selector does not match, and in the case the selctor does not match and the module is not a restricted one - also should not be synced

Tomasz-Smelcerz-SAP (approved):

PR #3267, opened 2026-05-13
chore: Claude hook enforcing FIPS on Go commands

### LINKED ISSUE #3241: Spike: Agentic AI coding support
## Objective
Define a pragmatic agentic AI coding approach for, building on the initial guidance introduced in PR #3139.

## Acceptance Criteria
- [x] All Jellyfish Repos have a CLAUDE.md to give claude-code more context on the current repo it operates on
  - [x] claude init command can help to start, but the created mds need to be reviewed and extended on 
- [ ] Eventual follow-up ideas for context engineering / skills ideas are created

## Timebox
2d

## Related Initiative
- PR #3139 () adds baseline AI-assistant guidance and should be treated as the starting point.

c-pius:
**Description**

- tried to create a hook that will prevent Claude from making changes to our Go commands removing FIPS mode
- not entirely happy with that yet, would also want to look into rules as an alternative
  - UPDATE:
    - the problem with rules is that they don't guarantee that Claude will adhere to them
    - hooks would be better since they are deterministic
    - what I don't like about it is the rather verbose configuration in settings.json. But I would swallow that pill

image

**Related issue(s)**


tests/e2e/Makefile
```diff
@@ -36,8 +36,9 @@ GINKGO_VERSION ?= v2.12.0
 # FIPS140 Module version
 FIPS140_MODULE_VERSION := v1.0.0
 
-set-fips:
-	export GOFIPS140=$(FIPS140_MODULE_VERSION)
```
c-pius, line 40: This seems to be an obsolete leftover

Makefile
```diff
@@ -111,7 +114,7 @@ build: generate fmt vet ## Build manager binary.
 
 .PHONY: run
 run: manifests generate fmt vet ## Run a controller from your host.
-	go run ./cmd/main.go
+	$(GO) run ./cmd/main.go
```
c-pius, line 117: Here we missed to set fips earlier

medmes (approved):

PR #3337, opened 2026-06-08
chore: Migrate E2E tests with trivial KLM patches to Makefile

LeelaChacha:
Ports the following E2E tests to the self-contained Makefile approach:

- `purge-controller`
- `purge-metrics`
- `kyma-metrics`
- `self-signed-certificate-rotation`
- `skr-image-pull-secret-sync`

Each test now lives in its own `*_test.mk` file. The `klm-patch` target writes a kustomize JSON patch appending the test-specific `--flag=value` to the deployment args (filename and flag match what `deploy-lifecycle-manager-e2e` does in CI). `module-setup` deploys 3 ModuleTemplates (current/older/newer) plus a ModuleReleaseMeta — single-channel `regular:OLDER` for the first four, two-channel `fast:NEWER regular:OLDER` for `skr-image-pull-secret-sync`.

The five tests are added to the self-contained list in `test-e2e.yaml` so the shared `Deploy lifecycle-manager` and `Deploy template-operator` actions are skipped for them.

The corresponding patch steps in `deploy-lifecycle-manager-e2e/action.yml` are intentionally kept (also present on `management-plane-charts`) and gated by the existing `self-contained == 'false'` condition on the deploy step.

Closes #3150


c-pius (dismissed):
Approved with one suggestion

> The corresponding patch steps in deploy-lifecycle-manager-e2e/action.yml are intentionally kept (also present on management-plane-charts) and gated by the existing self-contained == 'false' condition on the deploy step.

I would vote for directly removing them from deploy-lifecycle-manager-e2e/action.yml to ensure we don't have any residues left

c-pius (approved):

PR #3347, opened 2026-06-11
refactor: Extract ResourceRenderService from Manifest reconciler

LeelaChacha:
## Description

Extracts `renderTargetResources()` from the Manifest `Reconciler` into a dedicated `ResourceRenderService` living in `internal/service/manifest/render/`.

The `Reconciler` drops its `cachedManifestParser` and `resourceTransforms` fields and now holds a single `renderService ResourceRenderService` instead. `NewReconciler()` takes `ResourceRenderService` as a parameter (and no longer takes `skrImagePullSecretName` — transform composition moves to the wiring callers).

The consumer-side interface lives in `internal/declarative/v2` (per ADR 001), the implementation in the new render package (per ADR 004); both `RenderTargetResources` and `EvictCache` are exposed since `pruneDiff` invalidates the parser cache on in-OCI-layer diffs.

Existing behaviour is unchanged: same parser cache, same default transforms, same conditional `CreateSkrImagePullSecretTransform` when `--skr-image-pull-secret` is set, same SKR-coupled namespace normalisation in the reconciler.

## Changes

- New `internal/service/manifest/render` package with `Service` + unit tests
- New `ResourceRenderService` interface in `internal/declarative/v2`
- Reconciler fields and constructor signature simplified
- Wiring updated in `cmd/main.go`, `internal/controller/manifest/setup.go`, and the three manifest integration test suites

Resolves #3171


internal/declarative/v2/reconciler.go
```diff
@@ -546,31 +546,20 @@ func (r *Reconciler) renderTargetResources(ctx context.Context,
...
-
-	result := make([]client.Object, 0, len(targetResources.Items))
-	for _, unstrObj := range targetResources.Items {
-		converted := client.Object(unstrObj) // unstructured.Unstructured implements client.Object
-		err := normaliseNamespace(converted, apimetav1.NamespaceDefault, skrClient)
-		if err != nil {
-			recoverable := meta.IsNoMatchError(err)
-			if !recoverable {
+	for _, obj := range rendered {
+		if err := normaliseNamespace(obj, apimetav1.NamespaceDefault, skrClient); err != nil {
```
c-pius: I am rather surprised that we do this, but okay :) 

What about also moving this to the rendering? To me this even looks like another transform that runs at the end.

c-pius (dismissed):
Good change! One small suggestions

c-pius (approved):

PR #3358, opened 2026-06-16
refactor: Drop internal/declarative/v2 package

### LINKED ISSUE #2745: remove declarative `v2` package
### Description

after #2318 , #2329, we can deprecate the whole `v2` package, rewrite `internal/controller/manifest/controller.go` with a normal reconcile implementation.

### Reasons

_No response_

### Acceptance Criteria

- [ ] declarative `v2` package should be dropped completely, code inside the package relocated/deprecated/refactored completely
- [ ] stop using `declarative`, `v2` as new package name
- [ ] rewrite manifest controller to implement standard `reconcile`

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

LeelaChacha:
Relocates the contents of `internal/declarative/v2` into the layer each piece belongs to, then deletes the package.

- Reconciler + setup → `internal/controller/manifest/`
- `Spec` → `internal/manifest/spec/`
- Cached manifest parser → `internal/manifest/parser/`
- Resource transforms → `internal/service/manifest/render/`
- `ExistsStateCheck` → `internal/manifest/statecheck/`
- `ResourceList` → private to `internal/controller/manifest/`

Drops the `Object` abstraction (transforms now take `*v1beta2.Manifest` directly), removes the generated `mock_v2` package, consolidates duplicated constants onto `shared.OwnedByFormat` / `shared.DefaultRemoteNamespace`, lifts `labelsremoval.NewManagedByLabelRemovalService` to the composition root, turns `SpecResolver` into a consumer interface on the reconciler, inlines all collaborator interfaces alongside their consumer in `controller.go`, and pins the `manifestctrl` import alias in `.golangci.yaml`.

Closes #2745


internal/manifest/statecheck/exists_state_check.go
```diff
@@ -9,10 +9,9 @@ import (
 	"github.com/kyma-project/lifecycle-manager/api/shared"
 )
 
-type StateCheck interface {
```
LeelaChacha, line 12: removed to align with our decision of client-side interfaces

internal/controller/manifest/controller.go
```diff
@@ -159,82 +175,41 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-
-	skrClient, err := r.getTargetClient(ctx, manifest)
-	if err != nil {
-		manifest.SetStatus(manifest.GetStatus().WithState(shared.StateError).WithErr(err))
-		return r.finishReconcile(ctx, manifest, metrics.ManifestClientInit, manifestStatus, err)
-	}
-
-	if manifest.IsUnmanaged() {
-		return r.handleUnmanagedManifest(ctx, req, skrClient, manifest)
+	skrClient, result, done, err := r.prepareReconcile(ctx, req, manifest, manifestStatus)
```
Tomasz-Smelcerz-SAP: This is a very non-standard error handling. I think it's worth to refactor that into smaller chunks, so that no additional "done" boolean is required.

Other option would be to introduce a custom return type (that implements error) and internally wraps the `res` + `error` returned by `finishReconcile()`.
Then, if that "structuredError" is not nil, you just extract it's payload and return that. An example:

```
Instead of:

skrClient, result, done, err := r.prepareReconcile(ctx, req, manifest, manifestStatus)

You'd have:

skrClient, err := r.prepareReconcile(ctx, req, manifest, manifestStatus)
if err != nil {
        return unwrapStructuredError(err)
}

Example implementation (I haven't run it!)

// implements error
type structuredError struct {
       result ctrl.Result
       error
}

func structuredErrorFrom(result ctrl.Result, err error) *structuredError {
       res := structuredError{
               result: result,
               error:  err,
       }
       return &res
}

func unwrapStructuredError(err error) (ctrl.Result, error) {
       var recErr *structuredError
       if errors.As(err, &recErr) {
               return recErr.result, recErr.error
       }
       return ctrl.Result{}, err // this seems to be the "best option"
}


```

LeelaChacha: Agreed the 4-value return is awkward. I'd avoid the `structuredError` approach though — several of these early returns are successful requeues, not errors, and routing non-failures through the `error` channel makes `errors.Is/As`, log messages, and future wrappers misleading. A `*stepResult` pointer (nil = continue) gives the same call-site cleanup without overloading `error`, and avoids the `errors.As` fallthrough-to-zero-Result footgun. Happy to do that refactor in a follow-up.

internal/controller/manifest/controller.go
```diff
@@ -17,41 +16,64 @@ import (
...
+	errResourceSyncDiffInSameOCILayer = errors.New("resource syncTarget diff detected but in " +
 		"same oci layer, prevent sync resource to be deleted")
 )
 
-const (
-	DefaultInMemoryParseTTL = 24 * time.Hour
+// SyncedOCIRefAnnotation is the annotation key on a Manifest CR that records
+// the OCI ref last successfully synced to the SKR. The reconciler uses it to
+// detect re-renders triggered by an OCI ref change.
+const SyncedOCIRefAnnotation = "sync-oci-ref"
```
c-pius, line 44: I was this years old when I learned... :) 

Rhetorical question, why is this an annotation and not status? 🥹

internal/controller/manifest/controller.go
```diff
@@ -17,41 +16,64 @@ import (
...
+// (res, err) immediately. It exists so step bodies can compose directly with
+// helpers that return (ctrl.Result, error): stopReconcile(r.finishReconcile(…)).
+func stopReconcile(res ctrl.Result, err error) *stepResult {
+	return &stepResult{Result: res, Err: err}
 }
 
+// Consumer-defined collaborator interfaces (ADR 001). The reconciler depends
+// only on these interfaces; concrete implementations are injected by the
+// composition root in cmd/main.go.
+
```
c-pius:
```suggestion
```

Would prefer not to restate such things in code.

internal/controller/manifest/controller.go
```diff
@@ -60,29 +82,44 @@ type ManifestAPIClient interface {
...
 	DetectOrphanedManifest(ctx context.Context, manifest *v1beta2.Manifest) error
 }
 
+// SKRClientCache holds Kubernetes clients for SKR clusters keyed by the
+// manifest's cache key.
 type SKRClientCache interface {
 	GetClient(key string) *skrclient.SKRClient
 	AddClient(key string, client *skrclient.SKRClient)
 	DeleteClient(key string)
 }
```
c-pius, line 93: As the next follow up we can try to use the same SKR client cache for Kyma and Manifest reconciler :)

internal/controller/manifest/controller.go
```diff
@@ -60,29 +82,44 @@ type ManifestAPIClient interface {
...
 
+// SKRClient resolves a Kubernetes client for the SKR a Manifest installs to.
```
c-pius:
```suggestion
// SKRClient resolves the Kubernetes client for the SKR if the SKRClientCache doesn't hold it already.
```

Just to clarify as I had to lookup both.

internal/service/manifest/render/transform.go
```diff
@@ -0,0 +1,25 @@
...
+type ResourceTransform = func(ctx context.Context, manifest *v1beta2.Manifest,
+	resources []*unstructured.Unstructured) error
+
+// CachedManifestParser is the consumer-defined interface for the manifest
+// parser used by Service. Service owns the parser to keep the parse cache
+// scoped to the render lifecycle.
+type CachedManifestParser interface {
+	Parse(spec *spec.Spec) (internal.ManifestResources, error)
+	EvictCache(spec *spec.Spec)
+}
```
c-pius: I think this doesn't belong in a `transform.go` file. More valid options would be:

- have an own `cached_manifest_parser.go` file
- create a `types.go` file
- move both to `service.go` where it is consumed

internal/controller/manifest/controller.go
```diff
@@ -264,44 +259,101 @@ func (r *Reconciler) install(ctx context.Context, req ctrl.Request,
...
-//nolint:funlen // Declarative pkg will be removed soon
-func (r *Reconciler) delete(ctx context.Context, req ctrl.Request,
-	manifest *v1beta2.Manifest,
-) (ctrl.Result, error) {
-	manifestStatus := manifest.GetStatus()
-
+// prepareReconcile runs the steps shared by install: status init, mandatory-module
+// metric, SKR client resolution, unmanaged short-circuit, orphan detection, and
+// finalizer SSA. A non-nil *stepResult signals the caller must return immediately.
+func (r *Reconciler) prepareReconcile(ctx context.Context, req ctrl.Request,
```
c-pius: I don't understand what you mean with "shared by install" mean. Same as with "shared by delete" in `prepareDeleteReconcile`. Can you elaborate?

internal/controller/manifest/controller.go
```diff
@@ -348,6 +400,59 @@ func (r *Reconciler) delete(ctx context.Context, req ctrl.Request,
...
+	if err := status.Initialize(manifest); err != nil {
+		return nil, stopReconcile(r.finishReconcile(ctx, manifest, metrics.ManifestInit, manifestStatus, err))
+	}
+
+	recordMandatoryModuleState(manifest, r)
+
+	skrClient, err := r.getTargetClient(ctx, manifest)
+	if err != nil {
+		if errors.Is(err, accessmanager.ErrAccessSecretNotFound) {
+			return nil, stopReconcile(r.cleanupManifest(ctx, req, manifest, manifestStatus,
```
c-pius, line 418: I think this is a rather significant action and is now a bit hidden in a "prepare" func.

c-pius (commented):
Thanks, this is a great step forward! Consider my comments optional and decide yourself if you want to apply something. I think we will anyway need to continue polishing this.

I am a bit uneasy about this change though. It was super hard to review this because so many things moved. Also, this is one of the most important parts in the code of KLM. We should quickly align with the team how we plan to release and test this in the live landscapes. Also we may need to be cautious when delivering this together with other high-priority features.

internal/controller/manifest/controller.go
LeelaChacha: me too. missed that somehow.

c-pius:
Holding until the regular KLM release was created Friday. We will then merge it and create an additional patch release to isolate this change.

internal/controller/manifest/controller.go
LeelaChacha: Fair point — "shared by install" is misleading wording on my part. `prepareReconcile` only runs in the `install()` path; same for `prepareDeleteReconcile` in `delete()`. What I meant is *the preparatory steps that run at the start of `install()`* — status init, SKR-client resolution, unmanaged short-circuit, orphan detection, and finalizer SSA — before we touch the spec or render resources. I'll update both godocs to drop "shared" and say "prepares the install pipeline" / "prepares the delete pipeline".

internal/controller/manifest/controller.go
LeelaChacha: The access-secret cleanup (imo) belongs here because it's a gate, not a side-effect. `prepareDeleteReconcile` is the *can we proceed with deletion at all?* check, and `ErrAccessSecretNotFound` is one of the answers to that question: the SKR is gone, so there's nothing left to delete on the remote side — drive the finalizer cleanup and stop. It's the same shape as the other branches in this helper (unmanaged → cleanup, status init failure → finish): every path either returns a `*stepResult` or hands a usable `skrClient` back to `delete()`.

c-pius (dismissed):

lindnerby (commented):
Can you, if possible, move these into the service layer as well please:
* internal/manifest/parser
* internal/manifest/spec
* internal/manifest/statecheck

LeelaChacha:
> Can you, if possible, move these into the service layer as well please:
> 
> * internal/manifest/parser
> * internal/manifest/spec
> * internal/manifest/statecheck

Moved `parser` and `spec` to `internal/service/manifest/`. Kept `statecheck` where it is — it's a pure K8s readiness check helper with no orchestration, and its only interface (`StateCheck`) is already consumer-defined in the manifest controller per ADR 001. Happy to move it too if you'd rather

lindnerby (approved):

grego952 (approved):

Issue #3345, opened 2026-06-10
Support Image Pull Secrets for Restricted Default Modules


c-pius:
### Description

Implement a new resource transform injecting data into an image pull secret. The idea is that the Manifest of the restricted default module already defines the secret and its usage, only the data of the secret is a dummy value and not actual secret data. When KLM applies the manifest, the resource transform shall fetch the actual data from a matching secret on KCP and replace the data of the secret in the manifest.

A PoC has already confirmed this to work: https://github.com/kyma-project/lifecycle-manager/pull/3344

Constraints:

- only register the transform if KLM has restricted default module with name `deployer` configured
- only apply the transform if the Manifest belongs to module with name `deployer`
- the module team must define the secret and its usage in the Manifest, only the data is a dummy value
- secret in the manifest must be annotated with `operator.kyma-project.io/inject-data-from-kcp: "true"`
- secret on KCP resides in `kcp-system` namespace and name equals the name of the secret in the manifest
- secret on KCP MUST be labelled with `operator.kyma-project.io/module-name: <module-name>` matching the same label on the Manifest the resource transform is applied for (to prevent module A reading data from module B)
- only `.data` is injected, nothing else

Known Limitations:

- the Module Manifest will not work on its own, it requires processing by Lifecycle-Manager
- if two modules would define the same secret name there would be a conflict (reading from a wrong secret is still prevented by the label). However, the transform is only enabled for `deployer` right now and we don't expect it to be enabled for others for now.

### Reasons

The deployer module images must not be publicly accessible. Using rt-bootstrapper was considered as an alternative. However, the effort for rt-bootstrapper is significantly higher as they also must ensure that the secret is only synced and configured to Kymas that are enabled for the deployer module. 

### Acceptance Criteria

- [x] resource transform introduced with the constraints above
- [x] unit tested
- [x] e2e test covering
  - [x] replacement for restricted default module
  - [x] NO replacement for a regular module

c-pius:
To be clarified: how is the secret data kept up to date in KCP?

paulograbin:
A few questions from me:

- Who is required to register the secret value in KLM? Module developers or kyma team?

- How do we prevent someone unrelated from crafting a manifest that uses the same secret and namespace as ours from getting our value?

- How does the restricted module check work? it is fully KLM-controlled or can a tenant influence it?

c-pius:
> * Who is required to register the secret value in KLM? Module developers or kyma team?

[...]

PR #3361, opened 2026-06-18
chore: Use `deployer` as module name for restricted default modules E2E tests

c-pius:
**Description**

- #3345 will assert "deployer" as module name specifically, this prepares our E2E test suite to work with that

**Related issue(s)**

#3345

### DIFF
--- a/tests/e2e/restricted_module_normal_module_installation_test.go
+++ b/tests/e2e/restricted_module_normal_module_installation_test.go
@@ -24,7 +24,7 @@ var _ = Describe("Restricted Modules - Normal Module Installation", Ordered, fun
 			Eventually(EnableModule).
 				WithContext(ctx).
 				WithArguments(skrClient, defaultRemoteKymaName, RemoteNamespace,
-					NewDeployer(v1beta2.DefaultChannel)).
+					NewTemplateOperator(v1beta2.DefaultChannel)).
 				Should(Succeed())
 
 			By("And the Module CR has been installed on the SKR cluster")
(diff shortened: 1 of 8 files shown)


tests/e2e/restricted_module_normal_module_installation_test.mk
```diff
@@ -3,6 +3,8 @@
 
 include $(dir $(abspath $(lastword $(MAKEFILE_LIST))))e2e.common.mk
 
+MODULE_NAME := deployer
```
LeelaChacha: this is a normal module installation test - we should call the module template-operator here

tests/e2e/restricted_module_normal_module_installation_test.go
```diff
@@ -24,7 +24,7 @@ var _ = Describe("Restricted Modules - Normal Module Installation", Ordered, fun
 			Eventually(EnableModule).
 				WithContext(ctx).
 				WithArguments(skrClient, defaultRemoteKymaName, RemoteNamespace,
-					NewTemplateOperator(v1beta2.DefaultChannel)).
```
LeelaChacha, line 27: this is a normal module installation test - we should call the module template-operator here

LeelaChacha (changes requested):

tests/e2e/restricted_module_normal_module_installation_test.mk
c-pius: Right, didn't check throughly enough what the test does 👍🏻

LeelaChacha (approved):
LGTM

tests/e2e/restricted_module_normal_module_installation_test.mk
c-pius: https://github.com/kyma-project/lifecycle-manager/pull/3361/changes/fb4e6c6394cc6f342417dde00b040f0c6eccbe02

PR #3363, opened 2026-06-18
feat: Inject image pull secret data from KCP for deployer module

### LINKED ISSUE #3345: Support Image Pull Secrets for Restricted Default Modules

LeelaChacha:
## What

Adds a resource transform that replaces the `.data` of any `Secret` in a Manifest annotated with `operator.kyma-project.io/inject-data-from-kcp: "true"` with the `.data` of a same-named Secret in `kcp-system`.

## Constraints

- Transform registered only when `deployer` is in `--restricted-default-modules`.
- Per-Manifest no-op unless its `operator.kyma-project.io/module-name` label equals `deployer`.
- KCP source secret MUST carry `operator.kyma-project.io/module-name=deployer`; missing/mismatching label is rejected (prevents cross-module reads).
- Only `.data` is replaced; other fields untouched.

## Changes

- `api/shared/operator_annotations.go` — new `InjectDataFromKCPAnnotation` constant.
- `internal/declarative/v2/restricted_default_module_image_pull_secret_transform.go` — new transform + consumer-defined `SecretRepository` interface.
- `cmd/composition/service/manifest/render/render.go` — append the transform when `deployer` is configured.
- `cmd/main.go` — thread `secretRepo` and the restricted-modules list into render composition.

Related #3345

### DIFF
--- a/api/shared/operator_annotations.go
+++ b/api/shared/operator_annotations.go
@@ -13,7 +13,7 @@ const (
 	UnmanagedAnnotation       = OperatorGroup + Separator + "is-unmanaged"
 	// InjectDataFromKCPAnnotation, when set to "true" on a Secret in a module Manifest, signals
 	// that the Secret's .data MUST be replaced at apply-time with the .data of a matching Secret
-	// fetched from KCP. Only honored for restricted default modules (see the corresponding resource
+	// fetched from KCP. Only honored for restricted default module "deployer" (see the corresponding resource
 	// transform in internal/declarative/v2).
 	InjectDataFromKCPAnnotation = OperatorGroup + Separator + "inject-data-from-kcp"
 )
--- a/api/shared/operator_annotations.go
+++ b/api/shared/operator_annotations.go
@@ -11,4 +11,9 @@ const (
 	OwnedByFormat             = "%s/%s"
 	IsClusterScopedAnnotation = OperatorGroup + Separator + "is-cluster-scoped"
 	UnmanagedAnnotation       = OperatorGroup + Separator + "is-unmanaged"
+	// InjectDataFromKCPAnnotation, when set to "true" on a Secret in a module Manifest, signals
+	// that the Secret's .data MUST be replaced at apply-time with the .data of a matching Secret
+	// fetched from KCP. Only honored for restricted default modules (see the corresponding resource
+	// transform in internal/declarative/v2).
+	InjectDataFromKCPAnnotation = OperatorGroup + Separator + "inject-data-from-kcp"
 )
--- a/cmd/composition/service/manifest/render/render.go
+++ b/cmd/composition/service/manifest/render/render.go
@@ -2,21 +2,32 @@
 package render
 
 import (
+	"slices"
+
 	declarativev2 "github.com/kyma-project/lifecycle-manager/internal/declarative/v2"
 	"github.com/kyma-project/lifecycle-manager/internal/service/manifest/render"
 )
 
 // ComposeRenderService wires the manifest ResourceRenderService: the default
-// transform chain, plus the optional SkrImagePullSecret transform when
-// skrImagePullSecretName is non-empty.
+// transform chain, plus optional transforms enabled by configuration.
+//
+// The SkrImagePullSecret transform is appended when skrImagePullSecretName is non-empty.
+// The RestrictedDefaultModuleImagePullSecret transform is appended only when the
+// deployer module is configured as a restricted default module — see issue #3345.
 func ComposeRenderService(
 	parser declarativev2.CachedManifestParser,
 	skrImagePullSecretName string,
+	secretRepo declarativev2.SecretRepository,
+	restrictedDefaultModules []string,
 ) *render.Service {
 	transforms := declarativev2.GetDefaultResourceTransforms()
 	if skrImagePullSecretName != "" {
 		transforms = append(transforms,
 			declarativev2.CreateSkrImagePullSecretTransform(skrImagePullSecretName))
+	}
+	if slices.Contains(restrictedDefaultModules, declarativev2.RestrictedDefaultModuleDeployerName) {
+		transforms = append(transforms,
+			declarativev2.CreateRestrictedDefaultModuleImagePullSecretTransform(secretRepo))
 	}
 	return render.NewService(parser, transforms)
 }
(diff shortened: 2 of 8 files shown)


cmd/composition/service/manifest/render/render.go
```diff
@@ -2,21 +2,32 @@
...
 ) *render.Service {
 	transforms := declarativev2.GetDefaultResourceTransforms()
 	if skrImagePullSecretName != "" {
 		transforms = append(transforms,
 			declarativev2.CreateSkrImagePullSecretTransform(skrImagePullSecretName))
 	}
+	if slices.Contains(restrictedDefaultModules, declarativev2.RestrictedDefaultModuleDeployerName) {
+		transforms = append(transforms,
+			declarativev2.CreateRestrictedDefaultModuleImagePullSecretTransform(secretRepo))
+	}
```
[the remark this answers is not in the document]
LeelaChacha: e2e tests will be introduced in other PRs on this branch

api/shared/operator_annotations.go
```diff
@@ -11,4 +11,9 @@ const (
 	OwnedByFormat             = "%s/%s"
 	IsClusterScopedAnnotation = OperatorGroup + Separator + "is-cluster-scoped"
 	UnmanagedAnnotation       = OperatorGroup + Separator + "is-unmanaged"
+	// InjectDataFromKCPAnnotation, when set to "true" on a Secret in a module Manifest, signals
+	// that the Secret's .data MUST be replaced at apply-time with the .data of a matching Secret
+	// fetched from KCP. Only honored for restricted default modules (see the corresponding resource
+	// transform in internal/declarative/v2).
```
c-pius:
```suggestion
	// fetched from KCP. Only honored for restricted default module "deployer" (see the corresponding resource
	// transform in internal/declarative/v2).
```

cmd/main.go
```diff
@@ -307,7 +307,7 @@ func setupManager(flagVar *flags.FlagVar, cacheOptions cache.Options, scheme *ma
 		kymaMetrics, logger, maintenanceWindow, ociRegistry.GetReference(), kymaDeletionSvc, kymaLookupSvc,
 		mtEventHandlerMapFunc, mrmEventHandler, restrictedModuleDefaulter)
 	setupManifestReconciler(mgr, flagVar, options, sharedMetrics, mandatoryModulesMetrics, accessManagerService, logger,
-		eventRecorder, kymaRepo)
+		eventRecorder, kymaRepo, secretRepo)
```
c-pius, line 310: This `secretRepo` uses the uncached KCP client. Can see that we also use that in other services. one of them probably requires it uncached. Let's keep it as is, but we should check in a follow-up whether this is better to be replaced with a second repo with a cached KCP client.

internal/declarative/v2/restricted_default_module_image_pull_secret_transform.go
```diff
@@ -0,0 +1,118 @@
...
+	kcpSecret, err := secretRepo.Get(ctx, resource.GetName())
+	if err != nil {
+		return fmt.Errorf("failed to fetch kcp secret %q: %w", resource.GetName(), err)
+	}
+
+	if err := assertSecretBelongsToModule(kcpSecret, moduleName); err != nil {
+		return err
+	}
+
+	return replaceSecretData(resource, kcpSecret.Data)
```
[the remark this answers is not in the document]
c-pius: I think we can live without this check. We rely on all places that the composition root is configured correctly.

internal/declarative/v2/restricted_default_module_image_pull_secret_transform.go
```diff
@@ -0,0 +1,121 @@
...
+
+var (
+	ErrInjectFromKCPSecretMissingModuleLabel = errors.New(
+		"kcp source secret is missing the required " + shared.ModuleName + " label")
+	ErrInjectFromKCPSecretModuleLabelMismatch = errors.New(
+		"kcp source secret " + shared.ModuleName + " label does not match the manifest module")
+)
+
+// SecretRepository reads Secrets from a fixed namespace (kcp-system).
+// Defined at the consumer side per ADR 001.
```
c-pius:
```suggestion
```

internal/declarative/v2/restricted_default_module_image_pull_secret_transform.go
```diff
@@ -0,0 +1,121 @@
...
+// that replaces the .data of any Secret resource in the manifest annotated with
+// shared.InjectDataFromKCPAnnotation=true with the .data of a same-named Secret
+// fetched from the KCP control-plane namespace.
+//
+// To prevent module A from reading module B's secret data, the KCP secret MUST
+// carry the shared.ModuleName label matching the manifest's module-name label.
+//
+// The transform is a no-op for any module other than the hardcoded
+// RestrictedDefaultModuleDeployerName.
+func CreateRestrictedDefaultModuleImagePullSecretTransform(secretRepo SecretRepository) ResourceTransform {
```
c-pius: Let's name it `CreateDeployerModuleImagePullSecretTransform`, also the files as we run this specifically for deployer module only.

PR #3364, opened 2026-06-18
feat: Inject image pull secret data from KCP for deployer module

### LINKED ISSUE #3345: Support Image Pull Secrets for Restricted Default Modules

c-pius:
## What

Adds a resource transform that replaces the `.data` of any `Secret` in a Manifest annotated with `operator.kyma-project.io/inject-data-from-kcp: "true"` with the `.data` of a same-named Secret in `kcp-system`.

## Constraints

- Transform registered only when `deployer` is in `--restricted-default-modules`.
- Per-Manifest no-op unless its `operator.kyma-project.io/module-name` label equals `deployer`.
- KCP source secret MUST carry `operator.kyma-project.io/module-name=deployer`; missing/mismatching label is rejected (prevents cross-module reads).
- Only `.data` is replaced; other fields untouched.

## Changes

- `api/shared/operator_annotations.go` — new `InjectDataFromKCPAnnotation` constant.
- `internal/declarative/v2/restricted_default_module_image_pull_secret_transform.go` — new transform + consumer-defined `SecretRepository` interface.
- `cmd/composition/service/manifest/render/render.go` — append the transform when `deployer` is configured.
- `cmd/main.go` — thread `secretRepo` and the restricted-modules list into render composition.

## Known Limitations

- there is no E2E test verifying that the transform is only registered when "deployer" module is in the list of restricted default modules. There is a test verifying that the transform skips if the module name is not "deployer".
- an error in the transform stops updating the module resources and reports that error on the manifest => if the KCP secret is removed or re-labelled at some point, the Manifest goes to error state and the resources on the SKR remain in their state from that point of time. KLM does not revert them to the "un-injected" state.

Resolves https://github.com/kyma-project/lifecycle-manager/issues/3345

### DIFF
--- a/api/shared/operator_annotations.go
+++ b/api/shared/operator_annotations.go
@@ -13,7 +13,8 @@ const (
 	UnmanagedAnnotation       = OperatorGroup + Separator + "is-unmanaged"
 	// InjectDataFromKCPAnnotation, when set to "true" on a Secret in a module Manifest, signals
 	// that the Secret's .data MUST be replaced at apply-time with the .data of a matching Secret
-	// fetched from KCP. Only honored for restricted default module "deployer" (see the corresponding resource
-	// transform in internal/declarative/v2).
+	// fetched from KCP. The Secret's .stringData is also cleared to prevent it from overriding the
+	// injected .data. Only honored for restricted default module "deployer" (see the corresponding
+	// resource transform in internal/declarative/v2).
 	InjectDataFromKCPAnnotation = OperatorGroup + Separator + "inject-data-from-kcp"
 )
--- a/api/shared/operator_annotations.go
+++ b/api/shared/operator_annotations.go
@@ -11,4 +11,9 @@ const (
 	OwnedByFormat             = "%s/%s"
 	IsClusterScopedAnnotation = OperatorGroup + Separator + "is-cluster-scoped"
 	UnmanagedAnnotation       = OperatorGroup + Separator + "is-unmanaged"
+	// InjectDataFromKCPAnnotation, when set to "true" on a Secret in a module Manifest, signals
+	// that the Secret's .data MUST be replaced at apply-time with the .data of a matching Secret
+	// fetched from KCP. Only honored for restricted default module "deployer" (see the corresponding resource
+	// transform in internal/declarative/v2).
+	InjectDataFromKCPAnnotation = OperatorGroup + Separator + "inject-data-from-kcp"
 )
--- a/internal/declarative/v2/deployer_module_image_pull_secret_transform.go
+++ b/internal/declarative/v2/deployer_module_image_pull_secret_transform.go
@@ -67,6 +67,10 @@ func CreateDeployerModuleImagePullSecretTransform(secretRepo SecretRepository) R
 		}
 
 		for _, resource := range resources {
+			if resource == nil {
+				continue
+			}
+
 			if !isInjectableSecret(resource) {
 				continue
 			}
--- a/internal/declarative/v2/deployer_module_image_pull_secret_transform.go
+++ b/internal/declarative/v2/deployer_module_image_pull_secret_transform.go
@@ -1,0 +1,135 @@
+package v2
+
+import (
+	"context"
+	"encoding/base64"
+	"errors"
+	"fmt"
+
+	apicorev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/kyma-project/lifecycle-manager/api/shared"
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+)
+
+// DeployerModuleName is the only module for which
+// CreateDeployerModuleImagePullSecretTransform takes effect. The transform
+// MUST NOT silently apply to any other module: image-pull-secret data injection is
+// reserved for the deployer module per the feature design (issue #3345).
+const DeployerModuleName = "deployer"
+
+const secretKind = "Secret"
+
+var (
+	// ErrInjectDataFromKCP is the umbrella sentinel for every error returned by
+	// CreateDeployerModuleImagePullSecretTransform. The Manifest
+	// reconciler keys on it (errors.Is) to mark the manifest as StateError and
+	// short-circuit SSA — see renderResourcesForInstall in reconciler.go.
+	ErrInjectDataFromKCP = errors.New("inject-data-from-kcp transform failed")
+
+	ErrInjectFromKCPSecretMissingModuleLabel = fmt.Errorf(
+		"%w: kcp source secret is missing the required %s label", ErrInjectDataFromKCP, shared.ModuleName)
+	ErrInjectFromKCPSecretModuleLabelMismatch = fmt.Errorf(
+		"%w: kcp source secret %s label does not match the manifest module", ErrInjectDataFromKCP, shared.ModuleName)
+)
+
+// SecretRepository reads Secrets from a fixed namespace (kcp-system).
+type SecretRepository interface {
+	Get(ctx context.Context, name string) (*apicorev1.Secret, error)
+}
+
+// CreateDeployerModuleImagePullSecretTransform returns a ResourceTransform
+// that replaces the .data of any Secret resource in the manifest annotated with
+// shared.InjectDataFromKCPAnnotation=true with the .data of a same-named Secret
+// fetched from the KCP control-plane namespace.
+//
+// To prevent module A from reading module B's secret data, the KCP secret MUST
+// carry the shared.ModuleName label matching the manifest's module-name label.
+//
+// The transform is a no-op for any module other than the hardcoded
+// DeployerModuleName.
+func CreateDeployerModuleImagePullSecretTransform(secretRepo SecretRepository) ResourceTransform {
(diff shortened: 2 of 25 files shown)


tests/e2e/deployer_module_image_pull_secret_injection_test.go
```diff
@@ -0,0 +1,143 @@
...
+					skrClient,
+					imagePullResourceName,
+					TestModuleResourceNamespace,
+					".dockerconfigjson",
+					[]byte(`{"auths":{"e2e-test.example.com":{"username":"test","password":"test"}}}`+"\n"),
+				).
+				WithTimeout(ConsistentDuration).
+				WithPolling(interval).
+				Should(Succeed())
+		})
```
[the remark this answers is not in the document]
c-pius: The module is installed very early in the test when still other checks like Kyma and SKR Webhook Ready are processed. This is not a problem in practice.

tests/e2e/deployer_module_image_pull_secret_injection_test.go
```diff
@@ -0,0 +1,143 @@
...
+					skrClient,
+					"non-inject-pull-secret",
+					TestModuleResourceNamespace,
+					".dockerconfigjson",
+					[]byte(`{"auths": {}}`+"\n"),
+				).
+				WithTimeout(ConsistentDuration).
+				WithPolling(interval).
+				Should(Succeed())
+		})
```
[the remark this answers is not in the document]
c-pius: See above.

tests/e2e/block_non_deployer_module_image_pull_secret_injection_test.go
```diff
@@ -0,0 +1,57 @@
...
+					skrClient,
+					"image-pull-secret",
+					TestModuleResourceNamespace,
+					".dockerconfigjson",
+					[]byte(`{"auths": {}}`+"\n"),
+				).
+				WithTimeout(ConsistentDuration).
+				WithPolling(interval).
+				Should(Succeed())
+		})
```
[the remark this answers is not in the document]
c-pius: See comment above.

api/shared/operator_annotations.go
```diff
@@ -11,4 +11,9 @@ const (
 	OwnedByFormat             = "%s/%s"
 	IsClusterScopedAnnotation = OperatorGroup + Separator + "is-cluster-scoped"
 	UnmanagedAnnotation       = OperatorGroup + Separator + "is-unmanaged"
+	// InjectDataFromKCPAnnotation, when set to "true" on a Secret in a module Manifest, signals
+	// that the Secret's .data MUST be replaced at apply-time with the .data of a matching Secret
+	// fetched from KCP. Only honored for restricted default module "deployer" (see the corresponding resource
+	// transform in internal/declarative/v2).
+	InjectDataFromKCPAnnotation = OperatorGroup + Separator + "inject-data-from-kcp"
```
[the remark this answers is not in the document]
c-pius: https://github.com/kyma-project/lifecycle-manager/pull/3364/commits/bb9256344b989b4009715478f811f5429cb9fecb

scripts/tests/deploy_moduletemplate_e2e.sh
```diff
@@ -49,6 +51,11 @@ yq eval ".images[0].newTag = \"${RELEASE_VERSION}\"" -i config/manager/deploymen
 make build-manifests
 yq eval "(. | select(.kind == \"Deployment\") | .metadata.name) = \"${DEPLOYMENT_NAME}\"" -i template-operator.yaml
 
+for resource in "${ADDITIONAL_RESOURCES[@]}"; do
```
LeelaChacha, line 54: I'm not sure. do we need a `---` seperator here?

c-pius: The separators were part of the yaml files, but I was also not entirely happy with these and the newline hack. Changed to use yq instead now. https://github.com/kyma-project/lifecycle-manager/pull/3364/changes/81a976ba46a58e0c2b2b91da134b559680d4e349

c-pius:
Added one more verification that an update to the KCP secret data propagates to the SKR: https://github.com/kyma-project/lifecycle-manager/pull/3364/changes/58784a1d352e41fbfda57117be840dcd45e0e417

internal/declarative/v2/deployer_module_image_pull_secret_transform.go
```diff
@@ -0,0 +1,135 @@
...
+			if err := injectDataFromKCP(ctx, secretRepo, moduleName, resource); err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+}
+
+func isInjectableSecret(resource *unstructured.Unstructured) bool {
+	if resource.GetKind() != secretKind {
```
Tomasz-Smelcerz-SAP, line 86: I would also return false if the given `resource` is nil.
A bit paranoid check, but the signature cries for it (we're given a pointer, and pointers may be nil).

tests/e2e/restricted_default_module_catalog_sync_test.go
```diff
@@ -13,7 +13,7 @@ import (
 
 var _ = Describe("Restricted Default Module Catalog Sync", Ordered, func() {
 	kyma := NewKymaWithNamespaceName("kyma-sample", ControlPlaneNamespace, v1beta2.DefaultChannel)
-	module := NewTemplateOperator(v1beta2.DefaultChannel)
+	module := NewDeployer(v1beta2.DefaultChannel)
```
Tomasz-Smelcerz-SAP, line 16: Why we switched the module to "deployer"?

c-pius: Because this PR has a hard-coded that only "deployer" module can use this transform, so we needed the e2e tests to create a module named "deployer". And for consistency, we also named the module "deployer" in all other tests for the restricted default module story.

internal/declarative/v2/deployer_module_image_pull_secret_transform.go
c-pius: https://github.com/kyma-project/lifecycle-manager/pull/3364/changes/cde609f6727dd474df2f26bab2deaed740401cad

tests/e2e/block_non_deployer_module_image_pull_secret_injection_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+
+	. "github.com/kyma-project/lifecycle-manager/pkg/testutils"
+	. "github.com/kyma-project/lifecycle-manager/tests/e2e/commontestutils"
+)
+
+const notDeployerModuleName = "not-deployer"
+
+var _ = Describe("2 Block Non Deployer Module Image Pull Secret Injection", Ordered, func() {
```
Tomasz-Smelcerz-SAP, line 15: Where is "1 Block Non Deployer [...]"?
Just kidding. A typo?

c-pius: That cost me way too much time yesterday :) We have `Deployer Module Image Pull Secret Injection` and `Block Non Deployer Module Image Pull Secret Injection` tests. Gingko uses regex to determine what test suite to run, when calling `Deployer Module Image Pull Secret Injection`, it triggered both suites interfering with each other. The simple trick was to add 1 and 2 for deduplication. I also tried with a start of string `^` symbol, but that also didn't work because some stuff is prepended to the test name. Therefore kept the simple solution with 1 and 2 prefix.

Tomasz-Smelcerz-SAP (approved):
LGTM

LeelaChacha (approved):

PR #3371, opened 2026-06-19
refactor: Sync CRDs to SKR early via SSA with CRDsSync condition

### LINKED ISSUE #3186: Refactor the CRD Sync
### Description

Refactor the sync of CRDs to the SKR. The KymaCR should initialize with a condition expressing that CRDs are not synced yet. Early in the Kyma reconciliation, a service should be called that is syncing the CRDs. There is already a [CRD repo](https://github.com/kyma-project/lifecycle-manager/blob/main/internal/repository/skr/crd/repo.go) that can be extended. The implementation in the repo should use SSA so that we get upsert semantics. As we don't need to get the object state from the SKR, the SSA call can be simplified and use unstructured data instead of working with a typed ApplyConfiguration. Due to the SSA, we also don't need to check if generations changed. We just always send the SSA and leave it to the SKR API Server to determine if action is necessary.

If the CRD sync service returns without error, the condition should be updated to indicate that the CRDs have been synced successfully. If it returns an error, the condition should be updated to indicate that CRDs are not installed successfully.

If in the downstream processing (Kyma sync, remote catalog sync, ...) we receive a not match error, we requeue with error to have exponential backoff and re-enter the reconciliation where CRDs are tried to be installed again.

### Reasons

The existing CRD sync is hard to understand and has some smells. On installation, we create the CRDs "on-demand" when receiving "no match" errors, e.g. for [Kyma](https://github.com/kyma-project/lifecycle-manager/blob/a5238031a0aef57684460f0eee673fbc62f481f3/internal/remote/skr_context.go#L111-L116), for [MTs](https://github.com/kyma-project/lifecycle-manager/blob/a5238031a0aef57684460f0eee673fbc62f481f3/internal/remote/moduletemplate_syncworker.go#L76-L81) and for [MRMs](https://github.com/kyma-project/lifecycle-manager/blob/a5238031a0aef57684460f0eee673fbc62f481f3/internal/remote/modulereleasemeta_syncworker.go#L77-L82). At the same time in the controller there is a call to [`SkrSyncService.SyncCrds`](https://github.com/kyma-project/lifecycle-manager/blob/a5238031a0aef57684460f0eee673fbc62f481f3/internal/controller/kyma/controller.go#L318-L331), which is an indirection to a `SyncCrdsUseCase.Execute`, which then works on all CRDs. Within `Execute`, it calls a `fetchCrdsAndUpdateKymaAnnotations` function which doesn't only update annotations but patches the entire CRD... It also uses annotations to store generations and only patches if updates are necessary. Probably this is done to spare unnecessary patch calls, however, also the fetch is a network call. The "on-demand" creation even uses a `createCRDInRuntime` func which has another [check if the CRD exists and if it is the latest version](https://github.com/kyma-project/lifecycle-manager/blob/171ecba82f3dbbb093f45ce8f94e98c97fbac1cd/internal/remote/crd_upgrade.go#L294) which shouldn't be necessary.

Overall this needs to be simplified. There should be no reason to sync CRDs only "on-demand" (Kyma is synced immediately, also MTs and MRMs are synced immediately since there are default modules). Even if not, there is no harm of having the CRDs synced immediately.

### Acceptance Criteria

- [ ] CRD sync follows the layered architecture
- [ ] CRD sync is invoked from a central place early in the reconciliation
- [ ] Kyma initialized with a condition indicating that CRDs have not been installed yet
- [ ] Upon successful CRD sync, the condition is updated accordingly
- [ ] Upon error from CRD sync, the condition is updated accordingly
- [ ] no on-demand CRD creation anymore
- [ ] resource syncs receiving no match errors requeue with error
- [ ] remove the CRD sync annotations on the Kyma CR if not needed anylonger

### Dev Notes

- we discussed if this has performance implications. However it the `SyncCrdsUseCase.Execute` is also called upon every Kyma reconciliation and issues a Get for Kyma, MT, and MRM resources in every reconciliation. Effectively, we are trading 3 Get calls for 3 SSA calls.

LeelaChacha:
## Description

Refactors the CRD synchronization to the SKR cluster as outlined in #3186.

- Introduces a dedicated CRD sync service that is called early in the Kyma reconciliation, following the layered architecture (controller → service → repository).
- The SKR CRD repository performs **Server-Side Apply** with **unstructured** data, so no preceding `Get` against the SKR is required and the API server decides whether action is necessary. The reset to `NoneConverter` and the `managed-by` label are preserved.
- A new `CRDsSync` condition is initialized on the Kyma CR and updated to `True` on success or `False` on error.
- Removes on-demand CRD creation in `skr_context.go` and the catalog sync workers. If a downstream sync receives a `NoMatch` error, it propagates to the controller, which requeues with exponential backoff and re-enters reconciliation where CRDs are tried to be installed again.
- Drops the CRD generation tracking annotations on the Kyma CR (`*-{kcp,skr}-crd-generation`) and the obsolete `SyncCrdsUseCase` plus the `internal/crd` cache.

## Acceptance criteria

- [x] CRD sync follows the layered architecture
- [x] CRD sync is invoked from a central place early in the reconciliation
- [x] Kyma initialized with a condition indicating that CRDs have not been installed yet
- [x] Upon successful CRD sync, the condition is updated accordingly
- [x] Upon error from CRD sync, the condition is updated accordingly
- [x] No on-demand CRD creation anymore
- [x] Resource syncs receiving no-match errors requeue with error
- [x] Removed the CRD sync annotations on the Kyma CR

Closes #3186


tests/integration/controller/kcp/remote_sync_test.go
```diff
@@ -497,91 +442,5 @@ var _ = Describe("CRDs sync to SKR and annotations updated in KCP kyma", Ordered
 
 			return skrModuleReleaseMetaCrd.Spec.Versions[0].Schema
 		}, Timeout, Interval).Should(Equal(kcpModuleReleaseMetaCrd.Spec.Versions[0].Schema))
```
medmes, line 444: The previous suite had two specs — `Should regenerate Kyma CRD in SKR when deleted` and `Should regenerate ModuleTemplate CRD in SKR when deleted` — that covered the `delete CRD → reconcile → CRD re-exists` property. Both are gone without replacement.

SSA running every reconcile loop means the behaviour is correct, but it is now untested at integration level. A CRD deleted from SKR between two reconcile cycles would be re-applied on the next loop, but nothing asserts this. Given that on-demand re-creation was the primary behaviour being replaced, this is the one property that should have a regression guard.

Add an `It` block that deletes a CRD from the SKR client and then waits for it to reappear.

cmd/composition/service/skrsync/service.go
```diff
@@ -0,0 +1,70 @@
...
+	return func(kymaName types.NamespacedName) (client.Client, error) {
+		skrClient := skrClientCache.Get(kymaName)
+		if skrClient == nil {
+			return nil, fmt.Errorf("%w: Kyma %s", errorsinternal.ErrSkrClientNotFound, kymaName.String())
+		}
+		return skrClient, nil
+	}
+}
+
+func qualifiedCrdName(kind shared.Kind) string {
```
medmes, line 68: This function is byte-for-byte identical to `qualifiedCrdName` in `internal/service/skrsync/crd_sync.go:71`. Both compute `kind.Plural() + "." + shared.OperatorGroup`.

The composition package imports the service package, so it can't call the unexported helper directly. But the duplication can be eliminated at the root: `r.crdName` on `skrcrdrepo.Repository` is set to the same value as `kcpCrd.Name` that is passed into `Apply`. The `crdName` field is used only for error message formatting — switch those to `kcpCrd.Name` instead, drop the field and the second parameter on `skrcrdrepo.NewRepository`, and this function in the composition layer becomes unnecessary.

internal/remote/skr_context.go
```diff
@@ -105,16 +102,9 @@ func (s *SkrContext) CreateKymaNamespace(ctx context.Context) error {
 }
 
 func (s *SkrContext) CreateOrFetchKyma(
-	ctx context.Context, kcpClient client.Client, kyma *v1beta2.Kyma,
+	ctx context.Context, _ client.Client, kyma *v1beta2.Kyma,
```
medmes: The `kcpClient` parameter is now unused — intentionally discarded with `_` to avoid touching callers. There is one call site in the controller (`controller.go:392`) that still passes `r.Client`. The parameter adds noise to the signature and trains readers to expect the client is used here.

Remove the parameter from the function signature and update the single call site. If a follow-up is planned, add a `//nolint:unparam // tracked in #<issue>` annotation rather than keeping a silently-ignored `_`.

internal/remote/moduletemplate_syncworker.go
```diff
@@ -14,22 +14,20 @@ import (
...
 	namespace  string
 	patchDiff  func(ctx context.Context, obj *v1beta2.ModuleTemplate) error
 	deleteDiff func(ctx context.Context, obj *v1beta2.ModuleTemplate) error
-	createCRD  func(ctx context.Context) error
 }
 
 // newModuleTemplateConcurrentWorker returns a new moduleTemplateConcurrentWorker instance with default dependencies.
 func newModuleTemplateConcurrentWorker(
-	kcpClient, skrClient client.Client,
+	_, skrClient client.Client,
```
medmes: Same pattern as `CreateOrFetchKyma`: `kcpClient` is retained as `_` but the caller in `moduletemplate_syncer.go:38` still passes a real value. The parameter should be removed from the signature and the caller updated. Keeping it as `_` leaks the old interface and obscures that the dependency was actually removed.

api/v1beta2/condition_messages.go
```diff
@@ -24,62 +25,74 @@ const (
...
+		trueMessage:  ConditionMessageSKRImagePullSecretSynced,
+		falseMessage: ConditionMessageSKRImagePullSecretOutOfSync,
+	},
 }
 
 // GetRequiredConditionTypes returns all required ConditionTypes for a KymaCR.
 func GetRequiredConditionTypes(watcherEnabled, skrImagePullSecretSyncEnabled bool) []KymaConditionType {
 	requiredConditions := []KymaConditionType{ConditionTypeModules}
 	requiredConditions = append(requiredConditions, ConditionTypeModuleCatalog)
+	requiredConditions = append(requiredConditions, ConditionTypeCRDsSync)
```
medmes: `CRDsSync` is the first condition resolved in the reconcile loop — it runs before `Modules` or `ModuleCatalog`. The append order here (`Modules → ModuleCatalog → CRDsSync`) is the inverse of execution order, which means `status.conditions` in a Kyma CR will list `CRDsSync` last even though it is the earliest gate.

Operators and tooling that read conditions top-to-bottom will see an out-of-order picture. Move the `CRDsSync` append before `ModuleCatalog` so the array reflects the actual reconciliation sequence.

internal/repository/skr/crd/repo.go
```diff
@@ -67,3 +72,47 @@ func (r *Repository) Delete(ctx context.Context, kymaName types.NamespacedName)
...
+	}
+
+	desired := &unstructured.Unstructured{}
+	desired.SetGroupVersionKind(apiextensionsv1.SchemeGroupVersion.WithKind(
+		reflect.TypeFor[apiextensionsv1.CustomResourceDefinition]().Name()))
+	desired.SetName(kcpCrd.Name)
+	desired.SetLabels(map[string]string{shared.ManagedBy: shared.ManagedByLabelValue})
+	if specMap, err := machineryruntime.DefaultUnstructuredConverter.ToUnstructured(spec); err == nil {
+		_ = unstructured.SetNestedField(desired.Object, specMap, "spec")
+	}
```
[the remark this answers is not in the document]
medmes: Both errors are returned explicitly. `ToUnstructured` failure: lines 113–116 `return nil, fmt.Errorf("failed to convert CRD spec to unstructured: %w", err)`. `SetNestedField` failure: lines 121–123 `return nil, fmt.Errorf("failed to set CRD spec on unstructured object: %w", err)`. Nothing is silently ignored here — disregard this comment.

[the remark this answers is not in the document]
medmes: Correction to my previous reply: Copilot's observation was valid against the original commit where `buildSkrCrd` returned `*unstructured.Unstructured` (no error) and silently discarded both `ToUnstructured` and `SetNestedField` failures with `_ =`. The current HEAD refactored the function to return `(*unstructured.Unstructured, error)` and properly propagates both — the issue is addressed.

tests/integration/controller/kcp/remote_sync_test.go
```diff
@@ -26,12 +21,7 @@ import (
 	. "github.com/kyma-project/lifecycle-manager/pkg/testutils"
 )
 
-var (
-	ErrNotContainsExpectedCondition  = errors.New("kyma CR not contains expected condition")
-	ErrNotContainsExpectedAnnotation = errors.New("kyma CR not contains expected CRD annotation")
-	ErrContainsUnexpectedAnnotation  = errors.New("kyma CR contains unexpected CRD annotation")
-	ErrAnnotationNotUpdated          = errors.New("kyma CR annotation not updated")
-)
+var ErrNotContainsExpectedCondition = errors.New("kyma CR not contains expected condition")
```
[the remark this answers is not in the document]
medmes: Noise. This PR already produces the corrected string. The diff changes the old sentinel `"kyma CR not contains expected condition"` to `"kyma CR does not contain expected condition"` — the fixed version is live in the branch. Nothing to act on.

internal/remote/modulereleasemeta_syncworker.go
```diff
@@ -14,23 +14,21 @@ import (
 )
 
 var (
-	errModuleReleaseMetaCRDNotReady = errors.New("catalog sync: ModuleReleaseMeta CRD is not ready")
-	errModuleReleaseMetaCleanup     = errors.New("catalog sync: Failed to delete ModuleReleaseMeta")
-	errCatModuleReleaseMetaApply    = errors.New("catalog sync: Could not apply ModuleReleseMetas")
+	errModuleReleaseMetaCleanup  = errors.New("catalog sync: Failed to delete ModuleReleaseMeta")
+	errCatModuleReleaseMetaApply = errors.New("catalog sync: Could not apply ModuleReleseMetas")
```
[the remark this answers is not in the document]
medmes: Noise. This PR already fixes the typo. The diff replaces `"Could not apply ModuleReleseMetas"` with `"Could not apply ModuleReleaseMetas"` — the corrected string is already in the branch. Nothing to act on.

internal/repository/skr/crd/repo.go
```diff
@@ -67,3 +72,55 @@ func (r *Repository) Delete(ctx context.Context, kymaName types.NamespacedName)
...
+	skrClient, err := r.getSkrClient(kymaName)
+	if err != nil {
+		return err
+	}
+
+	desired, err := buildSkrCrd(kcpCrd)
+	if err != nil {
+		return fmt.Errorf("failed to build SKR CRD %q: %w", r.crdName, err)
+	}
+
```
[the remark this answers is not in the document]
medmes: Valid observation structurally, but the divergence cannot occur in the current wiring. `crdSync.applyOne` fetches the KCP CRD by `qualifiedCrdName(kind)` and passes the same `kind` to the `skrcrdrepo.Repository` constructor, so `r.crdName == kcpCrd.Name` is guaranteed by construction. The root fix is covered in the open thread on `cmd/composition/service/skrsync/service.go:68`: drop the `crdName` field from `Repository` and use `kcpCrd.Name` directly in error messages, which makes both the divergence risk and the field itself disappear.

internal/service/skrsync/crd_sync.go
```diff
@@ -0,0 +1,73 @@
...
+		if err := c.applyOne(ctx, kymaName, entry); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	if len(errs) == 0 {
+		return nil
+	}
+
+	return errors.Join(append(errs, ErrCrdSync)...)
```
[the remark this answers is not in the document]
medmes: Correct. Fix: `errors.Join(append([]error{ErrCrdSync}, errs)...)`. `errors.Is` semantics are unaffected by position; only the human-readable `.Error()` string changes — the high-level failure appears first in logs instead of last.

tests/integration/controller/kcp/remote_sync_test.go
LeelaChacha: added two It blocks that delete the kyma and moduletemplate CRDs on the SKR and Eventually wait for them to reappear. covers the same property the old specs did.

cmd/composition/service/skrsync/service.go
LeelaChacha: i'd push back on dropping the field. r.crdName isn't just for error messages, it's the actual object name used in Exists and Delete which have no kcpCrd argument. dropping it breaks those paths.

instead i killed the divergence concern itself: buildSkrCrd now takes the name explicitly and Apply passes r.crdName, so the SSA object name is always whatever the repo was constructed with. added a test that asserts this even when kcpCrd.Name differs.

the two qualifiedCrdName helpers are admittedly duplicated. could centralize on shared.Kind as a follow-up, didn't feel worth bloating this PR for two trivially identical lines.

internal/remote/skr_context.go
LeelaChacha: dropped. parameter is gone from the signature, single call site updated.

internal/remote/moduletemplate_syncworker.go
LeelaChacha: cascaded the removal through both workers, both syncer structs, both worker factory types, both outer catalog factory types, the closures, and the test factory signatures. RemoteCatalog.kcpClient stays since it still serves the actual KCP List calls.

api/v1beta2/condition_messages.go
LeelaChacha: reordered to [CRDsSync, Modules, ModuleCatalog, ...]. fair point on Modules/ModuleCatalog running concurrently in the errgroup, kept Modules ahead of ModuleCatalog since that's the order they're appended in the controller code, low stakes either way.

internal/repository/skr/crd/repo.go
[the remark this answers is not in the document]
LeelaChacha: agreed on it being structural rather than current-wiring. addressed by making r.crdName authoritative inside Apply itself: buildSkrCrd now takes the name explicitly, Apply passes r.crdName, the SSA object name is no longer derived from kcpCrd.Name. added a test that asserts the bound name wins when the two differ.

internal/service/skrsync/crd_sync.go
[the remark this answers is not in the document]
LeelaChacha: done, prepended ErrCrdSync.

medmes (approved):
All feedback addressed. Verification against HEAD fa84ed16 (commit d395837e):

LGTM!

IwonaLanger (approved):

PR #3376, opened 2026-06-22
chore: Use un-cached secret repo specifically

c-pius:
**Description**

- the shared secret repo uses an uncached KCP client which is not ideal
- there is one specific use where the KCP client needs to be uncached, all other usages should be cached

**Related issue(s)**


tests/e2e/deployer_module_image_pull_secret_injection_test.go
```diff
@@ -28,7 +28,7 @@ var _ = Describe("1 Deployer Module Image Pull Secret Injection", Ordered, func(
 		"%s: kcp source secret is missing the required %s label: secret=%s",
 		injectFromKCPTransform, shared.ModuleName, imagePullResourceName)
 	errMsgFetchKCPSecret := fmt.Sprintf(
-		"%s: failed to fetch kcp secret %q: failed to get secret %s-kcp-system: secrets %q not found",
+		"%s: failed to fetch kcp secret %q: failed to get secret %s-kcp-system: Secret %q not found",
```
c-pius, line 31: Since the PR switched the secret repo from an uncached to a cached client, the "not found" error is now produced by the cache layer rather than the API server, changing secrets "x" not found → Secret "x" not found.

LeelaChacha (dismissed):

LeelaChacha (approved):

PR #3382, opened 2026-06-24
test: Port maintenance-windows E2E tests to self-contained Makefile

### LINKED ISSUE #3339: [E2E From Makefile] Port Maintenance-Windows Tests
### Description

This ticket ports the E2E tests for maintenance-window behavior. They share the two-channel MRM shape from #3145 but require two extra steps: passing `--requires-downtime` to the newer-version module template, and generating a maintenance-window policy file that the lifecycle-manager Deployment mounts before it starts.

```
  - maintenance-windows
  - maintenance-windows-initial-installation
  - maintenance-windows-skip
```

### Reasons

We want easier local testing.

### Dev Notes

The three tests share identical setup — only the Ginkgo focus string differs. Use the shared-file pattern from #3140 (`e2e.standard_module_setup.mk`): a single `maintenance_windows_test.mk` parameterized by `GINKGO_FOCUS`, dispatched from three thin Makefile targets.

1. **Add a reusable target to `e2e.common.mk`** for the newer-version variant that requires downtime. The other tests pass `false` here; this is the only differentiator. Place it next to the existing `module-setup-in-newer-version`:

```makefile
.PHONY: module-setup-in-newer-version-requires-downtime
module-setup-in-newer-version-requires-downtime:
	@echo "::group::Module setup with $(MODULE_NEWER_VERSION) (requires-downtime)"
	@pushd $(TEMPLATE_OPERATOR_DIR) > /dev/null
	$(SCRIPTS_DIR)/deploy_moduletemplate_e2e.sh \
		--module-name $(MODULE_NAME) \
		--version $(MODULE_NEWER_VERSION) \
		--deployment-name $(MODULE_DEPLOYMENT_NEWER_VERSION) \
		--deployable-version $(MODULE_DEPLOYABLE_VERSION) \
		--requires-downtime
	@popd > /dev/null
	@echo "::endgroup::"
```

`deploy_moduletemplate_e2e.sh` already supports `--requires-downtime` — no script changes needed.

2. **Extract the maintenance-window policy generation to a shared script** so the GitHub action (`.github/actions/generate-maintenance-window-config-e2e/action.yml`) and the new Makefile call the same code. The current action body inlines `date -u -d` math; move it verbatim into:

```
scripts/tests/generate_maintenance_window_policy.sh <output-file-path>
```

Then update `.github/actions/generate-maintenance-window-config-e2e/action.yml` to call it. Keep the action in place — it's still used by tests not in this ticket.

**NOTE:** Make sure the script works on MacOS too!

3. **Create one shared `maintenance_windows_test.mk`** following the `e2e.standard_module_setup.mk` pattern. It defines `klm-patch` (writes the policy file before deploy) and `module-setup` (two-channel MRM, requires-downtime newer version):

```makefile
.DEFAULT_GOAL := test
.PHONY: test $(MAKECMDGOALS)

# klm-patch generates the maintenance-window policy file that the lifecycle-manager
# Deployment expects to find under config/maintenance_windows/. No kustomize patch needed —
# the file is consumed via the existing kustomize wiring.
.PHONY: klm-patch
klm-patch:
	@echo "::group::Maintenance-window policy"
	@$(SCRIPTS_DIR)/generate_maintenance_window_policy.sh \
		$(LIFECYCLE_MANAGER_DIR)/config/maintenance_windows/policy.json
	@echo "::endgroup::"

.PHONY: module-setup
module-setup: module-setup-latest module-setup-in-older-version module-setup-in-newer-version-requires-downtime
	@echo "::group::Test-specific module metadata setup"
	@pushd $(TEMPLATE_OPERATOR_DIR) > /dev/null
	$(SCRIPTS_DIR)/deploy_modulereleasemeta.sh $(MODULE_NAME) fast:$(MODULE_NEWER_VERSION) regular:$(MODULE_OLDER_VERSION)
	@popd > /dev/null
	@echo "::endgroup::"

.PHONY: test-run
test-run: log-tool-versions
	...

.PHONY: test
test: create-clusters klm-patch deploy-klm module-setup test-run
```

Note that `klm-patch` runs before `deploy-klm` so the policy file exists when KLM is deployed.

4. **Wire the three tests into the main `tests/e2e/Makefile`** the same way #3140 does:

```makefile
maintenance-windows:
	$(MAKE) -f $(MAKEFILE_DIR)/maintenance_windows_test.mk test "GINKGO_FOCUS=Maintenance Windows - Wait for Maintenance Window"

maintenance-windows-initial-installation:
	$(MAKE) -f $(MAKEFILE_DIR)/maintenance_windows_test.mk test "GINKGO_FOCUS=Maintenance Windows - No Wait for Maintenance Window on Initial Installation"

maintenance-windows-skip:
	$(MAKE) -f $(MAKEFILE_DIR)/maintenance_windows_test.mk test "GINKGO_FOCUS=Maintenance Windows - No Wait for Maintenance Widnow on Skip"
```

(The Ginkgo focus strings — including the existing typo `Widnow` — are taken verbatim from `tests/e2e/Makefile`. Don't fix the typo as part of this ticket.)

### Acceptance Criteria

- [x] These tests are ported to the E2E approach.
- [x] Add exceptions in `test-e2e.yaml` to not execute `Deploy lifecycle-manager` and `Deploy template-operator` GitHub actions for these tests.
  - [x] Add exceptions in `management-plane-charts` repo `.github/workflows/test-e2e.yaml` as well.
  - [x] in `management-plane-charts`, we only run a subset of the make targets - `module-setup` and `test-run`, the others like `create-clusters` and `deploy-klm` are already provided by helm
  - [x] **Note**: in `management-plane-charts`, the maintenance-window policy is wired differently — the existing `Create maintenance window policy file` and `Create the ConfigMap for the Maintenance-Window VolumeMount` steps generate the JSON via the `generate-maintenance-window-config-e2e` action and mount it as a ConfigMap consumed by the helm-managed Deployment. Those two steps must stay; do NOT call the new `klm-patch` target there (it writes the file under `config/maintenance_windows/` for kustomize, which is unused in the helm path).

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

medmes:
**Description**

- Add `scripts/tests/generate_maintenance_window_policy.sh` — cross-platform (Linux + macOS) script that generates the maintenance-window policy JSON. Replaces the inlined date math in the GHA action.
- Update `.github/actions/generate-maintenance-window-config-e2e/action.yml` to delegate to the new script, eliminating duplication.
- Add `module-setup-in-newer-version-requires-downtime` target to `e2e.common.mk` — identical to `module-setup-in-newer-version` but passes `--requires-downtime` to `deploy_moduletemplate_e2e.sh`.
- Add `tests/e2e/maintenance_windows_test.mk` — single `GINKGO_FOCUS`-parameterised file shared by all three maintenance-window tests. `klm-patch` generates the policy file before `deploy-klm` so kustomize picks it up at deploy time. `module-setup` wires the two-channel MRM with `fast:newer regular:older`.
- Wire the three `tests/e2e/Makefile` targets (`maintenance-windows`, `maintenance-windows-initial-installation`, `maintenance-windows-skip`) to the new `.mk` file.
- Add all three tests to the `self-contained=true` case in `.github/workflows/test-e2e.yml` so the `Deploy lifecycle-manager` and `Deploy template-operator` GHA steps are skipped for them.

**Related issue(s)**

Resolves #3339


.github/actions/generate-maintenance-window-config-e2e/action.yml
c-pius, line 1: This entire action can be dropped. Also it's usage https://github.com/kyma-project/lifecycle-manager/blob/727b14ef6dd31929372e1408f9b1bb7928c6783d/.github/actions/deploy-lifecycle-manager-e2e/action.yml#L66-L72

tests/e2e/e2e.common.mk
```diff
@@ -268,3 +268,17 @@ module-setup-in-newer-version:
 		--deployable-version $(MODULE_DEPLOYABLE_VERSION)
 	@popd > /dev/null
 	@echo "::endgroup::"
+
+.PHONY: module-setup-in-newer-version-requires-downtime
+module-setup-in-newer-version-requires-downtime:
```
c-pius: This is only used by `tests/e2e/maintenance_windows_test.mk`. We should rather define it in there if it is specific to this test.

tests/e2e/maintenance_windows_test.mk
```diff
@@ -0,0 +1,40 @@
...
+
+.PHONY: klm-patch
+klm-patch:
+	@echo "::group::Maintenance-window policy"
+	@$(SCRIPTS_DIR)/generate_maintenance_window_policy.sh \
+		$(LIFECYCLE_MANAGER_DIR)/config/maintenance_windows/policy.json
+	@echo "::endgroup::"
+
+.PHONY: module-setup
+module-setup: module-setup-latest module-setup-in-older-version module-setup-in-newer-version-requires-downtime
```
c-pius: `module-setup-latest` should not be needed

c-pius (changes requested):

c-pius (approved):

Issue #3188, opened 2026-04-15
Migrate Remaining Occurrences of SSA


c-pius:
### Description

Placeholder to refine when the following tickets are done:

- https://github.com/kyma-project/lifecycle-manager/issues/3184
- https://github.com/kyma-project/lifecycle-manager/issues/3186
- https://github.com/kyma-project/lifecycle-manager/issues/3187

### Acceptance Criteria

- [ ] Acceptance criterion 1
- [ ] Acceptance criterion 2

Issue #2706, opened 2025-09-02
Replace  `SubResourceWriter.Patch` with native apply


c-pius:
### Description

Replace deprecated `client.Apply` in `SubResourceWriter` with native SSA once available. E.g., replace `client.Status().Patch(ctx, obj, client.Apply)` with the native alternative once available.

### Reasons

`v0.22.0` of `controller-runtime` deprecated `client.Apply` in favor of "Native SSA Support" via `client.Client.Apply`. For `SubResourceWriter`, there is no replacement yet, see https://github.com/kubernetes-sigs/controller-runtime/issues/3183

### Acceptance Criteria

- [ ] `client.Apply` not used anymore on `SubResourceWriter`
- [ ] considered all projects
  - [ ] lifecycle-manager
  - [ ] template-operator
  - [ ] runtime-watcher
  - [ ] modulectl

lindnerby:
Closing with https://github.com/kyma-project/lifecycle-manager/issues/2707 as a replacement.

PR #3386, opened 2026-06-25
chore: Use typed ApplyConfigurations with native client.Apply

### LINKED ISSUE #3187: Use typed client.Apply
### Description

Blocked until https://github.com/kyma-project/lifecycle-manager/issues/3184 is done.

In parts of our existing SSA usage we already build the object to apply explicitly (instead of reading an objects, and re-applying). Those occurrences should be easy to migrate to the new native SSA with typed ApplyConfigurations. 

- https://github.com/kyma-project/lifecycle-manager/blob/171ecba82f3dbbb093f45ce8f94e98c97fbac1cd/internal/manifest/finalizer/handler.go#L54-L78
- https://github.com/kyma-project/lifecycle-manager/blob/171ecba82f3dbbb093f45ce8f94e98c97fbac1cd/internal/remote/skr_context.go#L137-L170
- https://github.com/kyma-project/lifecycle-manager/blob/171ecba82f3dbbb093f45ce8f94e98c97fbac1cd/internal/repository/skr/kyma/status/repo.go#L54-L90
- https://github.com/kyma-project/lifecycle-manager/blob/171ecba82f3dbbb093f45ce8f94e98c97fbac1cd/internal/repository/watcher/certificate/gcm/certificate/certificate_repo.go#L62-L107
- https://github.com/kyma-project/lifecycle-manager/blob/171ecba82f3dbbb093f45ce8f94e98c97fbac1cd/internal/repository/watcher/certificate/certmanager/certificate/certificate_repo.go#L55-L112
- https://github.com/kyma-project/lifecycle-manager/blob/171ecba82f3dbbb093f45ce8f94e98c97fbac1cd/pkg/testutils/certificate.go#L15-L51
- https://github.com/kyma-project/lifecycle-manager/blob/171ecba82f3dbbb093f45ce8f94e98c97fbac1cd/pkg/testutils/kyma.go#L306-L325

The creation of the Kyma System is also a SSA even though `client.Apply` is not used. We construct the raw patch and use `types.ApplyPatchType` (== `client.Apply.Type()`). Also replace this one with a native SSA and typed ApplyConfiguration.

- https://github.com/kyma-project/lifecycle-manager/blob/171ecba82f3dbbb093f45ce8f94e98c97fbac1cd/internal/remote/skr_context.go#L78-L105

### Reasons

Remove use of deprecated `client.Apply` patch.

### Acceptance Criteria

- [ ] all occurrences above use the native Apply function and a typed ApplyConfiguration

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

medmes:
**Description**

- Replace all 8 occurrences of the deprecated `client.Apply` patch type with the native `client.Apply` method and typed `ApplyConfiguration` objects, as introduced in controller-runtime v0.24.1
- `internal/manifest/finalizer/handler.go` — `EnsureCRFinalizer` now uses `ManifestApplyConfiguration` instead of `PartialObjectMetadata` + `Patch`
- `internal/remote/skr_context.go` — `CreateKymaNamespace` replaced hand-rolled JSON + `RawPatch` with `corev1apply.Namespace`; `SynchronizeKymaMetadata` replaced `unstructured.Unstructured` + `Patch` with `KymaApplyConfiguration`
- `internal/repository/skr/kyma/status/repo.go` — `SetStateDeleting` uses `KymaApplyConfiguration` status subresource apply
- `internal/repository/watcher/certificate/certmanager/certificate/certificate_repo.go` — `Create` uses the generated `certmanagerapplyv1.Certificate` builder chain
- `internal/repository/watcher/certificate/gcm/certificate/certificate_repo.go` — `Create` uses `client.ApplyConfigurationFromUnstructured` (GCM has no generated apply configs)
- `pkg/testutils/certificate.go` and `pkg/testutils/kyma.go` — test utilities migrated to typed apply configs
- Remove all `//nolint:staticcheck // issues: #2706, #2707` suppressions from the migrated call sites
- Register 4 new `importas` aliases in `.golangci.yaml` (`certmanagerapplyv1`, `certmanagerapplymetav1`, `apiconfigsv1beta2`, `corev1apply`)

**Related issue(s)**

 #3187
Full Epic, see also #3188


internal/repository/watcher/certificate/gcm/certificate/certificate_repo.go
```diff
@@ -92,14 +94,12 @@ func (r *Repository) Create(ctx context.Context, name, commonName string, dnsNam
...
 
 	// Patch instead of Create + IgnoreAlreadyExists for cases where we change the config of certificates, e.g. duration
-	err := r.kcpClient.Patch(ctx,
-		cert,
-		//nolint: staticcheck // issues: #2706, #2707
-		client.Apply,
-		client.ForceOwnership,
-		fieldowners.LifecycleManager,
-	)
+	unstructuredCert, err := machineryruntime.DefaultUnstructuredConverter.ToUnstructured(cert)
```
LeelaChacha, line 97: why use unstructured here?

medmes: Gardener cert-management ([v0.24.1](https://github.com/gardener/cert-management/tree/v0.24.1/pkg)) does not generate typed `ApplyConfiguration` builders — there is no `applyconfigurations` package in the module at all. `client.Apply()` requires a `runtime.ApplyConfiguration` interface, which `*gcertv1alpha1.Certificate` does not implement, so `client.ApplyConfigurationFromUnstructured` is the only available bridge for this type.

Re the `// Do not use Unstructured objects here that were generated from API objects` doc warning in controller-runtime: that warning targets the GET-then-apply antipattern — fetching an existing resource and re-applying it, which would silently include server-set fields (resource version, status, managed fields) in the SSA managed-field set. Here we construct the `Certificate` from scratch, so only the fields we explicitly set are present in the unstructured map.

This is safe because every field in `CertificateSpec` and `CertificatePrivateKey` is a pointer or slice/map with `json:"...,omitempty"` — no bare value types. `DefaultUnstructuredConverter.ToUnstructured` will therefore omit all nil/unset fields, and the SSA managed-field set will contain exactly the 8 fields we assign (commonName, dnsNames, secretName, secretLabels, issuerRef, privateKey, duration, renewBefore).

The previous `client.Patch(ctx, cert, client.Apply, ...)` call had the same JSON-with-omitempty serialization semantics, so this is not a behavioural regression — it is the same SSA field ownership, just using the non-deprecated API.

LeelaChacha (approved):

medmes:
As follow up for the GCM: https://github.com/kyma-project/lifecycle-manager/issues/3394

PR #3389, opened 2026-06-25
feat(E2E from Makefile): Port Module API Upgrade Under Blocked Deletion Test

### LINKED ISSUE #3340: [E2E From Makefile] Port Module API Upgrade Under Blocked Deletion Test
### Description

This ticket ports the E2E test that shares the two-channel MRM shape from #3145 but requires extra `yq` surgery on the newer-version template only — adding `v1beta1` to the Sample CRD versions, switching storage to `v1beta1`, and bumping `default-sample-cr.yaml` to match.

```
  - module-api-upgrade-under-blocked-deletion
```

### Reasons

We want easier local testing.

### Dev Notes

1. Same approach as #3140 — copy `e2e_test_template.mk`, rename it, fill in the placeholders. No KLM patches needed.

2. **Version pair.** This test calls `UpdateChannelVersionInModuleReleaseMeta(...,NewerVersion)` where `NewerVersion = "2.4.2-e2e-test"` is hard-coded in `tests/e2e/utils_test.go`. That value differs from `MODULE_NEWER_VERSION` (`2.4.1-smoke-test`) in `e2e.common.mk`, which is the version pair the **mandatory-module test path** depends on (and which must NOT change). The two pairs coexist on purpose — see also `OlderVersion=1.1.1-e2e-test` / `NewerVersion=2.4.2-e2e-test` in `.github/workflows/test-e2e.yml`'s env block, which the existing CI action publishes alongside the smoke-test pair.

   Add a second pair to `e2e.common.mk` for tests that align with the Go `NewerVersion`/`OlderVersion` constants:

   ```makefile
   # Versions used by tests that exercise channel switching / MRM-driven upgrades.
   # Aligned with the `NewerVersion`/`OlderVersion` Go constants in tests/e2e/utils_test.go
   # and the env vars in .github/workflows/test-e2e.yml.
   MODULE_OLDER_VERSION_E2E         := 1.1.1-e2e-test
   MODULE_NEWER_VERSION_E2E         := 2.4.2-e2e-test
   ```

   Use these throughout the new `.mk` file — NOT the smoke-test pair.

3. The latest template uses the standard `module-setup-latest` from `e2e.common.mk`. The older and newer setups must use the e2e-test pair, so the common `module-setup-in-older-version` / `module-setup-in-newer-version` targets cannot be reused — they reference the smoke-test pair. Build the older inline (no yq patch beyond the standard rename), and the newer inline with the CRD-upgrade yq patches:

```makefile
.PHONY: module-setup
module-setup: module-setup-latest
	@echo "::group::Module setup with $(MODULE_OLDER_VERSION_E2E)"
	@pushd $(TEMPLATE_OPERATOR_DIR) > /dev/null
	$(SCRIPTS_DIR)/deploy_moduletemplate_e2e.sh \
		--module-name $(MODULE_NAME) \
		--version $(MODULE_OLDER_VERSION_E2E) \
		--deployment-name $(MODULE_DEPLOYMENT_OLDER_VERSION) \
		--deployable-version $(MODULE_DEPLOYABLE_VERSION)
	@popd > /dev/null
	@echo "::endgroup::"

	@echo "::group::Module setup with $(MODULE_NEWER_VERSION_E2E) and CRD v1beta1 upgrade"
	@pushd $(TEMPLATE_OPERATOR_DIR) > /dev/null
	cp $(SCRIPTS_DIR)/deploy_moduletemplate.sh .
	yq eval '.images[0].newTag = "$(MODULE_NEWER_VERSION_E2E)"' -i config/manager/deployment/kustomization.yaml
	make build-manifests
	yq eval '(. | select(.kind == "Deployment") | .metadata.name) = "$(MODULE_DEPLOYMENT_NEWER_VERSION)"' -i template-operator.yaml

	# Introduce v1beta1 in the CRD to simulate the API upgrade scenario
	crd_selector='select(.kind == "CustomResourceDefinition" and .metadata.name == "samples.operator.kyma-project.io")'
	yq eval "$${crd_selector} |= (.spec.versions += [{\"name\": \"v1beta1\", \"served\": true, \"storage\": true, \"schema\": .spec.versions[0].schema}])" -i template-operator.yaml
	yq eval "($${crd_selector} | .spec.versions[] | select(.name == \"v1alpha1\")).storage = false" -i template-operator.yaml
	yq eval '.apiVersion = "operator.kyma-project.io/v1beta1"' -i config/samples/default-sample-cr.yaml

	./deploy_moduletemplate.sh $(MODULE_NAME) $(MODULE_NEWER_VERSION_E2E) $(MODULE_DEPLOYABLE_VERSION) true false true false
	@popd > /dev/null
	@echo "::endgroup::"

	@echo "::group::Test-specific module metadata setup"
	@pushd $(TEMPLATE_OPERATOR_DIR) > /dev/null
	$(SCRIPTS_DIR)/deploy_modulereleasemeta.sh $(MODULE_NAME) fast:$(MODULE_NEWER_VERSION_E2E) regular:$(MODULE_OLDER_VERSION_E2E)
	@popd > /dev/null
	@echo "::endgroup::"
```

The CRD-upgrade yq patches (`v1beta1` add, `v1alpha1` storage flip, `default-sample-cr.yaml` apiVersion bump) are byte-equivalent to the ones in `deploy-template-operator-with-modulereleasemeta`.

### Acceptance Criteria

- [x] This test is ported to the E2E approach.
- [x] Add an exception in `test-e2e.yaml` to not execute `Deploy lifecycle-manager` and `Deploy template-operator` GitHub actions for this test.
  - [x] Add an exception in `management-plane-charts` repo `.github/workflows/test-e2e.yaml` as well.
  - in `management-plane-charts`, we only run a subset of the make targets - `module-setup` and `test-run`, the others like `create-clusters` and `deploy-klm` are already provided by helm

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

amritanshusikdar:
**Description**

This PR ports the module API upgrade under blocked deletion test to a dedicated `module_api_upgrade_under_blocked_deletion_test.mk` file, enabling local execution.

**Related Issues**
Resolves #3340.


tests/e2e/e2e.common.mk
```diff
@@ -216,6 +216,11 @@ MODULE_DEPLOYMENT_NEWER_VERSION   := template-operator-v2-controller-manager
 MODULE_MANDATORY_OLDER_VERSION := 1.1.0-smoke-test
 MODULE_MANDATORY_NEWER_VERSION := 2.4.1-smoke-test
 
+# Explicit aliases for tests that align with the Go NewerVersion/OlderVersion constants
+# in tests/e2e/utils_test.go and the env vars in .github/workflows/test-e2e.yml.
+MODULE_OLDER_VERSION_E2E := 1.1.1-e2e-test
+MODULE_NEWER_VERSION_E2E := 2.4.2-e2e-test
+
```
c-pius: The are already defined, see above: `MODULE_OLDER_VERSION`, `MODULE_NEWER_VERSION`

tests/e2e/module_api_upgrade_under_blocked_deletion_test.mk
```diff
@@ -0,0 +1,66 @@
...
+	crd_selector='select(.kind == "CustomResourceDefinition" and .metadata.name == "samples.operator.kyma-project.io")'
+	yq eval "$${crd_selector} |= (.spec.versions += [{\"name\": \"v1beta1\", \"served\": true, \"storage\": true, \"schema\": .spec.versions[0].schema}])" -i template-operator.yaml
+	yq eval "($${crd_selector} | .spec.versions[] | select(.name == \"v1alpha1\")).storage = false" -i template-operator.yaml
+	yq eval '.apiVersion = "operator.kyma-project.io/v1beta1"' -i config/samples/default-sample-cr.yaml
+	./deploy_moduletemplate.sh $(MODULE_NAME) $(MODULE_NEWER_VERSION_E2E) $(MODULE_DEPLOYABLE_VERSION) true false true false
+	@popd > /dev/null
+	@echo "::endgroup::"
+
+.PHONY: module-setup
+module-setup: module-setup-latest module-setup-in-older-version-e2e module-setup-in-newer-version-e2e-with-crd-upgrade
```
c-pius: `module-setup-latest` not needed I think.

c-pius: Also, there are `module-setup-in-older-version` defined in common already. Can you check if it can be re-used?

`module-setup-in-newer-version` from common can probably not be re-used because of the crd upgrade.

tests/e2e/module_api_upgrade_under_blocked_deletion_test.mk
```diff
@@ -0,0 +1,66 @@
...
+	@popd > /dev/null
+	@echo "::endgroup::"
+
+.PHONY: module-setup-in-newer-version-e2e-with-crd-upgrade
+module-setup-in-newer-version-e2e-with-crd-upgrade:
+	@echo "::group::Module setup with $(MODULE_NEWER_VERSION_E2E) and CRD v1beta1 upgrade"
+	@export PATH=$(LOCALBIN):$$PATH
+	@pushd $(TEMPLATE_OPERATOR_DIR) > /dev/null
+	yq eval '.apiVersion = "operator.kyma-project.io/v1alpha1"' -i config/samples/default-sample-cr.yaml
+	cp $(SCRIPTS_DIR)/deploy_moduletemplate.sh .
```
c-pius: This is don't entirely understand why necessary. Couldn't see it in the previous action-based setup. Can you elaborate?

c-pius (changes requested):

tests/e2e/e2e.common.mk
amritanshusikdar: I saw those, but the very first AC specifically had this requirement. If you say, then I could remove it. I agree it's a duplication, but I don't know the thought process behind this AC.

LeelaChacha: sheesh, outdated AC. gonna remove it

tests/e2e/module_api_upgrade_under_blocked_deletion_test.mk
amritanshusikdar: To the first one: Yes `module-setup-latest` isn't needed, thanks for pointing it out.

tests/e2e/e2e.common.mk
amritanshusikdar: Oki, thanks for clarification.

tests/e2e/module_api_upgrade_under_blocked_deletion_test.mk
amritanshusikdar: Second one: After also the AC clarification, I've removed this new alias, and it's now using the preexisting `module-setup-in-older-version` target from `e2e.common.mk`. For the other target with the CRD upgrade, it has to stay because the `module-setup-in-newer-version` target only is the copy of the older version, just with the newer variable, without the CRD upgrade.

tests/e2e/module_api_upgrade_under_blocked_deletion_test.mk
amritanshusikdar: Nvm, works without this as well. I was experimenting with something. Has been removed now.

c-pius (approved):

PR #3401, opened 2026-07-01
deps: Update modulectl to 3.2.1

### LINKED ISSUE #3289: Use modulectl 3.2.1
### Description

Update to modulectl 3.2.1 and fix usage. E.g., there is at least the `deploy_moduletemplate.sh` script using the removed `--disable-ocm-registry-push` flag.

https://github.com/kyma-project/lifecycle-manager/blob/5ef254e9d91f6627f3d418ae6f8e87e3e4db2357/scripts/tests/deploy_moduletemplate.sh#L95

### Reasons

modulectl 3.2.1 dropped OCM library and registry push code. We need to adapt accordingly.

### Acceptance Criteria

- [x] modulectl 3.2.1 used
- [x] all e2e tests working

medmes:
**Description**

- Bump modulectl from `3.1.0` to `3.2.1` in `versions.yaml`
- Remove the `--disable-ocm-registry-push` flag from `scripts/tests/deploy_moduletemplate.sh` — this flag was removed in modulectl 3.2.0 when the OCM library and registry push code were dropped

**Related issue(s)**
Resolves #3289


tests/e2e/manifest_reconciliation_test.go
```diff
@@ -122,6 +122,11 @@ var _ = Describe("Manifest Skip Reconciliation Label", Ordered, func() {
 					shared.StateDeleting).
 				Should(Succeed())
 			By("And Manifest CR has message about waiting for finalizers")
+			Eventually(ManifestStatusOperationContainsMessage).
```
LeelaChacha, line 125: why this change in this PR?

medmes: The modulectl 3.2.1 output format change (dropping the OCM library) causes the component constructor to be structured slightly differently. When `ocm add componentversions `processes it, the resulting CTF takes marginally longer to transfer and register in the local registry. This delays the point at which KLM can resolve and begin deleting the SKR module CRs on the Manifest controller's first reconciliation pass.

The Manifest controller has two sequential Deleting sub-states:

- `"waiting for module crs deletion" `: SKR Sample CR still being removed
- `"waiting as other finalizers are present"` : blocking finalizer holding deletion

The original test went` Consistently(state=Deleting) → Consistently(message=2)`, which worked before because the timing always landed in sub-state 2 by that point. The modulectl change consistently exposes the race by landing in sub-state 1 instead.
The `Eventually` is the minimal fix to make the assertion correct regardless of which sub-state the controller is in when the check runs. it belongs here because the `modulectl` change is what makes 
the timing-dependent assumption observable.

LeelaChacha: The fix itself is fine — `Eventually` → `Consistently` is the right shape for bridging two sequential sub-states, and the two Deleting sub-states you describe (`"waiting for module crs deletion"` at `controller.go:561` and `"waiting as other finalizers are present"` at `controller.go:503`) are real.

But the causal story doesn't hold up. The modulectl step runs during **setup** — `modulectl create` → `ocm add componentversions` → push CTF → deploy ModuleTemplate — and completes before the test ever reaches the deletion assertions. The Manifest controller doesn't re-fetch or re-transfer CTFs during deletion; it reconciles against the already-installed SKR state. A "marginally longer CTF transfer" in setup would just shift the whole test later on the wall clock, not change which Deleting sub-state the assertion catches. Sub-state 1 → sub-state 2 timing is governed by how fast the SKR deletes the `Sample` CR and the Manifest controller's requeue cadence — neither depends on how long `ocm add componentversions` took earlier.

The real explanation is one of:

1. **modulectl 3.2.1 changes the rendered ModuleTemplate content** (different descriptor shape, resource set, or Sample CR), and *that* affects how long SKR deletion takes. A content effect, not a setup-timing effect.
2. **The race was always latent** — the pre-existing `Consistently(message==finalizer-msg)` was never robust against sub-state 1, and CI just got unlucky in a run that happened to include this bump.

Either way, the fix stands on its own merits.

LeelaChacha (approved):

PR #3407, opened 2026-07-03
chore: Align E2E Test Operator Versions

### LINKED ISSUE #3377: [E2E from Makefile] Align e2e test-operator versions and centralize their definition
## Description

Follow-up from PR review on #3370 ([thread](https://github.com/kyma-project/lifecycle-manager/pull/3370#discussion_r0)).

Once all e2e tests have been ported to the self-contained `.mk` approach, align on the test-operator version names and where they are defined.

### Current state

`tests/e2e/e2e.common.mk` declares two parallel version tracks as plain Make variables:

```makefile
MODULE_OLDER_VERSION              := 1.1.1-e2e-test
MODULE_NEWER_VERSION              := 2.4.2-e2e-test
MODULE_MANDATORY_OLDER_VERSION    := 1.1.0-smoke-test
MODULE_MANDATORY_NEWER_VERSION    := 2.4.1-smoke-test
```

The literals are also hardcoded in several Go test files, including:

- `tests/e2e/utils_test.go` — `NewerVersion = "2.4.2-e2e-test"`
- `tests/e2e/module_upgrade_channel_switch_test.go` — expects `"2.4.2-e2e-test"` in an error message
- `tests/e2e/modulereleasemeta_sync_test.go` — `v1Version`, `v2Version`
- `tests/e2e/mandatory_module_test.go` — `"1.1.0-smoke-test"`, `"2.4.1-smoke-test"`

### Problems

1. The version numbers (`1.1.1`, `2.4.2`, `1.1.0`, `2.4.1`) carry no inherent meaning — they're synthetic OCM component-descriptor versions wrapping the same `template-operator:$(MODULE_DEPLOYABLE_VERSION)` image. The numeric ordering is the only thing that matters; the specific values are confusing.
2. The `-e2e-test` vs `-smoke-test` suffix split is historical and only exists because the mandatory tests assert on different literals than the general-track tests.
3. The versions are duplicated between `e2e.common.mk` and the Go test files, with no enforcement that they stay aligned.

### Proposal

Define the versions in a central YAML (e.g. `tests/e2e/versions.yaml` or extend the existing `versions.yaml`) and have both the Make-level setup and the Go-level assertions parse from this single source of truth. Simplify the version naming convention while we're at it.

### AC
- [x] Align all versions used in the E2E tests

### Parent

Sub-issue of #2275.

amritanshusikdar:
**Description**
- Introduced a new file `tests/e2e/versions.yaml` which now standardizes which version to use for tests in terms of OLD and NEW.
- Adjusted rest of the codebase.
- Adapted the GitHub Workflow `test-e2e.yml`.  

**Related issue(s)**
Fixes #3377


.github/workflows/test-e2e.yml
```diff
@@ -193,22 +193,31 @@ jobs:
...
-          NewerVersion: 2.4.2-e2e-test
-          OlderVersion: 1.1.1-e2e-test
-          VersionForStatefulSetInWarning: 1.0.0-warning-statefulset
-          VersionForDeploymentInWarning: 1.0.0-warning-deployment
-          VersionForMisconfiguredDeploymentImage: 1.0.0-misconfigured-deployment
-          VersionForMisconfiguredStatefulSetImage: 1.0.0-misconfigured-statefulset
-          VersionForNoDefaultCR: 1.0.0-no-default-cr
+          ModuleDeploymentNameInNewerVersion: template-operator-${{ steps.e2e-versions.outputs.newer }}-controller-manager
+          ModuleDeploymentNameInOlderVersion: template-operator-${{ steps.e2e-versions.outputs.older }}-controller-manager
+          NewerVersion: ${{ steps.e2e-versions.outputs.newer }}-e2e
```
medmes, line 218: These env vars control the K8s Deployment name that `deploy_moduletemplate_e2e.sh` patches into the module template manifest:

```bash
yq eval "(. | select(.kind == \"Deployment\") | .metadata.name) = \"${DEPLOYMENT_NAME}\"" -i template-operator.yaml
```

Every Go test that asserts on deployment readiness uses the constants from `pkg/testutils/modulecr.go:25-26`:

```go
ModuleDeploymentNameInNewerVersion = "template-operator-v2-controller-manager"
ModuleDeploymentNameInOlderVersion = "template-operator-v1-controller-manager"
```

After this change CI deploys `template-operator-1.0.0-controller-manager`, but Go tests look for `template-operator-v1-controller-manager`. Every test still on the non-self-contained CI path that calls `DeploymentIsReady(..., ModuleDeploymentNameInOlderVersion, ...)` will time out. At least 15 test files are affected (`mandatory_module_test.go`, `module_upgrade_channel_switch_test.go`, `modulereleasemeta_*`, `maintenance_windows_*`, `restricted_modules_test.go`, etc.).

The Makefile local path is unaffected — `MODULE_DEPLOYMENT_OLDER_VERSION` and `MODULE_DEPLOYMENT_NEWER_VERSION` in `e2e.common.mk` remain `v1`/`v2` stable names. CI and local are now inconsistent on deployment identity.

`MODULE_DEPLOYMENT_*` are structural K8s identifiers, not version labels. Revert these two lines:

```yaml
ModuleDeploymentNameInNewerVersion: template-operator-v2-controller-manager
ModuleDeploymentNameInOlderVersion: template-operator-v1-controller-manager
```

tests/e2e/utils_test.go
```diff
@@ -73,6 +82,22 @@ func mustReadTemplateOperatorVersion() string {
 	return version
 }
 
+func readE2EVersionFromFile(key string) string {
+	content, err := os.ReadFile("versions.yaml")
+	if err != nil {
```
medmes: Resolves relative to process working directory at test runtime. The repo root also has a `versions.yaml` with a completely different schema (image-name → semver tag). If `go test` is invoked from any directory other than `tests/e2e/`, `os.ReadFile` succeeds on the wrong file and panics with a misleading `"module-version-older" not found` message.

The existing `mustReadTemplateOperatorVersion` in this same file uses an explicit relative path (`../../versions.yaml`) precisely to avoid this. Apply the same principle here, or use `//go:embed` to anchor resolution to the source file at compile time and eliminate the runtime path dependency entirely:

```go
//go:embed versions.yaml
var e2eVersionsYAML []byte

func readE2EVersionFromFile(key string) string {
    var versions map[string]string
    if err := yaml.Unmarshal(e2eVersionsYAML, &versions); err != nil {
        panic(fmt.Sprintf("failed to parse versions.yaml: %v", err))
    }
    v, ok := versions[key]
    if !ok {
        panic(fmt.Sprintf("%q not found in versions.yaml", key))
    }
    return v
}
```

medmes:
.github/workflows/test-e2e-with-gcm.yml was not updated. It still has all version strings hardcoded at the old values:

```yaml
NewerVersion: 2.4.2-e2e-test
OlderVersion: 1.1.1-e2e-test
OlderVersionForMandatoryModule: 1.1.0-smoke-test
NewerVersionForMandatoryModule: 2.4.1-smoke-test
VersionForStatefulSetInWarning: 1.0.0-warning-statefulset
VersionForDeploymentInWarning: 1.0.0-warning-deployment
VersionForMisconfiguredDeploymentImage: 1.0.0-misconfigured-deployment
VersionForMisconfiguredStatefulSetImage: 1.0.0-misconfigured-statefulset
VersionForNoDefaultCR: 1.0.0-no-default-cr
```

The centralization goal from #3377 is not complete until this workflow reads from `tests/e2e/versions.yaml` the same way `test-e2e.yml` now does. Any future bump to `versions.yaml` will propagate to `test-e2e.yml` but silently not to this one, re-introducing the divergence the PR is meant to eliminate.

This workflow also exposes `OlderVersionForMandatoryModule` / `NewerVersionForMandatoryModule` as separate env vars — `test-e2e.yml` does not have these at all (neither before nor after this PR). Both workflows need to be aligned on which keys they read and which env vars they expose to the `deploy-template-operator-with-modulereleasemeta` action.

tests/e2e/utils_test.go
```diff
@@ -73,6 +82,22 @@ func mustReadTemplateOperatorVersion() string {
...
+	var versions map[string]string
+	if err := yaml.Unmarshal(content, &versions); err != nil {
+		panic(fmt.Sprintf("failed to parse versions.yaml: %v", err))
+	}
+	v, ok := versions[key]
+	if !ok {
+		panic(fmt.Sprintf("%q not found in versions.yaml", key))
+	}
+	return v
+}
```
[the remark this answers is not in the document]
amritanshusikdar: Not relevant anymore. Removed the `tests/e2e/versions.yaml`. Now we have a single root `versions.yaml` file.

tests/e2e/utils_test.go
amritanshusikdar: Same as https://github.com/kyma-project/lifecycle-manager/pull/3407#discussion_r3552397784

.github/workflows/test-e2e.yml
amritanshusikdar: Kept the same versioning by extracting the major version from the `versions.yaml` file. Cleaner setup with minimal changes.

.github/workflows/test-e2e.yml
```diff
@@ -193,22 +193,31 @@ jobs:
...
-          ModuleDeploymentNameInOlderVersion: template-operator-v1-controller-manager
-          NewerVersion: 2.4.2-e2e-test
-          OlderVersion: 1.1.1-e2e-test
-          VersionForStatefulSetInWarning: 1.0.0-warning-statefulset
-          VersionForDeploymentInWarning: 1.0.0-warning-deployment
-          VersionForMisconfiguredDeploymentImage: 1.0.0-misconfigured-deployment
-          VersionForMisconfiguredStatefulSetImage: 1.0.0-misconfigured-statefulset
-          VersionForNoDefaultCR: 1.0.0-no-default-cr
+          ModuleDeploymentNameInNewerVersion: template-operator-${{ steps.e2e-versions.outputs.newer }}-controller-manager
+          ModuleDeploymentNameInOlderVersion: template-operator-${{ steps.e2e-versions.outputs.older }}-controller-manager
```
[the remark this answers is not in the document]
amritanshusikdar: Has been fixed now.

medmes:
**`pkg/testutils/modulecr.go:37-41`**

`ModuleDeploymentNameInOlderVersion` and `ModuleDeploymentNameInNewerVersion` were string constants. Promoting them to package-level vars initialized by `MustReadVersionFromFileOf` means the file read runs at package init for every importer of `pkg/testutils` — not just E2E tests. 7 of 9 integration test suites (`kyma`, `manifest`, `modulereleasemeta`, `moduletemplate`, `purge`, `withwatcher`, `kcp`) use `import . "github.com/kyma-project/lifecycle-manager/pkg/testutils"` and none of them use these vars.

Practical consequences:
- Every integration test run now reads `versions.yaml` at init as a side effect, for values those tests never use.
- If `versions.yaml` is missing a key or `runtime.Caller(0)` resolves to a stale compile-time path, the panic surfaces as an integration test failure with no indication of what went wrong.

`OlderVersion`, `NewerVersion`, `MandatoryModule*Version` already live in `tests/e2e/utils_test.go` — the right layer for E2E-specific concerns. The deployment names should follow the same boundary:

```go
// tests/e2e/utils_test.go — alongside OlderVersion/NewerVersion
var (
    ModuleDeploymentNameInOlderVersion = fmt.Sprintf("template-operator-v%s-controller-manager",
        extractMajorVersionFromSemVer(MustReadVersionFromFileOf("module-version-older")))
    ModuleDeploymentNameInNewerVersion = fmt.Sprintf("template-operator-v%s-controller-manager",
        extractMajorVersionFromSemVer(MustReadVersionFromFileOf("module-version-newer")))
)
```

Restore the constants in `pkg/testutils/modulecr.go` to their stable `v1`/`v2` values. `pkg/testutils` is a shared package — it should not carry E2E-specific init-time side effects.

amritanshusikdar:
> **`pkg/testutils/modulecr.go:37-41`**
> 
> `ModuleDeploymentNameInOlderVersion` and `ModuleDeploymentNameInNewerVersion` were string constants. Promoting them to package-level vars initialized by `MustReadVersionFromFileOf` means the file read runs at package init for every importer of `pkg/testutils` — not just E2E tests. 7 of 9 integration test suites (`kyma`, `manifest`, `modulereleasemeta`, `moduletemplate`, `purge`, `withwatcher`, `kcp`) use `import . "github.com/kyma-project/lifecycle-manager/pkg/testutils"` and none of them use these vars.
> 
> Practical consequences:
> 
> * Every integration test run now reads `versions.yaml` at init as a side effect, for values those tests never use.
> * If `versions.yaml` is missing a key or `runtime.Caller(0)` resolves to a stale compile-time path, the panic surfaces as an integration test failure with no indication of what went wrong.
> 
> `OlderVersion`, `NewerVersion`, `MandatoryModule*Version` already live in `tests/e2e/utils_test.go` — the right layer for E2E-specific concerns. The deployment names should follow the same boundary:
> 
> ```go
> // tests/e2e/utils_test.go — alongside OlderVersion/NewerVersion
> var (
>     ModuleDeploymentNameInOlderVersion = fmt.Sprintf("template-operator-v%s-controller-manager",
>         extractMajorVersionFromSemVer(MustReadVersionFromFileOf("module-version-older")))
>     ModuleDeploymentNameInNewerVersion = fmt.Sprintf("template-operator-v%s-controller-manager",
>         extractMajorVersionFromSemVer(MustReadVersionFromFileOf("module-version-newer")))
> )
> ```
> 
> Restore the constants in `pkg/testutils/modulecr.go` to their stable `v1`/`v2` values. `pkg/testutils` is a shared package — it should not carry E2E-specific init-time side effects.

Reverting it to hardcoded strings would only bring more inconsistencies across the idea that we want a centralized versioning solution. But I see that these 2 variables (`ModuleDeploymentNameInOlderVersion` & `ModuleDeploymentNameInNewerVersion`) are only referenced in the `tests/e2e` directory.

I could refactor these to the `tests/e2e` directory. What do you think?

medmes:
Yes, move them to `tests/e2e/utils_test.go`. Remove `ModuleDeploymentNameInOlderVersion`, `ModuleDeploymentNameInNewerVersion`, `semVerPartsCount`, and `extractMajorVersionFromSemVer` from `pkg/testutils/modulecr.go` entirely — note that keeping them as constants there alongside same-named vars in `utils_test.go` would cause a compile error due to the dot-import.

medmes (approved):

PR #3414, opened 2026-07-06
feat: Introduce ApplyConfigurations for Gardener's CertManager types

### LINKED ISSUE #3394: Replace GCM unstructured SSA with generated typed ApplyConfigurations
## Context

`internal/repository/watcher/certificate/gcm/certificate/certificate_repo.go` currently uses `ApplyConfigurationFromUnstructured` as a bridge to `client.Apply()` because `github.com/gardener/cert-management` does not ship typed `ApplyConfiguration` builders. This was the only available path when migrating from the deprecated `client.Patch(..., client.Apply, ...)` in #3386.

## Goal

Generate typed GCM apply configurations into our repo, mirroring the existing setup under `api/applyconfigurations/`, and replace the `ToUnstructured → ApplyConfigurationFromUnstructured` path in the GCM certificate repository with the typed builder.

## Acceptance Criteria

~~- [ ] `applyconfiguration-gen` runs against `github.com/gardener/cert-management/pkg/apis/cert/v1alpha1` as part of `make generate`, outputting to `api/applyconfigurations/gcm/cert/v1alpha1/`~~ (Decided to use `.../gardener/...` instead of `.../gcm/...` in the package path)
- [x] `applyconfiguration-gen` runs against `github.com/gardener/cert-management/pkg/apis/cert/v1alpha1` as part of `make generate`, outputting to `api/applyconfigurations/gardener/cert/v1alpha1/`
- [x] `certificate_repo.go` uses the typed `CertificateApplyConfiguration` builder instead of the unstructured path — symmetric with the cert-manager implementation
- [x] The revocation type is excluded from generation if it causes generator issues (no issues, `CertificateRevocationApplyConfiguration` type is generated)
- [x] Existing unit tests continue to pass and the stub is updated to use the typed config
- [ ] If types are generatable, open an issue on `gardener/cert-management` to add `--with-applyconfig` to their own codegen.

## References

- #3386

Tomasz-Smelcerz-SAP:
**Description**

Fixes #3394 

- Introduce ApplyConfigurations generator for the types defined in the `cert.gardener.cloud/v1apha1` Group/Version.
- The generated files are located in the `api` module, in the **`./applyconfigurations/gardener`** subdirectory. This keeps all the applyconfigurations in the same location, yet there is a distinct location for Gardener's types. This is a hard requirement of the generator, it can't output types for different K8s API Groups into the same Go package.
- The directory for the generated files: **`./applyconfigurations/gardener/...`** is different than what was originally suggested in the issue: `./applyconfigurations/gcm/...`. This is because `gcm` is at best an unofficial alias - it's better to use explicit identifiers instead.

**Related issue(s)**
#3394


Makefile
```diff
@@ -53,7 +53,7 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=controller-manager crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases output:rbac:dir=config/rbac/common
+	$(CONTROLLER_GEN) rbac:roleName=controller-manager crd webhook paths="./..." paths="!./api/applyconfigurations/cert/..." output:crd:artifacts:config=config/crd/bases output:rbac:dir=config/rbac/common
```
Tomasz-Smelcerz-SAP: [Note for reviewer]: We have to exclude these (generated) files from `controller-gen` scanning, as these files contain no useful generation-driving markers (annotations) for `controller-gen`. In addition, `controller-gen` would have problems processing these files as they import types from Gardener project that are not listed in `lifecycle-manager` main module (because they are not needed to build the `lifecycle-manger`).

api/go.mod
```diff
@@ -17,26 +17,27 @@ require (
 	github.com/emicklei/go-restful/v3 v3.13.0 // indirect
 	github.com/evanphx/json-patch/v5 v5.9.11 // indirect
 	github.com/fxamacker/cbor/v2 v2.9.0 // indirect
+	github.com/gardener/cert-management/pkg/apis v0.25.0 // indirect
```
LeelaChacha: Shouldn't it be a direct dependency now?
Perhaps we're missing `make generate; go mod tidy`

hack/generate-cert-applyconfiguration.sh
```diff
@@ -0,0 +1,41 @@
...
+
+echo "Using k8s.io/code-generator@${CODE_GEN_VERSION}"
+
+APPLYCONFIGURATION_GEN_BIN="$(go env GOPATH)/bin/applyconfiguration-gen"
+if [[ ! -x "${APPLYCONFIGURATION_GEN_BIN}" ]]; then
+  echo "Building applyconfiguration-gen..."
+  GOBIN="$(go env GOPATH)/bin" go install "k8s.io/code-generator/cmd/applyconfiguration-gen@${CODE_GEN_VERSION}"
+fi
+
+echo "Generating ApplyConfigurations into ${OUTPUT_DIR} ..."
```
LeelaChacha: `hack/generate-cert-applyconfiguration.sh` looks unused — nothing invokes it, and `make generate` already runs an identical `applyconfiguration-gen` call for the gardener types. It's a duplicate generation path that also diverges from the Makefile: version from `client-go` (v0.36.2) vs `versions.yaml` pin (0.36.0), installs to `GOPATH/bin` vs `bin/`, and no `cd api`. I suggest dropping it, or have it call `make generate-applyconfiguration` instead of reimplementing the invocation.

LeelaChacha (changes requested):

hack/generate-cert-applyconfiguration.sh
Tomasz-Smelcerz-SAP: You are right, it was used initially and then, after a few bugfixes, the logic was merged into Makefile and I forgot to remove the script.

api/go.mod
Tomasz-Smelcerz-SAP: Good catch, fixed.

hack/generate-cert-applyconfiguration.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+# Generates ApplyConfiguration helper types for cert-management API types into
+# api/applyconfigurations/cert/gardener/. Re-run this script whenever cert-management is upgraded.
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: This script is removed.

Makefile
```diff
@@ -71,10 +71,16 @@ generate-applyconfiguration: applyconfiguration-gen manifests generate-openapi-s
 		--go-header-file ../hack/boilerplate.go.txt \
 		--openapi-schema ../hack/crd-to-openapi/openapi.json \
 		./v1beta2
+	cd api && $(APPLYCONFIGURATION_GEN) \
+		--output-dir ./applyconfigurations/cert/gardener \
+		--output-pkg github.com/kyma-project/lifecycle-manager/api/applyconfigurations/cert/gardener \
+		--go-header-file ../hack/boilerplate.go.txt \
+		--external-applyconfigurations k8s.io/api/core/v1.SecretReference:k8s.io/client-go/applyconfigurations/core/v1 \
+		github.com/gardener/cert-management/pkg/apis/cert/v1alpha1
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: The AC in the issue was edited to reflect the new path: **`./applyconfigurations/gardener/cert/v1alpha1`**
This decision is also explicitly mentioned in the PR's description.

hack/generate-cert-applyconfiguration.sh
Tomasz-Smelcerz-SAP: Removed.

LeelaChacha (approved):

PR #3415, opened 2026-07-06
test: Port watcher-zero-downtime E2E test to GCM setup

### LINKED ISSUE #2570: [Gardener CM] Port zero-down-time E2E test to GCM
### Description

Execute the `watcher-zero-downtime` E2E test in the GCM setup.

### Reasons

We also need to ensure that rotation is properly working without downtime for restricted markets

### Acceptance Criteria

- [ ] `watcher-zero-downtime` tests executed successfully in `/lifecycle-manager/.github/workflows/test-e2e-with-gcm.yml`

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

medmes:
## Changes proposed in this pull request

- Add `watcher_zero_downtime_gcm_test.mk` — GCM-specific Makefile that patches the serving certificate to `duration: 1464h / renewBefore: 1463h55m0s` (satisfies GCM's minimum 1440h duration and `renewBefore <= duration - 5m` constraint, rotating every 5 minutes) and speeds up KLM requeue intervals
- Add `watcher-zero-downtime` to the matrix in `test-e2e-with-gcm.yml` with a dedicated setup/deploy path and 30-minute job timeout
- Migrate deploy scripts from env-var-based GCM detection to an explicit `--use-gcm` flag
- Extend the observation window from 4 to 12 minutes and bump a cert annotation each poll to force immediate GCM reconciliation (GCM does not auto-requeue within short intervals)
- Add `ErrCertValidityNotAvailable` sentinel error so KLM does not put a Kyma CR into `StateError` during startup while GCM is still issuing the serving certificate
- Guard `bumpServingCertAnnotation` with `meta.IsNoMatchError` so the shared test function runs correctly on both cert-manager and GCM clusters

## Related issue(s)

Closes #2570


tests/e2e/watcher_zero_downtime_gcm_test.mk
```diff
@@ -0,0 +1,31 @@
...
+klm-patch: kustomize-install
+	@echo "::group::KLM patch"
+	@export PATH=$(LOCALBIN):$$PATH
+	@pushd $(LIFECYCLE_MANAGER_DIR)/config/watcher_local_test_gcm > /dev/null
+	# op: replace at index 11 targets --kyma-requeue-success-interval added by config/watcher_local_test_gcm/kustomization.yaml.
+	# Verify the arg index there if this patch breaks silently after changing the base args list.
+	kustomize edit add patch --kind Deployment --patch \
+		'[{"op":"replace","path":"/spec/template/spec/containers/0/args/11","value":"--kyma-requeue-success-interval=1s"},{"op":"add","path":"/spec/template/spec/containers/0/args/-","value":"--istio-gateway-server-cert-switch-grace-period=30s"},{"op":"add","path":"/spec/template/spec/containers/0/args/-","value":"--istio-gateway-secret-requeue-success-interval=1s"}]'
+	@popd > /dev/null
+	@echo "::endgroup::"
```
c-pius, line 22: This misses the described patch of the certificate:

> Add watcher_zero_downtime_gcm_test.mk — GCM-specific Makefile with klm-patch (patches cert.gardener.cloud/v1alpha1 Certificate to duration: 1h / renewBefore: 59m and speeds up requeue intervals) and test-run targets

Duration and Renew before are supposed to be configured in a way that certs rotate every minute. Only with such a change we force the rotation to happen multiple times during test so that we can observe if it is working. See `config/gardener-certmanager/certificate_watcher.yaml`.  The defaults are `duration: 2160h # 90d`,  `renewBefore: 1464h # 61d`. It just never rotates in the test and therefore it is of course not observing downtime. It may be that Gardener doesn't support such short rotation intervals, this needs to be proven by the running test. See related comment https://github.com/gardener/cert-management/issues/492#issuecomment-3061548507. I think there are some limitations on duration and renewBefore.

See also the `gardener_cert_manager_renewal_window: ${{ env.E2E_GARDENER_CERT_MANAGER_RENEWAL_WINDOW }}` variable set as part of deploying. I think this is a relict from before GCM introducing support for renewBefore, can probably be dropped here and from the scripts: https://github.com/gardener/cert-management/issues/492. Now it seems that both needed feature `duration` and `renewBefore` can be set on the cert resources same as if using cert-manager.

.github/workflows/test-e2e-with-gcm.yml
```diff
@@ -112,6 +117,8 @@ jobs:
             make -f tests/e2e/e2e.no_module_setup.mk test-run "GINKGO_FOCUS=Enqueue Event from Watcher"
           elif [[ "${{ matrix.e2e-test }}" == "modulereleasemeta-watch-trigger" ]]; then
             make -f tests/e2e/modulereleasemeta_watch_trigger_test.mk test-run
+          elif [[ "${{ matrix.e2e-test }}" == "watcher-zero-downtime" ]]; then
+            make -f tests/e2e/watcher_zero_downtime_gcm_test.mk test-run
           else
             make -C tests/e2e ${{ matrix.e2e-test }}
           fi
```
c-pius, line 124: These should call the entire `test` target eventually. No need to keep the else fallback.

.github/workflows/test-e2e-with-gcm.yml
```diff
@@ -80,6 +81,10 @@ jobs:
           k8s_version: ${{ steps.configuration.outputs.k8s_version }}
           gardener_cert_manager_version: ${{ steps.configuration.outputs.gardener_cert_manager_version }}
           gardener_cert_manager_renewal_window: ${{ env.E2E_GARDENER_CERT_MANAGER_RENEWAL_WINDOW }}
```
c-pius, line 85: This should be dropped here and replaced by a `test` target in the makefile that forwards to a `create-clusters-gcm` target. The `create-clusters-gcm` would need to setup the test clusters with the variables above. See below https://github.com/kyma-project/lifecycle-manager/pull/3415/changes#r3543818851. I think `gardener_cert_manager_renewal_window` can be dropped entirely.

.github/workflows/test-e2e-with-gcm.yml
```diff
@@ -80,6 +81,10 @@ jobs:
...
+      - name: KLM patch for watcher-zero-downtime
+        if: ${{ matrix.e2e-test == 'watcher-zero-downtime' }}
+        working-directory: lifecycle-manager
+        run: make -f tests/e2e/watcher_zero_downtime_gcm_test.mk klm-patch
```
c-pius: This should be dropped here and replaced by a `test` target in the makefile that forwards to a `klm-patch` target

.github/workflows/test-e2e-with-gcm.yml
c-pius, line 91: This should be dropped here and replaced by a `test` target in the makefile that forwards to a `deploy-klm` target. We may be able to re-use the existing target in `e2e.common.mk`, but we need to supply it with a "USE GCM" var so that it adds the proper kustomization as in `.github/actions/deploy-lifecycle-manager-e2e/action.yml`.

c-pius, line 111: This should be dropped here and replaced by a `test` target in the makefile that forwards to a `module-setup` target. I think module-setup is only needed for `modulereleasemeta-watch-trigger`, see the related existing makefile.

c-pius (changes requested):
See the detailed commends below. The test still misses the actual configuration so that certs are renewed every minute.

It seems that by now, GCM and Cert-Manager have feature parity on the required features (duration, renewBefore). Please confirm but with that, the klm-patches would look the same. That leaves the only differences to account for:

- `tests/e2e/e2e.common.mk#deploy-klm`: needs to choose the correct Kustomize component as implemented in `.github/actions/deploy-lifecycle-manager-e2e/action.yml`
- `tests/e2e/e2e.common.mk#create-clusters`: needs to call `create_test_clusters.sh` with `--gardener-cert-manager-version` instead of `--cert-manager-version`

My proposal would be to to make these existing targets parametrizable with a `USE_GCM=true` make variable so that they do the correct setup. With this, it may be possible to use the exact same e2e test makefile for both flavors.

---

We should strive to finish this topic within this sprint. If something is unclear or you need help please ping me and I will try to support further.

tests/e2e/watcher_zero_downtime_gcm_test.mk
medmes: Done — added the Certificate CR patch in `klm-patch` using the `cert.gardener.cloud/v1alpha1` API group with `duration: 1h` / `renewBefore: 59m`. This confirms GCM/CM feature parity on these fields.

The open question about GCM minimum cert duration (ref: gardener/cert-management#492) will be settled by the CI run. If GCM rejects `duration: 1h`, the cert controller will surface a clear error and we can adjust to whatever the actual minimum is.

`gardener_cert_manager_renewal_window` is kept in the workflow for `watcher-enqueue` and `modulereleasemeta-watch-trigger` — those still use the shared setup steps. It'll be removed when those are migrated (tracked in #3417).

.github/workflows/test-e2e-with-gcm.yml
medmes: Done — `watcher-zero-downtime` now calls `make -f tests/e2e/watcher_zero_downtime_gcm_test.mk test`, which runs the full lifecycle (create-clusters → klm-patch → deploy-klm → test-run). The `else` fallback is removed.

Migrating `watcher-enqueue` and `modulereleasemeta-watch-trigger` to self-contained `test` targets is deferred — tracked in #3417.

.github/workflows/test-e2e-with-gcm.yml
medmes: Done for `watcher-zero-downtime` — step is gated on `matrix.e2e-test != 'watcher-zero-downtime'`. The `create-clusters` target in `e2e.common.mk` now branches on `USE_GCM`: passes `--gardener-cert-manager-version` when `true`, `--cert-manager-version` otherwise.

`gardener_cert_manager_renewal_window` stays for the remaining two tests for now; removed in the follow-up (#3417).

.github/workflows/test-e2e-with-gcm.yml
medmes: Done — the standalone KLM patch step is removed from the workflow. `klm-patch` is now called as part of `make test` in `watcher_zero_downtime_gcm_test.mk`.

.github/workflows/test-e2e-with-gcm.yml
medmes: Done — step gated on `!= watcher-zero-downtime`. The `deploy-klm` target in `e2e.common.mk` uses `USE_GCM=true` (set in the `.mk` file) to select `config/watcher_local_test_gcm` as the kustomize dir and prepends `E2E_USE_GARDENER_CERT_MANAGER=1` when calling `deploy_klm_from_registry.sh`, which then picks `local-deploy-with-watcher-gcm`.

medmes: Updated — condition changed to `matrix.e2e-test == 'modulereleasemeta-watch-trigger'` so the step only runs for the one test that actually needs it. `watcher-enqueue` had a no-op `module-setup` target in `e2e.no_module_setup.mk` so it was always safe to skip.

scripts/tests/deploy_klm_from_registry.sh
```diff
@@ -44,10 +44,16 @@ case "${KLM_IMAGE_REGISTRY}" in
     ;;
 esac
 
+if [[ -n "${E2E_USE_GARDENER_CERT_MANAGER:-}" ]]; then
+  DEPLOY_TARGET=local-deploy-with-watcher-gcm
+else
+  DEPLOY_TARGET=local-deploy-with-watcher
+fi
+
```
c-pius: Would prefer to have an explicit argument as for the images.

scripts/tests/deploy_klm_from_sources.sh
```diff
@@ -17,10 +17,16 @@ export TAG=$(date +%Y%m%d%H%M%S)
 make docker-build IMG=${LOCAL_IMG}:${TAG}
 make docker-push IMG=${LOCAL_IMG}:${TAG}
 
+if [[ -n "${E2E_USE_GARDENER_CERT_MANAGER:-}" ]]; then
+  DEPLOY_TARGET=local-deploy-with-watcher-gcm
+else
+  DEPLOY_TARGET=local-deploy-with-watcher
+fi
+
```
c-pius: See above, would prefer an explicit script argument

c-pius (changes requested):
There still seems to be an error on the test run

```
    lastOperation:
      lastUpdateTime: "2026-07-09T09:15:21Z"
      operation: 'failed to renew SKR certificate: failed to determine if SKR certificate
        needs renewal: failed to determine SKR client certificate validity: certificate
        status does not contain issuanceDate or expirationDat
```

Also the test still doesn't verify. In the GCM setup, we now rotate every 5 mins, but the `tests/e2e/watcher_zero_downtime_test.go` only observes 4 minutes. This should be increased to let's say 12 mins to be sure it survives two rotations.

c-pius:
The serving cert is going to error and we can't even issue skr certs

```
Name:         klm-watcher-serving
Namespace:    istio-system
Labels:       app.kubernetes.io/created-by=argo-cd
              app.kubernetes.io/instance=kcp-lifecycle-manager
              app.kubernetes.io/managed-by=kustomize
              app.kubernetes.io/name=lifecycle-manager
              app.kubernetes.io/part-of=kcp
Annotations:  <none>
API Version:  cert.gardener.cloud/v1alpha1
Kind:         Certificate
Metadata:
  Creation Timestamp:  2026-07-09T10:05:32Z
  Generation:          1
  Resource Version:    1201
  UID:                 a774857e-b04c-40f6-9e9d-c1e9370531b5
Events:
  Type     Reason     Age    From                     Message
  ----     ------     ----   ----                     -------
  Warning  reconcile  3m30s  cert-controller-manager  fetching issuer failed: issuers.cert.gardener.cloud "klm-watcher-root" not found
  Warning  reconcile  90s    cert-controller-manager  certificate duration must be at least 1440h0m0s

```

c-pius:
Locally this config leads to ready serving and skr certs `'[{"op":"replace","path":"/spec/renewBefore","value":"1463h55m0s"},{"op":"replace","path":"/spec/duration","value":"1464h"}]'`

However, one can observe that the serving cert is not rotated every 5 mins. It seems that the cert controller has longer requeue intervals. Adding an annotation to the cert worked to get it queued up and rotated. We may need to add an annotation counter in `triggerWatcherAndCheckDowntime`

internal/service/watcher/certificate/certificate_service.go
```diff
@@ -202,6 +203,9 @@ func (c *Service) constructDNSNames(kyma *v1beta2.Kyma) ([]string, error) {
 func (s *Service) skrCertificateNeedsRenewal(ctx context.Context, certName string) (bool, error) {
 	certNotBefore, _, err := s.certRepo.GetValidity(ctx, certName)
 	if err != nil {
+		if errors.Is(err, certerror.ErrCertValidityNotAvailable) {
+			return false, nil
+		}
```
c-pius, line 208: What is the reason for not propagating this error upwards anymore?

c-pius (commented):
Running this test locally unfortunately reported downtime:

```
kyma (kyma-sample-fbhvpykw) in cluster: Spec: {Channel:regular SkipMaintenanceWindows:false Modules:[]}, Status: {LastOperation:{Operation:kyma is ready LastUpdateTime:2026-07-09 16:14:42 +0200 CEST} State:Ready Conditions:[{Type:CRDsSync Status:True ObservedGeneration:1 LastTransitionTime:2026-07-09 16:14:42 +0200 CEST Reason:Ready Message:CRDs are synchronized to the SKR} {Type:Modules Status:True ObservedGeneration:1 LastTransitionTime:2026-07-09 16:14:42 +0200 CEST Reason:Ready Message:all modules are in ready state} {Type:ModuleCatalog Status:True ObservedGeneration:1 LastTransitionTime:2026-07-09 16:14:42 +0200 CEST Reason:Ready Message:module templates are synchronized} {Type:SKRWebhook Status:True ObservedGeneration:1 LastTransitionTime:2026-07-09 16:14:42 +0200 CEST Reason:Ready Message:skrwebhook is synchronized}] Modules:[] ActiveChannel:fast}
  [FAILED] in [It] - /Users/I507792/SAPDevelop/github.com/c-pius/lifecycle-manager/tests/e2e/watcher_zero_downtime_test.go:44 @ 07/09/26 16:15:02.613
  STEP: When delete KCP Kyma @ 07/09/26 16:15:02.613
  STEP: Then SKR Kyma deleted @ 07/09/26 16:15:02.618
  STEP: Then KCP Kyma deleted @ 07/09/26 16:15:02.619
• [FAILED] [583.719 seconds]
Watcher Zero Downtime Given SKR Cluster [It] Then no downtime errors can be observed
/Users/I507792/SAPDevelop/github.com/c-pius/lifecycle-manager/tests/e2e/watcher_zero_downtime_test.go:38

  [FAILED] Failed after 583.710s.
  Expected success, but got an error:
      <*errors.errorString | 0x3ab8455bb70>: 
      watcher is experiencing downtime
      {
          s: "watcher is experiencing downtime",
      }
  In [It] at: /Users/I507792/SAPDevelop/github.com/c-pius/lifecycle-manager/tests/e2e/watcher_zero_downtime_test.go:44 @ 07/09/26 16:15:02.613
------------------------------
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS
------------------------------
[AfterSuite] 
/Users/I507792/SAPDevelop/github.com/c-pius/lifecycle-manager/tests/e2e/suite_test.go:148
  STEP: Print out all remaining resources for debugging @ 07/09/26 16:15:02.622
  kyma: {{Kyma operator.kyma-project.io/v1beta2} {kyma-sample-fbhvpykw  kcp-system  1dce6b5e-abc6-4dd9-9881-4bed477080c1 3144 2 2026-07-09 16:04:42 +0200 CEST 2026-07-09 16:15:02 +0200 CEST 0x3ab8498d0c0 map[kyma-project.io/instance-id:test-instance kyma-project.io/runtime-id:yciaecdh operator.kyma-project.io/managed-by:lifecycle-manager] map[skr-domain:example.domain.com] [] [operator.kyma-project.io/purge-finalizer operator.kyma-project.io/Kyma] [{lifecycle-manager Apply operator.kyma-project.io/v1beta2 2026-07-09 16:15:02 +0200 CEST FieldsV1 {"f:metadata":{"f:annotations":{"f:skr-domain":{}},"f:deletionGracePeriodSeconds":{},"f:deletionTimestamp":{},"f:finalizers":{"v:\"operator.kyma-project.io/Kyma\"":{},"v:\"operator.kyma-project.io/purge-finalizer\"":{}},"f:labels":{"f:kyma-project.io/instance-id":{},"f:kyma-project.io/runtime-id":{},"f:operator.kyma-project.io/managed-by":{}}},"f:status":{"f:activeChannel":{},"f:conditions":{"k:{\"type\":\"CRDsSync\"}":{".":{},"f:lastTransitionTime":{},"f:message":{},"f:observedGeneration":{},"f:reason":{},"f:status":{},"f:type":{}},"k:{\"type\":\"ModuleCatalog\"}":{".":{},"f:lastTransitionTime":{},"f:message":{},"f:observedGeneration":{},"f:reason":{},"f:status":{},"f:type":{}},"k:{\"type\":\"Modules\"}":{".":{},"f:lastTransitionTime":{},"f:message":{},"f:observedGeneration":{},"f:reason":{},"f:status":{},"f:type":{}},"k:{\"type\":\"SKRWebhook\"}":{".":{},"f:lastTransitionTime":{},"f:message":{},"f:observedGeneration":{},"f:reason":{},"f:status":{},"f:type":{}}},"f:lastOperation":{"f:lastUpdateTime":{},"f:operation":{}},"f:state":{}}} status} {e2e.test Update operator.kyma-project.io/v1beta2 2026-07-09 16:04:42 +0200 CEST FieldsV1 {"f:metadata":{"f:annotations":{".":{},"f:skr-domain":{}},"f:labels":{".":{},"f:kyma-project.io/instance-id":{},"f:kyma-project.io/runtime-id":{}}},"f:spec":{".":{},"f:channel":{}}} } {manager Update operator.kyma-project.io/v1beta2 2026-07-09 16:04:47 +0200 CEST FieldsV1 {"f:metadata":{"f:finalizers":{".":{},"v:\"operator.kyma-project.io/Kyma\"":{},"v:\"operator.kyma-project.io/purge-finalizer\"":{}},"f:labels":{"f:operator.kyma-project.io/managed-by":{}}}} }]} {regular false []} {{waiting for modules to be deleted 2026-07-09 16:15:02 +0200 CEST} Deleting [{CRDsSync Unknown 2 2026-07-09 16:15:02 +0200 CEST Ready CRDs synchronization state is unknown} {Modules Unknown 2 2026-07-09 16:15:02 +0200 CEST Ready modules state is unknown} {ModuleCatalog Unknown 2 2026-07-09 16:15:02 +0200 CEST Ready module templates synchronization state is unknown} {SKRWebhook Unknown 2 2026-07-09 16:15:02 +0200 CEST Ready skrwebhook is out of sync and needs to be resynchronized}] [] regular}}
[AfterSuite] PASSED [0.003 seconds]
------------------------------

Summarizing 1 Failure:
  [FAIL] Watcher Zero Downtime Given SKR Cluster [It] Then no downtime errors can be observed
  /Users/I507792/SAPDevelop/github.com/c-pius/lifecycle-manager/tests/e2e/watcher_zero_downtime_test.go:44

Ran 2 of 321 Specs in 620.149 seconds
FAIL! -- 1 Passed | 1 Failed | 0 Pending | 319 Skipped
--- FAIL: TestAPIs (621.16s)
FAIL
exit status 1
FAIL    github.com/kyma-project/lifecycle-manager/tests/e2e     622.107s
::endgroup::
gmake: *** [tests/e2e/watcher_zero_downtime_gcm_test.mk:26: test-run] Error 1
```

Question is if the test was flaky, or if there was indeed downtime. Cannot tell unfortunately.

internal/service/watcher/certificate/certificate_service.go
medmes: `ErrCertValidityNotAvailable` signals a transient startup condition: the GCM serving certificate has been created but not yet issued, so its status has no `issuanceDate` / `expirationDate` yet. Propagating the error here would cause the reconcile loop to surface it as a failure on the Kyma CR, pushing it toward `StateError` — but there is nothing wrong; GCM simply hasn't finished issuance yet.

Returning `false, nil` means "we cannot determine whether renewal is needed right now, so skip it" — the reconciler will retry on the next cycle, by which point the cert will be ready. This was the error observed in the test run before this fix:

```
failed to renew SKR certificate: failed to determine if SKR certificate needs renewal:
failed to determine SKR client certificate validity: certificate status does not contain issuanceDate or expirationDate
```

c-pius (approved):
Let's keep an eye on it if it is flakey. Had two further local runs that worked.

PR #3446, opened 2026-07-20
chore: Update Plutono Self-Signed Cert Panel to have Instantaneous Tabular View

### LINKED ISSUE #2466: Self-signed Certificate Not Renew panel have false positive report
### Description

Self-signed Certificate Not Renew pannel report many self-signed certificate is not renewed, however, all this case seems just cluster 

Image

### Steps to reproduce

check lifecycle manager overview dashboard

### Environment Type

Managed

### Environment Info

Kubernetes Version: x.y.z
Docker Version: x.y.z
modulectl Version: x.y.z

### Attachments

_No response_

amritanshusikdar:
**Description**
- Adjusts the panel to show instantaneous data.
- Data is represented in tabular format.

**Related issue(s)**
Fixes #2466

### DIFF
--- a/config/grafana/overview.json
+++ b/config/grafana/overview.json
@@ -1619,17 +1619,20 @@
         "x": 0,
         "y": 83
       },
-      "id": 23763571993,
+      "id": 80,
+      "options": {
+        "showHeader": true
+      },
       "pluginVersion": "7.5.47",
       "targets": [
         {
-          "expr": "lifecycle_mgr_self_signed_cert_not_renew{service=\"klm-controller-manager-metrics\"}",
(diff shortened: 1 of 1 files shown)


amritanshusikdar:
image

Considering the [downtime of the DEV landscape](https://github.com/kyma-project/lifecycle-manager/issues/2466#issuecomment-5023303516), I setup the observability stack locally and can confirm that it works. The attached image shows what the table entry looks like in that instant, when a certificate is not valid anymore. Although it would've been nice to have it also tested on the DEV landscape, I think this is good enough to unblock the issue.

c-pius (approved):

PR #3448, opened 2026-07-21
refactor: Spread mandatory module requeues on ModuleReleaseMeta change

### LINKED ISSUE #3203: Spread Kyma requeues on New Mandatory Module Rollout
### Description

See:

- https://github.com/kyma-project/lifecycle-manager/pull/3185
- https://github.com/kyma-project/lifecycle-manager/pull/3200

Also update the handler for mandatory modules to rollout less aggressively. Consider to use one handler that determines the kidn of MRM (mandatory or regular) that then defers to injected services determining the Kymas to requeue

### Reasons

Follow-up to https://github.com/kyma-project/lifecycle-manager/issues/2767. We implemented less aggressive rollout for regular modules, but not mandatory ones. Mandatory modules should be rolled out in the same way.

### Acceptance Criteria

- [ ] mandatory module updates rolled out less aggressively

### Dev Notes

- this is the existing handler: https://github.com/kyma-project/lifecycle-manager/blob/main/internal/watch/modulereleasemeta/mrm_handler.go
  - currently the determination of affected Kymas is coded into the handler
  - this cold be separated out into functionality that is injected from the outside
  - we could just instantiate this handler two times, once with injected functionality to determine affected Kymas for regular moduels, once with injected functionality to determine Kymas for mandatory modules
  - it already uses a [requeueWithRandomDelay](https://github.com/kyma-project/lifecycle-manager/blob/14420ec01eb661665bd3f226a23bd0e82b2a0055/internal/watch/modulereleasemeta/mrm_handler.go#L101-L115) function that can also be applied to mandatory modules
- the second instance for mandatory modules then replaces the [mandatory mrm handler](https://github.com/kyma-project/lifecycle-manager/blob/main/internal/watch/modulereleasemeta/mandatory_mrm_handler.go) used here https://github.com/kyma-project/lifecycle-manager/blob/14420ec01eb661665bd3f226a23bd0e82b2a0055/internal/controller/mandatorymodule/setup.go#L42
- e2e testing for the existing handler is here: tests/e2e/modulereleasemeta_watch_trigger_test.go
  - it is configured with very slow reconciliation intervals
  - the test verifies that after a invalid change to mrm, the Kyma goes to error state before the long interval

LeelaChacha:
## Description

Regular module upgrades already spread Kyma requeues over a random delay window (`--module-upgrade-rollout-max-delay`) so SKRs don't all pull a new image at once. Mandatory modules did not benefit from this — a mandatory version change enqueued every managed Kyma immediately.

This PR extracts the affected-Kyma determination into an injectable `affectedKymasResolver`, so a single typed `ModuleReleaseMeta` event handler now serves both module kinds. The mandatory installation controller reuses the existing random-delay spreading on update events. The redundant mandatory handler is removed.

The existing `--module-upgrade-rollout-max-delay` flag (default 5m) is reused for both paths. Regular behavior is unchanged.

Resolves #3203


internal/watch/modulereleasemeta/mrm_handler.go
```diff
@@ -13,63 +13,52 @@ import (
...
+func (m TypedEventHandler[object, request]) Create(ctx context.Context, evt event.CreateEvent,
+	rli workqueue.TypedRateLimitingInterface[reconcile.Request],
 ) {
-	// NOOP - we don't want to react to Generic Events.
+	mrm, kymas, ok := m.lookup(ctx, evt.Object)
+	if !ok {
+		return
+	}
+	requeueKymas(rli, m.resolver.OnCreate(mrm, kymas))
 }
```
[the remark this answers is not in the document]
LeelaChacha: Good catch that `Create` lists Kymas even when `RegularResolver.OnCreate` returns nil. In practice this is a no-op cost: MRMs are created very rarely and `kymaRepo` uses the manager's cache-backed client, so `LookupByLabel` reads from the in-memory informer cache rather than the API server, and MRM creations are rare (version-publish events). Adding a `needsKymaList()`-style capability method would re-couple the handler to each resolver's decision logic — the opposite of what this refactor's resolver abstraction is for. Keeping the uniform `lookup → resolver` path as-is.

internal/watch/modulereleasemeta/mrm_handler.go
```diff
@@ -13,63 +13,52 @@ import (
...
-	_ workqueue.TypedRateLimitingInterface[reconcile.Request],
+func (m TypedEventHandler[object, request]) Create(ctx context.Context, evt event.CreateEvent,
+	rli workqueue.TypedRateLimitingInterface[reconcile.Request],
 ) {
-	// NOOP - we don't want to react to Generic Events.
+	mrm, kymas, ok := m.lookup(ctx, evt.Object)
+	if !ok {
+		return
+	}
+	requeueKymas(rli, m.resolver.OnCreate(mrm, kymas))
```
Tomasz-Smelcerz-SAP, line 47: I've read Copilot suggestion and I disagree with it, but for different reasons :)

In cases like that, where we're not really sure if the resolver needs Kymas or not (sometimes it does, sometimes it doesn't) the proper solution is NOT to ask the resolver "do you need kymas?", but to make kyma list lookup **lazy**. In other words, instead of giving the resolver the concrete list of kymas, it should be given a function-that-returns-the-list-of-kymas.
In that way it's then the resolver that decides when (and if) to evaluate this function.
In that way the contract between resolver and this handler remains decoupled and the small optimization mentioned by Copilot can still be implemented.

But I agree that in this particular case the "optimization" may be not worth the additional complexity of "lazy lookup" - while not much complicated, it's always more complicated than just passing around a concrete list of kyma objects. Up to you.

Tomasz-Smelcerz-SAP (dismissed):

Tomasz-Smelcerz-SAP (commented):
LGTM. Just left one comment. Not approving now because of auto-merge, please read the comment and either implement or dismiss it.

internal/watch/modulereleasemeta/mrm_handler.go
LeelaChacha: Agree the "does the resolver need kymas?" predicate is the wrong shape — it couples the handler to resolver internals. The lazy `func() (*KymaList, error)` is the correct decoupled form, but right now all three resolver methods always iterate the list, so there's no branch that would actually skip the fetch — it'd be generality for a caller that doesn't exist yet. I'd keep the concrete list for now and revisit if/when a resolver branch appears that can genuinely avoid the lookup. If fetch cost is the real worry, scoping `LookupByLabel` is the more direct lever than laziness.

YAGNI

Tomasz-Smelcerz-SAP (approved):

PR #3457, opened 2026-07-23
chore: Replace `skrAccessSecretRepo` with `ErrAccessSecretNotFound` explicit checks in Kyma Deletion

### LINKED ISSUE #3296: Drop explicit checks for SKR Kubeconfig Secrets
### Description

Drop the explicit checks whether the SKR Kubeconfig Access Secret still exists from the `IsApplicable` function of all SKR-related Use Cases in the Kyma Deletion Service.

```go
if exists, err := u.skrAccessSecretRepo.ExistsForKyma(ctx, kcpKyma.GetName()); !exists || err != nil {
    return false, err
}
```

Replace it with an explicit check in the Use Case for the [`ErrAccessSecretNotFound`](internal/service/accessmanager/manager.go). If this error is returned from the SKR-facing applicability check, return that the Use Case is not applicable (`false`).

### Reasons

There are no guarantees around the SKR Kubeconfig Access Secret. It was observed that is is gone shortly after setting the KCP Kyma CR to Deleting which made all SKR Use Cases not execute any longer. This change aligns the Deletion Service behavior with existing handling. As long as available, the cached client is used. Once there is no cached client anymore and no SKR Kubeconfig Access Secret to re-create the client, it is assumed that SKR-related cleanup is not possible anymore and the deletion progresses.

>[!Note]
>The cached client is evicted from the cache on network-related issues ([`IsConnectionRelatedError`](pkg/util/error.go)). This means that the cached client may be evicted on transient network errors even though it may still be valid. Also, it may happen that the TTL of cached client times out during deletion processing. This leads SKR-related cleanup not finishing gracefully. This is accepted as of now as Gardener forcefully finalizes the cluster after a timeout and we don't have stronger guarantees for the SKR Kubeconfig Access Secret.

### AC

- [x] `IsApplicable` functions of [SKR-related Use Cases](internal/service/kyma/deletion/usecases) do NOT use `u.skrAccessSecretRepo.ExistsForKyma` checks anymore
- [x] `IsApplicable` functions of [SKR-related Use Cases](internal/service/kyma/deletion/usecases) perform their check on the SKR. If the check returns [`ErrAccessSecretNotFound`](internal/service/accessmanager/manager.go), they return `false` (not applicable).

amritanshusikdar:
**Description**

- Cleaned codebase wide definition and usages of `skrAccessSecretRepo`
- Did explicit checks for `ErrAccessSecretNotFound` instead
- Adjusted unit tests

**Related issue(s)**
Fixes #3296


config/samples/component-integration-installed/crd/operator.kyma-project.io_kcpmodules.yaml
c-pius, line 1: Where is this change coming from?

internal/service/kyma/deletion/usecases/delete_skr_crd_test.go
c-pius, line 1: Image

The unexpected error case is uncovered.

internal/service/kyma/deletion/usecases/set_skr_kyma_state_deleting.go
c-pius, line 1: Image

internal/service/kyma/deletion/usecases/set_skr_kyma_state_deleting.go
```diff
@@ -36,14 +35,10 @@ func (u *SetSkrKymaStateDeleting) IsApplicable(ctx context.Context, kcpKyma *v1b
...
 
-	if exists, err := u.skrAccessSecretRepo.ExistsForKyma(ctx, kcpKyma.GetName()); !exists || err != nil {
-		return false, err
-	}
-
 	status, err := u.skrKymaStatusRepo.Get(ctx, kcpKyma.GetNamespacedName())
 
 	// SKR kyma is already gone
-	if util.IsNotFound(err) {
+	if errors.Is(err, accessmanager.ErrAccessSecretNotFound) || util.IsNotFound(err) {
```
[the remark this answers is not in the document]
c-pius: I think you can drop the comment entirely.

config/samples/component-integration-installed/crd/operator.kyma-project.io_kcpmodules.yaml
amritanshusikdar: I think it arises upon running the test suite. Unnecessary, has been reverted.

internal/service/kyma/deletion/usecases/set_skr_kyma_state_deleting.go
amritanshusikdar: New test added for the missing execution path.

internal/service/kyma/deletion/usecases/delete_skr_crd_test.go
amritanshusikdar: New test case for the missing execution path has been added.

c-pius (approved):

PR #3466, opened 2026-07-27
docs: ADR 010:  Introduce the behavioral contract documentation for `Module CR` handling

Tomasz-Smelcerz-SAP:
**Description**
Add ADR 009 (Proposal) describing how to document the internal contract for system behavior.


docs/contributor/adr/009-maintain-system-behavioral-contract.md
```diff
@@ -0,0 +1,148 @@
...
+
+
+## Decision
+
+### Format
+The contract is written in Markdown format. It has four sections:
+
+- **Scope** - defines the scope of the contract. It describes what is covered by the document and what is not.
+- **Decisions** - defines the contract items that are in effect
+- **Pending Decisions** - defines the contract items that are not yet in effect. These may be contract gaps or things that are still under discussion, but are important enough to be visible.
```
LeelaChacha, line 87: What do you think of not including pending decisions in the document? I have an intuition that it can be easily misinterpreted by an AI agent as an actual decision pending to be implemented, or in the worst case, when its context is full, and it somehow skips attention on the word "Pending", then it might think these are actual decisions too. Especially since we can't predict what files the AI agent loads in its context at runtime, it might be the case that this ADR is not in its context, and it doesn't understand the format of a contract and what each section means.
Of course, there will be a human in the loop, but if we want to develop efficiently, we should keep this contract as lean as possible and leave no margin for error. The contract MD should have everything the Agent needs to work with and nothing that it can misinterpret.
Similarly, we might skip the decision references, as it does not add any value to an agent and just fills up the context unnecessarily.

LeelaChacha (commented):
Theoretically, this is a solid proposal. I see the value in writing down all our current contracts in short and concise documents and using them as a source of truth for developments, discussions, and onboarding guides. I also agree with the ADR's suggestion for the contract location and reasoning for not using ADRs themselves as contracts.

In an ideal world, this would work perfectly. The contracts would be maintained and updated when anything changes, used to guide development, and efficiently consumed by AI agents. However, seeing that KLM is moving towards a maintenance mode, I feel like the overhead of creating the contracts now and maintaining them in case of changes is too big an overhead. This will involve a lot of interteam communication. This might end up as something that we work on slowly, like the refactor of the declarative reconciler, which took 2 years to finish, or it might just never be implemented to its full potential, like the `Discussions` tab feature in GitHub, which we agreed on using but never did. I am not against this but it is just food for thought that this undertaking might do more harm then good in the future.

lindnerby:
Thanks for the ADR proposal. Glad you're tackling the documentation gaps... the problem statement is true in my opinion: behavior is scattered across ADRs, issues, and implicit code, and that makes reasoning about the system difficualt.

I share some of the sustainability concern raised by @LeelaChacha, though not because of maintenance mode (KLM is under active development). My concern is that the proposed format may be too exhaustive to adopt and keep up-to-date, and that a simpler design would serve us better.

The ADR itself calls for a pragmatic approach, and I think we should apply that to the contract format itself. The proposed structure with formal IDs, a cross-reference table, a separate `Pending Decisions` section, a central directory, etc. poses a very formal documentation. The risk is that the ceremony becomes the obstacle.
Formal IDs like `R010` seem arbitray. Also documenting `Pending Decisions`, committing open questions and gaps to the repo sounds useful, but in practice these accumulate and never get resolved. **Documentation should describe behavior that is decided and in effect**; gaps and open questions belong in GitHub issues imo where they can be tracked and closed.
I also think that a central directory for this documentation is not a good idea.

My idea would be: bring the contract to the service layer, not the other way around.
ADR 004 already gives us the answer implicitly. It defines the service layer as the single home of all business logic:

> "Business code is confined to the Service layer, so no other layer will cause side-effects to the business logic."

If that holds, markdown sitting inside each service package *is* the behavioral contract for that scope. It doesn't need a separate home.

```
internal/service/modulereleasemeta/README.md    ← contract for MRM behavior
internal/service/mandatorymodule/README.md  ← contract for mandatory module behavior
```

The strongest argument against `docs/internal/contract/` is not format or tooling it would be *structural coupling*.

The Common Closure Principle states: *"Gather into components those classes that change for the same reasons and at the same times."* Markdown describing the behavior of `internal/service/mandatorymodule` changes for exactly the same reasons and at exactly the same times as the code in that package. By the CCP, they belong in the same component. When a behavior changes, the PR that changes the service code also touches the `README.md` in the same directory. The diff shows both and the reviewer sees both. No process rule or discipline is required to keep them in sync, like it would when using a central location.

The "packages move around" counterargument does not really apply here. The ADR argues that source code structure is unstable, so the contract's location shouldn't be tied to it. That's fair for general implementation packages. But service packages are named after domain concepts (`modulerelease`, `mandatorymodule`), not implementation details. They're as stable as the domain model itself. So if the domain changes significantly enough to rename a service package, updating the adjacent documentation is exactly the right thing to do.

Also keeping the format lightweight lowers the barrier for contributors to update it when behavior changes, which is the most important property a living document can have. A per-package markdown does not prescribe a fixed format. Format can be chosen where it fits best, if a table is needed or prose is up to the complexity. If a particular behavior is complex enough that a structured table genuinely helps, use that.

One concern with per-package documentation is cross-cutting behavior. for example, `Default-Module-CR` handling touches multiple places across the manifest reconciliation. But this isn't a problem if we follow the service dependency tree. ADR 004 defines a clear hierarchy: the manifest service is the top-level service in that tree. If a behavior is orchestrated across several lower-level services, the `README.md` of the top-level service is the right place to document the end-to-end contract. Readers who need detail can follow the dependency chain downward. This mirrors how the code itself works, and colocation means they always know where to look.

Would you be open to exploring that direction? Happy to help draft an updated proposal.

docs/contributor/adr/009-maintain-system-behavioral-contract.md
Tomasz-Smelcerz-SAP: > However, seeing that KLM is moving towards a maintenance mode, I feel like the overhead of creating the contracts now and maintaining them in case of changes is too big an overhead. 

If the KLM is moving towards maintenance mode, the effort is minimal: it's just a matter of figuring out what the contract is, because today nobody knows. It's too scattered across GH issues, ADRs, and lines of code.
If the KLM is NOT moving towards maintenance mode, but is actively developed - which I think it is - it's even more important to have the contract somewhere. Because again, today we don't know ourselves what the contract is. And more importantly - we don't know what gaps and how many gaps we have.

> This will involve a lot of interteam communication.

Not at all. The contract is **just for us**. Nobody else is allowed to contribute to it even! Other stakeholders are free to contribute to ADRs. This contract is our internal doc.

> This might end up as something that we work on slowly, like the refactor of the declarative reconciler, which took 2 years to finish,

It may be the case. But if we don't even start, there's no hope to finish. I am trying to address the biggest PITA that we have today: `we don't know ourselves what the contract is. And more importantly - we don't know what gaps and how many gaps we have.`

>I am not against this but it is just food for thought that this undertaking might do more harm then good in the future.

I think it will make our life easier in the future. I am tired of always looking into the source code to find the answer "how it should work". Because I never know if the code is implemented **correctly**. The contract serves as a authoritative reference for how the code should work. And it's much simpler to read then a test suite.
The contract for handling Module CRs (with CRDs) will probably be 30 lines long. 30 lines. That's all it takes. And that's all the maintenance effort - keeping 30 lines up-to-date.
The test suites for the same functionality, including e2e will be measured in thousands of lines...

Tomasz-Smelcerz-SAP: >What do you think of not including pending decisions in the document? I have an intuition that it can be easily misinterpreted by an AI agent 

But the contract is primarily for us! I want a mechanism to document missing pieces of the contract. 
We may have a different document for that, fine. But - at least for me - it's easier to take a look at both list: "known decisions" and "unknown decisions" at the same time to realize what the shape of the contract, overall, is. Otherwise, if I have 20 contract items (decided) in the document, and 200 contract items (undecided/unknown) somewhere else, I may have a false feeling that the contract is sound. But it's not.
Having that visible, in the painful way, will motivate us to fix the problem.
To use an analogy: If there's a left engine failure in an airplane, it's better to have it exposed so that "left engine failure" status light is even MORE visible than "right engine OK" status light.
Using this logic, perhaps we should have the "Pending decisions" section listed FIRST.
Because this section corresponds to a "failing engine" of our system, because this is probably where our system breaks.

>Similarly, we might skip the decision references, as it does not add any value to an agent and just fills up the context unnecessarily.

I disagree. The contract is not created out-of-thin-air. Every contract item must be backed up by some formal decision. It may be ADR, it may be Decision Log in the community repo... Finally it may be some Epic/Issue in Lifecycle-Manager repository. These resources provide more context and explanation for the contract item. The single line in the proposed document is just like an entry in the book index. It clearly shows that the item exists and tries to "name" the item in a reasonable way. The whole story behind the item is somewhere else - not in the contract document! This link must be "trackable".

Tomasz-Smelcerz-SAP:
>Formal IDs like R010 seem arbitray. 

So is ADR-007

>Also documenting Pending Decisions, committing open questions and gaps to the repo sounds useful, but in practice these accumulate and never get resolved. 

First: We don't have that practice yet, because we don't have the contract. What we observe is that GitHub issues tend to accumulate and never got resolved - that's why I wanted to avoid these.
Second: It will never be resolved until it's realized. We need to realize that we have open gaps in order to fix them.
When I am looking at list of items and I find some gap, it's much simpler to put the missing thing just near that particular item that made me realize the gap in the first place, as they are likely related.
I can put it in some other document, but the further it is, the harder it is later to "build the whole context" in one's head.

> Documentation should describe behavior that is decided and in effect; gaps and open questions belong in GitHub issues imo where they can be tracked and closed.

I simply don't agree. People were/are creating technical docs on different topics (like aerospace devices or vodka distillation) without having GH account. They just put the "missing pieces" alongside in their docs. Perhaps the most famous example is: "I have discovered a truly marvelous proof of this, which this margin is too narrow to contain." :)
Furthermore: When I want to drill a hole in the wall, I put the markings FIRST - they mark the point that I will drill in the future. This is my documentation. I am documenting gaps in my implementation. Everyone who has ever hanged a wall cabinet does this - start with documenting gaps. Holes-to-be-drilled are gaps.
So to me, documenting missing parts of our contract is like scribing markings on the wall for future holes, that are not existing now, but they should. Once drilled, the hole itself documents the existing contract, so that part is for free in the cabinet-hanging-business :)

Having said that I am not dogmatic about it. It's better to have it written somewhere, even in some GitHub issue (which tends to be forgotten so easily) written in Swahili (which is less of a concern since we have LLM translators) than not have it at all.

> My idea would be: bring the contract to the service layer, not the other way around.

The other way around would be to bring the service layer to the docs, which I never implied.
But I get the point. You want the contract closer to the implementation.

> But service packages are named after domain concepts (modulerelease, mandatorymodule), not implementation details. They're as stable as the domain model itself. So if the domain changes significantly enough to rename a service package, updating the adjacent documentation is exactly the right thing to do.

This is where our point of view diverges. I think as the contract as something high level. So that it's strictly "above" any implementation. So that, one day, I can ask LLM to generate a Lua code as an implementation. Or manually re-write it in Haskell. Chances for Haskell are small. But for Rust? Who knows?!
Go is just one possible choice. Not the worst one, not the best one. I am not personally tied to this choice in any particular way. Putting the contract into Go source packages couples it to the choice of using Go as an implementation - this is the part that I don't like. But I can accept that as a beareable solution.

> Would you be open to exploring that direction? Happy to help draft an updated proposal.

Yes, I am more interested in having something that rather nothing.

Tomasz-Smelcerz-SAP:
Team's decision:

We adopt the idea with the following remarks:

- Contract stored in the Service Layer (MD files)
- Let's start with documenting Module CR contract (Module-Default-CR, Module-CRD)
- Let's not decide on the format at this moment
- There is a summary of Gaps (let's not call it "Pending Decision")

LeelaChacha:
@Tomasz-Smelcerz-SAP Please rename to ADR 010

lindnerby (approved):

Issue #3493, opened 2026-07-31
Replace URL-scheme detection for OCI insecure layer pull with explicit startup wiring


medmes:
### Description

The current `pullLayer` function in `internal/manifest/img/pathextractor.go` decides whether to use plain HTTP (`crane.Insecure`) or HTTPS at **runtime** by checking whether the image reference starts with `http://`:

```go
isInsecureLayer, err := regexp.MatchString("^http://", imageRef)
if isInsecureLayer {
    crane.PullLayer(ref, crane.Insecure, ...)
}
```

This was surfaced during the investigation of #3471. The `crane.Insecure` option is currently used in local k3d test environments (e.g. `--oci-registry-host=http://k3d-kcp-registry.localhost:5000`) and is unreachable in production.

The problem: the transport-security decision is implicit, buried inside a string-parsing heuristic rather than an explicit configuration made at the composition root. This is hard to audit and easy to accidentally break in a future refactor.

Two options to evaluate:

1. **Startup wiring** — wire the secure/insecure choice as a startup decision in `cmd/main.go` (e.g. via `--oci-registry-insecure` flag), and inject it into `PathExtractor` via the existing `WithLayerPuller` option rather than detecting it from the URL.
2. **Eliminate the insecure path** — investigate whether the k3d test registry can be equipped with a self-signed certificate so the `http://` code path can be removed entirely. The `crane.Insecure` docstring hints at self-signed TLS support.

### Reasons

- The URL-scheme heuristic couples the security decision to string parsing, making it invisible at the composition root and hard to audit.
- The `crane.Insecure` option exposes authentication credentials over plaintext if the `http://` path is ever accidentally reached in a non-test environment.
- Explicit startup wiring follows ADR-002 (inject dependencies via constructor) and makes the security posture of the binary unambiguous.

### Acceptance Criteria

- [ ] The decision to use `crane.Insecure` is no longer driven by URL-scheme detection at call time
- [ ] Either: the insecure/secure choice is wired explicitly at startup (composition root in `cmd/main.go`) and injected via `WithLayerPuller`
- [ ] Or: the k3d test registry is equipped with TLS and the `crane.Insecure` / `http://` code path is removed entirely
- [ ] Local k3d test environments continue to function with whichever approach is chosen
- [ ] The selected approach is documented in `pullLayer`

### Feature Testing

- Unit tests
- Integration tests

### Testing approach

[...]

PR #3494, opened 2026-07-31
feat: Add explicit OCI digest re-verification before returning cached manifest path

### LINKED ISSUE #3471: Add explicit OCI digest re-verification before returning cached manifest path
### Description

In `internal/manifest/img/pathextractor.go`, `GetPathForFetchedLayer` returns the cached manifest path immediately when the file already exists on disk, without re-verifying the OCI digest:

```go
if dir != nil {
    return manifestPath, nil // returns cached path without re-verifying digest
}
```

While the current production setup mitigates practical exploitation (the cache path itself already encodes the SHA-256 content digest, and writing to `/tmp` inside the pod requires a prior container compromise), this is a defense-in-depth gap: there is no explicit integrity check at the point of use. A security audit ([HASI2026165-44](https://securityjira.wbs.only.sap/browse/HASI2026165-44)) flagged this as a pattern.

---

In general, we need to verify the verification of the OCI digest. In `internal/manifest/img/pathextractor.go`, we don't to any explicit verification. Verify whether this is implicitly handled by the layer pull using `crane` library. If so, add a short doc accordingly.

---

Also, we have separate code paths for `secure` and `insecure` layer pulling: https://github.com/kyma-project/lifecycle-manager/blob/bbdcf03f9f959b75801b5a703e889d47c6d5362a/internal/manifest/img/pathextractor.go#L179-L199
We need to understand what this means and also document that accordingly. If `insecure` implies security risks for production, we need to think about dropping this.

### Reasons

Adding an explicit digest re-verification before the early return closes the defense-in-depth gap with minimal effort. It makes the integrity guarantee explicit in code rather than relying on the implicit property that the cache path encodes the digest, which is easy to break accidentally in a future refactor.

### Acceptance Criteria

- [x] Before returning a cached manifest path, `GetPathForFetchedLayer` verifies that the on-disk content's digest matches the expected digest from `imageSpec.Ref`
- [x] If the digest does not match (tampered or corrupted file), the cached file is deleted and re-fetched from the OCI registry
- [x] Existing unit tests in `pathextractor_test.go` continue to pass
- [x] A unit test covering the tampered-cache scenario is added
- [x] Verified the OCI digest verification on initial layer pull by `crane` is happening
- [ ] Understood and documented the `secure` vs `insecure` layer pull and documented it accrdingly
  - [ ] created follow-ups if necessary

### Feature Testing

- Unit tests

### Testing approach

Add a unit test to `pathextractor_test.go` that writes a tampered file to the expected cache path before calling `GetPathForFetchedLayer`, and asserts that the fresh content from the registry is returned instead of the tampered file.

medmes:
**Description**

- Added explicit integrity verification for cached OCI manifest files in `GetPathForFetchedLayer`: on every cache hit a SHA-256 digest file (`.sha256`) is read and compared against the on-disk content; a mismatch evicts both files and triggers a fresh pull
- Stored the SHA-256 of the uncompressed manifest content in a digest file during the initial pull (computed via `io.MultiWriter` in a single pass, no extra I/O)
- Fixed a file handle leak: `os.Open` result was never closed on the cache-hit path
- Fixed a partial-file leak: on `io.Copy` failure the manifest file is now closed and removed before returning the error
- Made `PullLayer` injectable via direct constructor parameter on `PathExtractor` per ADR-002 (composition root wires the dependency explicitly); `PullLayer` is exported as the default implementation
- Documented crane's implicit OCI digest verification (lazy, triggered at EOF of the compressed stream by `verify.ReadCloser`) in `PullLayer`
- Documented the secure vs insecure layer pull: the `http://` / `crane.Insecure` path is not reached in production but is intentionally used by local k3d test environments; `crane.WithContext` is now passed consistently on the insecure path; follow-up tracked in #3493
- Added four unit tests covering: tampered cache (no digest file), digest mismatch (digest file present but content changed), valid cache reuse, and `io.Copy` failure cleanup — all using a fake `v1.Layer` with no network access

**Related issue(s)**

Resolves #3471


internal/manifest/img/pathextractor.go
```diff
@@ -30,12 +34,32 @@ var (
 	ErrInvalidArchiveStructure = errors.New("tar archive has invalid structure, expected a single file")
 )
 
+type layerPullerFunc func(
+	ctx context.Context, imageRef string, keyChain authn.Keychain,
+) (containerregistryv1.Layer, error)
+
+// PathExtractorOption configures a PathExtractor. Intended for testing.
+type PathExtractorOption func(*PathExtractor)
```
LeelaChacha: YAGNI. Introducing this complicates the PathExtractor unnecessarily. Why not just have layerPullFunc be constructor injected?

internal/manifest/img/pathextractor.go
```diff
@@ -85,40 +109,61 @@ func (p PathExtractor) GetPathForFetchedLayer(ctx context.Context,
...
-	if err != nil {
-		return manifestPath, fmt.Errorf("failed to close io: %w", err)
+	if err := io.Closer(outFile).Close(); err != nil {
+		return fmt.Errorf("failed to close io: %w", err)
 	}
-	return manifestPath, nil
+
+	digest := hex.EncodeToString(hasher.Sum(nil))
+	if err := os.WriteFile(sidecarDigestPath(manifestPath), []byte(digest), manifestFileMode); err != nil {
+		return fmt.Errorf("failed to write digest sidecar for layer %s: %w", imageRef, err)
```
LeelaChacha: sidecar is unconventional naming for the digest metadata. it is okay if you want to call it that in code but using that in error messages probably not a good idea

internal/manifest/img/pathextractor_test.go
```diff
@@ -115,6 +120,109 @@ func TestPathExtractor_FetchLayerToFile(t *testing.T) {
 	}
 }
 
+// TestPathExtractor_GetPathForFetchedLayer_TamperedCacheIsEvictedAndRefetched verifies
+// the defense-in-depth guarantee: a tampered cached manifest (no matching sidecar digest)
+// is evicted and replaced with freshly fetched content.
+func TestPathExtractor_GetPathForFetchedLayer_TamperedCacheIsEvictedAndRefetched(t *testing.T) {
```
LeelaChacha, line 129: Could you do me a favor and add this package to the unit-test-coverage.yaml file. The new code is at 64% coverage. The main gaps: the digest-mismatch eviction path (sidecar exists but content changed - kind of the whole point of the feature) isn't tested, and neither is the writeLayerToDisk error handling, especially the io.Copy partial-file cleanup you added in the review-fix commit. Both are mockable in-process via WithLayerPuller, so no registry needed

LeelaChacha (changes requested):

LeelaChacha (approved):
LGTM

Issue #3299, opened 2026-05-22
Drop the Purge Reconciler


c-pius:
### Description

Drop the purge reconciler and remove the purge finalizer from all Kymas. Options to be discussed:

- drop reconciler and have SRE run a script that removes the finalizer from all Kymas
- ~~two phases where first phase drops the regular purge logic and adding code to the purge reconciler dropping the finalizer if present; second phase dropping the purge reconciler entirely~~
- ~~two phases where first phase drops the purge reconciler entirely and somewhere else (Kyma reconciler) the purge finalizer is dropped; second phase where the purge finalizer dropping is removed~~

Going for option 1 as in restricted markets, we don't have full control of the update behavior and operators there may skip the version of lifecycle-manager that would removed the orphaned finalizer.

### Reasons

See https://github.com/kyma-project/community/issues/1023. It was aligned that Lifecycle-Manager does NOT purge resources anymore.

>[!Note]
>It is also unclear if it is even work as expected as of now. Filtering for `context.controller: Purge` in CLS, we only see the 5m requeue and reconciler error messages.

### AC

- [x] purge reconciler gone
- [x] purge-related code dropped from Kyma reconciler if any
- [x] purge finalizer gone from all Kymas
  - [x] create the shell script for dropping the finalizer and add a notable change to inform the restricted operator accordingly. Note explicitly that the script needs to be run directly after updating the KLM version as otherwise Kymas would be stuck in deletion. Consider to do a sanity check in the beginning of the script so that it applies only if the KLM version has been updated to the expected version or greater.
- [x] purge-related flags deprecated
- [x] purge-related documentation dropped
- [x] purge-related related metrics and dashboard visualization dropped

amritanshusikdar:
Because of the unreliability of the Purge Controller, and the not-so-harmonic collaboration with our new deletion service, [the integration tests were forced to run with the legacy deletion service](https://github.com/kyma-project/lifecycle-manager/pull/3504).

After this ticket has been implemented, make sure to _(additional ACs)_:

1. Adjust the existing integration tests to check if they're working reliably (as little flakiness as possible).
2. Create a new PR to remove the `ENABLE_LEGACY_KYMA_DELETION=true` flag from the integration tests.
---
Original idea by @c-pius:
[...]

PR #3501, opened 2026-08-03
feat: Re-align Kyma Deletion Controller with env Variable

### LINKED ISSUE #3298: Re-Align the Deletion Handling in Kyma Controller
### Description

In the [Kyma Controller](internal/controller/kyma/controller.go), re-align the deletion handling (#174-192). As of now, `handleDeletedSkr` is called when there is no SKR client and no SKR secret anymore. If the SKR Context fails to initialize, also check if the `ENABLE_LEGACY_KYMA_DELETION` varaible is set. If so, delegate to `processDeletion`. If unset, continue to use `handleDeletedSkr`.

Once the Kyma Deletion Service proofed to be working on the live environments, the env var check can be removed and with it the "legacy" implementations of the Kyma Deleting. The first step should look something like:

```go
err = r.SkrContextFactory.Init(ctx, kyma.GetNamespacedName())
if !kyma.DeletionTimestamp.IsZero() && errors.Is(err, accessmanager.ErrAccessSecretNotFound) {
    envValue, isDefined := os.LookupEnv("ENABLE_LEGACY_KYMA_DELETION")
    useLegacyKymaDeletion := isDefined && envValue == "true"
    if !useLegacyKymaDeletion {
        return r.processDeletion(ctx, kyma)
    }
    return r.handleDeletedSkr(ctx, req, kyma)
}

skrContext, err := r.SkrContextFactory.Get(kyma.GetNamespacedName())
if err != nil {
    r.Metrics.RecordRequeueReason(metrics.SyncContextRetrieval, queue.UnexpectedRequeue)
    setModuleStatusesToError(kyma, err.Error())
    return ctrl.Result{}, r.updateStatusWithError(ctx, kyma, err)
}

if !kyma.DeletionTimestamp.IsZero() {
    envValue, isDefined := os.LookupEnv("ENABLE_LEGACY_KYMA_DELETION")
    useLegacyKymaDeletion := isDefined && envValue == "true"
    if !useLegacyKymaDeletion {
        return r.processDeletion(ctx, kyma)
    }
}
```

Once the legacy processing can be dropped, it can be simplified to:

```go
// err from Init not handled explicitly as the consecutive processing uses the following errors from the client cache
_ = r.SkrContextFactory.Init(ctx, kyma.GetNamespacedName())
if !kyma.DeletionTimestamp.IsZero() {
    return r.processDeletion(ctx, kyma)
}

skrContext, err := r.SkrContextFactory.Get(kyma.GetNamespacedName())
if err != nil {
    r.Metrics.RecordRequeueReason(metrics.SyncContextRetrieval, queue.UnexpectedRequeue)
    setModuleStatusesToError(kyma, err.Error())
    return ctrl.Result{}, r.updateStatusWithError(ctx, kyma, err)
}
```

>[!Note]
> In `handleDeletedSkr`, there are explicit metric records. They don't seem to be entirely reasonable, but check explicitly if those need to be ported to `processDeletion`. => do a sanity check on the metrics (e.g. we also have a record on `metrics.SyncContextRetrieval` when receiving an error from `skrContext.CreateKymaNamespace(ctx)`
> Also, there is an `updateKyma` call which also doesn't seem to be necessary. Again, check explicitly if it needs to be ported. => do a sanity check if any other need for update Kyma is there other than removing the finalizers

>[!Caution]
>In `handleDeletedSkr` there is a `removeAllFinalizers` call. In the Kyma Deletion Service, only the Kyma finalizer is removed. The `purge` finalizer may still be present. We need to decide when to drop the Purge controller and how we proceed with the existing purge finalizers. => this will be handled in #3299. ‼️ the refactoring only becomes functional once #3299 is also implemented

### Reasons

Kyma Deletion Service also handles the "deleted SKR" case. For the transition period, we still want to easily switch back to legacy processing.

### AC

- [x] `ENABLE_LEGACY_KYMA_DELETION` controls legacy deletion processing vs. kyma deletion service processing for both, regular deletion and deletion when access to the SKR is gone
- [x] confirmed explicitly whether the metrics and `updateKyma` calls are reasonable and should be ported to the kyma deletion service processing
- [x] explicitly decided on the purge finalizer handling

amritanshusikdar:
**Description**
- Extends `ENABLE_LEGACY_KYMA_DELETION` guard to cover the `ErrAccessSecretNotFound` path, routing to `processDeletion()` upon enablement.

**Related Issues**
Resolves #3298

### DIFF
--- a/tests/integration/controller/kcp/remote_sync_test.go
+++ b/tests/integration/controller/kcp/remote_sync_test.go
@@ -316,18 +316,27 @@ var _ = Describe("Kyma sync default module list into Remote Cluster", Ordered, f
 	})
 
 	It("Default module list should be recreated if remote Kyma gets deleted", func() {
+		By("Wait for KCP Kyma to reach Ready state after previous spec")
+		Eventually(func() bool {
+			kcpKyma, err := GetKyma(ctx, kcpClient, kyma.GetName(), kyma.GetNamespace())
+			if err != nil {
+				return false
+			}
+			return kcpKyma.Status.State == shared.StateReady
+		}, Timeout, Interval).Should(BeTrue())
+
 		By("Delete remote Kyma")
 		Eventually(DeleteCR, Timeout, Interval).
 			WithContext(ctx).
 			WithArguments(skrClient, skrKyma).Should(Succeed())
 		By("Remote Kyma contains default module")
-		Eventually(ContainsModuleInSpec, Timeout, Interval).
+		Eventually(ContainsModuleInSpec, 2*Timeout, Interval).
 			WithContext(ctx).
 			WithArguments(skrClient, skrKyma.Name, skrKyma.Namespace, moduleInKCP.Name).
 			Should(Succeed())
 
 		By("KCP Manifest is being created")
-		Eventually(ManifestExists, Timeout, Interval).
+		Eventually(ManifestExists, 2*Timeout, Interval).
 			WithContext(ctx).
 			WithArguments(kcpClient, kyma.GetName(), kyma.GetNamespace(), moduleInKCP.Name).
 			Should(Succeed())
--- a/tests/integration/controller/kcp/remote_sync_test.go
+++ b/tests/integration/controller/kcp/remote_sync_test.go
@@ -329,13 +329,13 @@ var _ = Describe("Kyma sync default module list into Remote Cluster", Ordered, f
 			WithContext(ctx).
 			WithArguments(skrClient, skrKyma).Should(Succeed())
 		By("Remote Kyma contains default module")
-		Eventually(ContainsModuleInSpec, Timeout, Interval).
+		Eventually(ContainsModuleInSpec, 2*Timeout, Interval).
 			WithContext(ctx).
 			WithArguments(skrClient, skrKyma.Name, skrKyma.Namespace, moduleInKCP.Name).
 			Should(Succeed())
 
 		By("KCP Manifest is being created")
-		Eventually(ManifestExists, Timeout, Interval).
+		Eventually(ManifestExists, 2*Timeout, Interval).
 			WithContext(ctx).
 			WithArguments(kcpClient, kyma.GetName(), kyma.GetNamespace(), moduleInKCP.Name).
 			Should(Succeed())
--- a/tests/integration/controller/kcp/remote_sync_test.go
+++ b/tests/integration/controller/kcp/remote_sync_test.go
@@ -10,13 +10,15 @@ import (
 
 	"github.com/kyma-project/lifecycle-manager/api/shared"
 	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+
 	"github.com/kyma-project/lifecycle-manager/internal/descriptor/types/ocmidentity"
 	"github.com/kyma-project/lifecycle-manager/internal/pkg/flags"
 	"github.com/kyma-project/lifecycle-manager/pkg/testutils/builder"
 
-	. "github.com/kyma-project/lifecycle-manager/pkg/testutils"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+
+	. "github.com/kyma-project/lifecycle-manager/pkg/testutils"
 )
 
 var ErrNotContainsExpectedCondition = errors.New("kyma CR does not contain expected condition")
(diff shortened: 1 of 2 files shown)


amritanshusikdar:
Note for reviewer (reasoning for no test changes): The new `ErrAccessSecretNotFound` deletion path is already covered by the existing integration and e2e deprovision tests since `ENABLE_LEGACY_KYMA_DELETION` is never set in those environments; meaning they've been running through processDeletion all along. The branching logic itself is too simple to deserve a dedicated test for a mechanism that's explicitly designed to be removed once #3299 lands.

tests/integration/controller/kcp/remote_sync_test.go
amritanshusikdar, line 1: Although unrelated, this was needed to be added in order to make the integration tests pass. Without it just won't pass, no matter what. Funnily enough it worked every time locally, without any of these changes. But the pipeline is very adamant about it.

These values have been inflated, just to make the CI pass. I suggest we investigate the cause of this first.

internal/service/kyma/deletion/usecases/set_skr_kyma_state_deleting.go
```diff
@@ -36,7 +36,10 @@ func (u *SetSkrKymaStateDeleting) IsApplicable(ctx context.Context, kcpKyma *v1b
 
 	status, err := u.skrKymaStatusRepo.Get(ctx, kcpKyma.GetNamespacedName())
 
-	if errors.Is(err, accessmanager.ErrAccessSecretNotFound) || util.IsNotFound(err) {
+	skrUnreachable := errors.Is(err, accessmanager.ErrAccessSecretNotFound) ||
+		util.IsNotFound(err) ||
+		util.IsConnectionRelatedError(err)
+	if skrUnreachable {
```
LeelaChacha: I have been looking at the processDeletion() method, and something comes off weird to me.
Since this deletion service is supposed to be the new way to handle deleted SKRs, we need to make sure all errors related to deleted SKRs are covered by this so that it can actually be a viable replacement for the legacy deletion method. 
Currently, I see a problem and would like your opinion on it since you know more about the mechanism. Let's assume there is an SKR where the kubeconfig secret is gone and we're deleting it using the new deletion service. `SkrContextFactory.Init` returns `ErrAccessSecretNotFound` so we go into `r.processDeletion`.  in the second usecase of the deletion service, `SetSkrKymaStateDeleting.IsApplicable` uses the `skrClientRetrieverFunc` of the repo to get SKR kyma status. Which will return `errorsinternal.ErrSkrClientNotFound`. This is not considered in `skrUnreachable` bool here, so the bool stays false and we return an err and requeue. Wouldn't this cause a deadlock?

This makes me a bit uncomfortable and takes away my confidence in approving it.

LeelaChacha: If this is true, there might be more such cases which we are overseeing and therefore the deletion service would not be a good replacement for the legacy deletion func

internal/service/kyma/deletion/usecases/delete_skr_crd.go
```diff
@@ -36,7 +37,7 @@ func (u *DeleteSkrCrd) IsApplicable(ctx context.Context, kcpKyma *v1beta2.Kyma)
 
 	exists, err := u.skrCrdRepo.Exists(ctx, kcpKyma.GetNamespacedName())
 
-	if errors.Is(err, accessmanager.ErrAccessSecretNotFound) {
+	if errors.Is(err, accessmanager.ErrAccessSecretNotFound) || util.IsConnectionRelatedError(err) {
```
LeelaChacha: The DeleteSkrWebhookResources use case would probably also need connection errors resilience since it has SKR side operations, mind also adding this there?

internal/service/kyma/deletion/usecases/set_skr_kyma_state_deleting.go
amritanshusikdar: Hmm good catch. That's definitely a problem. This specific one has been fixed right now. Since I didn't implement the new deletion service mechanism myself, I cannot guarantee at this moment whether all the scenarios covered by the legacy is also covered by the new mechanism. I will, however, take the time today to go through and manually verify as much as possible.

amritanshusikdar:
Ok turns out there was a mix-up of which errors to consider inside the `IsApplicable()` of each SKR-related use case. So in view of [this update](https://github.com/kyma-project/lifecycle-manager/issues/3296#issuecomment-5201821548), I'll create a new PR as part of an [older issue](https://github.com/kyma-project/lifecycle-manager/issues/3296). These changes are outside the scope of this PR.

amritanshusikdar:
Here's the new PR: https://github.com/kyma-project/lifecycle-manager/pull/3507

All the necessary cases have been taken care of here.

c-pius:
## Deletion Service

### Deleting when SKR secret exists

All uses cases execute

image

### Deleting when SKR secret does NOT exist

All use cases execute because the client is cached

image

### Deleting when SKR secret does NOT exist and KLM does NOT have it cached (restarted KLM after deleting the secret)

The SKR use cases don't execute ✅ 
The EnsureManifestDeletion Use Case is not listed ❌ => this needs to be checked. Suspicion: the Manifest is already gone before the use case is invoked. Confirmed via debug log, use case is invoked but no manifest existed anymore. ✅ 

image


image


## Legacy Deletion

### Deleting when SKR secret exists

Manifest deletes in a few seconds, Kyma is pending for the purge finalizer removal. Purge finalizer dropped after timeout, Kyma gone.

### Deleting when SKR secret does NOT exist

Kyma and manifest delete in a few seconds.

### Deleting when SKR secret does NOT exist and KLM does NOT have it cached (restarted KLM after deleting the secret)

Kyma and manifest delete almost instantly

c-pius (approved):

amritanshusikdar:
> ## Deletion Service
> The SKR use cases don't execute ✅ The EnsureManifestDeletion Use Case is not listed ❌ => this needs to be checked. Suspicion: the Manifest is already gone before the use case is invoked. Confirmed via debug log, use case is invoked but no manifest existed anymore. ✅

Yeah I checked it manually, it does come up. Just had to set a custom temp finalizer on the Manifest, then it had enough time to trigger the `EnsureManifestDeletion` use case. It works, basically.

PR #3504, opened 2026-08-04
fix: Use ENABLE_LEGACY_KYMA_DELETION for Integration Tests

amritanshusikdar:
**Description**
Currently with the new setup including the new deletion service & unreliable purge controller, the existing integration tests are not being able to run on the GitHub CI pipeline. They pass at times locally (still flaky), but on the pipeline they skew more towards failing than passing.

Later on, when the new deletion service is all polished and ready to work, then we shall adjust the integration tests to work with it. Till then we're forcing the tests to use the old deletion method, which is also what our landscapes currently do.

Suggested Changes:
- added `ENABLE_LEGACY_KYMA_DELETION=true` to integration tests


amritanshusikdar:
After running the integration tests 16 consecutive successful times, with the legacy kyma deletion flag enabled, it's safe to go this route.

image

c-pius (approved):

PR #3507, opened 2026-08-06
fix: Replace `ErrAccessSecretNotFound` checks with `ErrSkrClientNotFound`

### LINKED ISSUE #3296: Drop explicit checks for SKR Kubeconfig Secrets

amritanshusikdar:
**Description**

There was a tiny mix-up in the description of the [original issue](https://github.com/kyma-project/lifecycle-manager/issues/3296) about which exact errors to check for and handle.

Context comment: https://github.com/kyma-project/lifecycle-manager/issues/3296#issuecomment-5201821548

- replace `ErrAccessSecretNotFound` checks with `ErrSkrClientNotFound`
- update the unit tests

**Related issue(s)**
Fixes #3296 and extends #3457

### DIFF
--- a/internal/service/kyma/deletion/usecases/delete_skr_crd.go
+++ b/internal/service/kyma/deletion/usecases/delete_skr_crd.go
@@ -7,8 +7,9 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
+	errorsinternal "github.com/kyma-project/lifecycle-manager/internal/errors"
 	"github.com/kyma-project/lifecycle-manager/internal/result"
-	"github.com/kyma-project/lifecycle-manager/internal/service/accessmanager"
+	"github.com/kyma-project/lifecycle-manager/pkg/util"
 )
 
 //nolint:iface // we accept the duplication for clarity
@@ -36,7 +37,7 @@ func (u *DeleteSkrCrd) IsApplicable(ctx context.Context, kcpKyma *v1beta2.Kyma) 
 
 	exists, err := u.skrCrdRepo.Exists(ctx, kcpKyma.GetNamespacedName())
 
-	if errors.Is(err, accessmanager.ErrAccessSecretNotFound) {
+	if errors.Is(err, errorsinternal.ErrSkrClientNotFound) || util.IsConnectionRelatedError(err) {
 		return false, nil
 	}
 
--- a/internal/service/kyma/deletion/usecases/delete_skr_crd.go
+++ b/internal/service/kyma/deletion/usecases/delete_skr_crd.go
@@ -9,7 +9,6 @@ import (
 	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
 	errorsinternal "github.com/kyma-project/lifecycle-manager/internal/errors"
 	"github.com/kyma-project/lifecycle-manager/internal/result"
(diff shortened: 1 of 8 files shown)


internal/service/kyma/deletion/usecases/delete_skr_crd.go
```diff
@@ -36,7 +37,7 @@ func (u *DeleteSkrCrd) IsApplicable(ctx context.Context, kcpKyma *v1beta2.Kyma)
 
 	exists, err := u.skrCrdRepo.Exists(ctx, kcpKyma.GetNamespacedName())
 
-	if errors.Is(err, accessmanager.ErrAccessSecretNotFound) {
+	if errors.Is(err, errorsinternal.ErrSkrClientNotFound) || util.IsConnectionRelatedError(err) {
```
c-pius: We explicitly don't want to check the `util.IsConnectionRelatedError(err)` here and in the other SKR use cases. This would "swallow" the error and we won't evict cache on transient network related issues any more.

We want such errors to bubble up to the controller. The controller checks if a connection related error is returned from the service. If so, it evicts it from the cache and requeues the Kyma. In the next reconciliation, the client cache will try to re-build the client. If it was a transient error, and we still have the skr secret, the client will be re-built and the check will work again in this time. If the skr secret is gone, we can't re-build the client and consecutive use case calls return the `errorsinternal.ErrSkrClientNotFound`.

c-pius (changes requested):

c-pius (approved):

PR #3534, opened 2026-08-12
fix: Reject enabling mandatory modules in Kyma spec

### LINKED ISSUE #3517: Enabling a mandatory module in SKR Kyma Spec lists it in status
### Description

Enabling a mandatory module in the SKR Kyma Spec lists this module in the Kyma status.modules. It should not be possible to enable mandatory modules. 
**Expected** behavior: The Kyma should rather go to Warning with "no such module" or an explicit message that mandatory modules cannot be enabled.

### Steps to reproduce

- enable Kim-snatch or warden in SKR Kyma Spec
- observe that the module is included in the status.modules

### Acceptance Criteria

- [x] Issue is fixed towards expected behavior and troubleshooting guides are updated

### Environment Type

Managed

### Environment Info

Kubernetes Version: x.y.z
Docker Version: x.y.z
modulectl Version: x.y.z

### Attachments

See comment https://github.com/kyma-project/lifecycle-manager/issues/3517#issuecomment-5239204906

LeelaChacha:
## Description

Enabling a mandatory module (e.g. `kim-snatch`, `warden`) in the Kyma spec previously resolved it like a regular module and listed it in `status.modules` as `Ready`. Mandatory modules are installed by a dedicated controller and must not be user-enableable.

The enabled module now surfaces a `Warning` module status with the message `mandatory module cannot be enabled`, which drives the Kyma to `Warning`. The guard keys on the `ModuleReleaseMeta` (where mandatory-ness is declared via `spec.mandatory`); the previous guard only checked the `ModuleTemplate` flag, which real mandatory modules leave unset. Also fixes a latent case where a warning-class error mapped to `Error` when the module had no prior status entry.

Covered by unit tests (template lookup + status generation) and an integration test that enables a mandatory module and asserts the Warning state.

Closes #3517


medmes (commented):
*(see inline comments below)*

internal/service/kyma/status/modules/generator/fromerror/generator.go
```diff
@@ -19,34 +19,32 @@ func GenerateModuleStatusFromError(err error, moduleName, desiredChannel, ocmCom
...
+			newModuleStatus.Message = err.Error()
+			newModuleStatus.State = shared.StateWarning
+			return newModuleStatus, nil
+		}
 	}
 
 	newStatus := newDefaultErrorStatus(moduleName, desiredChannel, ocmComponentName, err)
 
-	if errorIsTemplateNotFound(err) {
+	if errorIsTemplateNotFound(err) || errorIsMandatoryModuleEnabled(err) {
```
medmes: This refactoring also changes behavior for the pre-existing `ErrNoTemplatesInListResult` case: `nil status + template-not-found` was StateError before, is StateWarning now. The new `WhenCalledWithWarningClassError` table test only covers non-nil status. Worth adding a nil-status row there too.

pkg/templatelookup/regular.go
```diff
@@ -157,8 +164,7 @@ func (t *TemplateLookup) validateTemplateMode(template ModuleTemplateInfo,
 		return template
 	}
 	if template.Spec.Mandatory {
```
medmes: Is this path still hit in production, or has MRM `spec.mandatory` fully replaced the MT-flag? If it's dead, worth a follow-up to clean it up.

internal/service/kyma/status/modules/generator/fromerror/generator.go
medmes: This refactoring also changes behavior for the pre-existing `ErrNoTemplatesInListResult` case: `nil status + template-not-found` was StateError before, is StateWarning now. The new `WhenCalledWithWarningClassError` table test only covers non-nil status — worth adding a nil-status row there too.

internal/service/kyma/status/modules/generator/fromerror/generator.go
```diff
@@ -70,6 +68,10 @@ func errorIsTemplateNotFound(err error) bool {
 	return errors.Is(err, common.ErrNoTemplatesInListResult)
```
medmes, line 68: This refactoring also changes behavior for the pre-existing `ErrNoTemplatesInListResult` case: `nil status + template-not-found` was StateError before, is StateWarning now. The new `WhenCalledWithWarningClassError` table test only covers non-nil status — worth adding a nil-status row there too.

pkg/templatelookup/regular.go
```diff
@@ -157,8 +164,7 @@ func (t *TemplateLookup) validateTemplateMode(template ModuleTemplateInfo,
...
-		template.Err = fmt.Errorf("%w: for module %s in channel %s ",
-			common.ErrNoTemplatesInListResult, template.Name, template.DesiredChannel)
+		template.Err = fmt.Errorf("%w: %q", ErrMandatoryModuleCannotBeEnabled, template.Name)
 		return template
 	}
```
medmes, line 169: Is this MT-flag path still hit in production, or has MRM `spec.mandatory` fully replaced it? If it's dead, worth a follow-up to clean it up.

LeelaChacha: The `template.Spec.Mandatory` check pre-dates this PR and is unrelated to the fix. Removing it is a separate cleanup that should be taken care of in the deprecating ModuleTemplate epic

medmes (approved):

PR #3535, opened 2026-08-13
fix: Module Manifest Management Sync

### LINKED ISSUE #3470: Support re-managing a module even though Manifest was not deleted
### Description

In the situation described in `/kyma/backlog/issues/10122`, the user re-managed the module before it was fully unmanaged (Manifest gone) as it was stuck in dropping labels.

For the unmanaging, Lifecycle Manager sets the `operator.kyma-project.io/is-unmanaged: "true"` annotation on the Manifest. Given this annotation, the Manifest reconciler short-circuits. On installation, it ensures the `managed-by` labels are removed and then triggers the uninstallation. On uninstallation, it drops all finalizers to get the manifest deleted eventually.

Introduce a way so that KLM can properly handle the situation where the customer re-manages the module before the Manifest was gone

### Reasons

There is currently no way for KLM to drop this annotation again in between the un- and re-managing if the Manifest still exists. We need to handle this situation properly.

### Acceptance Criteria

- [x] re-managing the module works even if the Manifest was not already deleted
- [x] verified that there are no corner cases that would leave stuck resources or labels

### Feature Testing

_No response_

### Testing approach

_No response_

### Attachments

_No response_

amritanshusikdar:
**Description**

- When Manifest contains the `is-unmanaged=true` annotation && Module is unmanaged:
  - Setting Module back to managed, removes the annotation and restores Manifest reconciliation

**Related issue(s)**
Fixes #3470

### DIFF
--- a/pkg/module/sync/runner.go
+++ b/pkg/module/sync/runner.go
@@ -182,12 +182,8 @@ func (r *Runner) updateAvailableManifestSpec(ctx context.Context,
 func NeedToUpdate(manifestInCluster, newManifest *v1beta2.Manifest, moduleInStatus *v1beta2.ModuleStatus,
 	module *modulecommon.Module,
 ) bool {
-	if manifestInCluster == nil {
+	if manifestInCluster == nil || manifestInCluster.IsUnmanaged() {
 		return !(module.IsUnmanaged)
-	}
-
-	if manifestInCluster.IsUnmanaged() {
-		return !module.IsUnmanaged
 	}
 
 	if module.IsUnmanaged {
--- a/pkg/module/sync/runner.go
+++ b/pkg/module/sync/runner.go
@@ -187,7 +187,7 @@ func NeedToUpdate(manifestInCluster, newManifest *v1beta2.Manifest, moduleInStat
 	}
 
 	if manifestInCluster.IsUnmanaged() {
-		return false
+		return !module.IsUnmanaged
 	}
 
 	if module.IsUnmanaged {
--- a/pkg/module/sync/runner.go
+++ b/pkg/module/sync/runner.go
@@ -183,7 +183,7 @@ func NeedToUpdate(manifestInCluster, newManifest *v1beta2.Manifest, moduleInStat
 	module *modulecommon.Module,
 ) bool {
 	if manifestInCluster == nil || manifestInCluster.IsUnmanaged() {
-		return !(module.IsUnmanaged)
+		return !module.IsUnmanaged
 	}
 
 	if module.IsUnmanaged {
--- a/pkg/module/sync/runner_test.go
+++ b/pkg/module/sync/runner_test.go
@@ -179,7 +179,7 @@ func TestNeedToUpdate(t *testing.T) {
 			false,
 		},
 		{
-			"When cluster Manifest is unmanaged and module is managed, expect no update",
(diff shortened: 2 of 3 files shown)


pkg/module/sync/runner.go
```diff
@@ -187,7 +187,7 @@ func NeedToUpdate(manifestInCluster, newManifest *v1beta2.Manifest, moduleInStat
 	}
 
 	if manifestInCluster.IsUnmanaged() {
-		return false
```
c-pius, line 190: 4 lines higher it is `return !(module.IsUnmanaged)`, please align to one style.

c-pius (commented):
Have you considered the second AC?

Image

pkg/module/sync/runner.go
amritanshusikdar: I've adhered to the style above. I'll also merge these 2 guard clauses with an `||`.

amritanshusikdar: Actually, this is the only occurrence in the entire codebase and is not even appreciated by the official Go code styles. I'll revert back to my original style.

amritanshusikdar:
> Have you considered the second AC?

I definitely gave it a _best-effort_ shot but wasn't able to come across further issues locally. Upon asking Claude, here are 3 cases that could potentially be harmful:

1. **SSA patch on a terminating Manifest** — if label-removal-finalizer is already blocking deletion when the user re manages, patchManifest is called on an object with DeletionTimestamp set. SSA on a terminating object may fail silently, leaving the module permanently stuck.
2. **Partially stripped SKR labels not re-stamped** — label removal iterates status.synced and strips managed-by labels resource by resource. If re-manage happens mid-way, some SKR resources permanently lose their labels. The install path doesn't re-stamp them.
3. **Race between Kyma controller SSA and Manifest controller in-flight delete** — the Kyma controller removes is unmanaged via SSA, but the Manifest controller may have already read the stale object and be mid-way through handleUnmanagedManifest. The Delete call goes through despite the annotation being removed.

Out of these, the first one was the most severe. But then again, it wasn't that easy to replicate. Should something still surface later, we'll handle it in its own separate issue.

c-pius:
> 1. **SSA patch on a terminating Manifest** — if label-removal-finalizer is already blocking deletion when the user re manages, patchManifest is called on an object with DeletionTimestamp set. SSA on a terminating object may fail silently, leaving the module permanently stuck.

That's okay I think. Once there is a deletion timestamp, the object must finalize. If it doesn't the problem is somewhere else I think.

> 2. **Partially stripped SKR labels not re-stamped** — label removal iterates status.synced and strips managed-by labels resource by resource. If re-manage happens mid-way, some SKR resources permanently lose their labels. The install path doesn't re-stamp them.

This is incorrect I think. The resource transform applies the label to all resources installed for the manifest. They label will be re-applied.

> 3. **Race between Kyma controller SSA and Manifest controller in-flight delete** — the Kyma controller removes is unmanaged via SSA, but the Manifest controller may have already read the stale object and be mid-way through handleUnmanagedManifest. The Delete call goes through despite the annotation being removed.

That's okay. Once a deletion timestamp is there, the object must finalize. We will just create a new one.

c-pius (dismissed):
Approved given the unit test coverage is updated :)

c-pius (approved):

PR #3554, opened 2026-08-26
docs: Update the folder name per guidelines

mmitoraj:
**Description**

In the last release for the restricted markets, the notable change from Lifecycle Manager wasn't automatically pulled because of the wrong folder name. PTAL at the [issue](https://github.tools.sap/kyma/backlog/issues/10356) and the [Slack thread](https://sap-btp.slack.com/archives/CQF0V9075/p1787136214833449). 

- Updated the folder name per guidelines so that we have the correct pattern to follow in the future for the automation to work

**Related issue(s)**


NHingerl (dismissed):

c-pius (commented):
Maybe add here in the first point to explicitly not use `v` prefix?

https://github.com/kyma-project/lifecycle-manager/blob/main/docs/contributor/15-notable-changes.md#creating-a-notable-change

c-pius (approved):

IwonaLanger (approved):

PR #3562, opened 2026-08-31
fix: Prevent backoff bypass in Manifest controller error loops

### LINKED ISSUE #3523: Investigate aggressive Error Requeues Caused by persistent Module Errors
## Objective

Investigate how module error states cause disproportionate load on KLM and the KCP API server, and define a mitigation strategy that prevents a single misbehaving module (or a stuck cluster state) from degrading the central control plane. Eventually just a relaxed version of requeues on certain errors, that might appear more frequently.

### Background: Observed Incident (2026-06-25/26, Prod env)

A concrete incident was observed on 2026-06-25 in production (another incident in the comment below). The `kyma-system` namespace on one SKR became stuck in `Terminating` phase, blocked by finalizers from multiple module operators (btp-operator, cloud-manager, connectivity-proxy, istio, telemetry, warden). With the namespace terminating, every Manifest reconciliation attempted to patch resources into it and received:

```
forbidden: unable to create new content in namespace kyma-system because it is being terminated
```

7 Manifests for the affected Kyma were all stuck in Error state. With `manifest-requeue-error-interval` at 2s and `failure-max-delay` at 30s, this generated a sustained requeue storm:

- ~700,000 `Reconciler error` log entries in 12 hours
- ~687,000 `Invalidating manifest-controller client cache entry` log entries in 12 hours
- ~8,000–10,000 log events per 10-minute window (sustained, not a spike)
- KLM CPU climbing to **7.5 cores** (from a ~2.5 core baseline)
- KLM memory at **~10 GiB**
- KCP Kube API request rate spiking to **~2,500 ops/s**

The root cause was entirely outside KLM's control (a namespace stuck due to module operator finalizers), yet it caused significant resource pressure on the central KCP component. The Plutono dashboard showed the manifest controller reconcile errors as the dominant signal, with active manifest workers peaking at 75.
See evidence below.

## Acceptance Criteria

- [ ] The requeue path for Manifest in Error state is documented end-to-end, including how the rate limiter backoff interacts with `manifest-requeue-error-interval`
- [x] It is established whether the exponential backoff (`failure-base-delay` / `failure-max-delay`) is applied consistently for all error paths, or whether some paths bypass it and requeue with the fixed `manifest-requeue-error-interval` instead
- [x] The specific error category "namespace terminating" (and analogous permanent/semi-permanent error states) is identified: does KLM distinguish transient errors (retry fast) from persistent errors (back off aggressively)?
- [x] A mitigation strategy is defined: options may include a higher `failure-max-delay` cap, per-Manifest error budgets, error-type-aware requeue intervals, or a circuit breaker that throttles reconciliation of persistently-erroring Manifests
- [ ] If a safe configuration change is sufficient, it is applied; otherwise a follow-up implementation issue is created
- [ ] The impact on KLM resource usage of the chosen mitigation is validated (ideally against a reproduced scenario in the load test environment)

## Screenshots & logs

Image
Image
Image
Image
Image

[log.json](https://github.com/user-attachments/files/31418484/log.json)

[manifest-telemetry.yaml](https://github.com/user-attachments/files/31418494/manifest-telemetry.yaml)
[manifest-warden.yaml](https://github.com/user-attachments/files/31418496/manifest-warden.yaml)
[kyma-cr.yaml](https://github.com/user-attachments/files/31418495/kyma-cr.yaml)
[runtime-cr.yaml](https://github.com/user-attachments/files/31418493/runtime-cr.yaml)
[kyma-system-ns.yaml](https://github.com/user-attachments/files/31418497/kyma-system-ns.yaml)

LeelaChacha:
## Summary

When module resources fail concurrently (e.g. a namespace is Terminating), two issues combined to completely defeat the Manifest controller's exponential backoff:

**Root cause 1 — non-deterministic error string (`ssa.go`)**
`ConcurrentDefaultSSA.Run` collects per-resource errors from goroutines via a channel. The receive order is non-deterministic, so `errors.Join` produces a different combined error string on every reconcile cycle when ≥2 resources fail. `HasStatusDiff` compares `LastOperation.Operation` strings, so it returns `true` on every cycle, triggering a status patch each time.

**Root cause 2 — missing predicate (`setup.go`)**
`For(&Manifest{})` had no predicate. Status subresource patches don't increment `metadata.generation`, so every patch fired a `MODIFIED` watch event that called `queue.Add` directly — bypassing `AddRateLimited` and resetting the backoff timer.

Together, these produced a tight loop (~2–5 reconciles/second per manifest) instead of the intended 5–30s exponential backoff. In a production incident this caused ~700k error log entries in 12h from 7 stuck manifests, 7.5× CPU baseline, and ~2500 KCP API ops/s.

## Changes

- `ConcurrentDefaultSSA.Run`: sort per-resource errors before `errors.Join` so the combined string is deterministic from cycle 2 onward — `HasStatusDiff` returns false, no status patch fires, backoff accumulates normally.
- `SetupWithManager`: add `GenerationChangedPredicate` to the Manifest self-watch so status-only updates are filtered and never bypass the rate limiter.

## Verification

Reproduced locally with a k3d KCP+SKR setup: deployed template-operator, forced `template-operator-system` into Terminating, then bumped the module version to trigger concurrent SSA failures. With both fixes reverted the `manifest_unauthorized` requeue counter reached ~120–300 in 60 seconds (backoff bypassed). With both fixes applied it stayed ≤15 (proper 1s→10s backoff schedule).

closes #3523


LeelaChacha:
## Reproduction test

To verify locally, drop these two files into `tests/e2e/` and run:

```bash
make -f tests/e2e/backoff_bypass_test.mk test
```

With **both fixes reverted** the `manifest_unauthorized` counter hits ~120–300 in 60s (test fails). With the fixes it stays ≤15.

<details>
<summary><code>backoff_bypass_test.mk</code></summary>

```makefile
.DEFAULT_GOAL := test
.PHONY: test $(MAKECMDGOALS)

.PHONY: klm-patch
klm-patch: kustomize-install
	@echo "::group::KLM patch - short backoff delays for backoff-bypass test"
	@pushd $(LIFECYCLE_MANAGER_DIR)/config/watcher_local_test > /dev/null
	kustomize edit add patch --kind Deployment --patch \
		'[{"op":"add","path":"/spec/template/spec/containers/0/args/-","value":"--failure-base-delay=1s"},{"op":"add","path":"/spec/template/spec/containers/0/args/-","value":"--failure-max-delay=10s"}]'
	@popd > /dev/null
	@echo "::endgroup::"

.PHONY: module-setup
module-setup: module-setup-latest module-setup-in-newer-version
	@echo "::group::ModuleReleaseMeta setup for backoff-bypass test (initial: regular:$(MODULE_DEPLOYABLE_VERSION))"
	$(SCRIPTS_DIR)/deploy_modulereleasemeta.sh $(MODULE_NAME) regular:$(MODULE_DEPLOYABLE_VERSION)
	@echo "::endgroup::"

.PHONY: test-run
test-run: log-tool-versions
	@echo "::group::Setting kubeconfig variables"
	@export KCP_KUBECONFIG=$(shell k3d kubeconfig write kcp)
	@export SKR_KUBECONFIG=$(shell k3d kubeconfig write skr)
	@echo "::endgroup::"

	@echo "::group::E2E test: Namespace Terminating Backoff Bypass"
	@pushd $(E2E_TESTS_DIR) > /dev/null
	set +e; $(GO) test -timeout 20m -ginkgo.v -ginkgo.focus "Namespace Terminating Backoff Bypass"; status=$$?; set -e
	@popd > /dev/null
	@echo "::endgroup::"
	exit $${status}

.PHONY: test
test: create-clusters klm-patch deploy-klm module-setup test-run
```

</details>

<details>
<summary><code>backoff_bypass_test.go</code></summary>

```go
package e2e_test

import (
	"fmt"
	"time"

	apicorev1 "k8s.io/api/core/v1"
	"sigs.k8s.io/controller-runtime/pkg/client"

	"github.com/kyma-project/lifecycle-manager/internal/pkg/metrics"
	"github.com/kyma-project/lifecycle-manager/pkg/queue"

	. "github.com/kyma-project/lifecycle-manager/tests/e2e/commontestutils"
	. "github.com/onsi/ginkgo/v2"
	. "github.com/onsi/gomega"
)

const (
	backoffTestModuleName    = "template-operator"
	backoffBlockingFinalizer = "e2e.test/block-termination"
	moduleNamespace          = "template-operator-system"
	measurementDuration      = 60 * time.Second
	maxExpectedRequeueCount  = 15
)

var _ = Describe("Namespace Terminating Backoff Bypass", Ordered, func() {
	kyma := NewKymaWithNamespaceName("kyma-sample", ControlPlaneNamespace, v1beta2.DefaultChannel)

	InitEmptyKymaBeforeAll(kyma)
	CleanupKymaAfterAll(kyma)

	Context("Given a Kyma CR with template-operator enabled", func() {
		It("When the module is enabled on the SKR cluster", func() {
			Eventually(EnableModule).
				WithContext(ctx).
				WithArguments(skrClient, defaultRemoteKymaName, RemoteNamespace,
					NewTemplateOperator(v1beta2.DefaultChannel)).
				Should(Succeed())
		})

		It("And the Manifest reaches Ready state on KCP", func() {
			Eventually(ManifestExists).
				WithContext(ctx).
				WithArguments(kcpClient, kyma.GetName(), kyma.GetNamespace(), backoffTestModuleName).
				WithTimeout(5 * time.Minute).
				Should(Succeed())

			Eventually(func(g Gomega) {
				manifest, err := GetManifest(ctx, kcpClient,
					kyma.GetName(), kyma.GetNamespace(), backoffTestModuleName)
				g.Expect(err).NotTo(HaveOccurred())
				g.Expect(manifest.Status.State).To(Equal(shared.StateReady))
			}).WithTimeout(5 * time.Minute).Should(Succeed())
		})
	})

	Context("When the module namespace enters Terminating and the module version is bumped", func() {
		It("Adds a blocking finalizer to the module namespace and deletes it", func() {
			ns := &apicorev1.Namespace{}
			Expect(skrClient.Get(ctx, client.ObjectKey{Name: moduleNamespace}, ns)).To(Succeed())
			original := ns.DeepCopy()
			ns.Finalizers = append(ns.Finalizers, backoffBlockingFinalizer)
			Expect(skrClient.Patch(ctx, ns, client.MergeFrom(original))).To(Succeed())
			Expect(skrClient.Delete(ctx, &apicorev1.Namespace{ObjectMeta: ns.ObjectMeta})).To(Succeed())
		})

		It("Then the module namespace is in Terminating state", func() {
			Eventually(func(g Gomega) {
				ns := &apicorev1.Namespace{}
				g.Expect(skrClient.Get(ctx, client.ObjectKey{Name: moduleNamespace}, ns)).To(Succeed())
				g.Expect(ns.DeletionTimestamp).NotTo(BeNil())
			}).WithTimeout(30 * time.Second).Should(Succeed())
		})

		It("Bumps the MRM to the newer version to force a manifest upgrade attempt", func() {
			mrm := &v1beta2.ModuleReleaseMeta{}
			Expect(kcpClient.Get(ctx,
				client.ObjectKey{Name: backoffTestModuleName, Namespace: ControlPlaneNamespace},
				mrm)).To(Succeed())
			original := mrm.DeepCopy()
			for i, ch := range mrm.Spec.Channels {
				if ch.Channel == string(v1beta2.DefaultChannel) {
					mrm.Spec.Channels[i].Version = NewerVersion
					break
				}
			}
			Expect(kcpClient.Patch(ctx, mrm, client.MergeFrom(original))).To(Succeed())
		})
	})

	Context("Then the Manifest controller requeue rate stays within backoff bounds", func() {
		It("Verifies that backoff is not bypassed during repeated 403 Forbidden errors", func() {
			requeueReason := string(metrics.ManifestUnauthorized)
			requeueType := string(queue.UnexpectedRequeue)

			Eventually(func() (int, error) {
				return GetRequeueReasonCount(ctx, requeueReason, requeueType)
			}).WithTimeout(3 * time.Minute).WithPolling(5 * time.Second).
				Should(BeNumerically(">", 0))

			count1, err := GetRequeueReasonCount(ctx, requeueReason, requeueType)
			Expect(err).NotTo(HaveOccurred())
			time.Sleep(measurementDuration)
			count2, err := GetRequeueReasonCount(ctx, requeueReason, requeueType)
			Expect(err).NotTo(HaveOccurred())

			delta := count2 - count1
			GinkgoWriter.Printf("manifest_unauthorized requeues in %s: %d\n", measurementDuration, delta)

			Expect(delta).To(BeNumerically("<=", maxExpectedRequeueCount),
				fmt.Sprintf("backoff bypassed: %d requeues in %s (max expected: %d)",
					delta, measurementDuration, maxExpectedRequeueCount))
		})
	})
})
```

</details>

LeelaChacha:
Follow up issue created: https://github.com/kyma-project/lifecycle-manager/issues/3563

Tomasz-Smelcerz-SAP (approved):

